PR #20, opened 2022-03-23
Draft: Add form component

fikrikarim:


packages/runtime/src/prop-controllers/instances.ts
```diff
@@ -120,8 +126,43 @@ class RichTextPropController extends PropController<RichTextPropControllerMessag
...
+}
+
+export type TableFormFieldsMessage =
+  | TableLayoutTableFormFieldsMessage
+  | TableFieldLayoutTableFormFieldsMessage
+
+export class TableFormFieldsPropController extends PropController<TableFormFieldsMessage> {
+  recv(): void {}
+
+  subscribe() {}
```
migueloller, line 151: Forgot to remove this method?

packages/runtime/src/prop-controllers/instances.ts
```diff
@@ -120,8 +126,43 @@ class RichTextPropController extends PropController<RichTextPropControllerMessag
...
 }
 
+export const TableFormFieldsMessageType = {
+  TABLE_FORM_LAYOUT_CHANGE: 'TABLE_FORM_LAYOUT_CHANGE',
+  TABLE_FORM_FIELD_LAYOUT_CHANGE: 'TABLE_FORM_FIELD_LAYOUT_CHANGE',
+} as const
+
+type TableLayoutTableFormFieldsMessage = {
+  type: typeof TableFormFieldsMessageType.TABLE_FORM_LAYOUT_CHANGE
+  value: { id: string; layout: BoxModel }
```
migueloller, line 136: We don't have to nest `id` and `layout` inside `value` here

migueloller (changes requested):

packages/runtime/src/prop-controllers/instances.ts
fikrikarim: Updated

packages/runtime/src/prop-controllers/instances.ts
fikrikarim: Updated

packages/runtime/src/components/hooks/useTableFormFieldRefs.ts
```diff
@@ -0,0 +1,56 @@
...
+    let animationFrameHandle = requestAnimationFrame(handleAnimationFrameRequest)
+
+    return () => {
+      cancelAnimationFrame(animationFrameHandle)
+    }
+
+    function handleAnimationFrameRequest() {
+      if (propController == null) return
+
+      if (container != null) propController.tableFormLayoutChange({ id, layout: getBox(container) })
```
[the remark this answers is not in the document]
fikrikarim: Okay I removed the id, but keep the object the same, to match it with the other method.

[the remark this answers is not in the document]
migueloller: That sounds good to me!

packages/runtime/src/prop-controllers/instances.ts
migueloller: That sounds good to me

PR #22, opened 2022-03-23
feat: add PageId Context

joshuawootonn:


packages/runtime/src/runtimes/react.tsx
```diff
@@ -161,6 +161,22 @@ function useDocumentKey(): string | null {
   return useContext(DocumentContext)
 }
 
+const PageIdContext = createContext<string | null>(null)
+
+export function usePageIdOrNull(): string | null {
+  return useContext(PageIdContext)
+}
+
+export function usePageId(): string | null {
```
joshuawootonn, line 170: when I use `usePageIdOrNull` in renderer/page it is never null

when I use `usePageId` in renderer/page error is throw "PageId context must be provided."

I think this is because it apollo cache is being primed in getServerSideProps in `[[...path]].tsx` and `preview.tsx`, but even when I put the provider there it still errored out.

so I am using `usePageIdOrNull`

joshuawootonn:
this pageId is then used within page to get the default typography

packages/runtime/src/runtimes/react.tsx
alexhwoods: So should the return type on this just be `string`?

alexhwoods (approved):
Just the one comment!

Might not be the best person to review react code though 😅

packages/runtime/src/runtimes/react.tsx
joshuawootonn: Yes good catch!

migueloller: We shouldn't type the function as `string` unless we're actually ensuring that it's not a string. If the context default value is `null` and so the types are `null | string` then we need to check for presence of the value if the return type is `string`.

If an error is being thrown then in those cases the value is indeed `null`, so I'm not sure I understand what you're saying. You're saying "it's never null" but also saying that it throws an error but the error is only thrown when it's `null`? That doesn't make sense to me.

migueloller (approved):
This looks good to me!

PR #26, opened 2022-03-30
Add Typeahead Descriptor

fikrikarim:


migueloller (approved):
I have some thoughts on what the API should be for the typeahead control given functionality that I think we should support later on but we can make those changes later. What this PR introduces is perfect for what we need at the moment.

PR #27, opened 2022-03-30
Serialization for composable controls

migueloller:


fikrikarim:
Looking great! Tested a bit and it's working with `List` and `Shape`. Demo in Slack!

PR #33, opened 2022-04-13
Hide scrollbar when in builder

fikrikarim:


migueloller (changes requested):

packages/runtime/src/runtimes/react.tsx
```diff
@@ -116,15 +116,23 @@ export function RuntimeProvider({
 
     if (!isInBuilder) return
 
-    // TODO(miguel): Uncomment the lines below once we figure out how to make it so that we can hide
-    // the scrollbar in the builder without preventing scrolling in preview mode.
-    // const lastDocumentOverflow = window.document.documentElement.style.overflow
-    // window.document.documentElement.style.overflow = 'hidden'
+    // Hide scrollbar: https://stackoverflow.com/q/16670931/
+    window.document.head.insertAdjacentHTML(
```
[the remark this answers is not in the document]
fikrikarim: Yes I saw that commit. The change in this MR would hide the scroll bar both in builder and in preview, but it's still scrollable on the preview page; user just don't see the scroll bar.

The problem with `overflow = hidden` was that it disables scrolling on preview page. Instead, we hide the scroll bar but the user could still scroll the preview page.

I agree that the best case is to handle this just for the builder and not the preview page, but it'll take more time to implement that. If you think we should do that then i can try to implement it that way.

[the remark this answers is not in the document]
migueloller: Ok. I think that the requirements were probably improperly communicated. The problem here isn't just "hide the scrollbar". It's that we need different behavior when the builder is in "edit mode" vs when it is in "preview mode". Once of those behaviors is setting the `overflow` property to `hidden`.

I believe we already have an action that we send to the page when initializing the builder. It's in line 217 of `PreviewFrame.tsx` (at the moment of writing this) and it looks like this:

```typescript
broadcastPort?.postMessage(init())
```

We should just move the commented out code to the thunk where initialization and cleanup happen and that should take care of only doing it in the builder and not preview mode.

Later on, once preview mode is part of the builder and it's a "mode" you can switch between, then we can change the logic so that it happens depending on the "mode" you're in instead of on "init".

Does this make sense?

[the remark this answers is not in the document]
fikrikarim: Yes cool. Updated!

packages/runtime/src/state/react-builder-preview.ts
```diff
@@ -197,6 +197,36 @@ function getElementSize(element: HTMLElement): Size {
   }
 }
 
+function handleFocusEvent(): ThunkAction<() => void, unknown, unknown, Action> {
```
migueloller, line 200: Why make this a thunk if it doesn't rely on `dispatch` or `getState`?

migueloller: Also, the naming is a bit confusing. This is not an event handler, it's more of an effect since it returns a cleanup function.

migueloller: Some suggestions: `lockDocumentScroll`, `lockDocumentScrollEffect`, `setUpFocusListeners`, `setUpFocusListenersEffect`.

migueloller: Another one, `startHandlingFocusEvents`.

migueloller (approved):

PR #37, opened 2022-04-15
feat: add robots and sitemap

fikrikarim:
Example usage:

On `pages/sitemap.xml.tsx`:
```ts
export {
  sitemapXmlGetServerSideProps as getServerSideProps,
  SitemapXml as default,
} from "@makeswift/runtime/next";
```

### DIFF
--- a/packages/runtime/src/next.tsx
+++ b/packages/runtime/src/next.tsx
@@ -146,3 +146,116 @@ export function Page({ page, rootElement, makeswiftApiEndpoint, cacheData }: Pag
     </RuntimeProvider>
   )
 }
+
+export async function robotsTxtGetServerSideProps({
+  req,
+  res,
+}: GetServerSidePropsContext): Promise<GetServerSidePropsResult<Record<string, never>>> {
+  const baseUrl = new URL(`https://${req.headers.host}`)
+
+  res.setHeader('Content-Type', 'text/plain')
+
+  const sitemapUrl = new URL(`sitemap.xml`, baseUrl)
+
+  res.write(
+    `User-agent: *
+Allow: /
+Sitemap: ${sitemapUrl}`,
+  )
+
+  res.end()
+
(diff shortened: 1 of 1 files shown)


migueloller:
> > @fikrikarim and @joshuawootonn, maybe we should sync up and figure out what we're going to do for this PR?
> 
> Yea let's do that

Ok, so what I've been thinking about this stuff is that certain hosts can opt-in to certain features. For example, a multi-tenant host would opt-in to the feature of domains. Then, we can have an endpoint (e.g., `/api/makeswift/features`) that gives a list of what features are supported. I think that sitemaps and robots.txt can be such features.

So, the host could decide to generate a sitemap and be completely independent of Makeswift, or it could use the Makeswift API to help in generating the sitemap. If it does use the Makeswift API, then we would want to show sitemap related fields in the page settings, otherwise we don't. This is similar to where if a host is multi-tenant and supports domains we want to show domain settings, and otherwise we don't.

So once we've got that stuff set up, we then need to decide what the APIs look like for the sitemap and robots.txt files. We need both an API to get data into the Next.js app as well as an API to actually serve the files. Whatever API we come up with I think it should work well out of the box without any configuration while also be fully extensible so that if you wanted to have certain parts be local and use something like `next-sitemap`, while other parts you Makeswift data, you can do that.

Maybe one of you guys can create an RFC with some ideas?

joshuawootonn:
Happy to do this.I can draft something next week!

migueloller:
@fikrikarim, @joshuawootonn, once we merge #150 we should be able to enhance the new `Makeswift.getPages` API so that you get the data you need to generate a sitemap. We can then create some documentation as well as adding to examples using this API together with `next-sitemap` to generate a sitemap. Thoughts?

joshuawootonn:
I like that idea! @migueloller

PR #65, opened 2022-04-29
Set isInBuilder true after initialize

fikrikarim:


fikrikarim:
> Can we use the new reducer in the `react-page` state module?

Updated!

PR #73, opened 2022-05-16
fix: disable element registration inside unfocused element reference

fikrikarim:
Before, when you put two nested global components, the overlay is shown
on the wrong element.
This is happening because a nested global element have the same key,
and we're registering all elements.

This commit fix it by disabling element registration, when it's inside
an unfocused element reference. We know whether element reference
is focused or not, by checking whether the elementReferenceDocument
is null.

### DIFF
--- a/.changeset/yellow-wombats-peel.md
+++ b/.changeset/yellow-wombats-peel.md
@@ -1,0 +1,5 @@
+---
+"@makeswift/runtime": patch
+---
+
+Fix overlay for nested global components are showing the wrong selection, making you unable to edit.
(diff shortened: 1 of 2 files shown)


migueloller (approved):
This looks good to me. I'd like us to explore a more "elegant" way to achieve this in the future, though. This seems a bit "hacky".

fikrikarim:
> This looks good to me. I'd like us to explore a more "elegant" way to achieve this in the future, though. This seems a bit "hacky".

Yes I agree. 

What confused me was that I thought `ElementData` is purely only to render the data, without registering anything. But I just realized that `ElementData` will render `Box`, and `Box` will render `Element`, which would register the element inside it.

This is the reason that we'll still registering elements inside an unfocused element reference, and the cause of the overlay bug.

PR #77, opened 2022-05-26
fix: snippets issues

fikrikarim:
Problems:
1. Snippets aren't updating immediately when added, removed, or edited.
This is because we got the snippets value from getServerSideProps, which
doesn't get updated when there's any change in apollo cache.
This MR use the snippets value from apollo cache when it's in preview.

2. Snippet set to "Live page" is still showing in the builder.
We forgot to pass the preview props on the Page component.

I need to add the `typescript` to the codegen because we need to access the `SnippetLocation` enum, which isn't generated if we're just using the `typescript-operations`. Let me know if there's any config to add the `SnippetLocation` enum without generating the whole thing.

### DIFF
--- a/packages/runtime/src/api/react.tsx
+++ b/packages/runtime/src/api/react.tsx
@@ -60,6 +60,9 @@ const typePolicies: TypePolicies = {
       table(existingData, { args, toReference }) {
         return existingData ?? toReference({ __typename: 'Table', id: args?.id }, true)
       },
+      page(existingData, { args, toReference }) {
+        return existingData ?? toReference({ __typename: 'Page', id: args?.id }, true)
+      },
     },
   },
 }
--- a/.changeset/hip-wolves-talk.md
+++ b/.changeset/hip-wolves-talk.md
@@ -1,0 +1,5 @@
+---
+"@makeswift/runtime": patch
+---
+
+Fix snippets not updated immediately after changes
--- a/packages/runtime/codegen.yml
+++ b/packages/runtime/codegen.yml
@@ -6,6 +6,8 @@ generates:
     plugins:
       - typescript-operations
       - typed-document-node
+      - add:
+          content: export enum SnippetLocation { Body = 'BODY', Head = 'HEAD' }
 hooks:
   afterOneFileWrite:
     - prettier --write
--- a/packages/runtime/package.json
+++ b/packages/runtime/package.json
@@ -134,6 +134,7 @@
     "uuid": "^3.3.3"
   },
   "devDependencies": {
+    "@graphql-codegen/add": "^3.1.0",
     "@graphql-codegen/cli": "2.6.2",
     "@graphql-codegen/typed-document-node": "^2.2.7",
     "@graphql-codegen/typescript-operations": "2.3.4",
--- a/packages/runtime/src/api/constants.ts
+++ b/packages/runtime/src/api/constants.ts
@@ -7,6 +7,8 @@ import {
   PagePathnameSliceFragmentDoc,
   GlobalElementFragmentDoc,
   TableFragmentDoc,
+  SnippetFragmentDoc,
+  PageFragmentDoc,
 } from './generated/graphql'
 import type { APIResourceType } from './types'
 
@@ -17,4 +19,6 @@ export const Fragments: Record<APIResourceType, DocumentNode> = {
   PagePathnameSlice: PagePathnameSliceFragmentDoc,
   GlobalElement: GlobalElementFragmentDoc,
   Table: TableFragmentDoc,
+  Snippet: SnippetFragmentDoc,
+  Page: PageFragmentDoc,
 }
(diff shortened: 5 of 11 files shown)


packages/runtime/src/api/react.tsx
```diff
@@ -141,6 +144,12 @@ export function useMutation<
   return useApolloMutation(mutation, { client: client?.apolloClient, ...options })
 }
 
+export function useApolloClient() {
```
migueloller, line 147: This hook doesn't seem to be used anywhere.

packages/runtime/src/api/react.tsx
```diff
@@ -60,6 +60,9 @@ const typePolicies: TypePolicies = {
       table(existingData, { args, toReference }) {
         return existingData ?? toReference({ __typename: 'Table', id: args?.id }, true)
       },
+      page(existingData, { args, toReference }) {
+        return existingData ?? toReference({ __typename: 'Page', id: args?.id }, true)
+      },
```
migueloller, line 65: Do we need a type policy for snippets as well?

migueloller (changes requested):
This looks good to me. I have a few thoughts and concerns.

### Concerns
- I'm concerned that we're generating all the types for the GraphQL API. I purposefully avoided this because this project is OSS and our API is a private API not meant to be used publicly. I would rather we hard-code the type definitions for the enum instead of having to generate types for the _entire_ API schema just for the one enum.

### Thoughts/Questions
- I don't see why we shouldn't move _all_ data fetching to the page component and do _nothing_ in `getServerSideProps`/`getStaticProps`. We want things to be in sync, so this means that the page component should fetch and subscribe to all these API resources (snippets, site metadata, etc.). This is part of why I realized we won't need the REST API. If we shift all data requirements to be inside the page component, then really all that's needed from `getServerSideProps`/`getStaticProps` is a mapping from page path to a page ID so that the `Page` component can do its job. Even then, the `Page` component could just take a path and take it from there. This is something that we can properly implement later, though.
- the `preview` distinction is being hard-coded for `getServerSideProps` and `getStaticProps` but this is quite different from "is in builder" and "is in live page". Right now, preview mode uses the preview endpoint that uses `getServerSideProps` but it's not "in the builder". Then again, what's the behavior we've had in prod in the past? If a snippet is enabled in the builder, is it also enabled in preview mode?
- Do we have the necessary changes in the builder to send the API resource changes (i.e., `Page` and `Snippet`)?
- I left some inline comments in the code

packages/runtime/src/api/react.tsx
fikrikarim: We don't need it right now because we don't query the snippets directly.

packages/runtime/src/api/react.tsx
fikrikarim: Removed

fikrikarim:
> This looks good to me. I have a few thoughts and concerns.
> 
> ### Concerns
> * I'm concerned that we're generating all the types for the GraphQL API. I purposefully avoided this because this project is OSS and our API is a private API not meant to be used publicly. I would rather we hard-code the type definitions for the enum instead of having to generate types for the _entire_ API schema just for the one enum.
> 
> ### Thoughts/Questions
> * I don't see why we shouldn't move _all_ data fetching to the page component and do _nothing_ in `getServerSideProps`/`getStaticProps`. We want things to be in sync, so this means that the page component should fetch and subscribe to all these API resources (snippets, site metadata, etc.). This is part of why I realized we won't need the REST API. If we shift all data requirements to be inside the page component, then really all that's needed from `getServerSideProps`/`getStaticProps` is a mapping from page path to a page ID so that the `Page` component can do its job. Even then, the `Page` component could just take a path and take it from there. This is something that we can properly implement later, though.
> * the `preview` distinction is being hard-coded for `getServerSideProps` and `getStaticProps` but this is quite different from "is in builder" and "is in live page". Right now, preview mode uses the preview endpoint that uses `getServerSideProps` but it's not "in the builder". Then again, what's the behavior we've had in prod in the past? If a snippet is enabled in the builder, is it also enabled in preview mode?
> * Do we have the necessary changes in the builder to send the API resource changes (i.e., `Page` and `Snippet`)?
> * I left some inline comments in the code

- Yes I totally agree with that. Changed it to hardcode the enum.
- Again. Yess I agree. I just think it's a pretty big change to do that. But for now, just the snippet is enough, because that's the only thing that we need to automatically update if there's any change in builder. The others are seo, meta, and fonts. We might need to do the same thing here for fonts. Will test that tomorrow.
- In prod, just checked a bit and the preview page follows what's in the builder.
- I think so yea. I'm running leo on parity branch

migueloller (approved):
@fikrikarim, changes look good! Makes sure to test fonts as well. I would imagine that if a new font is added it won't be used in the page so the text will look unstyled.

fikrikarim:
> @fikrikarim, changes look good! Makes sure to test fonts as well. I would imagine that if a new font is added it won't be used in the page so the text will look unstyled.

Yes Andrew had comment on it, but I cannot reproduce it in dev. I'll try to reproduce it again to make sure we fix it. Will open another PR for that.

PR #79, opened 2022-05-27
fix: font not applied when added to site

fikrikarim:


packages/runtime/src/components/page/Page.tsx
```diff
@@ -77,6 +78,20 @@ function snippetToElement(snippet: Pick<Snippet, 'id' | 'code'>): (string | Reac
   })
 }
 
+const filterImportantSnippetProperty = ({
```
migueloller, line 81: It's not clear what "important" means here

migueloller (approved):

PR #80, opened 2022-05-27
fix: send onwheel event to builder

fikrikarim:
This will fix scrolling doesn't work in content mode
when hovering over a text block

### DIFF
--- a/packages/runtime/src/state/react-builder-preview.ts
+++ b/packages/runtime/src/state/react-builder-preview.ts
@@ -34,6 +34,7 @@ import {
   unregisterMeasurable,
   unregisterPropControllers,
   setIsInBuilder,
+  handleWheel,
 } from './actions'
 import { ActionTypes } from './actions'
 import { createPropController, PropController } from '../prop-controllers/instances'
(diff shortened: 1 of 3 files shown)


packages/runtime/src/state/react-builder-preview.ts
```diff
@@ -199,12 +200,20 @@ function getElementSize(element: HTMLElement): Size {
...
-  window.document.documentElement.style.overflow = 'hidden'
+function lockDocumentScroll(): ThunkAction<() => void, State, unknown, Action> {
+  return dispatch => {
+    const lastDocumentOverflow = window.document.documentElement.style.overflow
+    window.document.documentElement.style.overflow = 'hidden'
 
-  return () => {
-    window.document.documentElement.style.overflow = lastDocumentOverflow
+    const lastDocumentOnWheel = window.document.documentElement.onwheel
+    window.document.documentElement.onwheel = ({ deltaX, deltaY }) => {
```
migueloller, line 209: Why aren't we using `addEventListener` here?

migueloller: Is the reason we're not using the `scroll` event here that we've locked scrolling? Should we unlock scrolling instead  when in content mode? Perhaps this is something to worry about later.

migueloller (changes requested):

packages/runtime/src/state/react-builder-preview.ts
fikrikarim: In Leo we're handling it using onWheel event, I'm using just using the same event:

image

For why we're not using `addEventListener`, yea we could listen to `wheel` listener. I tried using `onwheel` instead and got error.

Even if we unlock the scroll, I think we still can't scroll left in content mode, because it's handled in Leo instead of scrolling the host.

fikrikarim: Cool just fixed that. Thanks for the review!

PR #87, opened 2022-05-31
feat: add MakeswiftBadge snippet to Page

fikrikarim:


migueloller:
Why are we adding a free badge to the custom host? I thought we had agreed that if a customer wasn’t in the plan that includes the custom host their requests to the API would just fail?

fikrikarim:
Ah this was made before the outage. We have to use this before because the renderer depend on this.

fikrikarim:
We might need to open this MR again when we use runtime for everyone.

migueloller:
> We might need to open this MR again when we use runtime for everyone.

The badge does not belong on the runtime, it belongs on the host. Whatever set up we have later, the badge should be added in the host and not the runtime.

PR #95, opened 2022-06-09
feat: use introspection for SSR instead of using getDataFromTree

fikrikarim:


packages/runtime/src/runtimes/react/index.tsx
```diff
@@ -62,11 +67,16 @@ export function createReactRuntime(store: ReactPage.Store): ReactRuntime {
         unregisterReactComponent()
       }
     },
+    getPropControllerDescriptors() {
```
migueloller, line 70: Is there a way we can avoid exposing this in the public `ReactRuntime` API?

packages/runtime/src/runtimes/react/index.tsx
```diff
@@ -85,17 +95,10 @@ export function RuntimeProvider({
...
       rootElements,
     })
-    const runtime = createReactRuntime(store)
-
-    registerBuiltinComponents(runtime)
 
     return store
   })
 
-  useEffect(() => {
```
migueloller, line 95: Won't removing this effect cause issues where the store is out of sync? We create a new store when we connect to the builder and if we don't register builtin components to that new store, how are builtin components rendered?

migueloller (changes requested):
@fikrikarim, this looks pretty good! Looks like we still need to handle the list and shape prop controllers and controls. Also, the changes we're making to the runtime feel like they could break some things. We should be careful there and see if we can avoid making a change like that. Finally, it seems that the introspection function should receive an extra argument to get access to the descriptors. Globally accessing `ReactRuntime` that way doesn't feel like a good practice, specially when you start thinking about how we'd go about testing it.

packages/runtime/src/api/react.tsx
```diff
@@ -94,6 +111,82 @@ export type MakeswiftClientOptions = {
...
+      getTypographyIds(descriptor, parent.props[propName]).forEach(typographyId =>
+        typographyIds.add(typographyId),
+      )
+
+      getTableIds(descriptor, parent.props[propName]).forEach(tableId => tableIds.add(tableId))
+
+      getPageIds(descriptor, parent.props[propName]).forEach(pageId => pageIds.add(pageId))
+
+      getElementChildren(descriptor, parent.props[propName]).forEach(children =>
+        remaining.push(children),
```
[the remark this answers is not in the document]
fikrikarim: Yep I renamed it to `child`

packages/runtime/src/state/react-page.ts
```diff
@@ -79,7 +79,7 @@ export function getComponentPropControllerDescriptors(
   )
 }
 
-function normalizeElement(
+export function normalizeElement(
```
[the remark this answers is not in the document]
fikrikarim: Fixedd

packages/runtime/src/runtimes/react/index.tsx
fikrikarim: So the way it's working is that when we initialize a new store, we get do `getState()` of the previous store, and use it in the `preloadedState`:
```tsx
  const [store, setStore] = useState(() => {
    const store = ReactPage.configureStore({
      preloadedState: contextDefaultValue.getState(),
      rootElements,
    })

    return store
  })
```

packages/runtime/src/runtimes/react/index.tsx
fikrikarim: Yea I thought so too. I moved it to a new runtime called `InternalReactRuntime`.

fikrikarim:
> @fikrikarim, this looks pretty good! Looks like we still need to handle the list and shape prop controllers and controls. Also, the changes we're making to the runtime feel like they could break some things. We should be careful there and see if we can avoid making a change like that. Finally, it seems that the introspection function should receive an extra argument to get access to the descriptors. Globally accessing `ReactRuntime` that way doesn't feel like a good practice, specially when you start thinking about how we'd go about testing it.

@migueloller 
- So far checking on the controls, it seems like I added every control that have any GraphQL query. Let me know if I miss a control or a PropController.
- Can you specify more where do you feel that they could break things?
- Okay I make the introspection function to accept descriptors.
- I make it so that we have `InternalReactRuntime`. But I just don't know how we could get the descriptors (after builtin and custom component registrations), without importing it globally. How?

packages/runtime/src/runtimes/react/index.tsx
migueloller: We already have internal APIs that we can access directly by using the Redux store and selectors. The runtime API was created specifically to wrap the internal API and expose a public API. So having an "internal" runtime API is a bit roundabout.

packages/runtime/src/runtimes/react/index.tsx
migueloller: Ah, that's right. We needed to get rid of the effect anyway because we want component registration to work statically so we can do things like introspection on an API endpoint (like we are now) and server components

packages/runtime/src/runtimes/react/index.tsx
fikrikarim: We need to access the store, which is the `contextDefaultValue`. Should we just access this variable directly?

Otherwise how could we access the store from `getServerSideProps`?

fikrikarim:
> > > @fikrikarim, this looks pretty good! Looks like we still need to handle the list and shape prop controllers and controls. Also, the changes we're making to the runtime feel like they could break some things. We should be careful there and see if we can avoid making a change like that. Finally, it seems that the introspection function should receive an extra argument to get access to the descriptors. Globally accessing `ReactRuntime` that way doesn't feel like a good practice, specially when you start thinking about how we'd go about testing it.
> > 
> > 
> > @migueloller
> > 
> > * So far checking on the controls, it seems like I added every control that have any GraphQL query. Let me know if I miss a control or a PropController.
> > * Can you specify more where do you feel that they could break things?
> > * Okay I make the introspection function to accept descriptors.
> > * I make it so that we have `InternalReactRuntime`. But I just don't know how we could get the descriptors (after builtin and custom component registrations), without importing it globally. How?
> 
> * I didn't see any new controls missing. And for the older prop-controllers it seems you brought code over from the old introspection so that should cover it.
> * Specifically I was thinking about removing `useEffect` from the provider. But we should be good there.
> * Ok, that'll work for now. I was thinking that we could just pass the store as an argument and then use selectors and potentially actions in the future in there. Passing the descriptors kind of leaks that shape to that function, tangling the two together.
> * We will definitely need "global" state somewhere. I think we should minimize that, though. So instead of creating a new runtime and using that, we could export `contextDefaultValue`, which is the Redux store and pass that as the third argument to `introspect`.
> 
> Have you thought about writing tests for this function? Setting up some JSON fixtures and then asserting that introspection results are correct would be a good thing to have.

- Nice. Changed it so that we pass the redux store as the third argument.
- Yeah I did. We also need to mock some Apollo query if we want to do that. Definitely possible. I was just trying to see whether this is working or not. I could write some test for this later?

PR #99, opened 2022-06-14
fix: forward link prop in ImageContainer

fikrikarim:
This is creating issue in live pages, where link is not working
in Image, because we're not forwarding the prop from ImageContainer
to Link component.


migueloller (approved):
Ship it! :rocket:

PR #104, opened 2022-06-20
WIP: Add new format for Image Control: With Dimensions

fikrikarim:


packages/runtime/src/controls/image.ts
```diff
@@ -2,10 +2,25 @@ export type ImageControlData = string
...
+  URL: 'URL',
+  WithDimensions: 'WithDimensions',
+} as const
 
-export type ImageControlDefinition = { type: typeof ImageControlType; config: ImageControlConfig }
+type ImageControlValueFormat =
+  typeof ImageControlValueFormats[keyof typeof ImageControlValueFormats]
 
-export function Image(config: ImageControlConfig = {}): ImageControlDefinition {
+type ImageControlConfig = { label?: string; format?: ImageControlValueFormat }
```
migueloller, line 13: For simplicity, we could make `format` required here and then default it to `ImageControlFormat.URL` in the factory function.

packages/runtime/src/controls/image.ts
```diff
@@ -2,10 +2,25 @@ export type ImageControlData = string
 
 export const ImageControlType = 'makeswift::controls::image'
 
-type ImageControlConfig = { label?: string }
+export const ImageControlValueFormats = {
```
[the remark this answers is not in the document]
fikrikarim: Okay got it!

packages/runtime/src/controls/image.ts
fikrikarim: Wait, it means that if they want to provide a config for a label, they also need to add the `format` right?

migueloller: The argument for the `Image` function itself is optional, but in the type for the config it's required. If this is too much of a pain, don't worry about it, though. It would make it so that in our code we always know there's a `format`, though.

packages/runtime/src/controls/image.ts
```diff
@@ -2,10 +2,24 @@ export type ImageControlData = string
 
 export const ImageControlType = 'makeswift::controls::image'
 
-type ImageControlConfig = { label?: string }
+export const ImageControlValueFormat = {
+  URL: 'URL',
```
migueloller, line 6: Something I was doing to make sure that enums are used and people don't hard-code a value and also to scope things was using strings like `makeswift::controls::image::format::url`, etc.

packages/runtime/src/controls/image.ts
```diff
@@ -2,10 +2,24 @@ export type ImageControlData = string
...
+  config: T
+}
+
+export function Image<T extends ImageControlConfig>(
+  config: T = {} as T,
+): ImageControlDefinition<T> {
   return { type: ImageControlType, config }
 }
+
+Image.Formats = ImageControlValueFormat
```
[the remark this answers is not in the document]
fikrikarim: Changed.

packages/runtime/src/controls/image.ts
fikrikarim: Very cool. Changed!

migueloller (approved):

PR #109, opened 2022-07-12
Component Code-Splitting

migueloller:


fikrikarim:
Looking great! Two things:
- Should we also update the examples with the new Next.js config, and maybe an example for the dynamic import for the component registration?
- Is it possible to create a `registerComponent` function with code splitting by default? I'm just thinking that having users copy `forwardNextDynamicRef(patch => dynamic(() => patch(import('./Box'))))` everytime they register component is a bit too much.

Is it even possible to do something like this?
```
runtime.registerComponent(import('./Box'), {
  type: './components/Box/index.js',
})
```

packages/runtime/src/next.tsx
```diff
@@ -158,3 +169,49 @@ export function Page({ page, rootElement, makeswiftApiEndpoint, cacheData, previ
     </RuntimeProvider>
   )
 }
+
+const FORWARDED_NEXT_DYNAMIC_REF_KEY = '__forwarded_next_dynamic_ref__'
```
joshuawootonn, line 173: Does this code preserve forwarded refs through the dynamic element of the components?

migueloller:
> Looking great! Two things:
> 
> * Should we also update the examples with the new Next.js config, and maybe an example for the dynamic import for the component registration?

100%! We'll want people to start using the plugin instead of manually adding the images domain. That way if we change it, people just need to upgrade. As for the dynamic import, I might make a separate guide for code-splitting but I'm not sure I want to make all examples use it since the API is so clunky. Once we have the static modules API, this will all happen automatically for the user.

I'll make a task to update the docs

> * Is it possible to create a `registerComponent` function with code splitting by default? I'm just thinking that having users copy `forwardNextDynamicRef(patch => dynamic(() => patch(import('./Box'))))` everytime they register component is a bit too much.
> 
> Is it even possible to do something like this?
> 
> ```
> runtime.registerComponent(import('./Box'), {
>   type: './components/Box/index.js',
> })
> ```

Unfortunately no. The Next.js webpack plugin that does the code splitting relies on explicitly importing `dynamic` from `next/dynamic` and then having an `import` statement inside the `dynamic` call. I wanted to do something like this:

```ts
import dynamic from '@makeswift/runtime/next/dynamic'

ReactRuntime.registerComponent(dynamic(() => import('./Comp'))
```

But even that doesn't work because the import _has_ to be from `next/dynamic`. And because the ref isn't forwarded, we also need the `forwardNextDynamicRef`. We might be able to solve this if we fork the `next/dynamic` webpack plugin and make it work with `registerComponent` but we're going to be deprecating this API anyway in favor of static component modules. All this stuff will just become compilation output once we do that.

packages/runtime/src/next.tsx
migueloller: Yeah. Basically, `next/dynamic` uses `useImperativeHandle` and exposes a ref that looks something like this:

```ts
useImperativeHandle(ref, () => ({
  refetch() {
    // refetch the JS chunk
  }
}), [])
```

But we need that ref to give us the DOM node (or have no ref at all so we can automatically find it). So instead we kind of drill it _through_ the dynamic component by using this special key (`__forwarded_next_dynamic_ref__`) and then doing `ref={props['__forwarded_next_dynamic_ref__']}` in the `patch` function.

PR #111, opened 2022-07-19
fix: text selection is preserved after blur

fikrikarim:
Also fix ENG-251

### DIFF
--- a/.changeset/mean-cups-call.md
+++ b/.changeset/mean-cups-call.md
@@ -1,0 +1,5 @@
+---
+"@makeswift/runtime": patch
+---
+
+Fix text selection is preserved even after we change the focus to other text.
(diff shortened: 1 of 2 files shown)


migueloller (commented):
Using `relatedTarget` is a very interesting solution. What is the blur event happens from hitting <kbd>Esc</kbd> or from clicking a non-focusable element like a `<div>`?

It seems that this would preserve the selection as _unless_ some other focusable element is focused. Is that what we want?

fikrikarim:
This code is mostly for handling the blur event when a user is clicking from a Text to another Text (from content mode to content mode), and when a user is in content mode, and then clicking the panel (the builder is still in content mode).

When a user is going from content mode to builder mode, like when user hitting `Esc` or other part of the builder that triggers builder mode, we blur the `RichText` on the `PropController` `recv`:

image

migueloller (approved):

PR #119, opened 2022-07-25
Add TextStyle option to Style control

fikrikarim:


packages/runtime/src/runtimes/react/controls/style.ts
```diff
@@ -63,6 +65,14 @@ function useStyleControlCssObject(
           borderBottomRightRadius: borderRadiusToString(borderRadius?.borderBottomRightRadius) ?? 0,
           borderBottomLeftRadius: borderRadiusToString(borderRadius?.borderBottomLeftRadius) ?? 0,
         }),
+        ...(properties.includes(StyleControlProperty.TextStyle) && {
+          fontFamily: textStyle?.fontFamily ?? '',
```
migueloller, line 69: Wouldn't it be better to default to `null` here so that the style is simply not added and it falls back to the default CSS?

migueloller (changes requested):
Looking good!

packages/runtime/src/runtimes/react/controls/style.ts
fikrikarim: I need to find official specification for this, but when I'm playing with browser, checking on any unstyled element will return empty string. For example:
image

So it might be that the default value for style with type string is an empty string.

This is what the dev tools show on a component without any TextStyle value:
image

migueloller: Yeah, but generally with CSS in JS a null means you’re not setting it, that way CSS can just inherit. We don’t want to force the values to be empty and instead want it to properly inherit in case there’s a snippet or something else applying a font, no?

fikrikarim: Okay cool updated!

packages/runtime/src/runtimes/react/controls/style.ts
```diff
@@ -63,6 +65,14 @@ function useStyleControlCssObject(
           borderBottomRightRadius: borderRadiusToString(borderRadius?.borderBottomRightRadius) ?? 0,
           borderBottomLeftRadius: borderRadiusToString(borderRadius?.borderBottomLeftRadius) ?? 0,
         }),
+        ...(properties.includes(StyleControlProperty.TextStyle) && {
+          ...(textStyle?.fontFamily && { fontFamily: textStyle.fontFamily }),
+          ...(textStyle?.letterSpacing && { letterSpacing: textStyle.letterSpacing }),
+          ...(textStyle?.fontSize && { fontSize: fontSizeToString(textStyle.fontSize) ?? '16px' }),
+          ...(textStyle?.fontWeight && { fontWeight: textStyle.fontWeight }),
+          textTransform: textStyle?.textTransform ?? [],
```
migueloller, line 73: Don't CSS style values always have to be a string, `null`, or sometimes a number?

packages/runtime/src/runtimes/react/controls/style.ts
```diff
@@ -63,6 +65,14 @@ function useStyleControlCssObject(
           borderBottomRightRadius: borderRadiusToString(borderRadius?.borderBottomRightRadius) ?? 0,
           borderBottomLeftRadius: borderRadiusToString(borderRadius?.borderBottomLeftRadius) ?? 0,
         }),
+        ...(properties.includes(StyleControlProperty.TextStyle) && {
+          ...(textStyle?.fontFamily && { fontFamily: textStyle.fontFamily }),
+          ...(textStyle?.letterSpacing && { letterSpacing: textStyle.letterSpacing }),
+          ...(textStyle?.fontSize && { fontSize: fontSizeToString(textStyle.fontSize) ?? '16px' }),
```
migueloller, line 71: Since we're already checking if `fontSize` is falsy here, why not just make `fontSizeToString` always return a string? The argument can be `NonNullable<FontSizePropertyData>`. This way there's not an arbitrary 16px value that's never used.

migueloller (changes requested):
Had a couple more comments. Also, a question: did you get a chance to test this locally in the custom host sandbox?

packages/runtime/src/runtimes/react/controls/style.ts
fikrikarim: `textTransform` and `fontStyle` could be an array:
image

packages/runtime/src/runtimes/react/controls/style.ts
fikrikarim: Yeah true I updated it.

fikrikarim:
> Had a couple more comments. Also, a question: did you get a chance to test this locally in the custom host sandbox?

Yep. Could show you on Around if you want to,

migueloller:
> > Had a couple more comments. Also, a question: did you get a chance to test this locally in the custom host sandbox?
> 
> Yep. Could show you on Around if you want to.

No need! I just wanted to make sure that we had tested it 😄 

Also, have you updated the docs? We should do that before we release the new version

packages/runtime/src/runtimes/react/controls/style.ts
migueloller: I didn't know that! That's cool

migueloller (approved):

fikrikarim:
> > > Had a couple more comments. Also, a question: did you get a chance to test this locally in the custom host sandbox?
> > 
> > 
> > Yep. Could show you on Around if you want to.
> 
> No need! I just wanted to make sure that we had tested it 😄
> 
> Also, have you updated the docs? We should do that before we release the new version

Not yet. I have a subtask for updating the docs. I need to merge the MR on cosmos, update the docs, then release a new version. Thanks for the review!

PR #120, opened 2022-07-27
feat: add Link Control

fikrikarim:
https://user-images.githubusercontent.com/13066728/181229560-28ec4b79-1449-40ca-9a74-37b92b7dfd5f.mp4


packages/runtime/src/controls/link.ts
```diff
@@ -0,0 +1,44 @@
...
+  | OpenURLLink
+  | SendEmailLink
+  | CallPhoneLink
+  | ScrollToElementLink
+
+export const LinkControlType = 'makeswift::controls::link'
+
+type LinkControlConfig = {
+  label?: string
+  defaultValue?: LinkControlData
```
migueloller, line 34: I'm not sure we want to expose the default value just yet since it's an internal data structure that we probably don't want people typing themselves. For example, for the `Color` control, I exposed `defaultValue` because I was able to make it any hex or CSS color without requiring the user to provide a responsive swatch ID. I think the default value here should probably be more like the "mapped" value we return or also just support a simple URL (i.e., `/about`).

I suggest we remove this option for now and we can try and figure out a better API for the default value later.

packages/runtime/src/runtimes/react/controls/link.ts
```diff
@@ -0,0 +1,102 @@
...
+          behavior: 'smooth',
+          block,
+        })
+
+        if (view.location.hash !== hash) view.history.pushState({}, '', hash)
+      }
+    }
+  }
+
+  return { href, target, onClick: handleClick }
```
migueloller, line 101: What do you think about memoizing this so that if `href` and `target` don't change, the click handler and the returned object also don't change?

migueloller (changes requested):

packages/runtime/src/controls/link.ts
fikrikarim: Yeah I agree. Updated.

packages/runtime/src/runtimes/react/controls/link.ts
fikrikarim: Good idea 👍

PR #121, opened 2022-07-28
feat: improve env var behavior

migueloller:


fikrikarim (dismissed):
Finally!

This should close https://linear.app/makeswift/issue/ENG-175/add-helpful-error-message-if-environment-variables-missing-in-custom.

Also we need to update docs to remove the `MAKESWIFT_API_HOST`

migueloller:
@fikrikarim, made some more change, can I get another review?

PR #126, opened 2022-08-02
Remove Styled Components

migueloller:
This pull request is an experiment to see how much we can improve performance, specifically during hydration, by using Emotion for styles instead of Styled Components.


migueloller:
I marked this PR as draft because the `useDeviceMode` hook always returns `'desktop'`.

migueloller:
> I marked this PR as draft because the `useDeviceMode` hook always returns `'desktop'`.

It seems that we only use it for dynamic control options, though, so it might no be a huge issue...

PR #129, opened 2022-08-04
example: add Bigcommerce

joshuawootonn:


examples/bigcommerce/README.md
```diff
@@ -0,0 +1,34 @@
+This is a [Next.js](https://nextjs.org/) project bootstrapped with [`create-next-app`](https://github.com/vercel/next.js/tree/canary/packages/create-next-app).
```
joshuawootonn, line 1: This file will be back filled with some marketing content from @apledger. No need for review

examples/bigcommerce/lib/productComponents.tsx
```diff
@@ -0,0 +1,56 @@
+import Image from 'next/image'
```
joshuawootonn, line 1: Exact styles here will change so feel free to not review them

migueloller:
@joshuawootonn, looks like the deployment is failing for the BigCommerce example

examples/bigcommerce/.env
```diff
@@ -0,0 +1,7 @@
+BIGCOMMERCE_STORE_NAME=
+BIGCOMMERCE_STORE_HASH=
+BIGCOMMERCE_CLIENT_SECRET=
+BIGCOMMERCE_ACCESS_TOKEN=
+BIGCOMMERCE_CLIENT_ID=
+
+MAKESWIFT_SITE_API_KEY=
```
migueloller, line 7: Missing newline at the end of the file. Also, `MAKESWIFT_SITE_API_KEY` and other env vars that should be secrets should probably be kept our of the `.env` file and instead be configured on the Vercel deployment.

examples/bigcommerce/lib/bigcommerce/bigcommerce.ts
```diff
@@ -0,0 +1,195 @@
+const STORE_NAME = process.env.BIGCOMMERCE_STORE_NAME
+const ACCESS_TOKEN = process.env.BIGCOMMERCE_ACCESS_TOKEN
+
+const getBearerToken = async () => {
+  const A_WEEK_FROM_NOW = Math.floor(
+    new Date(new Date().getTime() + 1000 * 60 * 60 * 24 * 7).getTime() / 1000,
+  )
+
+  const response = await fetch(
+    `https://api.bigcommerce.com/stores/${process.env.BIGCOMMERCE_STORE_HASH}/v3/storefront/api-token`,
```
migueloller, line 10: We probably want to be consistent in how we handle environment variables and configuration. I see that some are defined in module scope while others are defined in function scope. Also some of them are checked and others are not.

I would suggest making a `lib/config.ts` file that handles all configuration values and exports them in a clean interface that can then be used by other modules that need that configuration

examples/bigcommerce/lib/bigcommerce/bigcommerce.ts
```diff
@@ -0,0 +1,195 @@
...
+    {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+        'X-Auth-Token': ACCESS_TOKEN ?? '',
```
migueloller, line 15: Should this function fail early if the `BIGCOMMERCE_ACCESS_TOKEN` env var isn't defined?

examples/bigcommerce/lib/bigcommerce/bigcommerce.ts
```diff
@@ -0,0 +1,195 @@
...
+      },
+      body: JSON.stringify({
+        channel_id: 1,
+        expires_at: A_WEEK_FROM_NOW,
+        allowed_cors_origins: ['https://makeswift-example.mybigcommerce.com'],
+      }),
+    },
+  ).then(a => a.json())
```
migueloller, line 23: Since this in an `async` function I would suggest we use the `await` keyword instead of `.then`

examples/bigcommerce/lib/bigcommerce/bigcommerce.ts
```diff
@@ -0,0 +1,195 @@
+const STORE_NAME = process.env.BIGCOMMERCE_STORE_NAME
+const ACCESS_TOKEN = process.env.BIGCOMMERCE_ACCESS_TOKEN
+
+const getBearerToken = async () => {
```
migueloller, line 4: Since we're using function declarations in other places I suggest we make this a function declaration as well. What do you think?

examples/bigcommerce/lib/bigcommerce/bigcommerce.ts
```diff
@@ -0,0 +1,195 @@
...
+    throw new Error('BIGCOMMERCE_STORE_NAME env variable is needed')
+  }
+
+  if (ACCESS_TOKEN == null) {
+    throw new Error('BIGCOMMERCE_ACCESS_TOKEN env variable is needed')
+  }
+
+  const bearerTokenResponse = await getBearerToken()
+
+  if (bearerTokenResponse.errors) {
```
migueloller, line 38: Shouldn't errors be handled by the `getBearerToken` function? It seems the function is leaking internal implementation details if it's returning errors. Specially since they're just being thrown by the caller. I would suggest checking for `response.ok` in `getBearerToken` and throwing an error there if the value is `false`.

migueloller: We seem to do that for this `getProducts` function (a bit later in the function) so it would also help with consistency

examples/bigcommerce/lib/bigcommerce/bigcommerce.ts
```diff
@@ -0,0 +1,195 @@
...
+
+  if (response.errors) {
+    console.error(response.errors)
+    throw new Error('Error while fetching products')
+  }
+
+  return response.data.site.products.edges
+}
+
+export const DEFAULT_PRODUCT_ID = 114
```
migueloller, line 114: Maybe we can put this in a centralized configuration module?

examples/bigcommerce/lib/bigcommerce/utils.ts
```diff
@@ -0,0 +1,11 @@
+import { getProduct, getProducts } from './bigcommerce'
+
+export const getProductSlugPaths = async (): Promise<string[]> => {
```
migueloller, line 3: Let's use function declarations in this file too

examples/bigcommerce/lib/bigcommerce/utils.ts
```diff
@@ -0,0 +1,11 @@
...
+  const products = await getProducts()
+
+  return products?.map((edge: any) => edge.node.entityId)?.map((entityId: any) => entityId) ?? []
```
migueloller, line 6: Are we planning on leaving these types as `any`? We could manually define some types or use some code generation to leverage TypeScript and GraphQL

examples/bigcommerce/lib/bigcommerce/utils.ts
```diff
@@ -0,0 +1,11 @@
...
+}
+
+export const getProductBySlug = async (slug: string): Promise<number> => {
+  return await getProduct(parseInt(slug))
```
migueloller, line 10: Doesn't the `getProduct` function return a `Promise<Product>` and not a `Promise<number>`?

examples/bigcommerce/lib/header.tsx
```diff
@@ -0,0 +1,39 @@
+import Link from 'next/link'
+import ProductCart from './productCart'
+
+type Props = {
+  className?: string
+}
+
+const Header = ({ className }: Props) => {
```
migueloller, line 8: Can we use function declarations for React components? This could make it clearer that this is the default export since we can declare the component and export it all in one line. Also, I think we should leave arrow functions when using anonymous "lambdas", specially when we're not leveraging the arrow function's implicit return like we're doing here.

examples/bigcommerce/lib/makeswift/context.tsx
```diff
@@ -0,0 +1,149 @@
...
+  url: string
+  id: number
+  name: string
+  description: string
+  price: string
+  defaultImage: Image
+  images: Image[]
+  categories: Category[]
+}
+export type Category = 'Shop All' | 'Plants' | 'Pots'
```
migueloller, line 17: When using enum types like these I would suggest we actually define a value so that we're not hand-writing string values elsewhere. i.e.,

```ts
const ProductCategory = {
  ShopAll: 'shop-all',
  Plants: 'plants',
  Pots: 'pots',
} as const

type ProductCategory = typeof ProductCategory[keyof typeof ProductCategory]

// Use it
const category: Category = ProductCategory.Plants
```

examples/bigcommerce/lib/makeswift/context.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+export const ProductsContext = createContext<any>(null)
```
migueloller, line 19: Should this type be an array type and the default value an empty array?

examples/bigcommerce/lib/makeswift/context.tsx
```diff
@@ -0,0 +1,149 @@
...
+  }
+}
+export const ProductContext = createContext<any>(null)
+
+export const useProduct = (): {
+  product: Product
+} => {
+  const product = useContext(ProductContext)
+
+  return {
```
migueloller, line 69: We seem to use the exact same product mapping in the `useProducts` hook above. What do you think about creating a function that maps a `BigCommerceProduct` to this internal representation?

examples/bigcommerce/pages/_document.tsx
```diff
@@ -0,0 +1,23 @@
+import { Document } from '@makeswift/runtime/next'
+import { Html, Head, Main, NextScript } from 'next/document'
+
+class MyDocument extends Document {
+  render() {
+    return (
+      <Html>
+        <Head>
+          <link
```
migueloller, line 9: If we're using a Google Font, why not do that in the Makeswift builder?

examples/bigcommerce/pages/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+  ctx: GetStaticPropsContext,
+): Promise<GetStaticPropsResult<Props>> {
+  const makeswiftResult = await makeswiftGetStaticProps({
+    ...ctx,
+    params: { ...ctx.params, path: [''] },
+  })
+  const products = await getProducts()
+  if (!products) return { notFound: true }
+
+  // @ts-ignore
```
migueloller, line 27: Let's fix the `@ts-ignore`s in this file

examples/bigcommerce/pages/product/[slug].tsx
```diff
@@ -0,0 +1,50 @@
...
+  // @ts-ignore
+  return {
+    ...makeswiftResult,
+    // @ts-ignore
+    props: { ...makeswiftResult.props, product },
+  }
+}
+
+export default function Page({ product, ...restOfProps }: Props) {
+  if (!product) return
```
Comment (migueloller) [inline examples/bigcommerce/pages/product/[slug].tsx:43]: It seems that we're showing a 404 page if there is no product, so why return here? We really never want to render a blank page like this, no?

examples/bigcommerce/postcss.config.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+    plugins: {
+      tailwindcss: {},
+      autoprefixer: {},
+    }
+  }
```
migueloller, line 6: There's no newline at the end of this file and it seems the indentation is out of whack

migueloller (changes requested):
@joshuawootonn, this is looking good. In terms of functionality I don't have any comments. All of my comments are around coding style, conventions, and most importantly consistency. I think we want our examples to be an example (pun intended) of the level of quality and care we take to build our product. This is public facing content that we will put our names in so it's worth it to go the extra mile to make sure that all the nits are picked 😬 

I'm happy to go over these changes together, or even help with making them. Just let me know!

Here's some more comments that I didn't think fit with any particular line of code:
- Can we leave `lib` just for the server-side utilities and put components in a `components` folder next to `lib` and `pages`?
- A lot of functions are quite cramped without any whitespace between statements. Can we give those statements some breathing room? 😆 
- For component file names, can we use kebab-case? (e.g., `productCard.tsx` would be `product-cart.tsx`). For React components, the usual convention is to name components in PascalCase and have the files match the component name or kebab-case of the component name with a named export.
- Let's separate the bag of components in `productComponents.tsx` to individual component files.
- We shouldn't be using `any` in an example written in TypeScript. Let's add some types!

examples/bigcommerce/.env
joshuawootonn: Fair enough. My intention was for this to become their `.env.local` in the case that they cloned the repo. Should i rename it to `.env.example`

migueloller: Yes. Putting this like this in a `.env` will cause all of these env vars to be empty strings which might cause unintended side effects. Also documenting the env vars in the README and having good error messages when running the app works as well. Having a centralized config module can help with that.

examples/bigcommerce/README.md
migueloller: Got it. Do we have an ETA on that?

migueloller:
@joshuawootonn, still got a couple of files without newlines at the end of the file in them

joshuawootonn:
> @joshuawootonn, still got a couple of files without newlines at the end of the file in them

ok updated

examples/bigcommerce/README.md
joshuawootonn: I don't. I will come up with something tomorrow morning based on what Alan made for thirdweb and update the readme/merge.

examples/bigcommerce/.vscode/settings.json
```diff
@@ -0,0 +1,6 @@
+{
```
migueloller, line 1: I think we should drop this file since this only applies to specific extensions (e.g., cSpell) that users of the example probably won't have. If you want to have these options for yourself, you can set them in the user settings instead.

examples/bigcommerce/pages/[[...path]].tsx
```diff
@@ -0,0 +1,28 @@
...
+  if (!('props' in makeswiftResult)) return makeswiftResult
+
+  const products = await getProducts()
+  const product = await getProduct()
+
+  return { ...makeswiftResult, props: { ...makeswiftResult.props, products, product } }
+}
+
+export { getStaticPaths, Page as default } from '@makeswift/runtime/next'
+
```
Comment (migueloller) [inline examples/bigcommerce/pages/[[...path]].tsx:28]: Extra newline here

examples/bigcommerce/lib/bigcommerce/defaults.ts
```diff
@@ -0,0 +1,48 @@
+import { ProductFragment } from "./types";
```
joshuawootonn, line 1: @migueloller this was my best guess at extending the pattern established in the other files

migueloller (approved):
Looks great!

examples/bigcommerce/README.md
```diff
@@ -0,0 +1,198 @@
+# Demo
+
+[https://www.makeswift-examples-bigcommerce.vercel.app](https://www.makeswift-examples-bigcommerce.vercel.app)
```
joshuawootonn, line 3: CleanShot 2022-08-12 at 11 04 08@2x
This link will work when this is deployed. I can't change the domain so that it works till then

joshuawootonn:
@migueloller I pushed some changes that invalidated your approval. Can you reapprove?

migueloller (approved):

PR #133, opened 2022-08-16
example: add Shopify

joshuawootonn:
Potential questions you might have: 
* I didn't use https://shopify.github.io/js-buy-sdk/ because it was pretty restrictive 
* I didn't use Hydrogen because it's its own framework. There are work around for using some of it with next.js but I didn't want to use workarounds in an example


joshuawootonn:
I can't trigger a deploy(due to permissions) in Vercel to get the preview branch. So the link in the readme won't work, but it will shortly

examples/bigcommerce/.env.local.example
```diff
@@ -2,6 +2,5 @@ BIGCOMMERCE_STORE_NAME=
 BIGCOMMERCE_STORE_HASH=
 BIGCOMMERCE_CLIENT_SECRET=
 BIGCOMMERCE_ACCESS_TOKEN=
-BIGCOMMERCE_CLIENT_ID=
```
joshuawootonn, line 5: I never used this in the example so that is why it is removed

examples/bigcommerce/components/header.tsx
```diff
@@ -11,26 +11,36 @@ export function Header({ className }: Props) {
...
-              fill="#335510"
-            />
-          </svg>
+          <a>
```
joshuawootonn, line 14: just added these links because it was bothering me that the header didn't have cursor / wasn't focusabled

examples/bigcommerce/components/product-cart.tsx
```diff
@@ -9,7 +9,7 @@ export function ProductCart({ className }: ProductCartProps) {
   const { items } = useCart()
 
   return (
-    <div className={`${className} relative z-0 `}>
+    <button className={`${className} relative z-0 `}>
```
joshuawootonn, line 12: same reason as https://github.com/makeswift/makeswift/pull/133/files#r947059045

examples/bigcommerce/package.json
```diff
@@ -19,6 +19,7 @@
     "@types/react": "18.0.15",
     "@types/react-dom": "18.0.6",
     "autoprefixer": "^10.4.8",
+    "eslint": "^8.22.0",
```
joshuawootonn, line 22: eslint is included in next.js but using the config packages below warns that eslint isn't installed.

Just added eslint to remove the warning

migueloller: I'm surprised that's the case. Does this mean that a fresh `create-next-app` installation will have those warnings?

examples/shopify/.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "typescript.tsdk": "node_modules/typescript/lib"
+}
```
migueloller, line 3: I think this might be a bit too much for an example and that we should remove it. The GraphQL extension recommendation is okay because it's just a recommendation and it will enhance the example by making the GraphQL queries with comments get syntax highlighting

examples/shopify/README.md
```diff
@@ -0,0 +1,201 @@
...
+   and copy the Site API key. You will need it in step 3.
+
+   Then head over to [Shopify](https://www.shopify.com/) and create a store. Here is [a good starting point](https://www.shopify.com/online) to get you started.
+
+2. **Clone this template**
+
+   Instead of using `create-next-app`, run this command from the terminal:
+
+   ```bash
+   npx degit makeswift/makeswift/examples/shopify  shopify
```
migueloller, line 32: There's an extra space between `makeswift/makeswift/examples/shopify` and `shopify`

examples/shopify/README.md
```diff
@@ -0,0 +1,201 @@
...
+   cd shopify
+   ```
+
+   It will download this subdirectory of the `makeswift/makeswift` repo without including git history
```
migueloller, line 36: Missing a period at the end of the sentence

examples/shopify/README.md
```diff
@@ -0,0 +1,201 @@
...
+   + SHOPIFY_ACCESS_TOKEN=<YOUR_SHOPIFY_ACCESS_TOKEN>
+
+   + MAKESWIFT_SITE_API_KEY=<YOUR_MAKESWIFT_SITE_API_KEY>
+   ```
+
+4. **Start the dev server**
+
+   Run this command from the terminal:
+
+   ```bash
```
migueloller, line 64: "this command" isn't copy/paste-able or runnable in the terminal. We should probably just use a single type of package manager here. Which one should be based on what package manager the example is using. All our examples seem to have a Yarn lockfile so we should use Yarn

examples/shopify/README.md
```diff
@@ -0,0 +1,201 @@
...
+   yarn dev
+   # or
+   npm run dev
+   ```
+
+   Your next.js app should be up and running at http://localhost:3000.
```
migueloller, line 70: "next.js" -> "Next.js"

examples/shopify/README.md
```diff
@@ -0,0 +1,201 @@
...
+
+5. **Create a home page with a list of products**
+
+   1. Create a blank page
+
+      Look for the plus button in the left toolbar and specify "Blank page."
```
migueloller, line 76: For the purposes of our examples we probably want to put the period outside of the quotes even though that's the usual style. Specially because the button in the UI says "Blank page" and not "Blank page.". Also, below for "Edit URL" the period is outside the quotes so regardless of which style we pick, we should be consistent

examples/shopify/README.md
```diff
@@ -0,0 +1,201 @@
...
+
+   2. Edit page's URL to be `/__product__`
+
+      Hover over the "Untitled page" you just created, click on the ellipsis that appears, and click "Edit URL".
+
+      You should set this value to the `productTemplatePathname` from `/lib/config.ts` which comes predefined in this template as `/__product__`
+
+   3. Drop in Product specific components
+
+      Look for the ellipsis menu in the left toolbar again and drop the Product price, Product name, and Add to cart button into Makeswift.
```
migueloller, line 102: Are we using sentence case or uppercase for component labels? Above the "Product List" component is mentioned in uppercase, but here the "Product price", "Product name", and "Add to cart button" are in sentence case

examples/shopify/README.md
```diff
@@ -0,0 +1,201 @@
...
+
+---
+
+## Putting it all together
+
+With a home page and product template pages created it's probably a good time to explain what's going on.
+
+### How is Shopify product data getting to components?
+
+The `/pages/[[...path]].tsx` route uses Next.js' `getStaticProps` to get page data from Makeswift
```
migueloller, line 112: Missing period at the end of the sentence

migueloller: There's other sentences throughout the readme like this. I'll just point it out here this one time

examples/shopify/tailwind.config.js
```diff
@@ -0,0 +1,21 @@
+const defaultTheme = require("tailwindcss/defaultTheme");
+
+/** @type {import('tailwindcss').Config} */
```
migueloller, line 3: Does this type specifier work even though there's an empty line between the object definition and the comment? Also, does this work when doing just an assignment? I would've though it only worked when declaring new variables

migueloller (changes requested):
@joshuawootonn, looks good! Just have minor nit comments for the README and a couple of questions

examples/shopify/README.md
joshuawootonn: Going with sentence case

examples/shopify/tailwind.config.js
joshuawootonn: CleanShot 2022-08-16 at 14 27 28@2x
It does work, but I'll delete the line

joshuawootonn:
https://github.com/makeswift/makeswift/pull/133#discussion_r947114766

Actually, I was wrong. create-next-app installs eslint as a dev dependency. I thought it was a peer dependency and so the warning was actually legit. Either way no action is needed.

joshuawootonn:
This is ready for review again @migueloller

migueloller (approved):
LGTM! 👍🏻

PR #138, opened 2022-08-23
feat: add cart function to bigcommerce example

joshuawootonn:
This includes adding cart function and updating the readme accordingly.

### DIFF
--- a/examples/bigcommerce/components/product-cart.tsx
+++ b/examples/bigcommerce/components/product-cart.tsx
@@ -1,68 +1,243 @@
+import { Popover, Transition } from '@headlessui/react'
+import Image from 'next/image'
+import { useRouter } from 'next/router'
+import { useState } from 'react'
+
 import { useCart } from 'lib/cart-context'
 import { useProduct } from 'lib/product-context'
+import { Cart34, Minus28, Minus36, Plus28, Plus36, Spinner22, Spinner28 } from './icons'
+
+function formatPrice(value?: number) {
+  return value == null ? '$0.00' : `$${value.toFixed(2)}`
+}
+
+type CartState = 'initial' | 'loading' | 'redirecting' | 'error'
 
 type ProductCartProps = {
   className?: string
 }
 
 export function ProductCart({ className }: ProductCartProps) {
-  const { items } = useCart()
+  const { cart, updateItem, deleteItem, getCheckoutUrl } = useCart()
+  const router = useRouter()
+  const [cartState, setCartState] = useState<CartState>('initial')
 
+  const itemCount =
+    cart?.line_items.physical_items.reduce((acc, curr) => curr.quantity + acc, 0) ?? 0
   return (
-    <button className={`${className} relative z-0 `}>
-      <svg
-        width="34"
-        height="34"
-        viewBox="0 0 34 34"
-        fill="none"
-        xmlns="http://www.w3.org/2000/svg"
+    <Popover className={className}>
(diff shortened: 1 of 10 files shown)


examples/bigcommerce/pages/api/categories.ts
```diff
@@ -6,7 +6,6 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
     const categories = await getCategories()
     res.status(200).json(categories)
   } catch (error) {
-    console.error('Failed to fetch categories', error)
-    res.status(400).json({ message: 'Failed to fetch categories' })
+    res.status(400).json({ message: 'Failed to fetch categories.' })
```
joshuawootonn, line 9: Changed this to match the catch case of the `cart` and `checkout` routes

examples/bigcommerce/pages/api/categories.ts
```diff
@@ -2,11 +2,6 @@ import { getCategories } from 'lib/bigcommerce'
...
 
 export default async function handler(req: NextApiRequest, res: NextApiResponse) {
-  try {
-    const categories = await getCategories()
-    res.status(200).json(categories)
-  } catch (error) {
-    console.error('Failed to fetch categories', error)
-    res.status(400).json({ message: 'Failed to fetch categories' })
-  }
+  const categories = await getCategories()
```
joshuawootonn, line 5: This is already logging errors, and if it throws this should be a 500, so I moved this to be not include the try catch

examples/bigcommerce/components/product-cart.tsx
```diff
@@ -1,68 +1,243 @@
+import { Popover, Transition } from '@headlessui/react'
+import Image from 'next/image'
+import { useRouter } from 'next/router'
+import { useState } from 'react'
+
 import { useCart } from 'lib/cart-context'
 import { useProduct } from 'lib/product-context'
+import { Cart34, Minus28, Minus36, Plus28, Plus36, Spinner22, Spinner28 } from './icons'
+
+function formatPrice(value?: number) {
```
migueloller, line 10: Have you heard about the [`Intl.NumberFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat) API? I've never used it but it looks pretty cool!

examples/bigcommerce/pages/api/cart.ts
```diff
@@ -0,0 +1,146 @@
...
+    if (!response.ok) throw new Error(response.statusText)
+
+    if (response.status === 204) return res.status(204).json({})
+
+    const result: RestResponse<CartResponse | null> = await response.json()
+
+    return res.status(200).json(result)
+  }
+
+  return res.status(405)
```
migueloller, line 145: I think that if you don't send the response the request will just hang until the client times out. You might want to add `.send('Method Not Allowed')` or something to make sure the request doesn't hang

migueloller (dismissed):
Looking great!

examples/bigcommerce/pages/api/cart.ts
joshuawootonn: You are correct. Thanks for catching that.

examples/bigcommerce/components/product-cart.tsx
joshuawootonn: Going to pass on this add for now

migueloller (approved):

PR #139, opened 2022-08-24
fix: Twitter Cards meta tags

migueloller:


.changeset/nervous-suns-love.md
```diff
@@ -0,0 +1,5 @@
+---
```
joshuawootonn, line 1: I am not sure how this works / where it is consumed so I can't really speak to it

packages/runtime/src/components/page/Page.tsx
```diff
@@ -282,9 +282,9 @@ export function Page({ page, preview = false }: Props): JSX.Element {
         {socialImage && (
```
joshuawootonn, line 282: LGTM! 

Curious how you realized that this was a problem. Were you just sharing on twitter?

joshuawootonn (approved):

.changeset/nervous-suns-love.md
joshuawootonn: Probably not a blocker though.

migueloller: This is consumed by our GitHub Actions workflow when we release a new version of the runtime

packages/runtime/src/components/page/Page.tsx
migueloller: Andrew created an issue for it last week: https://linear.app/makeswift/issue/PRD-402/preview-card-isnt-displaying-in-twitter-when-sharing-a-makeswift-site

PR #143, opened 2022-08-26
fix: use process.cwd() instead of dirname

alexhwoods:


packages/makeswift/src/init.ts
```diff
@@ -14,7 +14,7 @@ async function init(
   name: string,
   { example = "basic-typescript" }: { example?: string }
 ): Promise<void> {
-  const nextAppDir = path.join(__dirname, name);
+  const nextAppDir = path.join(process.cwd(), name);
```
migueloller, line 17: Hmm, I wonder if this will work in all cases. I'm thinking that the best way we can ensure that Next.js behaves as it does right now is if we just pass the name as provided by the user, call it `projectName` (as Next.js calls it) instead of `dir` and then when we spawn the process it should just work assuming that the `cwd` of the spawned the process defaults to `process.cwd`.

I wonder if this would also result in Next.js showing cleaner output when it prints the project name since it wouldn't be an absolute path

migueloller: Did you find a way to reproduce the error happening with the public release locally? If so, I guess we could use that to verify whether the solutions work

alexhwoods: If I do `node dist/src/index.js init foo`, this works — and that kind of covers this issue as well, right?

alexhwoods:
```
```

alexhwoods: Let me experiment with not passing the directory

alexhwoods: It goes back to the original problem if we don't pass the directory

```ts
  spawn.sync("npx", ["--yes", "create-next-app", "--example", url, "bar"], {
    stdio: "inherit",
    cwd: __dirname,
  });
```

Resulted in 
```
Creating a new Next.js app in /Users/alexwoods/makeswift/makeswift/packages/makeswift/dist/src/bar.

```

migueloller: The issues that I was running into were only happening when running `npx makeswift init` with the released version of the CLI. These issue weren't happening when we were running it locally. So without reproducing the issue locally, first, we don't know if the changes you're making here will actually fix it

alexhwoods: Yeah that works as well. Same output as above

alexhwoods: I'm going to keep it with calculating the directory though, because we use that in other parts of the code as well

migueloller: > Yeah that works as well. Same output as above

I'm not sure what you're responding to here. Was it my comment about not using the directory or was it the one about reproducing the error locally?

alexhwoods: not using the directory does work, but I don't want to go with that solution because we use the directory variable elsewhere, so I think it's more consistent to use the directory variable

alexhwoods: And I have tested using `npx`, and that works as well

migueloller: > not using the directory does work, but I don't want to go with that solution because we use the directory variable elsewhere, so I think it's more consistent to use the directory variable

Sounds good 👍🏻

migueloller: > And I have tested using npx, and that works as well

There were no issues with that locally yesterday, but there seem to be issue _only_ when it's actually release to npm. So unless we reproduced the issue locally, there's really no way to know if that makes sense

alexhwoods: That's fair — that's why I've been testing with `node dist/src/index.js`. That command _does_ produce the issue, because the working directory and file directory are different

migueloller: Ok, cool! I just wanted to make sure we had a reproduction going. I think the issue before was probably that `npx` cached the last "working" run

alexhwoods: Yeah I think so to

migueloller (approved):

PR #146, opened 2022-08-29
feat: add shopify cart functionality

joshuawootonn:


examples/shopify/README.md
```diff
@@ -24,6 +24,10 @@ This example includes a home page for listing products by category and a product
 
    Then head over to [Shopify](https://www.shopify.com/) and create a store. Here is [a good starting point](https://www.shopify.com/online) to get you started.
 
+   Note: This custom storefront example assumes all products are in stock, so make sure all products have "Track quantity" unchecked as pictured below.
+
+      <img width="600" src="https://user-images.githubusercontent.com/20950876/187238713-22fd0c65-8d9b-4eab-b94e-1e498053f270.png">
+
```
joshuawootonn, line 30: other edits to this file are from prettier

examples/bigcommerce/components/cart.tsx
```diff
@@ -42,74 +121,23 @@ export function ProductCart({ className }: ProductCartProps) {
         </Transition>
       </Popover.Button>
       <Transition
-        className={'absolute z-10'}
+        className={'fixed left-5 right-5 sm:left-[unset] sm:right-[unset] sm:absolute z-10'}
```
joshuawootonn, line 124: mobile styles for the cart to look like:
CleanShot 2022-08-29 at 11 20 25@2x

examples/shopify/README.md
migueloller: For future PRs, it would be great if you could separate `prettier --write .` runs in their own commit. It allows the reviewer to avoid that noise and focus on relevant changes 😄

examples/bigcommerce/components/header.tsx
```diff
@@ -26,22 +26,13 @@ export function Header({ className }: Props) {
             </svg>
           </a>
         </Link>
-        <Link href={'/'}>
+        <Link href={'/#featured'}>
```
migueloller, line 29: Should this link come from the Makeswift builder directly? I'm assuming the `featured` HTML ID was assigned in the Makeswift builder.

Maybe the header can be added as a component and the links can be a `List` of `Link`s?

examples/shopify/README.md
```diff
@@ -24,6 +24,10 @@ This example includes a home page for listing products by category and a product
 
    Then head over to [Shopify](https://www.shopify.com/) and create a store. Here is [a good starting point](https://www.shopify.com/online) to get you started.
 
+   Note: This custom storefront example assumes all products are in stock, so make sure all products have "Track quantity" unchecked as pictured below.
```
migueloller, line 27: What happens if a product isn't in stock?

examples/shopify/lib/config.ts
```diff
@@ -1,30 +1,58 @@
...
 }
 
-function getEnvVarOrThrow(key: string): string {
-  const value = process.env[key]
+export type Config = PublicConfig & PrivateConfig
 
+/**
+ * public env values must be inlined for the value to be exposed to the browser
+ * https://nextjs.org/docs/basic-features/environment-variables#exposing-environment-variables-to-the-browser
+ * @param key -> name of env variable
```
migueloller, line 20: There's a specific JSDoc syntax for defining parameters: https://jsdoc.app/tags-param.html

Don't know if you were going for that, but if you were, you might want to avoid the `->`.

examples/shopify/lib/config.ts
```diff
@@ -1,30 +1,58 @@
...
+  return verifyPublicEnvValueOrThrow(key, value)
+}
+
+export function getPublicConfig(): PublicConfig {
   return {
     shopify: {
-      accessToken: getEnvVarOrThrow('SHOPIFY_ACCESS_TOKEN'),
-      storeName: getEnvVarOrThrow('SHOPIFY_STORE_NAME'),
+      accessToken: verifyPublicEnvValueOrThrow(
+        'NEXT_PUBLIC_SHOPIFY_ACCESS_TOKEN',
```
migueloller, line 40: The key here doesn't seem to be used except for writing the error message. This functionality is a very generic "make sure this value exists, otherwise throw". This is commonly known as an "invariant", popularize by Facebook OSS libraries. There's a few variants of it out there, this one being the most popular (I think): https://github.com/alexreardon/tiny-invariant

examples/shopify/lib/shopify/types.ts
```diff
@@ -4,6 +4,11 @@ export type GraphQLResponse<T> =
   | { data: T; errors: undefined }
   | { data: null; errors: GraphQLError[] }
 
+export type RestResponse<T> = {
```
migueloller, line 7: Is this used in this example?

migueloller (changes requested):

migueloller (approved):

examples/shopify/README.md
joshuawootonn: you can't add it to the cart so "add to cart" looks like it succeeds, but it doesn't

migueloller: Got it. Is there a way to get information about if it's in stock from the API response and then disable the "add to cart" button in that case? Maybe you considered this and it wasn't worth the effort but I'm just curious

examples/shopify/lib/shopify/types.ts
joshuawootonn: nope thanks for catching that

examples/shopify/lib/config.ts
joshuawootonn: > The key here doesn't seem to be used except for writing the error message

True. This is because Next.js wont include the env in the browser unless I write out `process.env.actualname` but I also want the variable name for error handling so I am passing in both.

https://nextjs.org/docs/basic-features/environment-variables#exposing-environment-variables-to-the-browser

joshuawootonn: Thank you for explaining that BTW. I had seen that before and been confused. Finally connected the dots by reading your comment and reading that package readme

examples/shopify/lib/config.ts
joshuawootonn: because I am now using tiny-invariant this comment is no longer needed / I removed it

examples/bigcommerce/components/header.tsx
joshuawootonn: Done

examples/bigcommerce/lib/makeswift/register-components.tsx
```diff
@@ -88,6 +88,16 @@ ReactRuntime.registerComponent(Header, {
   label: 'Header',
   props: {
     className: Style({ properties: Style.All }),
+    links: List({
+      type: Shape({
```
migueloller, line 92: There's a `Link` control: https://www.makeswift.com/docs/controls/link

migueloller (changes requested):

migueloller (approved):

examples/shopify/README.md
joshuawootonn: Yeah this is possible. I just don't think it's worth it. 

Shopify requires you to enter a SKU and Barcode for items, so I just thought I would avoid it all together.
CleanShot 2022-08-29 at 14 43 25@2x

PR #147, opened 2022-08-29
chore: update bigcommerce default product image

joshuawootonn:
My PR earlier included a low quality image for the default product, which causes the preview in Makeswift to look poor for no reason


migueloller (dismissed):

examples/bigcommerce/lib/bigcommerce/api.ts
```diff
@@ -27,7 +27,7 @@ export async function getApiToken(): Promise<string> {
       body: JSON.stringify({
         channel_id: 1,
         expires_at: A_WEEK_FROM_NOW,
-        allowed_cors_origins: [`https://${config.bigcommerce.storeName}.mybigcommerce.com`],
+        allowed_cors_origins: [config.deploymentURL],
```
migueloller, line 30: You might just want to make the configuration the allowed origins, i.e., `config.bigcommerce.allowedCorsOrigins`. That way instead of setting an empty string in dev, you can do an empty array

migueloller (dismissed):
Looks good!

examples/bigcommerce/lib/bigcommerce/api.ts
migueloller: Also, the allowed origins is really all configuration and we probably want to put any logic of figuring out that config value in the `getConfig` function itself, since that's the point of it, to abstract away getting a valid configuration. The `getApiToken` really shouldn't care about the contents of that array

migueloller (approved):

PR #150, opened 2022-08-30
v0.2.0

migueloller:
Closes: PRD-50, PRD-53, PRD-409, PRD-444


packages/runtime/src/next/client.ts
```diff
@@ -0,0 +1,135 @@
...
+    { preview = false }: { preview?: boolean } = {},
+  ): Promise<MakeswiftPageSnapshot | null> {
+    const searchParams = new URLSearchParams({ preview: String(preview) })
+    const response = await this.fetch(`/v1/pages/${pageId}/document?${searchParams}`)
+
+    if (!response.ok) {
+      if (response.status === 404) return null
+
+      throw new Error(`Failed to get snapshot with error: "${response.statusText}"`)
+    }
```
alexhwoods, line 116: Can this receive a 401?

Thinking of a misconfigured API key - that might return a 404 though

alexhwoods (approved):

packages/runtime/src/next/client.ts
migueloller: In that scenario it would throw an error with the message `Failed to get a snapshot with error: "Unauthorized"`, which is what I was going for. I only handle 404 so that the function returns `null` if the snapshot is not found

alexhwoods: Ok, cool. Just wanted to make sure you had considered that

migueloller:
Keeping this PR in a draft state because the following issues have to be fixed:
- Handle when `sessionStorage` API is disabled
- Handle when third-party cookies are disabled
- Handle infinite redirects due to Preview Mode not being set (likely due to cookies being disabled but can also happen due to misconfigured cookies like `SameSite=Lax`)
- We can't set `Secure` on a cookie if we're using localhost without HTTPS

Some thoughts on how to move forward:
- We can check `navigator.cookieEnabled` to determine whether we can use Preview Mode
- We can move from using `sessionStorage` to save the revalidation secret to using a session cookie
- When cookies aren't enabled the Preview Mode script should gracefully handle the situation and take the user somewhere else, perhaps the Makeswift docs?
- To avoid infinite redirects we can provide a flag to the `return_to` param and if we see we're trying to go to Preview Mode again and that param is in the URL, then we know we're going to infinitely redirect and we can fail gracefully
- The definition of a "Site" in cookie's `SameSite` is the combination of the public suffix (e.g., `com`,`co.uk`) with the second level domain. This means that `some-host.makeswift.com` is considered as the same site as `app.makeswift.com`!

migueloller:
- We should consider handling cookies not working and a timeout if Preview Mode isn't detected and giving the user helpful messages instead of proxying hosts to ensure first-party cookies. There's a lot of potential issues with proxying, including security issues.

packages/runtime/src/next/api-handler.ts
```diff
@@ -64,6 +64,24 @@ export function MakeswiftApiHandler(
...
+
+        try {
+          await res.revalidate(req.query.path)
+
+          return res.json({ revalidated: true })
+        } catch (error) {
+          return res.status(500).json({ message: 'Error Revalidating' })
+        }
+      }
+
```
alexhwoods, line 84: Woohoo!

alexhwoods (dismissed):

.changeset/tiny-maps-study.md
```diff
@@ -0,0 +1,171 @@
...
+  -export { getServerSideProps, Page as default } from '@makeswift/runtime/next'
+  ```
+
+- Go to your Makeswift site settings and update the host URL to be just your host's
+  origin. For example, change `https://www.makeswift.com/makeswift` to just
+  `https://www.makeswift.com` or `http://localhost:3000/makeswift` to just `http://localhost:3000`.
+
+If you have any questions about the migration or run into any issues, please don't hesitate to chat
+with us. [We're on Discord!](https://discord.gg/PkrUsFnMUn)
+
```
fikrikarim, line 77: Should we add a divider here or something more obvious to separate between two sections? At a first glance, it's kinda hard to read because I was assuming it was still part of the migration process.

I kinda see it like the first part is more of a TL;DR that anyone needs to read, and the second part is more of an optional detailed explanation.

.changeset/tiny-maps-study.md
```diff
@@ -0,0 +1,171 @@
...
+```js
+import { MakeswiftApiHandler } from '@makeswift/runtime/next'
+
+export default MakeswiftApiHandler(process.env.MAKESWIFT_SITE_API_KEY)
+```
+
+The API handler not only enables Next.js Preview Mode, allowing you to remove your API route, but it
+also adds support for automatic [on-demand revalidation](https://nextjs.org/docs/basic-features/data-fetching/incremental-static-regeneration#on-demand-revalidation)! Whenever you publish a page, Makeswift will
+automatically send a request to `/api/makeswift/revalidate` and take care of on-demand ISR. This
+means that you can leave off the `revalidate` option in `getStaticProps` and trust your pages will
```
fikrikarim, line 105: Maybe add the `revalidate` option to the migration process list?

fikrikarim (approved):
Looks great! Ship ship ship 🚢

.changeset/tiny-maps-study.md
```diff
@@ -0,0 +1,171 @@
...
+-export { getServerSideProps, Page as default } from '@makeswift/runtime/next'
+```
+
+```js
+import { MakeswiftApiHandler } from '@makeswift/runtime/next'
+
+export default MakeswiftApiHandler(process.env.MAKESWIFT_SITE_API_KEY)
+```
+
+The API handler not only enables Next.js Preview Mode, allowing you to remove your API route, but it
```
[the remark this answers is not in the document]
migueloller: Thanks!

.changeset/tiny-maps-study.md
```diff
@@ -0,0 +1,176 @@
...
+you're in the builder. Read more about the feature in the
+[RFC](https://github.com/makeswift/makeswift/discussions/142).
+
+To migrate from the old preview route API, delete your preview route:
+
+```diff
+-export { getServerSideProps, Page as default } from '@makeswift/runtime/next'
+```
+
+The create a new file at `pages/api/makeswift/[...makeswift].ts` with the following content:
```
fikrikarim, line 99: One last typo: The -> Then?

PR #151, opened 2022-08-31
fix: revert builtin components width control to use styled components

fikrikarim:


migueloller (approved):
Looks good 👍🏻

fikrikarim:
I tested every built-in component's width locally, and they look great.

migueloller:
> I tested every built-in component's width locally, and they look great.

There were some components—I think Navigation was one of them—that re-used some of the builtin components like Button and Image. When I moved these to a class I accidentally shipped some bugs because it was very hard to see where those props were being passed (some of them were just spread). Did you check for that too @fikrikarim?

fikrikarim:
> > I tested every built-in component's width locally, and they look great.
> 
> There were some components—I think Navigation was one of them—that re-used some of the builtin components like Button and Image. When I moved these to a class I accidentally shipped some bugs because it was very hard to see where those props were being passed (some of them were just spread). Did you check for that too @fikrikarim?
> 
> It's possible that for those my solution was moving it back to Styled Components so this might not be an issue

Ah you're right. I missed those. I updated the `Image` and `Button`. There should be no more components that use `responsiveWidth` directly.

migueloller (approved):

PR #155, opened 2022-09-01
feat: add on demand product page revalidation

joshuawootonn:
Regardless of the 404'ing, I don't think we are going to get passed on demand revalidation for the product pages

As for reproducing the problem of 404 on the home page: I found that when I created the Makeswift site and published pages before I deployed the custom host there wasn't really any issues with there being a 404. I think I am going to advice users during the video to take that route


migueloller:
> As for reproducing the problem of 404 on the home page: I found that when I created the Makeswift site and published pages before I deployed the custom host there wasn't really any issues with there being a 404. I think I am going to advice users during the video to take that route

That seems like a very weird requirement. Are you sure that actually means people won't run into the issue?

PR #166, opened 2022-09-12
fix: remove throw that is preventing product 404

joshuawootonn:
Currently this line is causing `/product/foobar` to 500 instead of 404.

### DIFF
--- a/examples/bigcommerce/lib/bigcommerce/api.ts
+++ b/examples/bigcommerce/lib/bigcommerce/api.ts
@@ -99,9 +99,7 @@ export async function getCategories(): Promise<Category[]> {
   return result.data.site.categoryTree
 }
 
-export async function getProduct(id?: number): Promise<ProductFragment> {
-  if (id == null) return DEFAULT_PRODUCT
-
+export async function getProduct(id: number): Promise<ProductFragment | null> {
   const config = getConfig()
   const apiToken = await getApiToken()
   const response = await fetch(
@@ -130,7 +128,5 @@ export async function getProduct(id?: number): Promise<ProductFragment> {
 
   const [productEdge] = result.data.site.products.edges
 
-  if (productEdge == null) throw new Error(`Product with ID "${id}" not found.`)
-
-  return productEdge.node
+  return productEdge?.node ?? null
 }
--- a/examples/bigcommerce/pages/[[...path]].tsx
+++ b/examples/bigcommerce/pages/[[...path]].tsx
@@ -6,7 +6,7 @@ import {
 } from '@makeswift/runtime/next'
 import { GetStaticPropsContext, GetStaticPropsResult } from 'next'
 
-import { getProduct, getProducts } from 'lib/bigcommerce'
+import { DEFAULT_PRODUCT, getProducts } from 'lib/bigcommerce'
 import { PageProps } from 'lib/types'
 
 type Props = MakeswiftPageProps & PageProps
@@ -19,7 +19,7 @@ export async function getStaticProps(
   if (!('props' in makeswiftResult)) return makeswiftResult
 
   const products = await getProducts()
-  const product = await getProduct()
+  const product = DEFAULT_PRODUCT
 
   return { ...makeswiftResult, props: { ...makeswiftResult.props, products, product } }
 }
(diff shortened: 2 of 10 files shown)


joshuawootonn:
Sorry for alerting you @migueloller . Going to just keep this PR open as a catchall for updates that I will be making based on my own feedback from making the videos

migueloller (changes requested):

examples/bigcommerce/lib/bigcommerce/api.ts
```diff
@@ -130,7 +130,5 @@ export async function getProduct(id?: number): Promise<ProductFragment> {
 
   const [productEdge] = result.data.site.products.edges
 
-  if (productEdge == null) throw new Error(`Product with ID "${id}" not found.`)
```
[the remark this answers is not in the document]
joshuawootonn: You right

examples/bigcommerce/lib/bigcommerce/api.ts
```diff
@@ -130,7 +130,5 @@ export async function getProduct(id?: number): Promise<ProductFragment> {
...
-
-  return productEdge.node
+  return productEdge?.node
```
migueloller, line 133: The type of this function is now wrong. The real type is now `Promise<ProductFragment | undefined>`. I try to avoid `undefined`, though, and would suggest you make the type `Promise<ProductFragment | null>` and have the return type be `productEdge?.node ?? null`.

The reason TypeScript didn't catch this was because of the array access. Since TypeScript doesn't check for array bounds, array access is "unsafe"

migueloller (changes requested):

examples/bigcommerce/lib/bigcommerce/api.ts
migueloller: We should turn on `noUncheckedIndexedAccess`. It seems that it's a new option since TypeScript 4.1 that checks for this

migueloller: Check it out in action here: https://www.typescriptlang.org/play?noUncheckedIndexedAccess=true#code/MYewdgzgLgBGCuBbARgUwE4RgXhgbQEYAaGAJhIGYBdAKBtElmQEMATHOJNTPAFlvrgIIADaoAdCJABzABQt2AahgEAlDSA

migueloller: Yet more info: https://github.com/microsoft/TypeScript/pull/39560

examples/shopify/lib/shopify/api.ts
```diff
@@ -80,7 +80,7 @@ export async function getCollections(): Promise<Collection[]> {
   return result.data.collections.edges.map(edge => edge.node)
 }
 
-export async function getProduct(handle?: string): Promise<ProductFragment> {
+export async function getProduct(handle?: string): Promise<ProductFragment | null> {
```
joshuawootonn, line 83: This was mis typed before. Tested it out and the api returns nulls as the product if you don't pass a valid handle.

examples/shopify/pages/[[...path]].tsx
```diff
@@ -19,7 +20,7 @@ export async function getStaticProps(
   if (!('props' in makeswiftResult)) return makeswiftResult
 
   const products = await getProducts()
-  const product = await getProduct()
+  const product = (await getProduct()) ?? DEFAULT_PRODUCT
```
Comment (joshuawootonn) [inline examples/shopify/pages/[[...path]].tsx:23]: In catchall routes and /makeswift i think it makes sense to use the default product here.

In the /product dynamic route if getProduct is null then we 404 as expected

Comment (migueloller) [reply examples/shopify/pages/[[...path]].tsx:23]: https://github.com/makeswift/makeswift/pull/166#discussion_r970869479

migueloller (dismissed):

examples/bigcommerce/pages/[[...path]].tsx
```diff
@@ -19,7 +19,7 @@ export async function getStaticProps(
   if (!('props' in makeswiftResult)) return makeswiftResult
 
   const products = await getProducts()
-  const product = await getProduct()
```
Comment (joshuawootonn) [inline examples/bigcommerce/pages/[[...path]].tsx:22]: We know that this will be default product so we might as well be explicit about it

migueloller (approved):

PR #169, opened 2022-09-14
feat: update ecom example to correctly use context

joshuawootonn:
I tried to take a nap and it haunted me :)


migueloller (approved):
Looks good!

You can probably delete the `_app.tsx` files now as well as the `PageProps` type in the `lib/types.ts` file (or at least you dont't need to share that type anymore since each page has its own types and there's no `_app` implicitly using them).

joshuawootonn:
If I delete my `__app.tsx` component then where would I put my cart context?

migueloller:
> If I delete my `__app.tsx` component then where would I put my cart context?

Ahh, good point. I missed the cart context.

PR #183, opened 2022-09-27
Output a components manifest

alexhwoods:


alexhwoods:
We may need to do something about adding `.makeswift` to the `.gitignore` file

migueloller:
> We may need to do something about adding `.makeswift` to the `.gitignore` file

Are we not putting the manifest in the `.next` folder?

packages/makeswift-next-plugin/index.js
```diff
@@ -3,14 +3,23 @@
...
 const NEXT_TRANSPILE_MODULES_MODULES = ['@makeswift/runtime']
 
+const defaultMakeswiftConfig = {
+  components: [],
+}
+
 /** @type {(options: MakeswiftNextPluginOptions) => (nextConfig: NextConfig) => NextConfig} */
-module.exports = ({ resolveSymlinks } = {}) => (nextConfig = {}) => {
-  /** @type {NextConfig} */
+module.exports = ({ resolveSymlinks } = {}) => async (nextConfig = {}) => {
```
migueloller, line 21: Our "enhance" function should still be sync, what should be async is what we return from this function, which would be the function-form of a Next.js config

packages/makeswift-next-plugin/index.js
```diff
@@ -56,7 +65,115 @@ module.exports = ({ resolveSymlinks } = {}) => (nextConfig = {}) => {
...
-  })(enhancedConfig)
+  await createComponentsManifest()
+
+  return withTmInitializer(NEXT_TRANSPILE_MODULES_MODULES, { resolveSymlinks })(enhancedConfig)
+}
+
+async function createComponentsManifest() {
+  let makeswiftConfig
+  try {
+    // @todo: When will this assumption fail? Monorepos?
```
migueloller, line 76: Yeah, probably. We likely want a more robust way to find the config file

packages/makeswift-next-plugin/index.js
```diff
@@ -56,7 +65,115 @@ module.exports = ({ resolveSymlinks } = {}) => (nextConfig = {}) => {
...
+    const filename = path.join(process.cwd(), 'makeswift.config.js')
+    makeswiftConfig = require(filename)
+  } catch (err) {
+    if (err instanceof Error) {
+      if (err.code === 'ENOENT') {
+        makeswiftConfig = defaultMakeswiftConfig
```
migueloller, line 82: We should probably _always_ require a Makeswift config, since at the very least they should be providing an API key

packages/makeswift-next-plugin/index.js
```diff
@@ -56,7 +65,115 @@ module.exports = ({ resolveSymlinks } = {}) => (nextConfig = {}) => {
...
+
+  // @note: The .next directory keeps getting deleted after this runs
+  const outDir = path.join(process.cwd(), '.makeswift')
+  await createDotMakeswiftDirectory(outDir)
+
+  const outFile = path.join(outDir, 'components-manifest.json')
+  fs.promises.writeFile(outFile, JSON.stringify({ components }))
+}
+
+async function createDotMakeswiftDirectory(outDir) {
```
migueloller, line 107: I thought we were putting the manifest file under `.next`?

migueloller (changes requested):
Great start!

By the way, there is now a global Prettier installation, config, and ignore file to enforce a consistent style across the entire monorepo so we shouldn't need individual configuration files on specific packages.

alexhwoods:
> > We may need to do something about adding `.makeswift` to the `.gitignore` file
> 
> Are we not putting the manifest in the `.next` folder?

@migueloller See this comment — https://github.com/makeswift/makeswift/pull/183/files#diff-19f9b2611ef669911a4219dfef480818b939bbfdb19116162d7fdad0f6aaee79R99

Perhaps that's because of me exporting a function that gets called at `require` time and not by Next when it means to. Let me see if that fixes that

packages/makeswift-next-plugin/index.js
alexhwoods: This is what you mean, right?
```javascript
module.exports = ({ resolveSymlinks } = {}) => (nextConfig = {}) => async () => {
  // implement
}
```

alexhwoods:
> https://github.com/makeswift/makeswift/pull/183#pullrequestreview-1122757268

Cool. I think this got merged while I was working on this. I'll wait till the end to remove this, before I clean up the commits. I don't want striped commits of Prettier, actual work, Prettier, actual work. That's super annoying to rebase

packages/makeswift-next-plugin/index.js
migueloller: Right, and just to be more explicit:

```ts
type MakeswiftPluginOptions = {
  resolveSymlinks?: boolean
}

function makeswiftNextPlugin(pluginOptions: MakeswiftNextPluginOptions) {
  return (nextConfig: NextConfig) => async (phase, { defaultConfig }) => {
    // apply Makeswift enhancements and custom logic here...

    /**
     * @type {import('next').NextConfig}
     */
    const nextConfig = defaultConfig

    return nextConfig
  }
}
```

migueloller: If you're not getting the `phase` argument and the second argument bag which has the `defaultConfig` then that's an easy way to know whether it's in the right place. More info here: https://nextjs.org/docs/api-reference/next.config.js/introduction

packages/makeswift-next-plugin/index.js
migueloller: We just need a reliable way to find the project root, i.e., where the `package.json` and `.next` folder are located. It's possible `process.cwd()` will always be the right one if running `next dev` or `next start` makes that always the case, but I would assume not. In a monorepo, doing `yarn workspace next-app dev` would be fine since Yarn workspaces sets the current working directory to the the workspace. The only case I can think of is the binary being executed from `node_modules` like `./node_modules/.bin/next dev`, but in that scenario, my guess is `next dev` (or `next start`) would require `process.cwd()` to be the directory with the `package.json` anyway so `process.cwd()` might be the right assumption...

alexhwoods: Ok cool. I've centralized this logic so we can isolate the assumption and have an easy place to change it if it's wrong. Once we settle everything else I can test some of these different scenarios further

packages/makeswift-next-plugin/index.js
alexhwoods: Ok cool. I will leave it the way I have it for now, but once we move this thing to TS, I'll go with the above refactor

alexhwoods: (I just verified what is there based on your `phase` comment)

packages/makeswift-next-plugin/index.js
alexhwoods: I think "pit of success" > "supporting many different flavors of configuration".

So, 1 way to do things with very obvious error messages

packages/makeswift-next-plugin/index.js
migueloller: > I will leave it the way I have it for now, but once we move this thing to TS, I'll go with the above refactor

Are you think we would leave the PR open until then or were you thinking about merging it?

alexhwoods:
> > > > We may need to do something about adding `.makeswift` to the `.gitignore` file
> > > 
> > > 
> > > Are we not putting the manifest in the `.next` folder?
> > 
> > 
> > @migueloller See this comment — https://github.com/makeswift/makeswift/pull/183/files#diff-19f9b2611ef669911a4219dfef480818b939bbfdb19116162d7fdad0f6aaee79R99
> > Perhaps that's because of me exporting a function that gets called at `require` time and not by Next when it means to. Let me see if that fixes that
> 
> Yeah, this is probably happening because of that

Just checked, it's not. That folder is just wiped out after the plugins are run.

migueloller:
> @migueloller, on the topic of `.next`. vs. `.makeswift`, check this out. This is with me having updated the function to be a function that returns a function which returns an async function https://user-images.githubusercontent.com/7907782/192803775-52f9f7b6-f8f8-4eeb-bc27-91340404d16e.mov

Well, that's unfortunate. Given that JJ from the Vercel team suggested we do this it might be worth mentioning this to him. I _really_ wanted to avoid having the user do one more thing for configuring Makeswift

alexhwoods:
Note, we can also use `.next/cache`, as that does not get wiped out. Note that that does get uploaded to Vercel's build cache

alexhwoods:
With that info, I'm going to consider this closed.

migueloller:
> Note, we can also use `.next/cache`, as that does not get wiped out. Note that that does get uploaded to Vercel's build cache

That's nice. It's ok that's uploaded, we'll be overwriting the manifest file anyway whenever we write. Also, isn't the entirety of the `.next` folder uploaded? Or is it just `.next/cache`?

PR #187, opened 2022-09-27
Update CLI to support preview mode

alexhwoods:


packages/makeswift/src/integrate-next-app.ts
```diff
@@ -33,6 +37,24 @@ function installMakeswiftRuntime({ dir }: { dir: string }): void {
...
 
+function addMakeswiftApiRoute({ dir }: { dir: string }): void {
+  const pagesFolder = getPagesFolder({ dir })
+  const extension = getExtension({ dir })
+
+  // If Makeswift API folder does not exist, create
+  createFolderIfNotExists(path.join(pagesFolder, 'api'))
+  createFolderIfNotExists(path.join(pagesFolder, 'api', 'makeswift'))
+
+  const apiRoute = `import { MakeswiftApiHandler } from '@makeswift/runtime/next'
```
migueloller, line 48: The contents of the file should be different if it's a TypeScript installation vs plain JavaScript. This is specially marked in the `[[...path]].js` file since there's a lot of types we want to include there. We might want to have a `typescript: boolean` option we can pass for all these functions that generate files so that we can detect TypeScript when running the CLI and then can act based on that.

packages/makeswift/src/integrate-next-app.ts
```diff
@@ -59,20 +59,89 @@ function addMakeswiftPages({ dir }: { dir: string }): void {
...
+
+  if (snapshot == null) return { notFound: true }
+
+  return { props: { snapshot } }
+}
+
+export default function Page({ snapshot }) {
+  return <MakeswiftPage snapshot={snapshot} />
+}`
+      case 'ts':
```
migueloller, line 96: The extension here _has_ to be `tsx` otherwise there will be TypeScript errors with the JSX used in the file.

migueloller (changes requested):

packages/makeswift/src/integrate-next-app.ts
alexhwoods: Sorry, I think what I did is confusing here — this should be called `language`, not `extension`.

It is `${language}x` when I generate the file; I[ had caught that](https://github.com/makeswift/makeswift/pull/187/files#diff-038313253614925d50e16bd9d8bf678fbf73bd1e30bae3b4756ed381a12252e5R144). I think it's better if we operate on "language" than extension though, because some files will be `.jsx` and others `.js`, but for each file there's only 2 options — the JS or the TS variety. Whether it has React JSX is another dimension.

packages/makeswift/src/integrate-next-app.ts
alexhwoods: We do have that, but I agree that it should be called once and be a parameter to the functions. I'm going to add that.

Also, do we have an example somewhere of what the TS version should look like, if it's different? I was going by [this](https://www.makeswift.com/docs/guides/manual-setup#add-the-makeswift-api-handler).

packages/makeswift/src/integrate-next-app.ts
migueloller: Ah, got it. Yeah, we should probably go with the grain here and do what `create-next-app` does and have a boolean for whether we're using TypeScript or not.

packages/makeswift/src/integrate-next-app.ts
alexhwoods: https://github.com/makeswift/makeswift/blob/07b469e0ed6ad8290d36f42add86526b4047954b/examples/basic-typescript/pages/api/makeswift/%5B...makeswift%5D.ts

```typescript
import { MakeswiftApiHandler } from '@makeswift/runtime/next'

export default MakeswiftApiHandler(process.env.MAKESWIFT_SITE_API_KEY!)
```

Aren't these the same?

packages/makeswift/src/integrate-next-app.ts
alexhwoods: sounds good

packages/makeswift/src/integrate-next-app.ts
migueloller: The one that's mostly different is the dynamic catch-all route. For this one the only different is the non-null assertion in the API key so that TypeScript doesn't error

migueloller (approved):
Looks good!

migueloller (approved):

PR #195, opened 2022-10-05
Simple CLI Fixes

alexhwoods:


packages/makeswift/src/utils/is-next-app.ts
```diff
@@ -12,7 +12,9 @@ function hasNextDependency(appDirectory: string): boolean {
 
     return 'next' in dependencies
   } catch (error) {
-    if (error instanceof Error) console.log(error.message)
+    if (error instanceof Error) {
+      /* do nothing */
+    }
   }
```
migueloller, line 18: Is the `instanceof` check here necessary if we're not doing anything with it?

```suggestion
  } catch {
    // do nothing
  }
```

or

```suggestion
  } catch {
    return false
  }
```

or perhaps even better would be to check for the not found error of `require` and in that case return `false` and avoid a try/catch here. Or even using `fs.exists` before `require`, etc.

packages/makeswift/src/integrate-next-app.ts
```diff
@@ -5,8 +5,40 @@ import path from 'path'
 import { manipulateNextConfig } from './utils/manipulate-next-config'
 import { yarnOrNpm } from './utils/yarn-or-npm'
 import { createFolderIfNotExists } from './utils/create-folder-if-not-exists'
+import inquirer from 'inquirer'
+import chalk from 'chalk'
+
+async function getApprovalToIntegrate(dir: string): Promise<boolean> {
+  const projectName = dir.split('/').reduce((prev, curr) => curr)
```
migueloller, line 12: `reduce` will throw an error if the array is empty and an initial value is not given. Should we pass an initial value? Or will this never be of length 0? Also, is this just trying to get the last item in the array? Seems convoluted instead of doing something like this:

```suggestion
  const projectName = dir.split('/').slice(-1)[0]
```

or

```suggestion
  const projectName = dir.split('/').at(-1)
```

packages/makeswift/src/integrate-next-app.ts
```diff
@@ -5,8 +5,40 @@ import path from 'path'
...
+          projectName,
+        )} is an existing Next.js app — would you like to integrate it?`,
+      },
+    ]
+
+    inquirer.prompt(questions).then(answers => {
+      if (typeof answers.approval == 'boolean') {
+        resolve(answers.approval)
+      } else {
+        throw Error('Something went wrong')
```
migueloller, line 29: Do we want a better error message here? I'm assuming this will be caught and the message will be presented to the user?

packages/makeswift/src/init.ts
```diff
@@ -11,6 +11,7 @@ import { getProjectName } from './utils/get-name'
 import isNextApp from './utils/is-next-app'
 
 const MAKESWIFT_APP_ORIGIN = process.env.MAKESWIFT_APP_ORIGIN || 'https://app.makeswift.com'
+const MAKESWIFT_API_HOST = process.env.MAKESWIFT_API_HOST
```
migueloller, line 15: Can we change this variable to `MAKESWIFT_API_ORIGIN`. It used to be `MAKESWIFT_API_HOST` in the runtime but I realized that "host" was the wrong term and it should be "origin". I changed it in the runtime so we use `MAKESWIFT_API_ORIGIN` and `MAKESWIFT_APP_ORIGIN`

packages/makeswift/src/utils/check-for-conflicting-files.ts
```diff
@@ -0,0 +1,8 @@
+import * as glob from 'glob'
+
+const CONFLICTING_FILES = ['pages/api/makeswift.*', 'pages/[...path].*', 'pages/_document.*']
```
migueloller, line 3: What about `[[...path]].*`?

migueloller: Does `'pages/api/makeswift.*'` match `pages/api/makeswift.ts` as well as `pages/api/makeswift/index.ts` or any file under a `pages/api/makeswift` directory?

migueloller: Would it be useful to write test for this?

packages/makeswift/src/utils/__tests__/makeswift-plugin.test.ts
```diff
@@ -102,3 +102,21 @@ module.exports = withMakeswift(withFoo(withCSS({
   console.log(output)
   expect(output).toBe(expected)
 })
+
+test('already integrated, should not reintegrate', () => {
```
migueloller, line 106: ❤️

packages/makeswift/package.json
```diff
@@ -33,7 +33,7 @@
     "@babel/traverse": "^7.18.13",
     "chalk": "^4.1.2",
     "commander": "^9.4.0",
-    "create-next-app": "^12.2.5",
+    "create-next-app": "^12.3.1",
```
migueloller, line 36: Should we make this `*` so that we don't have to upgrade all the time? Or would you rather we keep a stable version? we could also do `12` so that it automatically gets minor versions in addition to patch

migueloller (changes requested):
This is looking good. I think we should go over some of the user-facing outputs at some point. I don't know if this PR is when you would want to do that or if you'd want to address that in a separate PR

packages/makeswift/src/integrate-next-app.ts
alexhwoods: Yeah, going to do more more work on error messages and how that's presented to the user in another PR

packages/makeswift/src/init.ts
alexhwoods: Yeah of course. I was wondering why `MAKESWIFT_API_HOST` was having no effect

packages/makeswift/package.json
alexhwoods: I suppose we should stick within a major version. But yes I don't want to be upgrading all the time

packages/makeswift/src/utils/check-for-conflicting-files.ts
alexhwoods: Just `pages/api/makeswift.ts` or `pages/api/makeswift.js` (because of the dot). Do we want to support the `pages/api/makeswift/index.ts` use case?

And I'm not sure this is the kind of thing that leans itself well to tests — I don't want to keep around some fake files just for that case

alexhwoods: And ah, thanks, the optional catch-all route should be on this list as well

packages/makeswift/package.json
alexhwoods: Is that `12.x`?

I will also say — I don't care that much if we're out of date by a few minor versions. I think it's about whether or not the CLI works, not if that warning is in the output.

packages/makeswift/src/init.ts
migueloller: Well, it will have no effect on the runtime right now, you'll need to pass options to the client as `new Makeswift(apiKey, { apiOrigin: process.env.MAKESWIFT_API_ORIGIN })`

packages/makeswift/src/utils/check-for-conflicting-files.ts
migueloller: We _don't_ want them to have a folder named `pages/api/makeswift` or a file named `pages/api/makeswift.*` So I guess right now we're checking for the file, we want to also check for the folder

packages/makeswift/package.json
migueloller: Hmm ok. If you don't care about the warning, then the caret version should be enough. We could also suppress the warning

alexhwoods: if the caret version is enough, why did the previous caret version get a warning? I guess the lock file?

packages/makeswift/src/utils/check-for-conflicting-files.ts
alexhwoods: Ok, adjusted for that

alexhwoods: (updated)

packages/makeswift/package.json
```diff
@@ -31,8 +31,9 @@
     "@babel/generator": "^7.18.13",
     "@babel/template": "^7.18.10",
     "@babel/traverse": "^7.18.13",
+    "chalk": "^4.1.2",
     "commander": "^9.4.0",
-    "create-next-app": "^12.2.5",
+    "create-next-app": "12.x",
```
migueloller, line 36: `^12.3.1` was fine, it means "12.x" already

migueloller (changes requested):

packages/makeswift/package.json
migueloller: Because of caches/lockfiles. When you run `npx` it installs a version and caches it. Next time you run it it will use that cached version and the installed version could be outdated. If you do `npx makeswift@latest init` then the warning shouldn't be there

packages/makeswift/package.json
alexhwoods: done

migueloller (approved):

PR #198, opened 2022-10-06
Better Workflow for Deciding the Project Name

alexhwoods:


migueloller (dismissed):
Looks good! Do you want to add a changeset? Also, does inquirer not support a default value for the input? Try out `create-next-app` and see how it behaves. I was hoping that we could provide similar behavior.

PR #200, opened 2022-10-10
Integrate Next App in a staging environment

alexhwoods:
This leaves the filesystem clean if they decide to cancel on the site selection page.

There is a bit of a race condition that occurs now, as shown in this video (use sound)

https://user-images.githubusercontent.com/7907782/194921856-e85acd5c-4a86-4c6a-84cc-21119b05022c.mov

### DIFF
--- a/packages/makeswift/src/integrate-next-app.ts
+++ b/packages/makeswift/src/integrate-next-app.ts
@@ -1,6 +1,8 @@
+import * as os from 'os'
 import spawn from 'cross-spawn'
 import glob from 'glob'
 import * as fs from 'fs'
+import * as fse from 'fs-extra'
 import path from 'path'
 import { isAlreadyIntegrated, manipulateNextConfig } from './utils/manipulate-next-config'
 import { yarnOrNpm } from './utils/yarn-or-npm'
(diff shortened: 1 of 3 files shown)


migueloller (changes requested):
The way we stage changes is a bit weird. It will override changes they make in any page throughout the integration even most of those files won’t even be modified.

I feel like we should be building up only the changed files in the temporary folder and then moving those over, and only those files which need changing

packages/makeswift/src/integrate-next-app.ts
```diff
@@ -9,21 +11,45 @@ import inquirer from 'inquirer'
...
-  // Step 4 - adding the Makeswift Next.js plugin
-  addMakeswiftNextjsPlugin({ dir })
+  // Step 3 - adding the Makeswift Next.js plugin
+  addMakeswiftNextjsPlugin({ dir: temporaryDirectory })
+
+  // Overwrite pages and next.config.js with output from temporary directory
+  overwriteIntegratedFiles({ dir, temporaryDir: temporaryDirectory })
+
+  // Step 4 - install the runtime
+  installMakeswiftRuntime({ dir })
```
alexhwoods: flip order of these

migueloller (approved):
Looks good 👍🏻

PR #204, opened 2022-10-12
CLI Bug Fixes

alexhwoods:


packages/makeswift/src/errors/MakeswiftError.ts
```diff
@@ -3,7 +3,7 @@
  * where we want the program to simply exit.
  */
 export default class MakeswiftError extends Error {
-  constructor(message: string) {
+  constructor(message: string | undefined = '') {
```
migueloller, line 6: When a default value is used, it's not necessary to add `| undefined` as an argument. Also, when you want to make an argument optional, you can do `message?: string` instead of using `undefined` explicitly

migueloller, line 6: Should we make the default message something more descriptive? Perhaps "An unexpected error occurred"?

packages/makeswift/src/init.ts
```diff
@@ -64,7 +65,14 @@ async function init(
   const selectSiteUrl = new URL(`${MAKESWIFT_APP_ORIGIN}/cli/select-site`)
   selectSiteUrl.searchParams.set('project_name', projectName)
   selectSiteUrl.searchParams.set('callback_url', callbackUrl)
-  await open(selectSiteUrl.toString())
+
+  const selectSiteUrlString = selectSiteUrl.toString()
+  console.log(
+    `\nRedirecting you to ${chalk.blue(
```
migueloller, line 71: What is instead of "Redirecting" we said "opening your browser at"? "Redirecting" sounds weird since we're not doing a redirect here and are instead opening their browser

packages/makeswift/src/integrate-next-app.ts
```diff
@@ -214,16 +216,26 @@ export default function Page({ snapshot }: Props) {
...
+      !fs.existsSync(path.join(pagesFolder.absolute, `index.${extension}`)) &&
+      !fs.existsSync(path.join(pagesFolder.absolute, `index.${extension}x`))
+
+    let catchAllRouteFilename
+    if (useOptionalCatchAllRoute) {
+      catchAllRouteFilename = `[[...path]].${extension}x`
+    } else {
+      catchAllRouteFilename = `[...path].${extension}x`
+
+      console.log(
```
migueloller, line 231: We might want to add a URL here to the Next.js docs on dynamic routes. This is an advanced feature and I guess most people won't know what a normal vs non-normal catch-all route is

packages/makeswift/src/integrate-next-app.ts
```diff
@@ -214,16 +216,26 @@ export default function Page({ snapshot }: Props) {
...
-  const catchAllRouteFilename =
-    fs.existsSync(path.join(pagesFolder, `index.${extension}`)) ||
-    fs.existsSync(path.join(pagesFolder, `index.${extension}x`))
-      ? `[...path].${extension}x`
-      : `[[...path]].${extension}x`
-
   // catch-all-route does not exist
-  if (glob.sync(path.join(pagesFolder, `\\[*\\].${extension}*`)).length === 0) {
-    fs.writeFileSync(path.join(temporaryPagesFolder, catchAllRouteFilename), catchAllRoute)
+  if (glob.sync(path.join(pagesFolder.absolute, `\\[*\\].${extension}*`)).length === 0) {
```
migueloller, line 220: What's this glob checking? If any sort of dynamic route exists? This check might be a bit too restrictive since there might be a dynamic route but it might not be a catch-all route. To check if it's a catch-all, we'd want to check if any file starts with `[[...` or `[...`

migueloller: We also don't have to worry if it's nested inside a folder since those catch-all routes won't conflict so a regular expression or a simple `startsWith` might be enough here

packages/makeswift/src/integrate-next-app.ts
```diff
@@ -232,24 +244,80 @@ export default function Page({ snapshot }: Props) {
 
   // custom document
   if (
-    fs.existsSync(path.join(pagesFolder, `_document.${extension}`)) ||
-    fs.existsSync(path.join(pagesFolder, `_document.${extension}x`))
+    fs.existsSync(path.join(pagesFolder.absolute, `_document.${extension}`)) ||
+    fs.existsSync(path.join(pagesFolder.absolute, `_document.${extension}x`))
   ) {
     throw new MakeswiftError(
       'A custom document already exists, you will have to manually integrate: https://www.makeswift.com/docs/guides/manual-setup#set-up-custom-document',
```
migueloller, line 251: We probably want to link to this instead: https://www.makeswift.com/docs/guides/advanced-setup#extending-the-custom-document

The link that's currently there is not very helpful as it's just a 1 liner export that doesn't explain how to use a custom document together with Makeswift

packages/makeswift/src/integrate-next-app.ts
```diff
@@ -232,24 +244,80 @@ export default function Page({ snapshot }: Props) {
...
+  // hello world component
+  fs.mkdirSync(path.join(temporaryDir, 'components'))
+  if (isTypeScript) {
+    fs.writeFileSync(
+      path.join(temporaryDir, 'components', 'hello-world.tsx'),
+      `import React from 'react'
+
+export default function HelloWorld(props: { className?: string }) {
+  return <p {...props}>Hello, world!</p>
+}
```
migueloller, line 292:
```suggestion
type Props = {
  className?: string
}

export function HelloWorld({ className }: Props) {
  return <p className={className}>Hello, world!</p>
}
```

migueloller: Note that the component is a named export so we'll need to update the `lib/makeswift/register-component.ts` file

packages/makeswift/src/integrate-next-app.ts
```diff
@@ -232,24 +244,80 @@ export default function Page({ snapshot }: Props) {
...
   }
   const customDocument = `export { Document as default } from '@makeswift/runtime/next'`
-  fs.writeFileSync(path.join(temporaryPagesFolder, `_document.${extension}`), customDocument)
+  fs.writeFileSync(path.join(pagesFolder.temporary, `_document.${extension}`), customDocument)
+
+  // register components file
+  fs.mkdirSync(path.join(temporaryDir, 'lib', 'makeswift'), { recursive: true })
+  const registerComponentsFile = `import { Style } from '@makeswift/runtime/controls'
+import { ReactRuntime } from '@makeswift/runtime/react'
+import HelloWorld from '../../components/hello-world'
```
migueloller, line 261:
```suggestion
import { HelloWorld } from '../../components/hello-world'
```

packages/makeswift/src/integrate-next-app.ts
```diff
@@ -232,24 +244,80 @@ export default function Page({ snapshot }: Props) {
...
+    className: Style({ properties: Style.All }),
+  },
+})
+`
+  fs.writeFileSync(
+    path.join(
+      temporaryDir,
+      'lib',
+      'makeswift',
+      `register-components.${isTypeScript ? 'ts' : 'js'}x`,
```
migueloller, line 278:
```suggestion
      `register-components.${isTypeScript ? 'ts' : 'js'}`,
```

packages/makeswift/src/utils/get-name.ts
```diff
@@ -79,7 +85,7 @@ async function askApprovalToIntegrateIfNeeded(dir: string): Promise<void> {
     const approval = await askApproval()
 
     if (!approval) {
-      throw new MakeswiftError('Will not integrate Next app.')
+      throw new MakeswiftError()
```
migueloller, line 88: This is an interesting way to abort. Could this not just be a `return`?

migueloller (changes requested):

packages/makeswift/src/errors/MakeswiftError.ts
alexhwoods: Cool, good feedback

alexhwoods: I think I'm actually going to make this a required string. There is only 1 use case that I've had for an empty string, and this is making me realize I should've just done `process.exit(0)`

packages/makeswift/src/init.ts
alexhwoods: Yeah sure

packages/makeswift/src/integrate-next-app.ts
alexhwoods: Yeah I agree with that, I think when I wrote this I thought `/foo/bar/baz/[[...slug]].tsx` — would conflict with us somehow.

Thinking now, it won't.

alexhwoods: - it only checks the pages folder, no sub-folders
- you're right about dynamic routes that are not catch-all routes — do we want to not catch that? Like if they have `pages/[slug].tsx`? Would Next then only match things like `/foo/bar` for our Makeswift page?

If we want to find catch-all routes that are not also top level dynamic routes, we could use `glob.sync("\\[?(\\[)...*\\]?(\\]).*")`

And with `startsWith` or a regex — I'd also be fine with a regex (I do think they're cleaner than globs), but we also need something that searches recursively through all files. We're not just matching a string

alexhwoods: If we keep the glob as is, I should change the error message to say "we detected a dynamic route"

packages/makeswift/src/errors/MakeswiftError.ts
migueloller: Yeah, that makes sense

packages/makeswift/src/utils/get-name.ts
alexhwoods: This is a mistake, but I think return would be too (it's pretty deep in the stack!)

I've gone with `process.exit(0)`

packages/makeswift/src/integrate-next-app.ts
migueloller: > you're right about dynamic routes that are not catch-all routes — do we want to not catch that? Like if they have pages/[slug].tsx? Would Next then only match things like /foo/bar for our Makeswift page?

Yeah, that's correct. This is actually the setup Recurrency has. They have `[slug].tsx` and `[[...path]].tsx`. The former does WordPress pages and the latter Makeswift pages. They're also slowly making `[slug].tsx` pull pages from Makeswift as they incrementally move over.

> And with startsWith or a regex — I'd be fine with a regex (I do think they're cleaner than globs), but we also need something that iterates through the files in the pages folder. We're not just matching a string. Glob does that nicely, in my opinion

Ah, yes, we'd have to list the top-level files in the `pages` directory. Up to you whether you wanna use a glob or a regular expression, etc.

> If we keep the glob as is, I should change the error message to say "we detected a dynamic route"

Sounds good.

I think that it's probably safer to catch _any_ dynamic route at the top level of `pages` even if something like `[slug].tsx` wouldn't conflict. We generate paths at the top-level in the builder so it would be confusing if we integrated but things just didn't work. I'd rather they do things manually or reach out to us at that point.

packages/makeswift/src/utils/get-name.ts
migueloller: Ok, cool. Are there any issues with `process.exit(0)` not behaving like a `throw` or `return`? As in, do we risk some code running after the stack of this function unwinds while Node.js is attempting to exit, or does it immediately stop executing when `process.exit` is called?

migueloller (approved):
This is looking good, man. I would just say test out the happy paths and some of the latest changes and if it's working well, let's ship it and move on to the UI changes and marketing content!

packages/makeswift/src/integrate-next-app.ts
alexhwoods: Ok cool. I will check for any top level dynamic route (the current glob), and change the error message to be clear

packages/makeswift/src/utils/get-name.ts
alexhwoods: I don't think we'll have any issues. https://stackoverflow.com/questions/37818846/what-happens-when-process-exit-used-in-nodejs-program

I'm not certain no other lines of code will be run, but just from quick testing it was fine.

We'd probably have more problems if we had background async stuff happening.

migueloller: Ok, cool. I tested these two scripts and they run as expected:

```js
process.addListener('exit', () => {
  console.log('Exiting...')
})
process.exit(0)
```

```
process.exit(0)
console.log('Unreachable.')
```

PR #214, opened 2022-10-20
Use legacy-peer-dependencies

alexhwoods:


migueloller (approved):
Nice! Now we should make it a goal to remove these ASAP

PR #215, opened 2022-10-21
fix: CORS issue when loading local font

fikrikarim:
We're using a query instead of a header because we load the font using `new FontFace().load()` that doesn't support adding a custom header.

### DIFF
--- a/packages/makeswift-next-plugin/index.js
+++ b/packages/makeswift-next-plugin/index.js
@@ -1,7 +1,7 @@
 // @ts-check
 
 /** @typedef {import('next').NextConfig} NextConfig */
-/** @typedef {{ resolveSymlinks?: boolean }} MakeswiftNextPluginOptions */
+/** @typedef {{ resolveSymlinks?: boolean; appOrigin?: string }} MakeswiftNextPluginOptions */
 
 const withTmInitializer = require('next-transpile-modules')
 
@@ -10,7 +10,7 @@ const NEXT_TRANSPILE_MODULES_MODULES = ['@makeswift/runtime']
 
 /** @type {(options: MakeswiftNextPluginOptions) => (nextConfig: NextConfig) => NextConfig} */
 module.exports =
-  ({ resolveSymlinks } = {}) =>
+  ({ resolveSymlinks, appOrigin = 'https://app.makeswift.com' } = {}) =>
   (nextConfig = {}) => {
     /** @type {NextConfig} */
     const enhancedConfig = {
@@ -57,6 +57,29 @@ module.exports =
             : rewrites?.afterFiles ?? [],
           fallback: Array.isArray(rewrites) ? [] : rewrites?.fallback ?? [],
         }
+      },
+      async headers() {
+        const headers = await nextConfig.headers?.()
+
+        return [
+          ...headers,
+          {
+            source: '/:path*',
+            has: [
+              {
+                type: 'query',
+                key: 'x-makeswift-allow-origin',
+                value: 'true',
+              },
+            ],
+            headers: [
+              {
+                key: 'Access-Control-Allow-Origin',
+                value: appOrigin,
+              },
+            ],
+          },
+        ]
       },
     }
 


packages/makeswift-next-plugin/index.js
```diff
@@ -58,6 +58,31 @@ module.exports =
...
+                type: 'query',
+                key: 'x-makeswift-load-fonts',
+                value: 'true',
+              },
+            ],
+            headers: [
+              {
+                key: 'Access-Control-Allow-Origin',
+                value:
+                  process.env.MAKESWIFT_APP_ORIGIN ||
```
migueloller, line 79: We probably want to make the app origin an option for the plugin, not an environment variable. I think environment variables should be handled at an application level, but this package is a library

packages/makeswift-next-plugin/index.js
```diff
@@ -58,6 +58,31 @@ module.exports =
...
+        const headers = await nextConfig.headers?.()
+
+        return [
+          ...headers,
+          {
+            source: '/:path*',
+            has: [
+              {
+                type: 'query',
+                key: 'x-makeswift-load-fonts',
```
migueloller, line 71: This rewrite seems to apply to all files and doesn't see to do anything font specific, so why make the query parameter `x-makeswift-load-fonts`?

Also, what's to prevent anybody that's sending a request to the Next.js app from using this query parameter?

And, will there be a conflict with us adding this header if there's some functionality in Next.js or perhaps Vercel that sets the `Access-Control-Allow-Origin` header to a different value, say `*`?

migueloller:
@fikrikarim, from what I'm seeing in these changes, there's nothing font specific so putting "fonts" in the query parameter constrains the functionality in a weird way. For example, if we needed to set CORS headers for something else, would we then make _another_ query parameter for that thing even though it just sets the same header to the exact same value? Also, there doesn't seem to be anything preventing anybody from using this special header. In the case of Preview Mode, we guarantee it's only done by a trusted party by requiring the query param to have a secret. Also, similar to how CORS negotiation works in the browser we might want the origin to be specified by the requester (it's currently done as an option/env var), which if we do, then there would be even more reason to have a secret.

Just some thoughts from look at this.

packages/makeswift-next-plugin/index.js
fikrikarim: Okay just updated it to be an option instead of an env.

packages/makeswift-next-plugin/index.js
fikrikarim: > This rewrite seems to apply to all files and doesn't see to do anything font specific, so why make the query parameter `x-makeswift-load-fonts`?

Yes. I totally agree with this. I was thinking to make a generic query that enables CORS for `app.makeswift.com`. Updated to `x-makeswift-allow-origin`.

> Also, what's to prevent anybody that's sending a request to the Next.js app from using this query parameter?

Nothing. Everyone can use this query param. This should be safe because we're setting the `Access-Control-Allow-Origin` to be `app.makeswift.com`, and not `*`. So, even if someone is using this query from other origin, it won't be allowed, because we're just allowing `app.makeswift.com` and not the request origin.

> And, will there be a conflict with us adding this header if there's some functionality in Next.js or perhaps Vercel that sets the Access-Control-Allow-Origin header to a different value, say *?

I thought about it. But, we're only setting the `Access-Control-Allow-Origin` header only on a request with a special query param, so it shouldn't override any existing functionality. 

Users can still intentionally or unintentionally override this behavior by setting the `Access-Control-Allow-Origin` header on their `next.config.js` file. I don't think there's a way to prevent that.

fikrikarim:
> @fikrikarim, from what I'm seeing in these changes, there's nothing font specific so putting "fonts" in the query parameter constrains the functionality in a weird way. For example, if we needed to set CORS headers for something else, would we then make _another_ query parameter for that thing even though it just sets the same header to the exact same value? Also, there doesn't seem to be anything preventing anybody from using this special header. In the case of Preview Mode, we guarantee it's only done by a trusted party by requiring the query param to have a secret. Also, similar to how CORS negotiation works in the browser we might want the origin to be specified by the requester (it's currently done as an option/env var), which if we do, then there would be even more reason to have a secret.
> 
> Just some thoughts from look at this.

Cool, thanks! Just replied to the previous comments.

packages/makeswift-next-plugin/index.js
migueloller: Ok, this all sounds good! Let's go ahead with it and gather data/learn if there's any issues

migueloller (approved):

PR #219, opened 2022-10-26
Invert handshake to select example

alexhwoods:


alexhwoods:
demo:
https://user-images.githubusercontent.com/7907782/198069290-b387e612-73b3-42e4-9624-38506a3a2235.mp4

packages/makeswift/src/init.ts
```diff
@@ -33,85 +33,103 @@ export default async function wrappedInit(name: string | undefined, args: InitAr
...
-  const envLocal = buildLocalEnvFile(
-    {
-      MAKESWIFT_SITE_API_KEY: siteApiKey,
-      MAKESWIFT_API_ORIGIN,
-    },
-    env,
-  )
   fs.writeFileSync(`${nextAppDir}/.env.local`, envLocal)
 
   spawn.sync('yarn', ['dev', '--port', nextAppPort.toString()], {
```
migueloller, line 119: What happens here is they don't have Yarn installed and are using NPM?

alexhwoods: Good catch. This could be what Lindsay was hitting. I'll address this in another PR

PR #226, opened 2022-11-01
Update Bigcommerce example to use their Vercel integration ENVs

fikrikarim:
Resolves PRD-638


examples/bigcommerce/lib/bigcommerce/api.ts
```diff
@@ -17,12 +16,12 @@ const A_WEEK_FROM_NOW = Math.floor(
 export async function getApiToken(): Promise<string> {
```
fikrikarim, line 16: Is `getApiToken` still needed @joshuawootonn? Maybe we can remove it if we don't use it anymore.

joshuawootonn: Yeah don't need it! I will remove it.

FWIW this is the reason I didn't use their env before. But I will update the readme to have a curl request for getting the storefront token in the case of them not using the integration

alexhwoods (dismissed):

alexhwoods (approved):

alexhwoods (approved):

PR #231, opened 2022-11-07
fix: text not properly blurred when clicking panel

fikrikarim:


migueloller (approved):
@fikrikarim, these changes look good to me! I would do as much QA as possible to make sure that there isn't something else that's obvious that broke. But if we can't find anything, then let's just ship it. We can always fix future bugs and we've got plans to rewrite all of this logic anyway.

PR #236, opened 2022-11-09
fix(next-plugin): `locale` to `false` in rewrites

migueloller:


joshuawootonn:
@migueloller would you consider this ready to merge? I pulled it into a repro and tried it for myself. Seem great to me. 

I used this page and verified that paths for `/`, `/foo`, `/fr`, `/fr/foo` were resulting in the pages and locals we were expecting
```
export function getStaticProps(ctx: any) {
  return {
    props: {
      ctx: JSON.stringify(ctx),
    },
  };
}

export default function About(props: any) {
  const router = useRouter();

  return (
    <div>
      it is working {props.ctx} {JSON.stringify(router.query, null, 2)}
    </div>
  );
}
```

One thing we could change is conditionally adding `locale: false` based on whether or not they provide the key `i18n`.

joshuawootonn:
> When leveraging [i18n support](https://nextjs.org/docs/advanced-features/i18n-routing) with rewrites each source and destination is automatically prefixed to handle the configured locales unless you add locale: false to the rewrite. If locale: false is used you must prefix the source and destination with a locale for it to be matched correctly. 

It seems that this sentence is incorrect "If locale: false is used you must prefix the source and destination with a locale for it to be matched correctly."

From my repro it seems that setting locale to false just enabled the root route. But sub routes already work

https://user-images.githubusercontent.com/20950876/206020605-af10efe1-e73e-446c-b30c-c92f07c01029.mp4

migueloller:
@joshuawootonn, if you've tested it and it works well, then sure, let's ship it!

joshuawootonn (approved):

PR #246, opened 2022-11-17
fix: update thirdweb example to new version of sdk

joshuawootonn:
This PR is hard to review because you can't see the deployed version, so I made a demo and compared the now working code to another thirdweb example. 

https://user-images.githubusercontent.com/20950876/202516950-05f6969d-2a22-4bdb-8ff6-8877b8e75d4d.mp4

In the video, i failed to mint both times. I believe this is because the person that made that NFT doesn't have enough money in their account for the mint transaction cost. I think this is still enough to call this fixed because upgrading the thirdweb sdks fixed the 429s we were getting. 

Related thirdweb slack: https://makeswifthq.slack.com/archives/C03NWNGPG5P/p1668639336987069


migueloller (approved):
I'm not super familiar with this code, but changes look good to me. Maybe @apledger has more thoughts on this?

PR #250, opened 2022-11-23
feat: update deploy buttons to include template

joshuawootonn:


migueloller (approved):
I don't understand what's going on here since I don't know what the `external-id` query param is used for but I guess this looks good? 🤷🏻 

Here's some thoughts that come to mind:
- The fact I didn't know what this variable was for means there might be a better name needed here
- For URL query params, snake_case is what's usually used but here we're using kebab-case which seems to be consistent with the other variables Vercel's using. I like being consistent with Vercel here, so it's a good reason to deviate from our usual style with snake_case
- Are we planning on changing the name (in the public API) of the templates to drop the `ecommerce-` prefix? i.e., `shopify` instead of `ecommerce-shopify`?

joshuawootonn:
CleanShot 2022-11-29 at 10 58 05@2x

`external-id` is passed via the deploy links to the integration (our `/vercel/select-site`) so we aren't in control of the casing. I agree. It's a bad name :P 


> Are we planning on changing the name (in the public API) of the templates to drop the ecommerce- prefix? i.e., shopify instead of ecommerce-shopify?

I was not planning on it. Feel free to create a ticket for this work if you think we should. Originally we made the `slug` in our templates reflect the `name`. And the names were prefixed with `ecommerce-` since we weren't changing the template selection page to include categories.

PR #251, opened 2022-11-27
fix: column with a dot not recording data

fikrikarim:
https://user-images.githubusercontent.com/13066728/204128016-2b3971ae-557e-415e-b646-f9f7eb075efc.mp4

The malformed form (no pun intended) was causing the `CreateTableRecord` mutation to send the wrong data for the field containing a dot.

I tested the fix and it's properly added to the database.


migueloller (approved):
Looks great! Nice and clean fix.

PR #252, opened 2022-11-29
Create an endpoint for creating a new element tree from an existing one

alexhwoods:


packages/runtime/src/api/templates.ts
```diff
@@ -0,0 +1,102 @@
...
+import {
+  getComponentPropControllerDescriptors,
+  getPropControllerDescriptors,
+  State,
+} from '../state/react-page'
+
+type ElementData = {
+  type: string
+  key: string
+  props: any
```
alexhwoods, line 12: Note - there is **a lot** of types I could bring into the runtime.

I want some, but I don't want to go overboard. This is where I drew that line, having `props` be of type `any`. I believe this is the right call, because for code components, we have no type information.

packages/runtime/src/api/templates.ts
```diff
@@ -0,0 +1,102 @@
...
+  replacementContext: ReplacementContext
+}) {
+  const copy = JSON.parse(JSON.stringify(elementTree)) as ElementData
+  const store = ReactRuntime.__store()
+  const state = store.getState()
+
+  return createNewElementTreeNode(copy, { replacementContext, state })
+}
+
+function createNewElementTreeNode(
```
alexhwoods, line 80: Note - I'm a recursive function! DFS

packages/runtime/src/api/templates.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  for (const propKey of Object.keys(node.props)) {
+    const descriptors = getComponentPropControllerDescriptors(state, node.type)
+    if (descriptors == null) continue
+
+    const descriptor = descriptors[propKey]
+
+    const copyFunction = getCopyFunction(descriptor)
+
+    node.props[propKey] = copyFunction(node.props[propKey], replacementContext)
```
alexhwoods, line 98: This really does feel like it should be polymorphism here, e.g.

```typescript
const control = createControlFromDescriptor(descriptor)

node.props[propKey] = control.copy(node.props[propKey])
```

That's a bit of a debate for another time though

packages/runtime/src/next/api-handler.ts
```diff
@@ -175,6 +178,30 @@ export function MakeswiftApiHandler(
...
+
+        if (elementTree == null) {
+          return res.status(400).json({ message: 'elementTree must be defined' })
+        }
+
+        if (serializedReplacementContext == null) {
+          return res.status(400).json({ message: 'replacementContext must be defined' })
+        }
+
+        const replacementContext = createReplacementContext(serializedReplacementContext)
```
alexhwoods, line 195: A question I think you might have — "Why don't we need a replacement context endpoint?"

When we were doing templates back in January, we made the choice that we would copy over all **site resources**, and use them to seed the replacement context. So orion can manage that on its own.

In hindsight, I think this was the right call; it's making this work simpler.

packages/runtime/src/prop-controllers/copy.ts
```diff
@@ -0,0 +1,15 @@
...
+}
+
+export function getCopyFunction(descriptor: Descriptor) {
+  switch (descriptor.type) {
+    case 'Backgrounds':
+      return BackgroundsCopy
+    default:
+      return defaultCopy
+  }
+}
```
alexhwoods, line 15: Don't have a strong opinion as to the folder structure of how we manage these copy functions.

Although, definitely 1 copy function per file. They're a bit big. You might know if there's a more natural place for them to reside though

packages/runtime/src/api/templates.ts
migueloller: Most types should already be defined in the runtime

packages/runtime/src/prop-controllers/copy.ts
```diff
@@ -0,0 +1,15 @@
+import { Descriptor } from './descriptors'
+import { copy as BackgroundsCopy } from './copy/Backgrounds'
```
migueloller, line 2: camelCase?

packages/runtime/src/prop-controllers/descriptors.ts
```diff
@@ -204,7 +204,7 @@ export type BackgroundsValue = ResponsiveValue<Background[]>
 
 type BackgroundsOptions = Options<Record<string, never>>
 
-type BackgroundsDescriptor<_T = BackgroundsValue> = {
+export type BackgroundsDescriptor<_T = BackgroundsValue> = {
```
migueloller, line 207: We might not need to expose this type, just `Descriptor`

packages/runtime/src/api/templates.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  return createNewElementTreeNode(copy, { replacementContext, state })
+}
+
+function createNewElementTreeNode(
+  node: ElementData,
+  { replacementContext, state }: { replacementContext: ReplacementContext; state: State },
+): ElementData {
+  if (node.props?.children != null) {
+    node.props.children.elements = node.props.children.elements.map((child: ElementData) =>
```
migueloller, line 85: Can't assume `props.children`

packages/runtime/src/prop-controllers/copy/Backgrounds.ts
```diff
@@ -0,0 +1,61 @@
+import { ReplacementContext } from '../../api/templates'
```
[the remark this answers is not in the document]
alexhwoods: Will break up big `descriptors.ts` file in a subsequent PR

packages/runtime/src/state/react-page.ts
```diff
@@ -168,6 +169,116 @@ export function getElementId(state: State, documentKey: string, elementKey: stri
...
+   * themselves, to enable mutual recursion.
+   *
+   * Consider the slot control. It has to iterate through its elements, and for each of them,
+   * call some version of the below function.
+   *
+   * That is how the recursing through the tree happens.
+   */
+  function copyElementTreeNode(state: State, replacementContext: ReplacementContext) {
+    return function (node: Documents.Element) {
+      if ('props' in node) {
```
migueloller, line 257: Seems like we also want to handle element references and not just element data, right? For example, if it's an element reference we want to replace the global element ID with the one in the replacement context?

Also, we've got guard functions to refine element references and element data. I believe it's `isElementReference` and `isElementData`. They're set up so that if you do `if(isElementReference(anyValue))` the `if` block refines `anyValue` to the checked type.

migueloller (approved):
Looks good! Just had a minor comment about the copy logic only handling element data but not element references

packages/runtime/src/state/react-page.ts
alexhwoods: Yeah I'm aware of that, gonna handle this when I get to global elements

PR #257, opened 2022-12-01
fix: [ ] ' in table column causing data to not be recorded.

fikrikarim:
## Background

Previously, we had an issue where a dot `.` in a table column name was causing the data for that column to not be recorded. This was caused by [Formik nested object behavior](https://formik.org/docs/guides/arrays#avoid-nesting). We fixed that by wrapping the value to be `['${tableColumn.name}']`.

## New problem

This causes a new problem where an apostrophe `'` in a table column name was causing the data for that column to not be recorded again. This is happening because Formik is wrapping the name inside a quote`'`. For example, if we have a column name of `hello'world`, formik will change it to `'hello'world`.

## Fix

We fix this by escaping the `'` character, before passing it to Formik. We also need to escape `[` and `]` because it's used by [Lodash set array](https://lodash.com/docs/4.17.15#set) that Formik is using.

## Proof

To make sure that no other characters are causing the same problem, I test it by making a table column name with all ASCII characters:
```
!"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\]^_`abcdefghijklmnopqrstuvwxyz{|}~
```
I also tested a random Unicode string:
```
⽡≱⎒⥬⇍┅ⅎ⽭⊥⼅⸠⢒␙ⅉ⬅₝⧣ⷃ⪇⏘⌔⣊Ⲧ▧⦺⑕⦩⎢⥌➘ⓢ⵾Ⓙ⨮ ⃋⇻ⵎ⓶⮦⧽⮢⾔ⳤ┒␩⑲⣂ⶽ⣃⃎⸅⢷⋗␡⃖➙⧮⎣⾑⤠⤗⨨≇⇊❘✝⿽✧⛮⾆₫◉⾐ℌ┈⧰⾂⦀⩌✄⏵⢥↪⛧‎⸫℥ⳡ⃐⨤⮦⸬☧⾴⛧⃔⬌ⲱ␤
```

https://user-images.githubusercontent.com/13066728/205008308-8794a89e-d26e-4df5-b3dc-10e2f8f9f5d5.mp4

### DIFF
--- a/packages/runtime/src/components/builtin/Form/components/Field/index.tsx
+++ b/packages/runtime/src/components/builtin/Form/components/Field/index.tsx
@@ -85,6 +85,11 @@ type Props = {
   tableFormField: TableFormField
 }
 
+// Escape characters: [ ] '
+function escapeCharacters(string: string) {
+  return string.replace(/[[\]']/g, '\\$&')
+}
+
 export default function Field({
   tableColumn,
   tableFormField: {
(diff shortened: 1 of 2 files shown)


packages/runtime/src/components/builtin/Form/components/Field/index.tsx
```diff
@@ -85,6 +85,11 @@ type Props = {
   tableFormField: TableFormField
 }
 
+// Escape characters: [ ] '
+function escapeCharacters(string: string) {
+  return string.replace(/[[\]']/g, '\\$&')
```
migueloller, line 90: Doesn't the `[` need to be escaped as well? The syntax highlighting seems to be giving that off as well.

migueloller (approved):

packages/runtime/src/components/builtin/Form/components/Field/index.tsx
fikrikarim: Shouldn't be. I got this function from [here](https://stackoverflow.com/questions/3561493/is-there-a-regexp-escape-function-in-javascript/3561711#3561711), and it looks like it's working on [regExr](https://regexr.com/)

PR #282, opened 2022-12-15
Fix content-mode overlay doesn't appear properly.

fikrikarim:
https://user-images.githubusercontent.com/13066728/207834340-10a8d408-ce6c-4aa4-a870-517e7d08f4f6.mp4

### DIFF
--- a/.changeset/fluffy-dingos-train.md
+++ b/.changeset/fluffy-dingos-train.md
@@ -1,0 +1,5 @@
+---
+'@makeswift/runtime': patch
+---
+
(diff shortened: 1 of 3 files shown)


migueloller (approved):
Looks good!

I'm guessing that we need the pointer because in content mode we "cut out" parts of the SVG overlay causing the pointer state in the builder to stop being tracked while the pointer is over a content rect? If so, this will also be useful for when we have interaction mode so that we know where the pointer is, even if we don't have an overlay.

fikrikarim:
Yes, exactly 💯

PR #285, opened 2022-12-16
chore: upgrade examples to use runtime v0.2.19

fikrikarim:


migueloller (commented):
@fikrikarim, changes look pretty straightforward here. I'm guessing there's something you want to me to comment on with regards to the example review apps failing? To fix those we could either update the configuration in the Vercel projects to do what it was doing before explicitly so that it doesn't do the default Turbo command. We could also make it so that the examples are part of the monorepo and Turbo commands work out of the box as well.

The former will require no changes in code, just in Vercel. The latter would require to make it so that examples are part of the Yarn workspaces and that we remove the lockfiles (I'm not sure if Turbo will work with them without us making them part of Yarn workspaces). I'm looking at Next.js examples and they don't seem to have lockfiles so maybe that's ok? Then again, their examples aren't part of their monorepo packages. Also, we don't want to be building these examples every time that we make a change that's not associated with them, so perhaps we need to re-think how we run pipelines for this repo as far as examples go...

Would love to hear your thoughts on each of these things. I'm thinking that the ideal end state is of course that we fix it in Vercel, but also that these examples aren't part of workspaces so we're not having to install a bunch of example dependencies when using the repo, and also whenever we push changes we're not deploying the examples unless we actually made a change to the example.

fikrikarim:
I don't think it's a good idea to include the examples into our monorepo. As you said, we could have 100 examples, and we don't want to install all their dependencies.

I'm leaning toward updating the configuration in the Vercel projects. I think this is a CI problem, so we just need to fix the CI configuration. 

I don't have permission to change the build settings in the example project. So either I need to have access to change it, or I need help from someone who can change it 😄

migueloller:
> I don't think it's a good idea to include the examples into our monorepo. As you said, we could have 100 examples, and we don't want to install all their dependencies.

👍🏻 

> I'm leaning toward updating the configuration in the Vercel projects. I think this is a CI problem, so we just need to fix the CI configuration.

👍🏻 

> I don't have permission to change the build settings in the example project. So either I need to have access to change it, or I need help from someone who can change it 😄

Just added you as an admin of all the example projects

fikrikarim:
All pipelines passed! Merging this.

PR #292, opened 2022-12-31
Remove Apollo dependency from the runtime

migueloller:


packages/runtime/src/api/constants.ts
```diff
@@ -1,26 +0,0 @@
...
-  Swatch: SwatchFragmentDoc,
-  File: FileFragmentDoc,
-  Typography: TypographyFragmentDoc,
-  PagePathnameSlice: PagePathnameSliceFragmentDoc,
-  GlobalElement: GlobalElementFragmentDoc,
-  Table: TableFragmentDoc,
-  Snippet: SnippetFragmentDoc,
-  Page: PageFragmentDoc,
-  Site: SiteFragmentDoc,
-}
```
migueloller, line 26: This file can't be deleted because the `Fragments` export is depended on by the builder to extract the data from the cache that it sends to the builder.

packages/runtime/src/api/graphql/client.ts
```diff
@@ -0,0 +1,32 @@
+export class GraphQLClient {
+  private endpoint: string
+
+  constructor(endpoint: string) {
+    this.endpoint = endpoint
+  }
+
+  async request<TData, TVariables = Record<string, never>>(
```
migueloller, line 8: The generics aren't required so it's possible to miss them. If it's possible to make them required that would reduce the probability of making a mistake

packages/runtime/src/api/graphql/client.ts
```diff
@@ -0,0 +1,32 @@
...
+    if (!response.ok && response.status !== 400) {
+      throw new Error(`${response.status} ${response.statusText}`)
+    }
+
+    const body = await response.json()
+
+    if (body.errors != null) {
+      console.error(body)
+
+      throw new Error('GraphQL response contains errors, check the console.')
```
migueloller, line 27: A more descriptive error message would probably be better here

packages/runtime/src/api/index.ts
```diff
@@ -1,2 +1,2 @@
-export * from './constants'
-export * from './types'
+// TODO(miguel): This is a public API ;(
```
migueloller, line 1: Handle this TODO

packages/runtime/src/api/react.tsx
```diff
@@ -1,157 +1,152 @@
...
-        return existingData ?? toReference({ __typename: 'Page', id: args?.id }, true)
-      },
-      site(existingData, { args, toReference }) {
-        return existingData ?? toReference({ __typename: 'Site', id: args?.id }, true)
-      },
-    },
-  },
+export type MakeswiftClientOptions = {
+  uri: string
+  serializedState?: MakeswiftApiClient.SerializedState
```
migueloller, line 28: Maybe we want to abstract this behind a `cacheData` name and a type that doesn't expose the store's internal serialized state?

packages/runtime/src/api/react.tsx
```diff
@@ -1,157 +1,152 @@
...
 }
 
-type CreateApolloClientParams = {
-  uri?: string
-  cacheData?: NormalizedCacheObject
-}
+// NOTE(miguel): MakeswiftClient -> HostClient
```
migueloller, line 31: Expand on this note or remove it. As it is it will probably be confusing in a couple of months.

packages/runtime/src/api/react.tsx
```diff
@@ -1,157 +1,152 @@
...
+    return {
+      serializedState: MakeswiftApiClient.getSerializedState(this.makeswiftApiClient.getState()),
+    }
+  }
 
-export class MakeswiftClient {
-  apolloClient: ApolloClient<NormalizedCacheObject>
+  // updateCacheData(cacheData: NormalizedCacheObject): void {
+  //   this.apolloClient.cache.restore(cacheData)
+  // }
```
migueloller, line 64: Need to handle this case with the new data cache

packages/runtime/src/api/react.tsx
```diff
@@ -1,157 +1,152 @@
...
-export function createApolloClient({ uri, cacheData }: CreateApolloClientParams) {
-  const cache = new InMemoryCache({ typePolicies })
+  async prefetch(element: Element): Promise<{
+    serializedState: MakeswiftApiClient.SerializedState
+  }> {
+    const introspectionData = await introspect(element, this, storeContextDefaultValue)
 
-  if (cacheData) cache.restore(cacheData)
+    await this.makeswiftApiClient.dispatch(
+      MakeswiftApiClient.preloadIntrospectionResources(introspectionData),
```
migueloller, line 52: To be consistent might just want to name this as a normal query since all queries "preload" normalized resources?

packages/runtime/src/api/react.tsx
```diff
@@ -1,157 +1,152 @@
...
-  mutation: DocumentNode | TypedDocumentNode<TData, TVariables>,
-  options?: MutationHookOptions<TData, TVariables, TContext>,
-): MutationTuple<TData, TVariables, TContext, TCache> {
+const Context = createContext<MakeswiftClient | undefined>(undefined)
+
+export function useMakeswiftClient(): MakeswiftClient {
   const client = useContext(Context)
 
-  return useApolloMutation(mutation, { client: client?.apolloClient, ...options })
+  if (client == null) throw new Error('Makeswift API client must be provided.')
```
migueloller, line 147: Could we provide a default value to the context to avoid this?

packages/runtime/src/components/builtin/Text/components/RichTextEditor/components/Mark/hooks/useTypographyMark.ts
```diff
@@ -66,26 +72,15 @@ export const overrideTypographyStyle = <A>(
...
-  const { swatches = [] } = colorData
-
-  if (colorError != null) return null
+  const swatches = useSwatches(swatchIds)
 
   return overrideTypographyStyle(
-    (typography && typography.style.map(withColor(swatches))) || [],
-    (value && value.style.map(withColor(swatches))) || [],
+    (typography && typography.style.map(withColor(swatches.filter(isNonNullable)))) || [],
+    (value && value.style.map(withColor(swatches.filter(isNonNullable)))) || [],
```
migueloller, line 84: Optional chaining + nullish cialescing operation

packages/runtime/src/next/client.ts
```diff
@@ -44,13 +44,14 @@ export type MakeswiftPageDocument = {
...
   apiOrigin: string
-  cacheData: NormalizedCacheObject
+  serializedState: SerializedState
 }
 
 type MakeswiftConfig = {
   apiOrigin?: string
 }
 
+// NOTE(miguel): Makeswift -> MakeswiftClient
```
migueloller, line 54: Expand on this NOTE

packages/runtime/src/next/index.tsx
```diff
@@ -125,13 +125,13 @@ export const Page = memo(({ snapshot }: PageProps) => {
...
+        serializedState: snapshot.serializedState,
       }),
   )
 
-  useEffect(() => {
-    client.updateCacheData(snapshot.cacheData)
-  }, [client, snapshot])
+  // useEffect(() => {
+  //   client.updateCacheData(snapshot.cacheData)
+  // }, [client, snapshot])
```
migueloller, line 134: Handle this with new data cache

packages/runtime/src/prop-controllers/descriptors.ts
```diff
@@ -222,7 +222,11 @@ export function Backgrounds(options: BackgroundsOptions = {}): BackgroundsDescri
 
 type BorderSideStyle = 'dashed' | 'dotted' | 'solid'
 
-type BorderSide = { width: number | null | undefined; style: BorderSideStyle; color?: Color | null }
+export type BorderSide = {
+  width: number | null | undefined
+  style: BorderSideStyle
+  color?: Color | null
+}
```
migueloller, line 229: Unnecessary change?

packages/runtime/src/runtimes/react/index.tsx
```diff
@@ -275,16 +272,10 @@ const ElementReference = memo(
...
-      variables: { id: elementReference.value },
-    })
-    const globalElementData = data?.globalElement?.data as ReactPage.ElementData | undefined
+    const globalElement = useGlobalElement(elementReference.value)
+    const globalElementData = globalElement?.data as ReactPage.ElementData | undefined
     const elementReferenceDocument = useDocument(elementReference.key)
 
-    if (error != null) {
-      return <FallbackComponent ref={ref as Ref<HTMLDivElement>} text="Something went wrong!" />
-    }
```
migueloller, line 286: We could update the hook to keep this error handling if we want to avoid changing behavior too much

packages/runtime/src/state/modules/api-resources.ts
```diff
@@ -0,0 +1,298 @@
...
+              action.payload.resource,
+            ),
+          }
+
+        default:
+          return state
+      }
+    }
+
+    case ActionTypes.EVICT_API_RESOURCE:
```
migueloller, line 294: Handle cache evictions

migueloller (commented):
- The `documents.ts` file isn't consistent in how it defines document variable and operation names
- Maybe should get rid of all the unused pending and rejected actions. No need to add them if they're not used
- Should we consolidate the state and/or actions into a generic "resources" normalized state instead of hard-coded individual maps for each type of resource?

packages/runtime/src/api/graphql/client.ts
migueloller: Not going to worry about this

packages/runtime/src/api/graphql/client.ts
migueloller: Whoops, never got around to this one. We can address it in a separate release.

PR #296, opened 2023-01-04
Update next-transpile-modules, next/link, and next/image for Next.js 13 support

fikrikarim:
Resolves PRD-906, PRD-907, PRD-908, PRD-911

Actually, we don't need `next-transpile-modules` [anymore from Next.js v13.1](https://github.com/martpie/next-transpile-modules/releases/tag/the-end), but we probably need to update it to support v13.0. 

So far, I haven't seen any issues from updating `next-transpile-modules` on Next v12.


migueloller (changes requested):
These changes look good, but I don't think we can ship them as-is since it would likely break users in Next.js v12, right? For example, the `useLegacyBehavior` prop is only in `next/link` when using Next.js v13?

I think that the way that we handle support for Next.js v13 is similar to how React introduced the `useSyncExternalStore` for React 18. They released a package `use-sync-external-store`, and that package had a shim `use-sync-external-store/shim`. A shim is a module that will transparently use the native library if it's available and otherwise it provides an appropriate replacement. In the case of `useSyncExternalStore`, if you're on React v18 is just uses the one from React, otherwise it falls back to a close implementation for React < v17.

https://github.com/facebook/react/blob/de7d1c90718ea8f4844a2219991f7115ef2bd2c5/packages/use-sync-external-store/src/useSyncExternalStoreShim.js#L21

We don't need to make a separate package, but we can create shim modules for Next.js features so that we can ship the same behavior across different versions of Next.js (e.g., `next-transpile-modules`, new `next/image`, and new `next/link`).

packages/runtime/src/components/shared/Link/index.tsx
```diff
@@ -123,10 +123,15 @@ export const Link = forwardRef<HTMLAnchorElement, Props>(function Link(
...
-      </NextLink>
+      <NextLink
+        {...restOfProps}
+        ref={ref}
+        target={target}
+        onClick={handleClick}
+        href={href}
+        // Next.js v12 has legacyBehavior set to true by default
+        legacyBehavior={false}
+      />
```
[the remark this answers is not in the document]
fikrikarim: We might need a shim for `next/image`, but we don't need it for `next/link`.

`legacyBehavior` is supported from [Next.js 12](https://github.com/vercel/next.js/pull/36436). So this should work for both Next.js 13 and 12.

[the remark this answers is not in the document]
fikrikarim: Thanks for linking the shim code from React! We probably need it for `next/image`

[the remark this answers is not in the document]
migueloller: It seems this was for Next.js v12.2: https://nextjs.org/blog/next-12-2#other-improvements

We might still need to shim it if the version is lower. Or do you think we only support Next.js v12.3 and v13+?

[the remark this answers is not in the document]
fikrikarim: Our runtime has peerDependency of `"next": ">=12.2.0"`, so I thought that we only support Next v12.2+?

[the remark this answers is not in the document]
migueloller: Ok, perfect. I think that we do 12.2+ for on-demand ISR. This means our peer dependencies should be `"next": ">=12.2.0 < 13.0.0 || ^13.0.0"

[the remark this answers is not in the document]
fikrikarim: Cool, I added the next.js version checking, and package transpile support on Next.js 13 using their built-in feature.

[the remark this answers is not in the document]
fikrikarim: If they are on version `13.0.*`, there would be a warning:
```
warn  - You have enabled experimental feature (transpilePackages) in next.config.js.
```
because `transpilePackages` is still experimental. This should be fine, as on their MR they called it ["Move transpilePackages out of experimental"](https://github.com/vercel/next.js/pull/44194), but we could just not support `13.0.*` if we want to.

packages/makeswift-next-plugin/index.js
```diff
@@ -81,7 +82,38 @@ module.exports =
...
     }
 
-    return withTmInitializer(NEXT_TRANSPILE_MODULES_MODULES, {
-      resolveSymlinks,
-    })(enhancedConfig)
+    const nextVersion = require('next/package.json').version
+
+    if (satisfies(nextVersion, '<12.2.0')) {
+      throw new Error(
+        'Makeswift requires a minimum Next.js version 12.2.0. Please update to Next.js version ^12.2.0 if you want to use Next.js 12, or version ^13.0.0 if you want to use Next.js v13.',
```
migueloller, line 89: Some copy suggestions:

Makeswift requires a minimum Next.js version of 12.2.0. Please upgrade to Next.js version ^12.2.0 if you want to use Next.js v12, or version ^13.0.0 if you want to use Next.js v13.

packages/runtime/src/components/shared/CompatNextImage/CompatNextImage.ts
```diff
@@ -0,0 +1,11 @@
...
+
+let NextImage = CurrentNextImage
+
+try {
+  // If on Next.js v13, we're using next/legacy/image
+  // TODO: This will throw a warning on a console: Module not found: Can't resolve 'next/legacy/image'
+  NextImage = require('next/legacy/image')
+} catch {}
+
+export const CompatNextImage = NextImage
```
migueloller, line 11: Maybe we could check the Next.js version instead of using try/catch?

packages/runtime/src/runtimes/react/index.tsx
```diff
@@ -122,8 +123,10 @@ export function RuntimeProvider({
     async function setReactBuilderPreviewStore(): Promise<void> {
       const ReactBuilderPreview = await import('../../state/react-builder-preview')
 
-      setStore(store =>
-        ReactBuilderPreview.configureStore({ preloadedState: store.getState(), client }),
+      startTransition(() =>
+        setStore(store =>
+          ReactBuilderPreview.configureStore({ preloadedState: store.getState(), client }),
+        ),
```
migueloller, line 129: If this causes a large delay in the builder we might want to consider other approaches to using the builder store

migueloller (changes requested):
Maybe instead of vendoring the SemVer utilities we can add the `semver` package as a dependency?

packages/runtime/src/runtimes/react/index.tsx
migueloller: One approach would be to lazily load different providers. Something like this:

```tsx
const PreviewProvider = dynamic(() => import('./path/to/preview-provider'))
const LiveProvider = dynamic(() => import('./path/to/live-provider'))

export function Provider({ isPreview }) {
  return isPreview ? <PreviewProvider /> : <LiveProvider />
}
```

packages/makeswift-next-plugin/index.js
fikrikarim: Updated

fikrikarim:
Currently having an error on the initial render in the builder for this commit:
image

packages/runtime/src/components/shared/CompatNextImage/CompatNextImage.ts
fikrikarim: I tried checking the Next.js version and dynamic import using this code:
```ts
export const CompatNextImage = dynamic(() =>
  nextMajorVersion >= 13 ? import('next/legacy/image') : import('next/image'),
) as typeof NextImage
```
But it still gives this error on compile time:
```
wait  - compiling...
error - .yalc/@makeswift/runtime/dist/CompatNextImage.es.js:5:113
Module not found: Can't resolve 'next/legacy/image'
```
Like our problem with the dynamic import for the new pages sidebar, I think Next.js is resolving the import during compile time, so we still can't use this approach.

I'm not sure how to solve this other than using the `require` with a `try catch`. I tried to search on how to make the `import()` function not throw an error, but I couldn't find anything yet.

packages/runtime/src/runtimes/react/index.tsx
fikrikarim: Updated.

But currently having an error on the initial render in the builder for this commit:
image

Will work on this again tomorrow.

migueloller: That looks like initialization isn't being done properly—specifically the effect that registers the document with the store.

migueloller:
> > Maybe instead of vendoring the SemVer utilities we can add the semver package as a dependency?
> 
> Added the semver as a dependency.

Ok. Will this only be used for the Next.js plugin? I wouldn't want to add `semver` as a dependency of the runtime itself

fikrikarim:
> > > Maybe instead of vendoring the SemVer utilities we can add the semver package as a dependency?
> > 
> > 
> > Added the semver as a dependency.
> 
> Ok. Will this only be used for the Next.js plugin? I wouldn't want to add `semver` as a dependency of the runtime itself

Yes, only for the Next.js plugin and not for the runtime itself.

packages/runtime/src/components/shared/CompatNextImage/CompatNextImage.ts
fikrikarim: The difference is that we can put `require` in a `try catch`, and basically we catch the error thrown when we try to import `next/legacy/image`.

From what I'm seeing, the Next.js will resolve the `import` first, before even building/compiling the code. 

Yes, I'm also thinking of _adding a layer of indirection with an extra module_. Let me try it.

packages/runtime/src/runtimes/react/index.tsx
fikrikarim: Cool. I got it fixed!

packages/runtime/src/components/shared/CompatNextImage/CompatNextImage.ts
migueloller: But why would require work any differently than import? At the end of the day webpack needs to create the chunks and it will use the import specifier or module ID to do that. When Next.js builds the browser bundle I would imagine the require module ID is also resolved at build time. Or maybe that's not the case?

fikrikarim: I'm not sure what's happening. I just can't catch the error if we're using `import`, but it's working if we're using `require`.

migueloller: Well that's interesting. I guess we could look at the build output if we want to see what webpack's doing

migueloller: Hmm, ok. We should be careful using `require`, then. Can you take analyze the bundle using `@next/bundle-analyzer` and make sure that using `require` doesn't result in a problem like a bunch of JavaScript being included that wasn't before?

Also, we should verify that Rollup (via Vite) isn't what's attempting to resolve the `require` at build time when building `@makeswift/runtime`

fikrikarim: >Also, we should verify that Rollup (via Vite) isn't what's attempting to resolve the require at build time when building @makeswift/runtime

I don't think this is from the Rollup. This is the output from the Rollup:
```js
import dynamic from "next/dynamic";
const nextVersion = require("next/package.json").version;
const nextMajorVersion = parseInt(nextVersion.split(".")[0]);
const CompatNextImage = dynamic(() => nextMajorVersion >= 13 ? import("next/legacy/image") : import("next/image"));
export { CompatNextImage as C };
//# sourceMappingURL=CompatNextImage.es.js.map
```

I tested using `yarn build` on both Next.js v12 and Next.js v13 using the `@next/bundle-analyzer`, and there was no increase in size or `First Load JS`.

But, in Next.js v12, there were these warnings during the build step:
```
./.yalc/@makeswift/runtime/dist/CompatNextImage.es.js
Module not found: Can't resolve 'next/legacy/image' in '/Users/fikrikarim/Desktop/workspace/custom-host-sandbox/.yalc/@makeswift/runtime/dist'

Import trace for requested module:
./.yalc/@makeswift/runtime/dist/CompatNextImage.es.js
./.yalc/@makeswift/runtime/dist/Root.es.js
./.yalc/@makeswift/runtime/dist/index.es.js
./.yalc/@makeswift/runtime/dist/next.es.js
```

Sigh..

[Relevant issue from the Next.js repo](https://github.com/vercel/next.js/issues/18819)

packages/runtime/src/runtimes/react/index.tsx
```diff
@@ -79,61 +79,20 @@ export const ReactRuntime = createReactRuntime(storeContextDefaultValue)
 
 registerBuiltinComponents(ReactRuntime)
 
-const Context = createContext(storeContextDefaultValue)
+export const StoreContext = createContext(storeContextDefaultValue)
```
migueloller, line 82: This export being shared by the two other modules that we're attempting to code-split might mess with how webpack builds chunks

packages/runtime/src/runtimes/react/components/PreviewProvider.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+import { StoreContext, storeContextDefaultValue } from '..'
+import * as ReactBuilderPreview from '../../../state/react-builder-preview'
+import * as ReactPage from '../../../state/react-page'
+import { MakeswiftProvider, MakeswiftClient } from '../../../api/react'
+import { registerDocumentEffect } from '../../../state/actions'
+
+type Props = {
+  client: MakeswiftClient
+  rootElements?: Map<string, ReactPage.Element>
```
migueloller, line 11: Is `Element` not available via `ReactBuilderPreview`?

packages/runtime/src/state/react-builder-preview.ts
```diff
@@ -287,15 +286,13 @@ export function initialize(): ThunkAction<() => void, State, unknown, Action> {
     const stopHandlingFocusEvent = startHandlingFocusEvents()
     const unlockDocumentScroll = dispatch(lockDocumentScroll())
     const stopHandlingPointerMoveEvent = dispatch(startHandlingPointerMoveEvent())
-    dispatch(setIsInBuilder(true))
```
migueloller, line 290: If this state is read-only and doesn't change anymore we probably want to get rid of the actions as well

migueloller (changes requested):
Looking good so far!

packages/runtime/src/components/builtin/Image/Image.tsx
```diff
@@ -125,22 +125,20 @@ const ImageComponent = forwardRef(function Image(
...
+      {nextMajorVersion < 13 ? (
         <NextLegacyImage
           layout="responsive"
           src={imageSrc}
           sizes={imageSizes(width)}
           alt={altText}
           width={dimensions.width}
           height={dimensions.height}
+          unoptimized={isInBuilder}
         />
```
fikrikarim, line 142: The reason why I changed `isInBuilder` yesterday was because of an SSR hydration error because we're rendering just a `` instead of `NextImage`.

When I looked into the commit why we're not using `NextImage` in the builder, it just says:
> Avoid using next/image in the builder

Can we still use `NextImage`, if we're using [`unoptimized` prop](https://nextjs.org/docs/api-reference/next/image#unoptimized)? @migueloller

packages/runtime/src/runtimes/react/components/PreviewProvider.tsx
fikrikarim: Not available, I think it's fine to load `ReactPage` from `PreviewProvider`, as that's how it behaved previously.

packages/runtime/src/state/react-builder-preview.ts
fikrikarim: I dropped the commit that changes `isInBuilder`, but let's see whether we still need to add it or not. It depends on the thread down below.

packages/runtime/src/runtimes/react/index.tsx
fikrikarim: I tested this using the bundle analyzer, and the `PreviewProvider` is only loaded for the Preview page, and not for Live page.

The `ReactPage` is loaded for both Preview and Live page. This is also how it behaves previously.

packages/runtime/src/state/react-builder-preview.ts
fikrikarim: I added the commit again, somehow I cannot edit the text without that commit.

packages/runtime/src/state/react-builder-preview.ts
```diff
@@ -287,15 +286,13 @@ export function initialize(): ThunkAction<() => void, State, unknown, Action> {
...
 
     return () => {
       stopMeasuringElements()
       stopMeasuringDocumentElement()
       stopHandlingFocusEvent()
       unlockDocumentScroll()
       stopHandlingPointerMoveEvent()
-      dispatch(setIsInBuilder(false))
```
fikrikarim, line 298: With the isInBuilder change commit, Texts are editable on the preview page.

packages/runtime/src/state/react-builder-preview.ts
migueloller: Got it. Is the plan still to rename this to `isPreview` and rework the Text component so that it starts as readonly on changes to editable when it receives a message from the builder?

packages/runtime/src/components/builtin/Image/Image.tsx
migueloller: Yeah, that works

packages/runtime/src/components/shared/Link/index.tsx
```diff
@@ -124,10 +124,15 @@ export const Link = forwardRef<HTMLAnchorElement, Props>(function Link(
...
-        <a {...restOfProps} ref={ref} target={target} onClick={handleClick} />
-      </NextLink>
+      <NextLink
+        {...restOfProps}
+        ref={ref}
+        target={target}
+        onClick={handleClick}
+        href={href}
+        // Next.js v12 has legacyBehavior set to true by default
+        legacyBehavior={false}
```
migueloller, line 134: So it's defaulted to `true` in Next v12 and we're hardcoding it to `false`, won't it be using the new behavior with Next v13?

packages/runtime/src/runtimes/react/components/LiveProvider.tsx
```diff
@@ -0,0 +1,42 @@
...
+}
+
+export default function LiveProvider({ client, children, rootElements }: Props): JSX.Element {
+  const store = useMemo(
+    () =>
+      ReactPage.configureStore({
+        preloadedState: storeContextDefaultValue.getState(),
+        rootElements,
+      }),
+    [],
```
migueloller, line 21: This is missing the `rootElements` dependency

packages/runtime/src/runtimes/react/components/PreviewProvider.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+export default function PreviewProvider({ client, children, rootElements }: Props): JSX.Element {
+  const store = useMemo(
+    () =>
+      ReactBuilderPreview.configureStore({
+        preloadedState: storeContextDefaultValue.getState(),
+        rootElements,
+        client,
+      }),
+    [client],
```
migueloller, line 23: `rootElements` is missing in dependencies

packages/runtime/src/state/react-builder-preview.ts
```diff
@@ -545,15 +545,26 @@ function makeswiftApiClientSyncMiddleware(
...
+  rootElements?: Map<string, Documents.Element>
   preloadedState?: PreloadedState<State>
   client: MakeswiftClient
 }): Store {
+  const initialState = {
+    ...preloadedState,
+    documents: Documents.getInitialState({ rootElements }),
+  }
+
+  if (typeof window === 'undefined') {
```
migueloller, line 559: I feel like this logic should be colocated with the browser API dependencies. It feels weird to magically put this check here since there's nothing about the other middleware that indicate they require browser APIs. I think the better thing to do would be to check for this right before a browser-only API is used

packages/runtime/src/next/api-handler.ts
```diff
@@ -100,7 +100,7 @@ export function MakeswiftApiHandler(
           return res.status(400).json({ message: 'Bad Request' })
         }
 
-        const revalidate = res.revalidate ?? res.unstable_revalidate
+        const revalidate = res.revalidate
```
migueloller, line 103: This change should come with a changeset that clearly explains what's happening and communicates this as a breaking change.

.changeset/three-eels-shave.md
```diff
@@ -0,0 +1,5 @@
+---
+'@makeswift/runtime': minor
+---
+
+BREAKING CHANGE: The Image component will use the new `next/image` if the host is using Next.js v13.
```
migueloller, line 5: I think that there's an opportunity here to go into more detail as to what exactly users can expect to "break". At the very least a link to the Next.js documentation or blog post announcing the new Image. But ideally we're explaining what has changed, why, and what they can expect, etc.

packages/runtime/src/components/builtin/Image/Image.tsx
```diff
@@ -17,6 +18,9 @@ import { cx } from '@emotion/css'
 import { useStyle } from '../../../runtimes/react/use-style'
 import { responsiveStyle, responsiveWidth } from '../../utils/responsive-style'
 import { useFile } from '../../../runtimes/react/hooks/makeswift-api'
+import { nextMajorVersion } from '../../../next/nextVersion'
```
migueloller, line 21: How come the filename is not following the naming conventions of other files?

packages/runtime/src/components/builtin/Image/Image.tsx
```diff
@@ -1,5 +1,6 @@
 import { useState, useEffect, Ref, forwardRef } from 'react'
 import NextImage from 'next/image'
+import _NextLegacyImage from 'next/legacy/image'
```
migueloller, line 3: To avoid actually importing it and including it in the output you can explicitly do `import typeof NextLegacyImageType from 'next/legacy/image'`

packages/runtime/src/next/nextVersion.ts
```diff
@@ -0,0 +1,5 @@
+import nextPackageJSON from 'next/package.json'
```
migueloller, line 1: Is it possible to import just the version directly so it doesn't bundle the entire package.json?

```ts
import { version } from 'next/package.json'

const [major, minor, patch] = version.split('.')

export { major, minor patch }
```

Also, maybe we should replace the `split('.')` with a regular expression at some point. Maybe have something a bit more robust haha

packages/runtime/src/components/builtin/Image/Image.tsx
```diff
@@ -153,6 +151,7 @@ const ImageComponent = forwardRef(function Image(
             width: '100%',
             height: 'auto',
           }}
+          unoptimized={isInBuilder}
```
migueloller, line 154: Is there an `unoptimized` prop for `next/image` in Next.js v12?

packages/runtime/src/state/react-builder-preview.ts
migueloller: Hmm, the commit still seems to be there on this latest version

migueloller (changes requested):
- Should we also update Next.js, React, etc. in the peer dependencies of the package.json for that first commit?
- Any reason why we don't have a changeset for each commit? It's ok not to have one for each commit, but we should make sure that the final changelog has a very clear explanation of what changed, why, and how it will impact a user if they upgrade as well as migration instructions if relevant

packages/runtime/src/state/react-builder-preview.ts
fikrikarim: Not sure whether we want to rename it to `isPreview` in this PR.

For the Text component, I'm not sure how we can send the editable message from the builder reliably @migueloller . We have this code on the `RichTextPropController` in cosmos:
```ts
    constructor(api: PropControllerAPI) {
      super(api)

      // eslint-disable-next-line @typescript-eslint/no-this-alias
      const controller = this

      this.Panel = function ConnectedPanel(props) {
        const [prop, setProp] = useProp<RichTextValue>(api)

        useEffect(() => {
          let editMode = api.builder.getEditMode()

          return api.builder.subscribe(() => {
            const nextEditMode = api.builder.getEditMode()

            if (nextEditMode !== editMode) {
              editMode = nextEditMode

              controller.messageHostPropController({
                type: RichTextPropControllerMessageType.CHANGE_BUILDER_EDIT_MODE,
                editMode,
              })
            }
          })
        }, [])

        return <RichTextPanel {...props} controller={controller} value={prop} onChange={setProp} />
      }
    }
```

But because it's attached to the `Panel`, the code is only running when the `Panel` is mounted. This is a problem because for example, let's say we're already in content mode, and we want to edit another Text, that's not possible because the other Text wasn't editable because that Panel wasn't mounted yet.

I don't know where to attach the subscribe function to the controller other than on the Panel.

We could have a global store for content mode, but again, it's gonna be similar to `isInBuilder`.

migueloller: Hmm, ok. I think that we would do is subscribe when the control itself is instantiated. I guess we can take care of that another time.

So what will we be doing instead. Was there another way to solve the issue?

packages/runtime/src/components/shared/Link/index.tsx
fikrikarim: Oh, I was wrong yesterday when I said we're using the `legacyBehavior`. We remove the `<a>` tag inside the `next/link`, so we're using the new behavior now for both Next.js v12 and Next.js v13.

packages/runtime/src/runtimes/react/components/LiveProvider.tsx
fikrikarim: Added

packages/runtime/src/state/react-builder-preview.ts
fikrikarim: So the reason why I put it here is that even though they might not use browser-only API, all other middlewares are only needed in the browser; they are not needed for the SSR. 

I was hoping that we could reduce the amount of code needed for the SSR/page generation even by a small amount.

packages/runtime/src/next/api-handler.ts
fikrikarim: Added the changeset. We can review the final changeset/release notes later once we want to publish the new version.

.changeset/three-eels-shave.md
fikrikarim: Yes, it's hard to write the complete release notes on different commits. This is just a draft, and we can go through the release notes together when we want to release the version.

packages/runtime/src/components/builtin/Image/Image.tsx
fikrikarim: Fixed

packages/runtime/src/next/nextVersion.ts
fikrikarim: Cool, that's cleaner. I think using `split('.')` is fine. If we want to be robust then we should use the `semver` package 😄 .

fikrikarim:
> * Should we also update Next.js, React, etc. in the peer dependencies of the package.json for that first commit?

I think we should update the peer dependencies on the last commit, as we need all the previous commits before we can support the Next.js v13.

> * Any reason why we don't have a changeset for each commit? It's ok not to have one for each commit, but we should make sure that the final changelog has a very clear explanation of what changed, why, and how it will impact a user if they upgrade as well as migration instructions if relevant

Yeah, I don't think we need a changeset for a refactor or anything that doesn't change the external behavior. Yes, I agree, let's do the final changelog together.

packages/runtime/src/next/nextVersion.ts
fikrikarim: Related links: 
- https://github.com/webpack/webpack/issues/13469, 
- https://stackoverflow.com/questions/64993118/error-should-not-import-the-named-export-version-imported-as-version

packages/runtime/src/next/next-version.ts
```diff
@@ -0,0 +1,7 @@
+import nextPackageJSON from 'next/package.json'
+
+const { version } = nextPackageJSON
+
+const [major, minor, patch] = version.split('.').map(parseInt)
```
migueloller, line 5: CleanShot 2023-01-12 at 12 14 47@2x

Oof

packages/runtime/src/state/react-builder-preview.ts
```diff
@@ -337,6 +337,8 @@ export function messageChannelMiddleware(): Middleware<Dispatch, State, Dispatch
       const messageChannel = new MessageChannel()
       let cleanUp = () => {}
 
+      if (typeof window === 'undefined') return () => {}
```
migueloller, line 340: Why not `return cleanUp`?

migueloller (changes requested):

migueloller (approved):

PR #298, opened 2023-01-05
fix reveal animation lag (moving box animation from @framer/motion to css)

joshuawootonn:
There is a companion PR in www.makeswift.com repo for comparing perf with the actual site.

companion PR: https://github.com/makeswift/www.makeswift.com/pull/96
companion preview deploy: https://www-makeswift-com-git-josh-prd-910-makeswift.vercel.app/

### DIFF
--- a/.changeset/quick-cows-notice.md
+++ b/.changeset/quick-cows-notice.md
@@ -1,0 +1,5 @@
+---
+'@makeswift/runtime': patch
+---
+
+Swap @framer/motion box animation for a light CSS version.
--- a/packages/runtime/src/components/builtin/Box/Box.tsx
+++ b/packages/runtime/src/components/builtin/Box/Box.tsx
@@ -1,10 +1,10 @@
-import { forwardRef, Ref, useImperativeHandle, useRef, useState } from 'react'
-import { motion } from 'framer-motion'
+import { forwardRef, Ref, useEffect, useImperativeHandle, useRef, useState } from 'react'
 import { cx } from '@emotion/css'
+import { v4 as uuid } from 'uuid'
 
 import { Element } from '../../../runtimes/react'
 import Placeholder from './components/Placeholder'
-import { useBoxAnimations } from './animations'
+import { areBoxAnimationPropsEqual, BoxAnimationProps, useBoxAnimation } from './animations'
 import {
   ElementIDValue,
   ResponsiveIconRadioGroupValue,
@@ -12,13 +12,10 @@ import {
   GapYValue,
   GapXValue,
   CheckboxValue,
-  ResponsiveSelectValue,
-  ResponsiveNumberValue,
   BackgroundsValue,
 } from '../../../prop-controllers/descriptors'
 import { BoxModelHandle, parse, createBox } from '../../../box-model'
 import BackgroundsContainer from '../../shared/BackgroundsContainer'
-import { BoxAnimateIn } from './constants'
 import { responsiveStyle } from '../../utils/responsive-style'
 import { GridItem } from '../../shared/grid-item'
 import { useStyle } from '../../../runtimes/react/use-style'
@@ -38,16 +35,9 @@ type Props = {
   boxShadow?: string
   rowGap?: GapYValue
   columnGap?: GapXValue
-  boxAnimateType?: ResponsiveSelectValue<BoxAnimateIn>
-  boxAnimateDuration?: ResponsiveNumberValue
-  boxAnimateDelay?: ResponsiveNumberValue
-  itemAnimateType?: ResponsiveSelectValue<BoxAnimateIn>
-  itemAnimateDuration?: ResponsiveNumberValue
-  itemAnimateDelay?: ResponsiveNumberValue
-  itemStaggerDuration?: ResponsiveNumberValue
   hidePlaceholder?: CheckboxValue
   children?: GridValue
-}
+} & BoxAnimationProps
 
 const Box = forwardRef(function Box(
   {
@@ -77,8 +67,6 @@ const Box = forwardRef(function Box(
 ) {
   const innerRef = useRef<HTMLDivElement | null>(null)
   const [boxElement, setBoxElement] = useState<HTMLElement | null>(null)
-  const hasAnimations = boxAnimateType != null || itemAnimateType != null
-  const Grid = hasAnimations ? motion.div : 'div'
 
   useImperativeHandle(
     ref,
@@ -123,22 +111,38 @@ const Box = forwardRef(function Box(
     responsiveStyle([verticalAlign], ([alignItems = 'flex-start']) => ({ alignItems })),
   )
 
-  const { initial, animate, variants, transition, key } = useBoxAnimations({
+  const [animationClassName, replayAnimation] = useBoxAnimation(
+    boxElement,
     boxAnimateType,
     boxAnimateDuration,
     boxAnimateDelay,
-    itemAnimateDelay,
+    itemAnimateType,
+  )
+
+  const [key, setKey] = useState(() => uuid())
+
+  const animationProps = {
+    boxAnimateType,
+    boxAnimateDuration,
+    boxAnimateDelay,
     itemAnimateType,
     itemAnimateDuration,
+    itemAnimateDelay,
     itemStaggerDuration,
-    boxElement,
-    elements: children?.elements,
-  })
+  }
+
+  const prevAnimationProps = useRef(animationProps)
+  useEffect(() => {
+    if (!areBoxAnimationPropsEqual(prevAnimationProps.current, animationProps)) {
+      replayAnimation()
+      setKey(uuid())
+      prevAnimationProps.current = animationProps
+    }
+  }, [replayAnimation, animationProps])
 
   return (
     <BackgroundsContainer
       ref={setBoxElement}
-      hasAnimations={hasAnimations}
       id={id}
       className={cx(
         width,
@@ -146,16 +150,13 @@ const Box = forwardRef(function Box(
         borderRadius,
         useStyle({ display: 'flex' }),
         useStyle(responsiveStyle([height], ([alignSelf = 'auto']) => ({ alignSelf }))),
+        animationClassName,
       )}
       backgrounds={backgrounds}
-      animate={animate?.container}
-      initial={initial?.container}
-      variants={variants?.container}
-      transition={transition?.container}
-      key={key?.container}
     >
-      <Grid
+      <div
         ref={innerRef}
+        key={key}
         className={cx(
           padding,
           boxShadow,
@@ -165,23 +166,19 @@ const Box = forwardRef(function Box(
             responsiveStyle([verticalAlign], ([alignContent = 'flex-start']) => ({ alignContent })),
           ),
         )}
-        animate={animate?.parent}
-        initial={initial?.parent}
-        transition={transition?.parent}
       >
         {children && children.elements.length > 0 ? (
           children.elements.map((child, index) => (
             <GridItem
-              as={hasAnimations ? motion.div : 'div'}
               key={child.key}
               className={gridItemClassName}
               grid={children.columns}
               index={index}
+              itemAnimateDuration={itemAnimateDuration}
+              itemAnimateDelay={itemAnimateDelay}
+              itemStaggerDuration={itemStaggerDuration}
               columnGap={columnGap}
               rowGap={rowGap}
-              // @ts-ignore: `variants` is not a prop for `div`, but it is for `motion.div`.
-              variants={variants?.child}
-              transition={transition?.child}
             >
               <Element element={child} />
             </GridItem>
@@ -189,7 +186,7 @@ const Box = forwardRef(function Box(
         ) : (
           <Placeholder hide={hidePlaceholder} />
         )}
-      </Grid>
+      </div>
     </BackgroundsContainer>
   )
 })
--- a/packages/runtime/src/components/builtin/Box/animations.tsx
+++ b/packages/runtime/src/components/builtin/Box/animations.tsx
@@ -1,201 +1,21 @@
+import { CSSObject } from '@emotion/css'
+import { useState, useEffect, useCallback } from 'react'
 import {
-  TargetAndTransition,
-  Transition,
-  useAnimation,
-  useReducedMotion,
-  Variants,
-  MotionProps,
-} from 'framer-motion'
-import { TransitionMap } from 'framer-motion/types/types'
-import { useCallback, useEffect, useState } from 'react'
-
-import { ResponsiveValue } from '../../../prop-controllers'
+  ResponsiveNumberValue,
+  ResponsiveSelectValue,
+  ResponsiveValue,
+} from '../../../prop-controllers'
+import { useStyle } from '../../../runtimes/react/use-style'
 import { useMediaQuery } from '../../hooks'
-import { Element } from '../../../state/react-page'
+import { gridItemIdentifierClassName } from '../../shared/grid-item'
 import {
   BoxAnimateIn,
   DEFAULT_BOX_ANIMATE_DELAY,
   DEFAULT_BOX_ANIMATE_DURATION,
   DEFAULT_BOX_ANIMATE_TYPE,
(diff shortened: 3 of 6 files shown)


joshuawootonn:
CleanShot 2023-01-09 at 17 43 33@2x
pricing page diff

joshuawootonn:
CleanShot 2023-01-09 at 17 44 32@2x
product page diff (+4)

joshuawootonn:
# Best of three
### CSS
CleanShot 2023-01-10 at 15 46 12@2x
### Framer Motion
CleanShot 2023-01-10 at 15 48 03@2x

packages/runtime/src/components/builtin/Box/animations.tsx
```diff
@@ -1,365 +1,133 @@
...
+  const className = useStyle({
+    ...exitedBoxAnimationProperties[animationType],
+    transition: `all 0ms`,
+    '&.show': {
+      ...enteredBoxAnimationProperties[animationType],
+      transition: `transform ${actualDuration}ms cubic-bezier(0.16, 0.84, 0.44, 1) ${actualDelay}ms,filter ${actualDuration}ms cubic-bezier(0.16, 0.84, 0.44, 1) ${actualDelay}ms, opacity ${actualDuration}ms ease ${actualDelay}ms`,
+    },
+    '&:not(.show) > div > div': {
+      ...exitedBoxAnimationProperties[itemAnimationType],
+      transition: `all 0ms !important`,
```
joshuawootonn, line 96: this has to be important because the other css class is more specific
This css has to be specified here since it is based on a class on the root.
The item css has to be specified there since it needs the index to know how much stagger to add.

.changeset/quick-cows-notice.md
```diff
@@ -0,0 +1,5 @@
+---
+'@makeswift/runtime': patch
+---
+
+swap out framer motion for box animations
```
migueloller, line 5: Can we use a full sentence here? This will make its way to the changelog and release notes so we want to make this as informative as possible

packages/runtime/src/components/builtin/Box/animations.tsx
```diff
@@ -1,365 +1,133 @@
...
-    transition: Transition
-    variants: Variants
-  }
-}
+const QUERY = '(prefers-reduced-motion: no-preference)'
 
-const useElementOnScreen = (element: HTMLElement | null, options: IntersectionObserverInit) => {
-  const [isVisible, setIsVisible] = useState(false)
+function usePrefersReducedMotion() {
+  const [prefersReducedMotion, setPrefersReducedMotion] = useState(false)
```
migueloller, line 18: We should use `useSyncExternalStore` for this

packages/runtime/src/components/shared/BackgroundsContainer/index.tsx
```diff
@@ -10,31 +9,54 @@ import {
...
+    ...restOfProps
+  }: Props,
   ref: Ref<HTMLDivElement | null>,
 ) {
   const [handle, setHandle] = useState<HTMLDivElement | null>(null)
-  const Component = hasAnimations ? motion.div : 'div'
 
   useImperativeHandle(ref, () => handle, [handle])
 
+  const animationClassName = useBoxAnimation(
```
migueloller, line 51: Could we not have this animation logic live in the Box component and pass the class as a prop to this one? It seems like this component should be unaware of the animation functionality

packages/runtime/src/components/shared/grid-item.tsx
```diff
@@ -1,32 +1,45 @@
...
-  const Component = as ?? 'div'
+  const animationClassName = useItemAnimation(
+    itemAnimateDuration,
+    itemAnimateDelay,
+    itemStaggerDuration,
+    index,
+  )
 
-  return <Component {...restOfProps} className={cx(gridItemClassName, className)} />
+  return <div {...restOfProps} className={cx(gridItemClassName, className, animationClassName)} />
```
migueloller, line 44: Similar to the backgrounds container, couldn't all this animation logic stay in the Box component and come in to this component just as a class? Seems like we don't need to introduce new animation logic here

packages/runtime/src/components/shared/grid-item.tsx
```diff
@@ -1,32 +1,45 @@
...
+  itemAnimateDelay?: ResponsiveNumberValue
+  itemStaggerDuration?: ResponsiveNumberValue
 }
 
-type Props<T extends ElementType> = BaseProps<T> &
-  Omit<ComponentPropsWithoutRef<T>, keyof BaseProps<T>>
+type Props = BaseProps & Omit<ComponentPropsWithoutRef<'div'>, keyof BaseProps>
 
-export function GridItem<T extends ElementType = 'div'>({
-  as,
```
migueloller, line 20: How come we're getting rid of the `as` prop? That didn't seem related to animations specifically

packages/runtime/src/components/builtin/Box/animations.tsx
```diff
@@ -1,365 +1,133 @@
...
+  responsiveAnimationType: ResponsiveValue<BoxAnimateIn> | undefined,
+  responsiveDuration: ResponsiveValue<number> | undefined,
+  responisveDelay: ResponsiveValue<number> | undefined,
+  itemResponsiveAnimationType: ResponsiveValue<BoxAnimateIn> | undefined,
+) {
+  const reducedMotion = usePrefersReducedMotion()
+  const animationType = useMediaQuery(responsiveAnimationType) || DEFAULT_BOX_ANIMATE_TYPE
+  const itemAnimationType = useMediaQuery(itemResponsiveAnimationType) || DEFAULT_ITEM_ANIMATE_TYPE
+  const duration = useMediaQuery(responsiveDuration) || DEFAULT_BOX_ANIMATE_DURATION
+  const delay = useMediaQuery(responisveDelay) || DEFAULT_BOX_ANIMATE_DELAY
```
migueloller, line 83: Can we also make sure that `useMediaQuery` uses `useSyncExternalStore`?

packages/runtime/src/components/builtin/Box/animations.tsx
```diff
@@ -1,365 +1,133 @@
...
+  const reducedMotion = usePrefersReducedMotion()
+  const duration = useMediaQuery(responsiveDuration) || DEFAULT_BOX_ANIMATE_DURATION
+  const delay = useMediaQuery(responisveDelay) || DEFAULT_BOX_ANIMATE_DELAY
+  const stagger = useMediaQuery(responsiveStagger) || DEFAULT_ITEM_STAGGER_DURATION
+  const delayFromStagger = responsiveStagger == null || index == null ? 0 : stagger * index
+  const actualDelay = (delay + delayFromStagger) * 1000
+  const actualDuration = duration * 1000
+
+  const className = useStyle({
+    transition: `transform ${actualDuration}ms cubic-bezier(0.16, 0.84, 0.44, 1) ${actualDelay}ms,filter ${actualDuration}ms cubic-bezier(0.16, 0.84, 0.44, 1) ${actualDelay}ms, opacity ${actualDuration}ms ease ${actualDelay}ms !important`,
```
migueloller, line 125: How come we're using "important" here? The `cx` and the order of the classes should take care of specificity and if the component gets its transition property overwritten, that should be ok

packages/runtime/src/components/builtin/Box/Box.tsx
```diff
@@ -123,22 +129,61 @@ const Box = forwardRef(function Box(
...
+    [
+      boxAnimateType,
+      boxAnimateDuration,
+      boxAnimateDelay,
+      itemAnimateType,
+      itemAnimateDuration,
+      itemAnimateDelay,
+      itemStaggerDuration,
+    ],
+  )
```
migueloller, line 143: This seems like a very weird way to use `useMemo` since none of the dependencies are used in computing the memoized value. Did you explore other ways to get a key?

migueloller (changes requested):
I love how much simpler the animation logic is now!

packages/runtime/src/components/shared/grid-item.tsx
joshuawootonn: I assumed that the as prop / the customizability of the GridItem and the BackgroundContainer were introduced so that we could use `motion.div : 'div'`. Just trying to leave it better than I found it :)

packages/runtime/src/components/builtin/Box/animations.tsx
joshuawootonn: You are right. This one slipped through. I'll double check both `!important` uses because I might have made the other one because I forgot I had added this one 😅

packages/runtime/src/components/builtin/Box/Box.tsx
joshuawootonn: Yeah I did. I don't know of a better way of watching props as a user changes them. and rerendering the children.

joshuawootonn:
> A comment on the Lighthouse reports. These aren't very helpful because the differences are so small as to not make them that meaningful. These scores vary wildly when ran so having a benchmark that stress tests the specific thing we're trying to optimize is likely the best way for us to measure the impact. 
I a perfect world I do agree with you. Are you asking me to make a benchmark for stress testing? 

> Also, I'm noticing a warning in the screenshots, were these tests run in Incognito Mode to avoid extensions and other browser features messing with the results?
I have 1pass enabled in incognito.

packages/runtime/src/components/shared/grid-item.tsx
migueloller: Cool, that makes sense. I'm guessing you checked to make sure that no other component using `GridItem` relied on that prop

packages/runtime/src/components/builtin/Box/Box.tsx
migueloller: How about being explicit about the different in the props and calling `setState` directly in render? This is fine and is something I've seen used in the React docs

migueloller:
Something I forgot to ask... can we remove `framer-motion` from the dependencies now?

joshuawootonn:
> Something I forgot to ask... can we remove framer-motion from the dependencies now?

No it is used in our carousel. I wish :P

migueloller:
> > Something I forgot to ask... can we remove framer-motion from the dependencies now?
> 
> No it is used in our carousel. I wish :P

Oof, ok. Next goal, make Carousel CSS only haha

joshuawootonn:
> Well, it's a matter of expectations. By posting screenshots of Lighthouse scores I'm assuming you think those are worthwhile and important but the ones posted don't really say anything. A delta of 1 or 4 can just be noise. So if you want to communicate using Lighthouse metrics you kind of have to do it right.

How else should I have communicated?

migueloller:
> > Well, it's a matter of expectations. By posting screenshots of Lighthouse scores I'm assuming you think those are worthwhile and important but the ones posted don't really say anything. A delta of 1 or 4 can just be noise. So if you want to communicate using Lighthouse metrics you kind of have to do it right.
> 
> How else should I have communicated?

I think posting Lighthouse scores is great! I'm just responding to your comment

>  In a perfect world I do agree with you. Are you asking me to make a benchmark for stress testing?

I don't think it was essential to have Lighthouse scores dropped in the PR, but if you're going to use them as a communication tool, you gotta make sure it does that job well. And what I was trying to say is that it's very hard to see that without a benchmark, or a page that's just showing a 1 or 4 delta since when it comes to Lighthouse that's basically just noise, so it really doesn't say much

packages/runtime/src/components/builtin/Box/Box.tsx
joshuawootonn: I think I am forgoing this solution. In my testing the `setState` in render doesn't cause the `useEffect(() => {}, [key])` to rerun.

packages/runtime/src/components/hooks/useMediaQuery.ts
```diff
@@ -1,33 +1,35 @@
-/* eslint-env browser */
-
-import { useState } from 'react'
-
+import { useSyncExternalStore } from 'react'
```
joshuawootonn, line 1: If you have comments here @migueloller that aren't trivial can we chat through this sync?

packages/runtime/src/components/builtin/Box/Box.tsx
migueloller: That sounds very weird to me and possibly a fluke where something else could be causing that (i.e., the `key` not changing). But if you feel like `useMemo` is the way to go, I'm not going to push for something else

packages/runtime/src/components/hooks/useMediaQuery.ts
migueloller: The import should come from `use-sync-external-store/shim` since we need to continue to support React v17

packages/runtime/src/components/hooks/useMediaQuery.ts
```diff
@@ -1,33 +1,35 @@
...
+        return curr.id
       }
+      return matchedDevice
+    }, DEVICE_QUERIES[0].id)
+    return findDeviceOverride(responsiveValue, deviceId)?.value
+  }
 
-      listener()
-      mediaQueryList.addListener(listener)
+  function subscribe(notify: () => void) {
```
migueloller, line 23: You want the identity of the subscribe function to be stable across renders

migueloller (changes requested):

packages/runtime/src/components/builtin/Box/Box.tsx
joshuawootonn: Actually I could just put. this into the render too 

```
  useEffect(() => {
    if (hasIntersected.current === false) return

    boxElement?.classList.remove('show')
    const timeout = setTimeout(() => boxElement?.classList.add('show'), 100)

    return () => clearTimeout(timeout)
  }, [key])
  ```

File: packages/runtime/src/components/hooks/useMediaQuery.ts (hunk 11 above)
Comment (joshuawootonn) [reply packages/runtime/src/components/hooks/useMediaQuery.ts:23]: Can you explain more

File: packages/runtime/src/components/hooks/useMediaQuery.ts [hunk 12]
```diff
@@ -1,33 +1,35 @@
...
 
-      const listener = () => {
-        if (mediaQueryList.matches) setDeviceId(device.id)
+  function getSnapshot() {
+    const deviceId: string = DEVICE_QUERIES.reduce((matchedDevice, curr) => {
+      if (window.matchMedia(curr.query).matches) {
+        return curr.id
       }
+      return matchedDevice
+    }, DEVICE_QUERIES[0].id)
```
migueloller, line 19: I find this use of `reduce` quite confusing

packages/runtime/src/components/hooks/useMediaQuery.ts
migueloller: https://beta.reactjs.org/reference/react/useSyncExternalStore#my-subscribe-function-gets-called-after-every-re-render

packages/runtime/src/components/builtin/Box/Box.tsx
joshuawootonn: Actually that doesn't work either. I think this is why doing stuff in a ref callback is an anti pattern :/

migueloller: Why would it be an anti-pattern?

packages/runtime/src/components/builtin/Box/Box.tsx
```diff
@@ -123,39 +121,52 @@ const Box = forwardRef(function Box(
...
+    itemAnimateDelay,
     itemStaggerDuration,
-    boxElement,
-    elements: children?.elements,
-  })
+  }
+
+  const prevAnimationProps = useRef(animationProps)
+  useEffect(() => {
+    if (JSON.stringify(prevAnimationProps.current) !== JSON.stringify(animationProps)) {
```
migueloller, line 146: Let's use a different way to compare. I think this might hurt performance a bit

migueloller, line 146: Let's use a different way to compare. I think this might hurt performance a bit

migueloller (changes requested):

migueloller (approved):

PR #301, opened 2023-01-07
Slimmer rich text, bundle size improvements, and more efficient runtime initialization

migueloller:
Fixes PRD-747 and PRD-901.

### DIFF
--- a/.changeset/tender-onions-leave.md
+++ b/.changeset/tender-onions-leave.md
@@ -1,0 +1,7 @@
+---
+'@makeswift/runtime': minor
+---
+
+Adds a slimmer rich text component for live pages so that large dependencies like Slate and Immutable aren't included in bundles for live pages and are only used in the Makeswift builder. This reduces the overhead of the Makeswift runtime for live pages and boosts performance.
+
+While behavior is intended to be the same, these changes modify the structure of the DOM for live pages, which could cause issues with existing sites if they're relying on the DOM structure of the Text component. For this reason we're releasing this in a minor update as a _breaking change_.
--- a/packages/runtime/src/components/builtin/Box/Box.tsx
+++ b/packages/runtime/src/components/builtin/Box/Box.tsx
@@ -1,4 +1,12 @@
-import { forwardRef, Ref, useEffect, useImperativeHandle, useRef, useState } from 'react'
+import {
+  forwardRef,
+  Ref,
+  useCallback,
+  useEffect,
+  useImperativeHandle,
+  useRef,
+  useState,
+} from 'react'
 import { cx } from '@emotion/css'
 import { v4 as uuid } from 'uuid'
 
@@ -66,7 +74,18 @@ const Box = forwardRef(function Box(
   ref: Ref<BoxModelHandle>,
 ) {
   const innerRef = useRef<HTMLDivElement | null>(null)
-  const [boxElement, setBoxElement] = useState<HTMLElement | null>(null)
+  const boxElementObjectRef = useRef<HTMLElement | null>(null)
+  const [animationClassName, replayAnimation, setElement] = useBoxAnimation(
+    boxAnimateType,
+    boxAnimateDuration,
+    boxAnimateDelay,
+    itemAnimateType,
+  )
+  const boxElementCallbackRef = useCallback((current: HTMLElement | null) => {
+    boxElementObjectRef.current = current
+
+    setElement(current)
+  }, [])
 
   useImperativeHandle(
     ref,
@@ -74,7 +93,7 @@ const Box = forwardRef(function Box(
       getBoxModel() {
         const paddingBoxElement = innerRef.current
         const borderBoxElement = innerRef.current
-        const marginBoxElement = boxElement
+        const marginBoxElement = boxElementObjectRef.current
         const borderBox = innerRef.current?.getBoundingClientRect()
         const paddingBoxComputedStyle =
           paddingBoxElement?.ownerDocument.defaultView?.getComputedStyle(paddingBoxElement)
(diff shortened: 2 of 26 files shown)


packages/runtime/src/components/builtin/Text/ReadOnlyText.tsx
```diff
@@ -0,0 +1,171 @@
...
+  ref: ForwardedRef<HTMLDivElement>,
+) {
+  const plaintext = text?.document == null ? '' : getNodeText(text.document)
+
+  return (
+    <div ref={ref} id={id} className={cx(width, margin)}>
+      {plaintext === '' ? <Placeholder /> : <Node {...text?.document} />}
+    </div>
+  )
+})
```
fikrikarim, line 32: The reader component is simpler than I expected. Did you copy-paste the rendering logic/markup from Slate or is this written from scratch?

I'm just worried that there will be a discrepancy between the reader component and the editable component.

migueloller: I wrote it from scratch. There _is_ a discrepancy. I purposefully kept the DOM structure a bit simpler. I looked at the Slate source code to attempt to make it as close as possible, but honestly, Slate's markup is trash and I want us to handcraft it moving forward, specially once we upgrade Slate and we have more control over that.

As for the difference in markup, I mention this in the changelog and it's also why I marked the change as a breaking one. People shouldn't be relying on the DOM structure of our Text component anyway

fikrikarim: Okay, nice. The code looks clean so should be easy if want to update it. 👍

.changeset/tender-onions-leave.md
```diff
@@ -0,0 +1,7 @@
+---
+'@makeswift/runtime': minor
+---
+
+Adds a slimmer rich text component for live pages so that large dependencies like Slate and Immutable aren't included in bundles for live pages and are only used in the Makeswift builder. This reduces the overhead of the Makeswift runtime for live pages and boosts performance.
+
+While behavior is intended to be the same, these changes modify the structure of the DOM for live pages, which could cause issues with existing sites if they're relying on the DOM structure of the Text component. For this reason we're releasing this in a minor update as a _breaking change_.
```
joshuawootonn, line 7: Is there anything that should clue a user into this being a breaking change other than the fact that we are calling it out? 

I am trying to understand why we are releasing a breaking change as a minor version as opposed to a major version. If you have a link to this please share.

packages/runtime/src/runtimes/react/index.tsx
```diff
@@ -158,6 +158,10 @@ export function useIsInBuilder(): boolean {
   return useSelector(state => ReactPage.getIsInBuilder(state))
```
joshuawootonn, line 160: I don't totally understand what this class represents. We should talk through it in pairing today

packages/runtime/src/components/builtin/Box/Box.tsx
```diff
@@ -66,15 +74,26 @@ const Box = forwardRef(function Box(
   ref: Ref<BoxModelHandle>,
 ) {
   const innerRef = useRef<HTMLDivElement | null>(null)
-  const [boxElement, setBoxElement] = useState<HTMLElement | null>(null)
+  const boxElementObjectRef = useRef<HTMLElement | null>(null)
+  const [animationClassName, replayAnimation, setElement] = useBoxAnimation(
```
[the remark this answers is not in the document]
migueloller: Yeah, that's right. The reference to the element doesn't affect the DOM (i.e., React JSX) so there's really no need to put it in state, so we use a ref. This avoids an extra render happening when this component mounts due to the HTML element being set in state

packages/runtime/src/runtimes/react/index.tsx
```diff
@@ -266,31 +262,46 @@ type ElementProps = {
...
     const dispatch = useDispatch()
     const documentKey = useDocumentKey()
-    const [handle, setHandle] = useState<unknown>(null)
+    const imperativeHandleRef = useRef(new ElementImperativeHandle())
+    const findDomNodeCallbackRef = useCallback((current: (() => Element | Text | null) | null) => {
+      imperativeHandleRef.current.callback(() => current?.() ?? null)
+    }, [])
+    const elementCallbackRef = useCallback((current: unknown | null) => {
+      imperativeHandleRef.current.callback(() => current)
+    }, [])
```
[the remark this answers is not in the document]
migueloller: This is a good question. If the component doesn't forward the ref, then `elementCallbackRef` will never be called. In that case, we rely in `findDOMNode` as a best-effort to get a measurable element so we can place our telegraphs. If the component _does_ forward the ref, then `elementCallbackRef` will be called, but _after_ `findDomNodeCallbackRef` because that component is higher up in the React element tree. So I'm relying on that React behavior here, otherwise they _would_ compete with each other.

I could avoid relying on React's behavior here and instead make it explicit by only using `findDOMNode` is the ref isn't forwarded. I wasn't sure that was possible to detect, but given my explanation above it should be doable if `elementCallbackRef` is called at all.

Will push up a commit to address this 🔜

.changeset/tender-onions-leave.md
migueloller: It's general practice in SemVer, that before a 1.0.0 release minor changes are breaking changes and patch changes are more or less like minor changes. For example, the `^` caret range specifier respects this de-facto standard by treating pre-1.0.0 versions differently. See [this](https://stackoverflow.com/a/22345808) StackOverflow answer. In essence, `0.6.0` is _not_ compatible with `^0.5.0`, but `1.6.0` _is_ compatible with `^1.5.0`

packages/runtime/src/runtimes/react/index.tsx
migueloller: Which class? If you mean `ReactPage`, that's not a class, it's just a module that has a bunch of named exports and I'm importing with that name like so:

```typescript
import * as ReactPage from './path/to/react-page.ts
```

.changeset/tender-onions-leave.md
joshuawootonn: 😮

packages/runtime/src/runtimes/react/index.tsx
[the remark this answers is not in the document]
joshuawootonn: Cool. Makes sense.

joshuawootonn (dismissed):
LGTM. Thanks for the answers!

joshuawootonn (approved):

PR #302, opened 2023-01-09
fix: falsy check on Style control utils

migueloller:
We did a falsy check instead of checking for `undefined` or `null` explicitly, causing `0` values to be ignored for margin, padding, border, and border-radius, due to `0` being falsy.


joshuawootonn (approved):
LGTM

PR #315, opened 2023-01-24
chore: move `MakeswiftComponentType` to main.js

fikrikarim:
We need MakeswiftComponentType on the builder, but we don't want to import it from `@makeswift/runtime/components` because we only need the constant, and not other things that's included in the components entrypoint.


migueloller (approved):
Looks good! Did you verify the bundle so that the main entry point didn't grow too much?

PR #316, opened 2023-01-24
fix: snippets don't run on client-side navigation

fikrikarim:
Because we're using a useState for snippets in Page, when we change the document/page, the Page component is still preserving the old state, which was for the old document/ page.

We add key to reset the snippets state, when the document/page is changed.

https://beta.reactjs.org/learn/preserving-and-resetting-state

https://user-images.githubusercontent.com/13066728/214278992-4affaa87-7b09-45f5-a66f-2c129d3f329b.mp4

### DIFF
--- a/.changeset/quick-carpets-fly.md
+++ b/.changeset/quick-carpets-fly.md
@@ -1,0 +1,5 @@
+---
+'@makeswift/runtime': patch
+---
+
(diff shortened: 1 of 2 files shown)


migueloller (approved):
Looks good!

PR #321, opened 2023-01-26
wip: hover telegraph is working

joshuawootonn:
The point of this review is for us to sync on my progress so far :) 

things still broken
* there are a bunch of panels showing for the different richtextcontrols
* entering content mode doesn't focus the correct rich text control

https://share.cleanshot.com/Xx51dDZM8vwXvjJstWCt


packages/runtime/src/components/builtin/Text/Text.tsx
```diff
@@ -47,11 +48,21 @@ const Text = forwardRef(function Text(
...
   const [editor, setEditor] = useState<Editor | null>(null)
   const [propControllers, setPropControllers] =
     useState<DescriptorsPropControllers<Descriptors> | null>(null)
   const controller = propControllers?.text
 
+  useEffect(() => {
+    if (element == null || controller == null) return
+
+    return pollBoxModel({
+      element,
```
joshuawootonn, line 61: this has a red line when I attempt to poll the box model on the `editor`. need to find a way of accessing the dom node of the slate editor

packages/runtime/src/runtimes/react/controls/slot.tsx
```diff
@@ -166,7 +166,7 @@ function SlotPlaceholder({ control }: SlotPlaceholderProps): JSX.Element {
   )
 }
 
-function pollBoxModel({
+export function pollBoxModel({
```
joshuawootonn, line 169: This is not what I will be doing in the final version. This PR is simply here for use to start the conversation.

packages/runtime/src/components/builtin/Text/Text.tsx
```diff
@@ -47,11 +48,21 @@ const Text = forwardRef(function Text(
...
+      onBoxModelChange: boxModel => controller.changeBoxModel(boxModel),
```
joshuawootonn, line 62: useFindDomNode on the actual editor

packages/runtime/src/runtimes/react/controls/rich-text.tsx
```diff
@@ -0,0 +1,206 @@
...
+  const [editor, setEditor] = useState<Editor | null>(null)
+
+  useEffect(() => {
+    if (element == null || control == null) return
+
+    return pollBoxModel({
+      element,
+      onBoxModelChange: boxModel => control.changeBoxModel(boxModel),
+    })
+  }, [element, control])
```
joshuawootonn, line 46: This could be a `register` and then the polling could instead be in the controller itself. Just food for thought. Both are fine. 

The question you need to ask yourself is "What do I want the API of the RichTextControl to look like?"

Passing an element or passing a boxmodel.

packages/runtime/src/controls/rich-text.ts
```diff
@@ -0,0 +1,151 @@
...
+
+  changeBoxModel(boxModel: BoxModel | null): void {
+    this.send({ type: RichTextControlMessageType.BOX_MODEL_CHANGE, payload: { boxModel } })
+  }
+}
+
+// todo:josh move all the rich text stuff from /prop-controllers/instances here and rename it to control rahter than prop-control
+
+// todo:josh replace this with something similar to copy in `/prop-controllers/copy/rich-text.ts`
+// export function copySlotData(
```
joshuawootonn, line 141: This should be similar or the same as the Text copy method

packages/runtime/src/runtimes/react/controls.tsx
```diff
@@ -218,6 +221,19 @@ export function PropsValue({ element, children }: PropsValueProps): JSX.Element
...
             )
 
+          case RichTextControlType:
+            const control = (propControllers?.[propName] ?? null) as RichTextControl | null
+            console.log({ control, propName })
+            return (
+              <RenderHook
+                key={descriptor.type}
+                hook={useRichText}
+                parameters={[props[propName], control, descriptor]}
```
joshuawootonn, line 231: Eventually, it would be great to have the descriptor in the control.

joshuawootonn: This si fine for now

PR #322, opened 2023-01-27
Add support for interaction mode

fikrikarim:


packages/runtime/src/state/react-builder-preview.ts
```diff
@@ -237,27 +237,33 @@ function startHandlingPointerMoveEvent(): ThunkAction<() => void, State, unknown
...
-      window.parent.focus()
+    function handleFocusOut(event: FocusEvent) {
+      if (ReactPage.getEditMode(getState()) === EditMode.EditModes.INTERACT) return
+
+      if (
+        !(event.relatedTarget instanceof window.HTMLElement) ||
+        !event.relatedTarget.isContentEditable
+      ) {
+        window.parent.focus()
+      }
```
fikrikarim, line 266: Right now, the reason we can capture the blur when in content mode when pressing `esc` is because we're attaching `esc` listener in the Text component:
```ts
if (isHotkey('escape')(event)) {
  lastController.current?.blur()

  return true
}
```

When the Text component blurs, this event listener will catch the `focusout`, and then focus the `window.parent`.

We cannot do that to any other components in interaction mode without attaching the `esc` event listener. Pressing `esc` won't trigger blur of any HTML element, so we cannot capture the `focusout` event.

packages/runtime/src/components/builtin/Text/EditableText.tsx
```diff
@@ -171,7 +174,22 @@ const EditableText = forwardRef(function EditableText(
     if (event.relatedTarget == null) editor.select(selection)
   }, [])
 
-  const isInBuilder = useIsInBuilder()
+  // If the user selects any text when they are in interaction mode or content mode,
+  // deselect any text when they go to the build mode.
+  useEffect(() => {
+    if (editMode === EditModes.BUILD) editor?.deselect().blur()
+  }, [editMode, editor])
```
migueloller, line 181: I don't think that we should have a code-path for this, and instead, switching between modes should blur the entire document. This is also related to how we handle exiting interaction mode to go back to build mode

packages/runtime/src/components/builtin/Text/EditableText.tsx
```diff
@@ -171,7 +174,22 @@ const EditableText = forwardRef(function EditableText(
...
+  // HACK: Slate holds on to the very first DOM event handlers passed in and doesn't update them
+  // even if they change.
+  const lastEditMode = useRef(editMode)
+  if (lastEditMode.current !== editMode) lastEditMode.current = editMode
+
+  const handleClick = (event: ReactMouseEvent, _editor: Editor, next: () => any) => {
+    next()
+
+    if (lastEditMode.current === EditModes.CONTENT) event.stopPropagation()
+  }
```
migueloller, line 192: This is going to cause problems with Slate when we attempt to extend our rich text control to support rich inline plugins. Also, this seems to be very specific to the "click" event when it seems this should be handled more generically, like when the contenteditable is focused? Are there other interactions like this one where we want to prevent stuff like this?

I don't have a better solution for this right now so maybe we just keep this, but we should add a comment as to why we have this for our future selves.

packages/runtime/src/next/preview-mode.tsx
```diff
@@ -54,6 +54,16 @@ if (originalUrl.searchParams.has(searchParamName)) {
...
+
+// Makeswift host connection check
+if (window.parent !== window) {
+  setInterval(() => {
+    window.parent.postMessage({ 
+      type: 'makeswift_connection_check',
+      payload: { currentUrl: window.location.href } 
+    }, appOrigin)
+  }, 20)
+}
```
migueloller, line 66: I feel like this should be part of a separate script, not the Preview Mode script, since it's not related to Preview Mode. Also, ideally we only send the heartbeat if we've established a connection instead of just sending a message every 20ms

packages/runtime/src/state/modules/edit-mode.ts
```diff
@@ -0,0 +1,25 @@
+import { Action, ActionTypes } from '../actions'
+
+export const EditModes = {
+  BUILD: 'build',
+  CONTENT: 'content',
+  INTERACT: 'interact',
+} as const
```
migueloller, line 7: This is a minor nit, but I'm mentioning it because it used to be `BuilderEditMode` before and it was changed, but for enums we want to keep it singular, so `EditMode`. We want the type and the value to be the same so you don't have to import `EditModes` and `EditMode`

packages/runtime/src/state/modules/edit-mode.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export type EditMode = typeof EditModes[keyof typeof EditModes] | null
```
migueloller, line 9: `null` shouldn't be a part of this type, it seems to be a part of the state not the underlying enum

packages/runtime/src/state/react-builder-preview.ts
migueloller: I would prefer if we didn't hard-code these assumptions into our components like this. We should discuss how we want to handle entering and exiting interaction mode without needing to add an `esc` listener

packages/runtime/src/state/modules/edit-mode.ts
migueloller: Also, why did you drop `Builder` from the name? I'm trying to be as explicit as possible with naming in the runtime so that it's clear what's builder-specific vs what's part of the host

migueloller (changes requested):
@fikrikarim, I think that we can handle exiting interaction mode without any keyboard listeners in a way that will have less assumptions and should work with most components people throw at us

packages/runtime/src/components/builtin/Text/EditableText.tsx
migueloller: Yeah, something like this makes sense. We can make a middleware that handles focus and selection

packages/runtime/src/next/preview-mode.tsx
migueloller: Yeah, that makes sense. I think I would like to keep the scripts separate internally for the purposes of code organization and then we can have a single script the API consumers use. That being said, it's ok if you want to leave it in there for now. We can come back later and split it up.

As for how we initialize, we can do something similar to Preview Mode, yes. You could even put the heartbeat inside the Preview Mode block if you want so we don't need two separate messages from the host (or we could have two listeners for that message)

migueloller (changes requested):

packages/runtime/src/state/react-builder-preview.ts
```diff
@@ -283,13 +297,34 @@ function startMeasuringDocumentElement(): ThunkAction<() => void, unknown, unkno
...
+    return () => {
+      window.document.body.removeEventListener('keydown', handle)
+    }
+
+    function handle(event: KeyboardEvent) {
+      if (ReactPage.getBuilderEditMode(getState()) !== BuilderEditMode.BuilderEditMode.INTERACT) {
+        return
+      }
+
+      if (isHotkey('escape')(event)) {
```
migueloller, line 313: We might also want to check if the default behavior has been prevented, and if so, not go back to build mode.

Something else to consider, instead of focusing the parent, do we want to send a message to the builder so that it switches to build mode immediately?

packages/runtime/src/state/react-builder-preview.ts
migueloller: Looks like we need a keyboard listener if we want to make this work. The reason focus worked before is that for content mode and `contenteditable`, hitting <kbd>Esc</kbd> does fire a `focusout` event, but with other focusable elements, like buttons, hitting <kbd>Esc</kbd> does not trigger the `focusout` event.

packages/runtime/src/components/builtin/Text/EditableText.tsx
```diff
@@ -171,7 +174,18 @@ const EditableText = forwardRef(function EditableText(
...
+  // HACK: Slate holds on to the very first DOM event handlers passed in and doesn't update them
+  // even if they change.
+  const lastEditMode = useRef(editMode)
+  if (lastEditMode.current !== editMode) lastEditMode.current = editMode
+
+  const handleClick = (event: ReactMouseEvent, _editor: Editor, next: () => any) => {
+    next()
+
+    // This is needed to prevent clicks from propagating in content mode
+    // https://linear.app/makeswift/issue/PRD-525/need-to-prevent-clicks-from-propagating-in-content-mode
```
migueloller, line 186: For an OSS repo I think we should try and keep comments to things that are publicly accessible in case somebody wants to contribute

migueloller (approved):
LGTM! 👍🏻🚢

PR #324, opened 2023-01-31
fix: input not rendered on a single select column

fikrikarim:
It was because we were not passing the children prop on the Container

### DIFF
--- a/.changeset/flat-squids-search.md
+++ b/.changeset/flat-squids-search.md
@@ -1,0 +1,5 @@
+---
+'@makeswift/runtime': patch
+---
+
+Fix the single select column not rendering radio buttons on the built-in Form component.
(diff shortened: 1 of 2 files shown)


packages/runtime/src/components/builtin/Form/components/Field/components/RadioButton/index.tsx
```diff
@@ -22,9 +22,10 @@ type ContainerBaseProps = Pick<Value, 'size'>
 type ContainerProps = ContainerBaseProps &
   Omit<ComponentPropsWithoutRef<'div'>, keyof ContainerBaseProps>
 
-function Container({ size }: ContainerProps) {
+function Container({ size, ...restOfProps }: ContainerProps) {
   return (
     <div
+      {...restOfProps}
```
migueloller, line 28: If we're gonna support all props and spread them instead of just adding `children`, we should also compose the class name prop

migueloller (changes requested):

.changeset/flat-squids-search.md
```diff
@@ -0,0 +1,5 @@
+---
+'@makeswift/runtime': patch
+---
+
+Fix radio buttons not rendered on a single select column.
```
migueloller, line 5: Can you be more descriptive for nee Makeswift users and specify this is for the built-in Form component?

migueloller (changes requested):

.changeset/flat-squids-search.md
fikrikarim: Cool, updated!

migueloller (approved):

PR #330, opened 2023-01-31
fix: SSR hydration mismatch in Image component

migueloller:


fikrikarim:
Whoops, Thanks! Somehow I didn't run into this SSR hydration during testing.

PR #333, opened 2023-02-01
Decouple content mode from element layout + RichText control

joshuawootonn:
Milestone 1: Decouple content mode from element layout + RichText control

https://user-images.githubusercontent.com/20950876/216170600-6358cb60-c591-4e67-b56a-b1b92f9ba0e2.mp4


.changeset/popular-roses-deliver.md
```diff
@@ -0,0 +1,5 @@
+---
+'@makeswift/runtime': patch
```
joshuawootonn, line 2: Should we release this? I think it is better for it to have a release entry than for it to be released under other work.

packages/runtime/src/runtimes/react/pollBoxModel.tsx
```diff
@@ -0,0 +1,32 @@
+import { BoxModel, getBox } from '../../state/modules/box-models'
```
joshuawootonn, line 1: open to suggestions as to where this should go 🤷🏼 

We don't really have a folder for utils like this that aren't hooks.

packages/runtime/src/prop-controllers/instances.ts
```diff
@@ -81,12 +96,11 @@ class RichTextPropController extends PropController<RichTextPropControllerMessag
...
-          case BuilderEditMode.CONTENT:
-            this.editor?.focus().moveToRangeOfDocument()
-            break
         }
-
+        break
+      }
+      case RichTextPropControllerMessageType.FOCUS: {
+        this.editor?.focus().moveToRangeOfDocument()
         break
```
joshuawootonn, line 104: Open to pushback on this change. 

I could also update the `BuilderEditMode.CONTENT` to pass a `elementkey` + `propname`. 

That felt like the implementation creeping across the API, so I did this instead.

.changeset/popular-roses-deliver.md
migueloller: I don't see why not! There's definitely value in being able to use a rich text control (or multiple) even if the panels appear in the palette in a non-optimal way (i.e., no sub-selection yet). Alan thinks it's fine to release it like this as well while we work on sub-selection, etc.

packages/runtime/src/runtimes/react/pollBoxModel.tsx
migueloller: We might already have some code that does something like this so we could re-use it. If it makes sense for it to be in a shared file then we can figure out some structure for it

packages/runtime/src/prop-controllers/instances.ts
migueloller: Is this behavior different than the current rich text prop controller? If so, why does the behavior need to be different?

.changeset/popular-roses-deliver.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Add RichText control with naive selection
```
migueloller, line 5: I think being a tad more descriptive would be helpful. It's not clear to me what "naive selection" is

packages/runtime/src/components/builtin/Text/EditableText.tsx
```diff
@@ -53,6 +54,15 @@ const EditableText = forwardRef(function EditableText(
     useState<DescriptorsPropControllers<Descriptors> | null>(null)
   const controller = propControllers?.text
 
+  useEffect(() => {
+    const element = editor?.findDOMNode([])
+    if (element == null || controller == null) return
+    return pollBoxModel({
+      element: element as unknown as HTMLElement,
```
migueloller, line 61: I would suggest checking the element is actually an HTML element instead of bypassing the types

packages/runtime/src/controls/rich-text.ts
```diff
@@ -0,0 +1,132 @@
...
+import { OnChangeParam, Editor } from 'slate-react'
+import { BoxModel } from '../box-model'
+
+export type RichTextControlData = IndexSignatureHack<Slate.ValueJSON>
+
+export const RichTextControlType = 'makeswift::controls::rich-text'
+
+type RichTextControlConfig = {
+  label?: string
+  defaultValue?: RichTextControlData
```
migueloller, line 14: I would like us not to leak the internal data structure in the public API. Since the prop is received as a `ReactNode`, this default value is really the only place where you have to think in terms of Slate data. Besides it exposing complexity it also makes it harder for us to change internals, like upgrading Slate and changing our data structure since it would be a public API change.

My suggestion would be to allow plaintext as a default value that we then convert to the data structure. For example, for the color control we support color strings even though the underlying data is a Makeswift swatch ID. We could also simply drop the default value option.

packages/runtime/src/controls/rich-text.ts
```diff
@@ -0,0 +1,132 @@
...
+import { PropController } from '../prop-controllers/instances'
+import { OnChangeParam, Editor } from 'slate-react'
+import { BoxModel } from '../box-model'
+
+export type RichTextControlData = IndexSignatureHack<Slate.ValueJSON>
+
+export const RichTextControlType = 'makeswift::controls::rich-text'
+
+type RichTextControlConfig = {
+  label?: string
```
migueloller, line 13: I thought that there were multiple panels for the rich text control. Is this label applied as a prefix to all panel labels?

packages/runtime/src/controls/rich-text.ts
```diff
@@ -0,0 +1,132 @@
...
+
+type BlurRichTextControlMessage = { type: typeof RichTextControlMessageType.BLUR }
+
+type UndoRichTextControlMessage = { type: typeof RichTextControlMessageType.UNDO }
+
+type RedoRichTextControlMessage = { type: typeof RichTextControlMessageType.REDO }
+
+type BoxModelChangeRichControlMessage = {
+  type: typeof RichTextControlMessageType.CHANGE_LAYOUT
+  payload: { layout: BoxModel | null }
```
migueloller, line 64: For consistency's sake, since we use the term "box model" in all of the runtime instead of "layout", can we change the payload field from `layout` to `boxModel`?

packages/runtime/src/prop-controllers/instances.ts
migueloller: Ok, so as I've reviewed the code I'm seeing that you've changed this across both the rich text control and the prop controller. Do you mind explaining why you introduced the change? I see you mentioned `elementKey` and `propName`, but I'm not seeing those used here so I don't understand the connection yet.

packages/runtime/src/runtimes/react/controls/rich-text.tsx
```diff
@@ -0,0 +1,161 @@
...
+  RichTextControlData,
+  RichTextControlDefinition,
+} from '../../../controls/rich-text'
+import { RichTextEditor } from '../../../components/builtin/Text/components/RichTextEditor'
+// @ts-expect-error: no types for 'slate-hotkeys'
+import Hotkeys from 'slate-hotkeys'
+import { isHotkey } from 'is-hotkey'
+
+import { Value } from 'slate'
+import { Editor, OnChangeParam } from 'slate-react'
```
migueloller, line 22: These imports are quite heavy and I'm worried this could be affecting live page bundles even when a site isn't using this control. Can you please check bundle sizes and figure out how to code-split if it's a problem?

packages/runtime/src/runtimes/react/controls/rich-text.tsx
```diff
@@ -0,0 +1,161 @@
...
+  return (
+    <RichTextEditor
+      ref={setEditor}
+      readOnly={!isInBuilder || control == null}
+      value={value}
+      onChange={handleChange}
+      onFocus={handleFocus}
+      onKeyDown={handleKeyDown}
+      onBlur={handleBlur}
+    />
```
migueloller, line 159: This doesn't seem to be leveraging the optimizations I shipped for v0.6 for live pages. Given the similarity between this and our built-in Text component, is there a reason you didn't use that component directly?

packages/runtime/src/runtimes/react/pollBoxModel.tsx
migueloller: Ok, it seems to have been extracted from the Slot control. I think it's good where it is right now, but can we follow the file-naming conventions from other utilities in that folder and rename to `poll-box-model.ts`? (note the lack of `x` in the extension 😄 )

migueloller (changes requested):
This looks pretty good! Nice work @joshuawootonn. Left various questions and comments requesting some changes. Also, here's a high-level question and a comment

- Looks like we're changing the existing rich text component. Do you feel like it was necessary to do this even though the prop controller isn't part of the public API and won't be used? It's not a big deal, but I tend to shy away from updating some of our legacy controls to avoid breaking our default host users in any way (via the Text component)
- I would recommend figuring out how to reuse the Text component in the `useRichText` hook so that we only have to maintain one component and we can leverage all perf improvements, etc.

.changeset/popular-roses-deliver.md
joshuawootonn: Cool cool

packages/runtime/src/prop-controllers/instances.ts
joshuawootonn: When you have multiple panels, and you double click to enter content mode on the first rich text controller, then focus flickers through all rich text instances and lands on the last one. 

This is because we hadn't had this "multiple richtext panels visible at once" edge case before. This change handles that edge case.

packages/runtime/src/runtimes/react/controls/rich-text.tsx
joshuawootonn: I am taking this comment to mean that I should look into reusing the Text component.

packages/runtime/src/controls/rich-text.ts
joshuawootonn: Great catch. I am not doing anything with this value.

packages/runtime/src/controls/rich-text.ts
joshuawootonn: In my mind, I was thinking "I will handle this in milestone 3," but starting out with a basic text default value means that we don't have to have a breaking change. Will update!

packages/runtime/src/prop-controllers/instances.ts
migueloller: Got it. Are there any edge cases that were covered by the focus even that aren't covered by the builder mode switching? There's some very small nuances in this UX so I want to make sure that nothing breaks

packages/runtime/src/controls/rich-text.ts
migueloller: In that case, we should remove it

migueloller: Also, what are we going to do about labels? It might be confusing when having multiple controls?

packages/runtime/src/runtimes/react/controls/rich-text.tsx
migueloller: That sounds like it could solve any bundle size problems, yeah. The Text component might still be included in a site even if the control isn't used, but if it's light enough, that might not be a big deal. So regardless, you probably want to check the bundle and verify

packages/runtime/src/prop-controllers/instances.ts
joshuawootonn: Not that I have found. I'll keep my eyes pealed!

packages/runtime/src/controls/rich-text.ts
joshuawootonn: Agreed. I think it will be confusing, but I think it is ok for now. Users won't be able to find this feature unless they dig for it. We can not release it till sub-selection comes out.

packages/runtime/src/runtimes/react/controls.tsx
```diff
@@ -221,17 +221,18 @@ export function PropsValue({ element, children }: PropsValueProps): JSX.Element
...
+
+          //   return (
+          //     <RenderHook
+          //       key={descriptor.type}
+          //       hook={useRichText}
+          //       parameters={[props[propName], control]}
+          //     >
+          //       {value => renderFn({ ...propsValue, [propName]: value })}
+          //     </RenderHook>
+          //   )
```
joshuawootonn, line 235: if you uncomment this and drop a text component this breaks

packages/runtime/src/runtimes/react/controls/rich-text.tsx
```diff
@@ -1,161 +1,26 @@
...
-
-      setValue(currentValue =>
-        currentValue.selection.isBlurred
-          ? Value.fromJSON(nextValue.toJSON({ preserveSelection: false }))
-          : nextValue,
-      )
-    }
-  }, [shouldCommit, data])
+export function useRichText(data: RichTextControlData, control: RichTextControl | null) {
+  const [current, ref] = useState<PropControllersHandle<Descriptors> | null>(null)
```
joshuawootonn, line 17: Were you expecting me to be able to do this with a `useRef` ?

I don't think it's possible, but I wanted to double check.

migueloller: Yes, we should be able to avoid using state here

joshuawootonn: I don't think we can avoid it actually. 

Tried again and I don't think we can on second thought. 

We want to set the controller in the effect so it has to be in state. Because `useEffect` wont rerun on ref change.

I could put the `set` in a ref callback, but then that is not reactive to my controller changing.

packages/runtime/src/runtimes/react/controls/rich-text.tsx
```diff
@@ -0,0 +1,39 @@
...
+  RichTextControlData,
+  RichTextControlDefinition,
+} from '../../../controls/rich-text'
+import { PropControllersHandle } from '../../../state/modules/prop-controller-handles'
+import { useIsPreview } from '..'
+import { BoxModelHandle } from '../../../box-model'
+import dynamic from 'next/dynamic'
+import { forwardNextDynamicRef } from '../../../next'
+
+const EditableText = forwardNextDynamicRef(patch =>
```
joshuawootonn, line 14: When I instead tried to import the `Text` component here I got some really strange errors.

packages/runtime/src/controls/rich-text.ts
```diff
@@ -0,0 +1,129 @@
...
+import { BuilderEditMode } from '../utils/constants'
+import { PropController } from '../prop-controllers/instances'
+import { OnChangeParam, Editor } from 'slate-react'
+import { BoxModel } from '../box-model'
+
+export type RichTextControlData = IndexSignatureHack<Slate.ValueJSON>
+
+export const RichTextControlType = 'makeswift::controls::rich-text'
+
+type RichTextControlConfig = {}
```
migueloller, line 12: Something to be careful with is that in TypeScript `{}` is almost the same as `any`. If you want to model an empty object, you should use `Record<string, never>` instead. But really, we should just not have any arguments if there aren't any yet.

.changeset/popular-roses-deliver.md
```diff
@@ -0,0 +1,5 @@
+---
+'@makeswift/runtime': patch
+---
+
+Add RichText control for usage within custom components
```
migueloller, line 5: This feels rather simple for a release as important as this one, don't you think?

packages/runtime/src/runtimes/react/controls/rich-text.tsx
```diff
@@ -0,0 +1,39 @@
...
+export type Descriptors = { text?: RichTextControlDefinition }
+
+export function useRichText(data: RichTextControlData, control: RichTextControl | null) {
+  const [current, ref] = useState<(PropControllersHandle<Descriptors> & BoxModelHandle) | null>(
+    null,
+  )
+
+  useEffect(() => {
+    if (current != null && control != null) {
+      current?.setPropControllers({ text: control as any })
```
migueloller, line 32: How come we're using `any` here?

.changeset/popular-roses-deliver.md
```diff
@@ -0,0 +1,41 @@
...
+  props: {
+    title: RichText(),
+    body: RichText(),
+    image: Image(),
+    className: Style(),
+  },
+})
+```
+
+Read more about the Richtext control in our [docs](https://www.makeswift.com/docs/controls/rich-text).
```
joshuawootonn, line 41: working on this doc. will hold on pushing this till it is ready. this shouldn't be a blocker to approval

packages/runtime/src/runtimes/react/controls/rich-text.tsx
joshuawootonn: This should be resolved now!

.changeset/popular-roses-deliver.md
joshuawootonn: Updated it with more details.

packages/runtime/src/controls/rich-text.ts
joshuawootonn: I decided to work that muscle and remove this since there are no props yet

packages/runtime/src/runtimes/react/controls/rich-text.tsx
joshuawootonn: I was wrong here.

Using a ref callback with `useCallback` actually nicely keeps this out of state yet reactive to the controller changing.

.changeset/popular-roses-deliver.md
```diff
@@ -0,0 +1,41 @@
+---
+'@makeswift/runtime': patch
+---
+
+Add RichText control.
+
+Until now, having rich text within your custom components required putting a Text component within a Slot control.
+While we are greatful for the flexibility of our primitives we wanted to enable specific component API design.
```
migueloller, line 8: typo: "greatful" -> "grateful"

On a separate note, it seems weird for us to be grateful for something we built ourselves for other people haha. Maybe we can say something like "While it's possible to add rich text by dropping a `Text` component in a slot, sometimes you've got a component where text is the only thing you want to be editable, not any arbitrary React node"

.changeset/popular-roses-deliver.md
```diff
@@ -0,0 +1,41 @@
...
+The RichText control allows for any number of rich text props, which are then individually editable within our bulder,
```
migueloller, line 9: typo: "bulder" -> "builder"

.changeset/popular-roses-deliver.md
```diff
@@ -0,0 +1,41 @@
...
+and just like the Slot control, it returns a ReactNode.
```
migueloller, line 10: Use back ticks for `Slot` and `ReactNode`?

.changeset/popular-roses-deliver.md
```diff
@@ -0,0 +1,41 @@
+---
+'@makeswift/runtime': patch
+---
+
+Add RichText control.
+
+Until now, having rich text within your custom components required putting a Text component within a Slot control.
```
migueloller, line 7: Use back ticks for `Text` and `Slot`?

.changeset/popular-roses-deliver.md
migueloller, line 41: typo: "Richtext" -> "RichText", also back ticks?

packages/runtime/src/runtimes/react/controls/rich-text.tsx
```diff
@@ -0,0 +1,32 @@
...
+  dynamic(() => patch(import('../../../components/builtin/Text'))),
+)
+
+export type RichTextControlValue = ReactNode
+
+export type Descriptors = { text?: RichTextControlDefinition }
+
+export function useRichText(data: RichTextControlData, control: RichTextControl | null) {
+  const refCallback = useCallback(
+    (ref: PropControllersHandle<Descriptors> | HTMLDivElement) => {
```
migueloller, line 24: minor nit, the argument provided to the callback ref isn't a ref since it's not mutable and there's no `current` field, it's actually the `current` field itself. Some suggestions for names:

- `refCallback` -> `textCallbackRef`
- `ref` -> `current`
- `ref` -> `handle`

migueloller (changes requested):
This is looking great! Just have a couple of minor things you should be able to take care of pretty easily

.changeset/popular-roses-deliver.md
joshuawootonn: I like that

PR #334, opened 2023-02-03
New element selection algorithm based on Document.elementFromPoint

migueloller:
- Use builder pointer information and DOM APIs to reliably determine the active element. This guarantees that the Makeswift builder can properly select elements even if their CSS box models overlap (e.g., absolute and fixed position elements).

Resolves PRD-1294 and PRD-1293

### DIFF
--- a/packages/runtime/src/components/builtin/Text/EditableText.tsx
+++ b/packages/runtime/src/components/builtin/Text/EditableText.tsx
@@ -69,6 +69,11 @@ const EditableText = forwardRef(function EditableText(
   useImperativeHandle(
     ref,
     () => ({
+      getDomNode() {
+        const el = editor?.findDOMNode([])
+
+        return el instanceof Element ? el : null
+      },
       getBoxModel() {
         const el = editor?.findDOMNode([])
 
--- a/.changeset/few-trains-stare.md
+++ b/.changeset/few-trains-stare.md
@@ -1,0 +1,5 @@
+---
+'@makeswift/runtime': patch
+---
+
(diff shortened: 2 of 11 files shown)


packages/runtime/src/components/builtin/Text/EditableText.tsx
```diff
@@ -69,6 +69,11 @@ const EditableText = forwardRef(function EditableText(
   useImperativeHandle(
     ref,
     () => ({
+      getDomNode() {
+        const el = editor?.findDOMNode([])
```
joshuawootonn, line 73: I noticed this morning that findDOMNode is deprecated. Does that worry you at all? 

https://beta.reactjs.org/reference/react-dom/findDOMNode

joshuawootonn (approved):
Skimmed. LGTM just have a question about findDOMNode

packages/runtime/src/components/builtin/Text/EditableText.tsx
migueloller: Nope! I've talked to Sebastian Markbåge about this and basically the reason this hasn't been fully removed from React is because there's no alternative to it right now.

Ideally we can do this:

```tsx
<Fragment ref={domNodeRef}>
  <SomeUknownComponent />
</Fragment>
```

But that's not possible right now. I think that by the time they remove `findDOMNode`, then something equivalent to that will be possible and we'll have a suitable replacement.

joshuawootonn: Cool. Good to know!

PR #336, opened 2023-02-06
Snapshot Data Format & Create Snapshot Endpoint

alexhwoods:
Will leave this in draft, since I can just build on it for snapshotting.

### DIFF
--- a/packages/runtime/src/components/page/Page.tsx
+++ b/packages/runtime/src/components/page/Page.tsx
@@ -9,7 +9,7 @@ import { createDocumentReference } from '../../state/react-page'
 import { useMakeswiftClient } from '../../api/react'
 import { useIsInBuilder } from '../../react'
 import deepEqual from '../../utils/deepEqual'
-import { MakeswiftPageDocument } from '../../next'
+import { MakeswiftPageData, MakeswiftPageDocument } from '../../next'
 import { Page as PageType, Site } from '../../api'
 
 const SnippetLocation = {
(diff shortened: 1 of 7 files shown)


alexhwoods:
If you log `cacheData` in your custom host:

```json
{
    "Swatch": [
      {
        "id": "U3dhdGNoOmFiZDViZDMyLTE0ZmItNDkzYy05MjdlLThlYTU2ZGQxMGZjNw==",
        "value": {
          "__typename": "Swatch",
          "id": "U3dhdGNoOmFiZDViZDMyLTE0ZmItNDkzYy05MjdlLThlYTU2ZGQxMGZjNw==",
          "hue": 319,
          "saturation": 95.19231,
          "lightness": 79.2
        }
      }
    ],
    "File": [],
    "Typography": [
      {
        "id": "VHlwb2dyYXBoeTpjZWQwNWE2OC0xZDU5LTRlYzEtYjZhZi0zYjk5NDlmMjQ0ODY=",
        "value": {
          "__typename": "Typography",
          "id": "VHlwb2dyYXBoeTpjZWQwNWE2OC0xZDU5LTRlYzEtYjZhZi0zYjk5NDlmMjQ0ODY=",
          "name": "foo",
          "style": [
            {
              "deviceId": "mobile",
              "value": {
                "fontFamily": null,
                "fontSize": {
                  "value": 16,
                  "unit": "px"
                },
                "color": null,
                "lineHeight": null,
                "letterSpacing": null,
                "fontWeight": null,
                "textAlign": null,
                "uppercase": null,
                "underline": null,
                "strikethrough": null,
                "italic": null
              }
            },
            {
              "deviceId": "desktop",
              "value": {
                "fontFamily": null,
                "fontSize": {
                  "value": 18,
                  "unit": "px"
                },
                "color": {
                  "swatchId": "U3dhdGNoOmFiZDViZDMyLTE0ZmItNDkzYy05MjdlLThlYTU2ZGQxMGZjNw==",
                  "alpha": 1
                },
                "lineHeight": 1.5,
                "letterSpacing": null,
                "fontWeight": 400,
                "textAlign": null,
                "uppercase": null,
                "underline": null,
                "strikethrough": null,
                "italic": null
              }
            }
          ]
        }
      }
    ],
    "PagePathnameSlice": [],
    "GlobalElement": [],
    "Table": [],
    "Snippet": [],
    "Page": [],
    "Site": []
  }
```

packages/runtime/src/next/client.ts
```diff
@@ -6,10 +7,29 @@ export type MakeswiftPage = {
...
 }
 
+// @note: open to whatever name
+export type MakeswiftPageData = {
+  pageId: string
+  siteId: string
+  snapshot: MakeswiftSnapshotV1
+  options: { preview: boolean; apiOrigin: string }
+}
+
```
alexhwoods, line 17: Used here, in place of the top level thing previously called "snapshot"

```typescript
export async function getStaticProps(
  ctx: GetStaticPropsContext<ParsedUrlQuery>
): Promise<GetStaticPropsResult<Props>> {
  const makeswift = new Makeswift(process.env.MAKESWIFT_SITE_API_KEY!);
  const path = "/" + (ctx.params?.path ?? []).join("/");
  const pageData = await makeswift.getPage(path, {
    preview: ctx.preview,
  });

  if (pageData == null) return { notFound: true };

  return { props: { pageData } };
}

export default function Page({ pageData }: Props) {
  return <MakeswiftPage pageData={pageData} />;
}
```

packages/runtime/src/next/client.ts
```diff
@@ -39,12 +59,12 @@ export type MakeswiftPageDocument = {
     isIndexingBlocked: boolean | null
   }
 }
+type MakeswiftResources = ElementTreeResources & NonElementTreeResources
```
alexhwoods: This distinction feels a bit artificial, but it has two real world effects:
1. `NonElementTreeResources` come from the `/document` endpoint, which we could clean up in the future, but we have enough for now.
2. `ElementTreeResources`, "cache data", and `SerializedState` are all the same. We should probably consolidate that, or maybe bring all of the resources into the Redux store for simplicity.

packages/runtime/src/next/client.ts
```diff
@@ -39,12 +59,12 @@ export type MakeswiftPageDocument = {
...
 
-export type MakeswiftPageSnapshot = {
-  document: MakeswiftPageDocument
-  apiOrigin: string
-  cacheData: CacheData
-  preview: boolean
+export type MakeswiftSnapshotV1 = {
+  resources: MakeswiftResources
+  elementTree: Element
+  runtimeVersion: string
```
alexhwoods, line 67: Using the `V1` suffix so we can support the old format in the meantime, get this merged, and break no ones app.

packages/runtime/src/components/page/Page.tsx
```diff
@@ -237,6 +237,167 @@ export function Page({ document: page }: Props): JSX.Element {
   )
 }
 
+type PropsV1 = { pageData: MakeswiftPageData }
+
+export function PageV1({ pageData }: PropsV1): JSX.Element {
```
alexhwoods, line 242: Just had an idea - let's not `export` this type so that the code can truly remain latent, but we can move forward developing snapshotting.

That's if we decide to merge. If we keep a long running branch separate from `main` and just regularly rebase, then there's no need for that.

I'm ok with either.

packages/runtime/src/next/client.ts
```diff
@@ -137,4 +157,77 @@ export class Makeswift {
...
+        if (response.status === 404) throw Error('bad')
+
+        throw new Error(`Failed to create snapshot with error: "${response.statusText}"`)
+      }
+
+      fetchedDocument = await response.json()
+    }
+
+    if (fetchedDocument == null) {
+      throw Error('asdf')
```
alexhwoods, line 216: fix this

alexhwoods:
Rename things to `unstable_`

PR #343, opened 2023-02-13
fix: rich text list block inconsistencies

migueloller:
See https://github.com/ConvertKit/slate-plugins/blob/master/packages/slate-lists/src/create-render-block.js for how `@convertkit/slate-lists` handles it

### DIFF
--- a/packages/runtime/src/components/builtin/Text/ReadOnlyText.tsx
+++ b/packages/runtime/src/components/builtin/Text/ReadOnlyText.tsx
@@ -1,5 +1,5 @@
 import { cx } from '@emotion/css'
-import { ForwardedRef, forwardRef, ReactNode } from 'react'
+import { ComponentPropsWithoutRef, ForwardedRef, forwardRef, ReactNode } from 'react'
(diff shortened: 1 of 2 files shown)


packages/runtime/src/components/builtin/Text/ReadOnlyText.tsx
```diff
@@ -84,6 +84,30 @@ function Document({ nodes }: DocumentJSON) {
   )
 }
 
+type OrderedListProps = ComponentPropsWithoutRef<'ol'>
+
+function OrderedList(props: OrderedListProps) {
+  return <ol className="ordered-list" {...props} style={{ listStylePosition: 'inside' }} />
```
fikrikarim, line 90: I'm guessing there's no need to spread the style and className props?

fikrikarim (approved):
LGTM! Thanks for attaching the source.

packages/runtime/src/components/builtin/Text/ReadOnlyText.tsx
migueloller: I normally would, but in this case I copied the implementation from `@convertkit/slate-lists` since we're going for as much parity as possible

PR #344, opened 2023-02-14
Client-side Navigation

fikrikarim:


packages/runtime/src/state/react-builder-preview.ts
```diff
@@ -406,6 +409,14 @@ export function messageChannelMiddleware(): Middleware<Dispatch, State, Dispatch
         }
       })
 
+      Router.events.on('routeChangeStart', () => {
+        messageChannel.port1.postMessage(changePathnameStart())
+      })
+
+      Router.events.on('routeChangeComplete', () => {
+        messageChannel.port1.postMessage(changePathnameComplete())
+      })
```
migueloller, line 418: This is adding Next.js specific code to the React runtime. I've been trying to keep these things separate so that we're not super tied into Next.js

migueloller (changes requested):

packages/runtime/src/state/react-builder-preview.ts
fikrikarim: I was aware of that, but I couldn't find any place where we could place a Next.js-specific code that has access to the `messageChannel`.

Where should we put this?

migueloller: There isn't a place for stuff like this just yet. The repo is quite new so these patterns come up as use cases like these ones happen. If you don't know how to approach this or would rather just add the code here, that's fine, but it's something I would like contributors to the repo to be aware of so that I'm not the only one pulling in that direction 😄

migueloller (approved):

packages/runtime/src/state/react-builder-preview.ts
fikrikarim: Haha yeah, I tried to follow the pattern and I tried to separate them, but it's hard to do it right now. We could refactor them later once we know as the codebase develop more.

PR #349, opened 2023-02-22
Use Google Storage URL for files

migueloller:


packages/runtime/src/api/graphql/documents/fragments.ts
```diff
@@ -20,7 +20,7 @@ export const FileFragment = /* GraphQL */ `
     __typename
     id
     name
-    publicUrl
+    publicUrlV2
```
migueloller, line 23: This is the change that is most worrisome to me. Could there be conflicts with data that's currently stored in the old format that don't be compatible with the new fragment selection set?

alexhwoods: What data is stored? No file fragments are, right?

Just ids in the element tree, which are then fetched using this fragment

alexhwoods (approved):

packages/runtime/src/api/graphql/documents/fragments.ts
migueloller: I think this is a fine change, yeah. We used to have the props cache that would save a snapshot of all this data but now it shouldn't be a problem

PR #358, opened 2023-02-28
Josh/prd 1168

joshuawootonn:


packages/runtime/src/components/builtin/Text/Text.tsx
```diff
@@ -16,7 +13,7 @@ type Props = {
   margin?: string
 }
 
-const Text = forwardRef(function Text(props: Props, ref: ForwardedRef<unknown>) {
+const Text = forwardRef(function Text(props: Props, ref: ForwardedRef<any>) {
```
joshuawootonn, line 16: I want to make sure I don't run into any weird issues with lazyloading while I work on this feature. So I am no using next/dynamic and this any type is a temporary artifact of that.

packages/runtime/src/controls/rich-text/fixtures/list.ts
```diff
@@ -0,0 +1,111 @@
+import { RichTextDAO } from '../types.old'
+
+export const listFixture: RichTextDAO = {
+  object: 'value',
+  annotations: undefined,
+  data: undefined,
```
joshuawootonn, line 6: the translation I have made ignores these properties and the `data` property on the `object: 'document'`

The construct of a Value and a Document are no longer part of slate, since I don't think we used these anyway I think it is say to not include them in the translation.

packages/runtime/src/controls/rich-text/rich-text.ts
```diff
@@ -74,31 +75,31 @@ export class RichTextControl extends PropController<RichTextControlMessage> {
...
           case BuilderEditMode.BUILD:
-            this.editor?.deselect().blur()
+            // this.editor?.deselect().blur()
             break
         }
         break
       }
       case RichTextControlMessageType.FOCUS: {
-        this.editor?.focus().moveToRangeOfDocument()
+        // this.editor?.focus().moveToRangeOfDocument()
```
joshuawootonn, line 84: I wanted to get the editor working in the most basic sense that it functions as a `textArea`. these comments will be handled in future PRs

joshuawootonn:
PRD-1168
PRD-1169
PRD-1184

PR #361, opened 2023-03-01
Create Leaf component for Text component

joshuawootonn:
I did look into putting this code into a plugin, but plugins are a way of editing the `Editor` and rendering is done by `Editable`.  After reading through the source/doc I don't think this is possible, but it also doesn't seem like a big loss.

### DIFF
--- a/packages/runtime/src/components/utils/types.ts
+++ b/packages/runtime/src/components/utils/types.ts
@@ -4,4 +4,4 @@ export type SwatchValue = {
   lightness: number
 }
 
-export type ColorValue = { swatch?: SwatchValue; alpha: number }
+export type ColorValue = { swatch?: SwatchValue; alpha?: number }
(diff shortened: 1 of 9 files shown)


packages/runtime/src/components/utils/types.ts
```diff
@@ -4,4 +4,4 @@ export type SwatchValue = {
   lightness: number
 }
 
-export type ColorValue = { swatch?: SwatchValue; alpha: number }
+export type ColorValue = { swatch?: SwatchValue; alpha?: number }
```
joshuawootonn, line 7: The type for this when you `useSwatch` is `number | null`. We avoided a type error by casting in the past. Since I am rewriting I thought it would be good to just update the type. Especially since it is low impact.

joshuawootonn:
@migueloller I thought I would throw this up. I am pretty confident in it, but I wanted to keep you in the loop about how I am imagining `renderElement` will look, and also just call out the comments above.

packages/runtime/src/components/utils/types.ts
migueloller: This type is used all over the place, not just with rich text. Is making this optional not causing any errors in other places?

PR #369, opened 2023-03-14
draft: create slate list plugin

joshuawootonn:


packages/runtime/src/components/builtin/Text/EditableText/ListPlugin/lib/setRootBlockType.ts
```diff
@@ -0,0 +1,31 @@
+import { Editor, Node, Path, Transforms } from 'slate'
```
joshuawootonn, line 1: This is essentially toggle list.

In slate list plugin though It just runs a command on all the blocks in selection based on the first block.

What this should really do it iterate the root blocks, and dynamically switch them wrap them or whatever to get them to the type the user is setting. That is not happening here yet, but it will happen when I get around to

joshuawootonn: https://linear.app/makeswift/issue/PRD-1426/upgrade-toggle-list-to-dynamically-change-blocks-based-on-their-type

PR #371, opened 2023-03-15
Make ListControl and ShapeControl composable

fikrikarim:


packages/runtime/src/controls/list.ts
```diff
@@ -32,6 +38,57 @@ export type ListControlItemData<T extends ListControlDefinition> = {
...
+}
+
+export type ListControlMessage = ListControlItemControlMessage
+
+export class ListControl<
+  T extends ListControlDefinition = ListControlDefinition,
+> extends PropController<ListControlMessage> {
+  controls: Map<string, AnyPropController>
+  descriptor: ListControlDefinition
+  send: any
```
migueloller, line 57: We don't want this to be any, right?

packages/runtime/src/prop-controllers/instances.ts
```diff
@@ -222,6 +227,9 @@ export function createPropController<T extends PropControllerMessage>(
     case RichTextControlType:
       return new RichTextControl(send as Send<RichTextControlMessage>)
 
+    case ListControlType:
+      return new ListControl(send as Send<unknown>, descriptor)
```
migueloller, line 231: Instead of `unknown`, shouldn't this be typed?

packages/runtime/src/runtimes/react/controls/list.tsx
```diff
@@ -11,17 +12,26 @@ type ListControlValueProps<T extends ListControlDefinition> = {
...
 }
 
 export function ListControlValue<T extends ListControlDefinition>({
   definition,
   data,
   children,
+  control,
 }: ListControlValueProps<T>): JSX.Element {
+  // Create new controls every time the user adds or removes an item.
+  const itemsControl = useMemo(() => control?.setItemsControl(data), [data?.length, control])
```
migueloller, line 25: How come we're doing this here instead of handling it in the control? It seems very weird to let this logic "spill out", it breaks encapsulation

migueloller: Don't you find it weird you're calling `setItemsControl` in `useMemo`, mutating state on a "getter"?

packages/runtime/src/controls/control.ts
```diff
@@ -28,6 +28,7 @@ export type ControlDefinition =
   | ShapeControlDefinition
   | ListControlDefinition<any>
   | LinkControlDefinition
+  | SlotControlDefinition
```
migueloller, line 31: This will allow `Slot` as a valid type for `Shape`, I think, which we might not want just yet.

migueloller (changes requested):
For this one I think we need to rework how we're keeping the composed controls up to date. Seems very weird it happens when reading the value in a React component.

packages/runtime/src/controls/list.ts
fikrikarim: Updated

packages/runtime/src/prop-controllers/instances.ts
fikrikarim: Updated

packages/runtime/src/runtimes/react/controls/list.tsx
fikrikarim: Yeah, I also didn't like it. We could discuss this further.

packages/runtime/src/controls/control.ts
migueloller: For this case we can make a new type, just for `Shape` that omits the `Slot` definition or something like that. It's also not a big deal, `Slot` in `Shape` should be released together with this anyway.

I don't like that we're avoiding writing change logs because it's tied to a release. Those two things should be decoupled from one another... I'm seriously considering removing the release pull request automation in favor of manual releases so that we can write change logs without worrying about it making its way to a release.

I say just write the change log for what you're doing and if you don't intend it to be used or don't want to release it then don't expose it in the public API or expose it under an `unstable_` prefix.

migueloller: Things could also be placed behind an experimental or unstable feature flag. Do whatever you think is best.

packages/runtime/src/runtimes/react/controls/list.tsx
fikrikarim: So right now we're creating the control from an action. Same as with the builder, we need to have a `subscribeToValue` function available to the `createPropController`. 

The problem is, currently, we don't have any way to subscribe to a value from inside an action. I don't think there's any official way to do `store.subscribe` from an action.

One way to solve this is to have a custom thunk that also passes the `store.subscribe` as the third argument for the action, similar to this [package](https://github.com/AugurProject/redux-thunk-subscribe/blob/master/src/index.js).

But I'm not 100% sure that would be the best approach.

fikrikarim: Decided to use useEffect for now. Added comments there as to how we can improve it.

PR #372, opened 2023-03-17
Make ShapeControl composable

fikrikarim:


migueloller (changes requested):
This one's pretty simple to do after #371 so let's resolve that one first and then have this one match.

PR #385, opened 2023-03-24
fix: text not editable inside List or Shape

fikrikarim:


fikrikarim:
@migueloller this fixes the problem of Text not editable inside List or Shape.

PR #394, opened 2023-03-30
Make RichTextControl composable

fikrikarim:


packages/runtime/src/state/modules/read-only-documents.ts
```diff
@@ -1,6 +1,15 @@
...
-export type Data = undefined | null | boolean | number | string | Data[] | { [key: string]: Data }
+export type Data =
+  | undefined
+  | null
+  | boolean
+  | number
+  | string
+  | Data[]
+  | { [key: string]: Data }
+  | ValueJSON
```
migueloller, line 12: We should _not_ do this. `ValueJSON` should already extend `Data` given how its defined. I've seen some weird issues with interfaces not matching with the type and there was a utility function I made to work around it called `IndexSignatureHack`, maybe that can help again? If not, we can use `// @ts-expect-error` comments in specific places.

packages/runtime/src/controls/rich-text/rich-text.ts
```diff
@@ -70,7 +70,7 @@ export type RichTextControlMessage =
 export class RichTextControl extends PropController<RichTextControlMessage> {
   private editor: Editor | null = null
 
-  recv(message: RichTextControlMessage): void {
+  recv(message: PropControllerMessage): void {
```
migueloller, line 73: Why would the rich text control receive a message that's not a rich text control message?

packages/runtime/src/prop-controllers/instances.ts
```diff
@@ -86,19 +86,17 @@ export abstract class PropController<T = PropControllerMessage> {
     this.send = send
   }
 
-  abstract recv(message: T): void
-}
-
-class DefaultPropController extends PropController {
```
migueloller, line 92: This change makes us lose type safety. Why change the generic `T`, which is specific for each control to a generic `PropControllerMessage`?

migueloller (changes requested):
Main rich text changes look good but it seems we're refactoring the prop controller instances in a way that makes us lose specificity of types. I'm curious, what motivated that change?

packages/runtime/src/controls/rich-text/rich-text.ts
fikrikarim: See my other comment on why.

packages/runtime/src/prop-controllers/instances.ts
migueloller: Hmm. I was thinking that we'd be able to define the base interface type as the more general type using generics and then we'd refine it in each control that implements the base control. I don't have a specific way in mind of how this would work since I don't have the latest state of the code in my head but I'm happy to look at it if you want.

packages/runtime/src/controls/rich-text/dto-types.ts
```diff
@@ -25,7 +26,8 @@ interface ValueJSON {
   annotations?: { [key: string]: AnnotationJSON } | undefined
   data?: { [key: string]: any } | undefined
 }
-export type RichTextDTO = ValueJSON
+
+export type RichTextDTO = IndexSignatureHack<ValueJSON>
```
migueloller, line 30: Wouldn't it be better to use this utility type in the places where `RichTextDTO` gets imported from this file? That way the "hack" is placed in the location where it's needed vs in this general types file, which it's not clear why it's needed here

migueloller (dismissed):
@fikrikarim, this looks good to me. The only thing that maybe I would look into is dropping the commit that reduces type specificity for prop controller messages. It's not a big deal, though

packages/runtime/src/controls/rich-text/dto-types.ts
fikrikarim: That's true. Moved that to a place where we need it.

packages/runtime/src/prop-controllers/instances.ts
fikrikarim: Okay, for now, I dropped the commit and I type cast the `recv` instead. I put a `TODO` there so we can fix it later.

fikrikarim:
> @fikrikarim, this looks good to me. The only thing that maybe I would look into is dropping the commit that reduces type specificity for prop controller messages. It's not a big deal, though

Got it. For now, I dropped the commit and I type cast the `recv` instead. I put a `TODO` there so we can fix it later.

PR #395, opened 2023-03-30
feat: slate typography and block

joshuawootonn:
This PR 

* Moves the List plugin into `@makeswift/runtime/slate` [link](https://github.com/makeswift/makeswift/pull/395/commits/7ecf82a51c5dc4e0b1c3bd4e674bbf6f9db7a33e)
* Moves some helpers that I created for the Listplugin to a more general plugin location so that they can be reused across plugins
* Removes the concept of `type` from Text. We now just rely on an optional `typography` property
* Rework the List plugin to follow the new conventions I created for the two new plugins


joshuawootonn:
@migueloller thought it would be good for you to see what the new rich-text code looks like. Feel review at this point if you like. The last commit is a WIP that I'll be finishing tomorrow.

packages/runtime/src/controls/rich-text/compare.ts
```diff
@@ -1,185 +0,0 @@
-import { Descendant, Selection } from 'slate'
```
joshuawootonn, line 1: Removed this comparison. I wasn't aware that we already had the deepEqual :)

packages/runtime/src/slate/tsconfig.json
```diff
@@ -0,0 +1,21 @@
...
+    "strict": true,
+    "sourceMap": true,
+    "resolveJsonModule": true,
+    "esModuleInterop": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "noImplicitReturns": true,
+    "skipLibCheck": true,
+    "rootDir": ".",
+    "jsx": "preserve",
```
joshuawootonn, line 18: This allows me to not have `//@ts-ignores` on my pragma imports

packages/runtime/src/plugins/TypographyPlugin/__tests__/clearActiveTypographyStyle.test.tsx
```diff
@@ -0,0 +1,77 @@
+/** @jsx jsx */
```
joshuawootonn, line 1: Went for a more flat structure. I like it!

packages/runtime/src/plugins/utils/filterForSubtreeRoots.ts
```diff
@@ -0,0 +1,11 @@
+import type { Node, NodeEntry } from 'slate'
```
joshuawootonn: These utils aren't all used here, but they are going to be used in the wip commit. They were c/p'd from the `ListPlugin/utils` folder

packages/runtime/src/plugins/utils/element.ts
```diff
@@ -0,0 +1,61 @@
+import { Node, Element } from 'slate'
```
joshuawootonn, line 1: FWIW I hate util folders. I don't think this is hot stuff. I don't like having `ElementUtils`. 

It's a temporary thing until I find best practices on this stuff.

packages/runtime/src/slate/ListPlugin/ListPlugin.ts
```diff
@@ -90,6 +90,7 @@ export function onKeyDown(e: KeyboardEvent, editor: Editor) {
 }
```
joshuawootonn, line 90: Missed this before. More to come in this commit

packages/runtime/src/plugins/utils/element.ts
migueloller: Slate seems to call them[ [custom helpers](https://kevinzakka.github.io/robopianist-demo/) so maybe we can just go for what they go for? Maybe look at the Slate source code to see how they organize their built-in helpers?

packages/runtime/src/slate/ListPlugin/ListPlugin.ts
migueloller: Missed what, exactly?

migueloller (approved):
So many tests 😍 

LGTM

packages/runtime/src/slate/ListPlugin/ListPlugin.ts
joshuawootonn: When I dropped the list plugin we only had one set of normalization so not calling `normalizeNode(entry` at the end wasn't an issue. For the normalization to cascade through all normalization plugins from the different plugins I have to call `normalizeNode` at the end like this diff changes.

> More to come in this commit
Refers to all the changes that I am going to make with the List plugin to get it to match the new plugins

packages/runtime/src/plugins/utils/element.ts
joshuawootonn: Seems like you shared the wrong link. Unless I am missing something :P 

Yeah, I have seen those. I didn't like the idea of wrapping the standard interfaces initially because there are already enough import options named Element, Editor, Range, etc. Now that I am thinking about it though I like the idea of having MakeswiftElement. It would clear things up so that you don't have to check the import statements as much. 

For another day regardless!

PR #405, opened 2023-04-06
Make StyleControl composable

fikrikarim:


migueloller (changes requested):
@fikrikarim, this looks good! There’s a request I have, though. The way we get the element, if the underlying element changes, then the ref won’t be set again, meaning the box model won’t be right. This way to fix this is to call the query selector in the animation frame callback on each animation frame request. You might have to write the animation frame polling manually in a single effect instead of using `pollBoxModel` to do this. There’s a couple of places we do this in so a search for `requestAnimationFrame` should give you some ideas.

packages/runtime/src/runtimes/react/controls/style.ts
```diff
@@ -115,8 +118,30 @@ export type StyleControlFormattedValue = string
 export function useFormattedStyle(
   styleControlData: StyleControlData | undefined,
   controlDefinition: StyleControlDefinition,
+  control: StyleControl | null,
 ): StyleControlFormattedValue {
   const style = useStyleControlCssObject(styleControlData, controlDefinition)
+  const elementRef = useRef<Element | null>(null)
+  const guid = useId().replaceAll(':', '')
```
migueloller, line 125: How come we had to replace colons here?

fikrikarim: We're removing the colons because `useId` returns a string wrapped with colons, e.g. ":r0:", and we cannot use a colon in a class name.

migueloller (approved):

PR #408, opened 2023-04-07
Makeswift API resource endpoints in Next.js API handler

migueloller:


.changeset/happy-items-turn.md
```diff
@@ -0,0 +1,7 @@
+---
+'@makeswift/runtime': minor
+---
+
+BREAKING: The `MakeswiftClient.prefetch()` and `MakeswiftClient.getTypographies` methods have been removed. These were internal APIs so there shouldn't be any changes required to upgrade.
```
fikrikarim, line 5: Should be `fetchTypographies`?

fikrikarim (commented):
Thanks for breaking up the API handler file 😄 

Why are we removing `fetchTypographies` but not other fetch methods? Are we removing them one step at a time?

.changeset/happy-items-turn.md
migueloller: Yes! Nice catch.

As to why we're only removing `fetchTypographies` it's because that one's only used for introspection so it's being moved to the Makeswift API client. All other fetch methods are still needed by the host API client to fetch data for controls like swatches, files, etc.

fikrikarim (approved):

alexhwoods (dismissed):
Looks good!

One thought I had as I was reviewing - long term, do you think we should change `Makeswift` and `MakeswiftClient` to something else. They both product "makeswift clients", right?

I think the one that interacts with just the host should perhaps be a `HostClient`, or a `NextClient`, or something like that.

migueloller:
> Looks good!
> 
> One thought I had as I was reviewing - long term, do you think we should change `Makeswift` and `MakeswiftClient` to something else. They both product "makeswift clients", right?
> 
> I think the one that interacts with just the host should perhaps be a `HostClient`, or a `NextClient`, or something like that.

Yes! I was thinking of naming them `HostClient` and `Makeswift`. So usage looks like this:

```
import { Makeswift } from '@makeswift/runtime/next'
```

`HostClient` would be internal for now and not part of the public API.

PR #416, opened 2023-04-12
Custom Breakpoints

fikrikarim:


packages/runtime/src/components/builtin/Image/Image.tsx
```diff
@@ -47,19 +48,24 @@ function loadImage(src: string): Promise<HTMLImageElement> {
   })
 }
 
-function imageSizes(width?: Props['width']): string {
-  const baseDevice = DEVICES.find(device => device.maxWidth == null)
-  const baseWidth = baseDevice && width && findDeviceOverride(width, baseDevice.id)
+function useImageSizes(width?: Props['width']): string {
```
migueloller, line 51: Note, there's no need to make _everything_ a hook. We can keep `imageSizes` and add a `breakpoints` arguments that replaces the `DEVICES` variable and then implement `useImageSizes` this way:

```ts
function useImageSizes(width) {
  const breakpoints = useBreakpoints()

  return imageSizes(width, breakpoints)
}
```

This is a general principle when writing React code. It's easy to shove everything into a hook, but then we lose the ability to reuse the `imageSizes` logic in other places. In this case it might not matter, but I've found myself repeating _a lot_ of code that was "embedded" in a hook this way when it didn't need to be

packages/runtime/src/components/hooks/useMediaQuery.ts
```diff
@@ -1,32 +1,47 @@
...
-    }, DEVICE_QUERIES[0].id)
-    return findDeviceOverride(responsiveValue, deviceId)?.value
+    const deviceId: string = getDeviceQueries(breakpoints).reduce(
+      (matchedDevice, deviceQueries) => {
+        if (window.matchMedia(deviceQueries.query).matches) {
+          return deviceQueries.id
+        }
+        return matchedDevice
+      },
+      DEFAULT_BREAKPOINTS[0].id,
```
migueloller, line 42: I think it's ok here not to assume an initial value anymore if the breakpoints are dynamic. In the edge case that there are no breakpoints, then the fallback strategy should take care of it.

Something we need to discuss is how we'll handle that. In the past i've thought of defining a `null` or "special" breakpoint ID which is the "fallback", but perhaps we don't need to use `null`. Here's an idea:

First, we never allow there to be 0 breakpoints, if 0 breakpoints are registered we throw an error, we need at least 1. Furthermore, we store in the state a "base" breakpoint. We can compute that automatically by sorting by width if we're doing a desktop-first design strategy but we could also have other strategies, like mobile-first design. We'll either keep that "base" endpoint in the state or if we're computing it, it can just be a selector.

Then, whenever we're wanting to get the "base" style, or what we call "desktop" right now, we can get the "base" style from state. In this way, we don't need to hardcode `DEFAULT_BREAKPOINTS[0].id` (i.e., `'desktop'`) and things can be dynamic based on the breakpoints

packages/runtime/src/prop-controllers/descriptors.ts
```diff
@@ -47,7 +47,7 @@ export type IndexSignatureHack<T> = T extends Record<string, any>
   ? { [K in keyof T]: IndexSignatureHack<T[K]> }
   : T
 
-export type Device = 'desktop' | 'tablet' | 'mobile'
+export type Device = 'desktop' | 'tablet' | 'mobile' | string
```
migueloller, line 50: I see we're doing this all over the place, I'm guessing the type will just be `string` eventually?

packages/runtime/src/runtimes/react/index.tsx
```diff
@@ -52,6 +54,8 @@ export interface ReactRuntime {
     elementTree: ReactPage.ElementData,
     replacementContext: ReactPage.SerializableReplacementContext,
   ): ReactPage.Element
+  registerBreakpoints(breakpoints: Breakpoints): void
```
migueloller, line 57: If we put this in the API handler, how will we retrieve breakpoints during SSR? Do we need the breakpoints for SSR rendering? We do to generate the classes with responsive styles, right? For the `getFonts` case it's just for the builder to request and show in the panel. But for the runtime, it's a bit more complicated than that.

Should breakpoints perhaps be a prop on the Makeswift page component? Or a piece of context?

packages/runtime/src/slate/BlockPlugin/setBlockKeyForDevice.ts
```diff
@@ -20,7 +20,13 @@ export function setBlockKeyForDevice(
   for (const [node, path] of rootElements) {
     if (ElementUtils.isBlock(node)) {
       const deviceValues = node[key] ?? []
-      const currentDeviceValue = findDeviceOverride(deviceValues, deviceId, v => v)
+      // TODO: Get from registered breakpoints instead of the default breakpoints
```
migueloller, line 23: I think this can be fixed by modifying the plugin so that it takes the list of devices/breakpoints as an argument 🤔

migueloller (commented):
@fikrikarim, this all looks pretty straightforward. Mainly wiring things so they can read from state instead of the hardcoded devices. Seems like the primary things to address is where the API for adding breakpoints will be as well as what we're going to share between the host and builder

migueloller:
> 1. Whether we want to make it an array or an object, with the key of the id.

I think I like using an object with IDs as keys better.

> 2. Whether we want to use just `width`, and we'll interpolate that. Or should we use `minWidth` and `maxWidth`?

I think that we should start with a more strict API that we can expand on later as needed without breaking anything. So let's start with just `width`? Maybe later we can add more options, like Tailwind?

> 3. Should we let users specify the `viewportWidth`, or should we just interpolate that?

I think we can start with a smaller API and expand from there as we learn? So for now just use `width`?

> 4. Is the label required or optional? The label is used in the breakpoint dropdown, and also on the panel. We can fallback to the `id` if the label is undefined, but it seems like the label is quite important.

I think it would be cool to make it optional so we can auto-generate from Tailwind. I'm thinking we can use the ID as the label but put it inside `<code>` so it's styled a bit differently in the UI. We'll want some mockups from @andrewreifman for that. The idea would be that if you provide a label then we just show "Label" but if you don't provide a label we show "`id`" in a way that's styled so it looks like "code" (i.e., monospace font and maybe with a different color or background)

migueloller:
> It's a bit weird because right now you don't have to pass the created runtime anywhere. But in the future, we will require to pass the runtime to the MakeswiftPage.

Will these be part of this PR or are you thinking it will be a future change? I'm thinking for now we can add it as an optional prop so that we can start using the new pattern but it's not a breaking change?

fikrikarim:
I also just realized that it might be impossible to interpolate the `viewportWidth`. For example, for the default mobile breakpoint, the `maxWidth` is 575 and the `viewportWidth` is 390. Let's say we interpolate from 575 and 0, it'll be `288` if we use the midpoint. 

But for the default tablet, the `minWidth` is 576, `maxWidth` is 768 and the `viewportWidth` is 760, so for tablet the `viewportWidth` is much closer to the `maxWidth`.

migueloller:
> I just realized that we cannot turn just width into a fixed range breakpoints system that we currently use.

I think you might be misunderstanding how Tailwind's API maps to our API. The example you provided would be incorrect, the correct use would be this:

```ts
export const runtime = new ReactRuntime({
  unstable_breakpoints: {
    'tablet': { width: 770 },
    'mobile': { width: 576 },
  },
})
```

The "base" breakpoint doesn't exist. In the same way in Tailwind, which is mobile-first, you don't specify the `min-width: 0` breakpoint, it's implicit. In Makeswift, which is desktop-first, there is no "desktop" breakpoint, it's the "base" value where everything inherits from.

migueloller:
> I also just realized that it might be impossible to interpolate the `viewportWidth`. For example, for the default mobile breakpoint, the `maxWidth` is 575 and the `viewportWidth` is 390. Let's say we interpolate from 575 and 0, it'll be `288` if we use the midpoint.
> 
> But for the default tablet, the `minWidth` is 576, `maxWidth` is 768 and the `viewportWidth` is 760, so for tablet the `viewportWidth` is much closer to the `maxWidth`.

Perhaps doing an interpolation doesn't give good results, but the mapping function doesn't have to be linear

packages/runtime/src/components/builtin/Text/components/Leaf/leaf.tsx
```diff
@@ -54,7 +54,7 @@ export function isDeviceOverride(value: {
   deviceId: string
   value: RichTextTypographyValue
 }): value is DeviceOverride<RichTextTypographyValue> {
-  return value.deviceId === 'desktop' || value.deviceId === 'tablet' || value.deviceId === 'mobile'
+  return value.deviceId != null
```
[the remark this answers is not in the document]
migueloller: It seems to be refining `string` to `'desktop' | 'tablet' | 'mobile'`. That might or might not be relevant anymore depending on the desired value of `deviceId`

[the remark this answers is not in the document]
migueloller: And by the way, checking for `deviceId` not being `null` would be incorrect given the function never receives `null` and `value.deviceId` is a `string`

packages/runtime/package.json
```diff
@@ -16,7 +16,8 @@
     "next/plugin.js",
     "builder.js",
     "controls.js",
-    "slate.js"
+    "slate.js",
+    "state"
```
migueloller, line 20: Is this missing a file extension?

packages/runtime/src/components/builtin/Image/Image.tsx
```diff
@@ -47,23 +49,33 @@ function loadImage(src: string): Promise<HTMLImageElement> {
...
+      return `(max-width: ${breakpoint.maxWidth}px) ${Math.min(
+        breakpoint.maxWidth,
+        override.value.value,
+      )}px`
+    })
     .filter((size): size is NonNullable<typeof size> => size != null)
     .reduce((sourceSizes, sourceSize) => `${sourceSize}, ${sourceSizes}`, baseWidthSize)
 }
 
+function useImageSizes(...args: DropFirst<Parameters<typeof imageSizes>>) {
```
migueloller, line 75: I think I would rather we be explicit with arguments here. This looks very fancy but I think it make things harder to follow

packages/runtime/src/components/builtin/Image/Image.tsx
```diff
@@ -132,15 +144,15 @@ const ImageComponent = forwardRef(function Image(
         <NextLegacyImage
           layout="responsive"
           src={imageSrc}
-          sizes={imageSizes(width)}
+          sizes={useImageSizes(width)}
```
migueloller, line 147: Breaking the rules of hooks

packages/runtime/src/components/builtin/Image/Image.tsx
```diff
@@ -132,15 +144,15 @@ const ImageComponent = forwardRef(function Image(
...
           alt={altText}
           width={dimensions.width}
           height={dimensions.height}
         />
       ) : (
         <NextImage
           src={imageSrc}
-          sizes={imageSizes(width)}
+          sizes={useImageSizes(width)}
```
migueloller, line 155: Breaking the rules of hooks

packages/runtime/src/components/builtin/Text/register.ts
```diff
@@ -36,11 +37,12 @@ export function registerComponent(runtime: ReactRuntime) {
                               id: null,
                               style: [
                                 {
-                                  deviceId: 'mobile',
+                                  // TODO: This will only work if they have the mobile breakpoint.
```
migueloller, line 40: Can we make this more robust by only adding the entry to the array if the breakpoints contain the desired hardcoded ID?

packages/runtime/src/runtimes/react/index.tsx
```diff
@@ -52,34 +58,58 @@ export interface ReactRuntime {
...
+    return ReactPage.getBreakpoints(storeContextDefaultValue.getState())
   }
-}
 
-export const ReactRuntime = createReactRuntime(storeContextDefaultValue)
+  constructor({ unstable_breakpoints }: { unstable_breakpoints?: BreakpointsInput } = {}) {
+    if (unstable_breakpoints) {
+      storeContextDefaultValue.dispatch(setBreakpoints(parseBreakpointsInput(unstable_breakpoints)))
+    }
+  }
```
migueloller, line 111: Constructing a runtime dispatches on the default context? We should instead return a new runtime, which its own store, and add all the static methods as instance methods to this class. What this is doing right now is _very_ confusing

packages/runtime/src/state/modules/breakpoints.ts
```diff
@@ -0,0 +1,240 @@
...
+  maxWidth?: number
+}
+
+export type Breakpoints = Breakpoint[]
+
+export type State = Breakpoints
+
+export const DEFAULT_BREAKPOINT_ID_DESKTOP = 'desktop'
+export const DEFAULT_BREAKPOINT_ID_TABLET = 'tablet'
+export const DEFAULT_BREAKPOINT_ID_MOBILE = 'mobile'
```
migueloller, line 29: Why not create an enum like we normally do?

```ts
export const DefaultBreakpointID = {
  Desktop: 'desktop',
  Table: 'tablet',
  Mobile: 'mobile'
} as const

type DefaultBreakpointID = typeof DefaultBreakpointID[keyof typeof DefaultBreakpointID]
```

migueloller (changes requested):

packages/runtime/src/components/builtin/Text/register.ts
fikrikarim: Updated

packages/runtime/src/components/builtin/Image/Image.tsx
fikrikarim: Updated

packages/runtime/src/runtimes/react/index.tsx
fikrikarim: Yes, but it's kinda hard to do this right now because we're using `storeContextDefaultValue` in some places, and we want to make this backward compatible.

This is just the very minimal version so we can start using the new API. I was thinking that we can update the internal implementation later.

fikrikarim: We decided to not refactor the `ReactRuntime` into class for now. The current unstable public API is:
```ts
ReactRuntime.unstable_setBreakpoints({
  desktop: {
    label: "Ultra screen",
    min: 1600,
  },
  max: {
    label: "Max screen",
    min: 1200,
    max: 1599,
    viewport: 1400,
  },
  pro: {
    min: 600,
    max: 1199,
    viewport: 800,
  }
});
```

PR #419, opened 2023-04-15
fix: path normalization

migueloller:
By pushing a relative path, Next.js client-side navigation was broken. We now ensure paths start with a "/".


migueloller:
@fikrikarim, I think this is why there was a 404 when using client-side navigation in the default host.

PR #426, opened 2023-04-27
fix: preserve dom selection for rich text

joshuawootonn:


packages/runtime/src/components/builtin/Text/EditableText/useSyncDOMSelection.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { useIsomorphicLayoutEffect } from '../../../hooks/useIsomorphicLayoutEffect'
+import { Editor, Range as SlateRange } from 'slate'
+
+/**
+ * Clicking outside of the host blurs our `<Editable />`.
+ * `<Editable />` only updates the DOM's selection to match slate when it is focused.
+ * In the case of a panel being clicked this hook updates the DOM selection to match slate. 
+ */
+export function useSyncDOMSelection(editor: Editor) {
+  useIsomorphicLayoutEffect(() => {
```
joshuawootonn, line 11: They use an effect in Slate's `<Editable />` so I decided to do the same. 

Where as before with the `setTimeout` I was sometimes losing the selection on rapid changes, I haven't seen anything like that with `useIsomorphicLayoutEffect`

packages/runtime/src/components/builtin/Text/EditableText/editable-text.tsx
```diff
@@ -82,6 +85,7 @@ export const EditableText = forwardRef(function EditableText(
 
   const handleFocus = useCallback(() => {
     controller?.focus()
+    setHasBeenSelected(true)
```
joshuawootonn, line 88: This prevents you from preserving selection where you have not previously focused. Ideally, this value would listen to "focus" and "blur" messages from the control and only preserve selection when you had currently selected the control. 

That is something I will include in the RichTextV2.

PR #432, opened 2023-05-03
Refactor ReactRuntime to be a class and move unstable_breakpoints there

fikrikarim:


packages/runtime/src/runtimes/react/index.tsx
```diff
@@ -44,66 +43,106 @@ import {
   parseBreakpointsInput,
 } from '../../state/modules/breakpoints'
 
-export const storeContextDefaultValue = ReactPage.configureStore()
+export class ReactRuntime {
+  // TODO: the static methods here are deprecated and only keep here for backward-compatibility purpose.
+  // We will remove them when we release a new breaking change.
+  // ------------------ Deprecated API ------------------ //
```
migueloller, line 49: We might even want to add a JSDoc deprecated tag so that it shows up in users' VSCode once we actually do release a stable version of breakpoints and deprecate this usage, like so:

```
/**
 * @deprecated The ReactRuntime singleton was deprecated in v0.9.0, instantiate your own runtime instead. See here for more details...
 */
```

packages/runtime/src/next/api-handler/handlers/manifest.ts
```diff
@@ -28,5 +29,6 @@ export default async function handler(
     interactionMode: true,
     clientSideNavigation: true,
     elementFromPoint: false,
+    customBreakpoints: true,
```
migueloller, line 32: How come we need this information in the manifest? Is it for the builder to determine whether it should wait for the host to send breakpoints?

migueloller (dismissed):
Looks great!

packages/runtime/src/runtimes/react/index.tsx
fikrikarim: I thought about that, but I intentionally did not put the deprecated tag because we don't the documentation ready for the new API pattern yet.

We can put the deprecated tag once we have all the documentation updated and also once we have a guide on how to migrate to the new API pattern.

PR #435, opened 2023-05-04
feat: create textarea RichTextV2

joshuawootonn:


joshuawootonn:
Hey @migueloller, here is what I took away from the conversation we had about v2 naming. 

Additionally, I added `unstable_` in front of `RichTextV2` just to be sure it was publicly obvious. 

What I am asking for here is a review of how I chose to name things :)

packages/runtime/src/components/builtin/TextV2/Text.tsx
migueloller, line 1: I'm confused as to why we've got a new Text component. I didn't think we wanted to release a new builtin component but instead just a new control

packages/runtime/src/controls/rich-text/types.ts
migueloller, line 1: How come we're changing the types for the legacy rich text control?

migueloller, line 1: How come we're changing the types for the legacy rich text control?

packages/runtime/src/runtimes/react/controls/rich-text-v2.tsx
```diff
@@ -0,0 +1,16 @@
...
+  RichTextV2ControlDefinition,
+} from '../../../controls'
+import TextV2 from '../../../components/builtin/TextV2'
+
+export type RichTextV2ControlValue = ReactNode
+
+export type Descriptors = { text?: RichTextV2ControlDefinition }
+
+export function useRichTextV2(data: RichTextV2ControlData, control: RichTextV2Control | null) {
+  return <TextV2 text={data} control={control} />
```
migueloller, line 15: I see now where the builtin component is being used. We probably want to build that editable from scratch in this file and avoid bringing any bad assumptions or old patterns from the legacy rich text component

packages/runtime/src/components/builtin/TextV2/Text.tsx
joshuawootonn: Great point! I'll just put this in the hook like your other comment mentioned.

packages/runtime/src/controls/rich-text/types.ts
joshuawootonn: If you are referring to `RichTextTypography`: Whoops. I thought this was a part of another commit that I had removed. I just realized (In that other commit) that I want to expand these types out to be nullable/optional to avoid type wrestling between graphql generated typography (nullable) and the resulting typography value (optional) 

If you are referring to the fact that I added `Text` here. That is because I want to initialize my component to something that is not a paragraph. This could have been included in the inline mode PR, but I'm going to leave it here. `Text` will be the inline mode initial type. `Paragraph` will be the block mode initial type.

joshuawootonn:
@migueloller I resolved your comments so that I could merge. Open to more feedback though, just want to iterate quickly. I can include changes in upcoming PRs.

joshuawootonn:
Actually ... just realized the next step is all in cosmos. We can continue discussing this here.

packages/runtime/src/controls/rich-text/types.ts
migueloller: My comment was specifically as to why we were modifying `src/controls/rich-text-/types.ts` when this PR was for introducing a new control, not modifying the legacy one.

Sounds like you just slipped in some changes unrelated to the new control

packages/runtime/src/runtimes/react/controls/rich-text-v2.tsx
```diff
@@ -0,0 +1,171 @@
...
+      } else {
+        domSelection?.removeAllRanges()
+      }
+    } catch (e) {
+      console.error(e)
+    }
+  })
+}
+
+export function useRichTextV2(data: RichTextV2ControlData, control: RichTextV2Control | null) {
```
migueloller, line 79: Minor point, but we might want to actually create a `RichTextV2` component in this file instead of embedding what looks like a component into this hook. Then, this hook can return `<RichTextV2 />` passing the relevant props that are outside the scope of the component, like getting the control instance, communicating with the builder control, etc.

packages/runtime/src/controls/rich-text/types.ts
migueloller, line 1: Minor note, it would be helpful during review if changes like these, which seem to be unrelated to creating the new rich text control, were made in a separate commit. When I was reviewing originally I had assumed that these changes were necessary for the new control, but it seems they're just changes you had meant to do before, in a separate PR. Is that right?

joshuawootonn: Not sure at this point what this is referring to, maybe adding the Text type to slate types? If so I will keep that in mind!

migueloller: Do you see that these changes are made in the old rich text control’s file/folder? I would have thought that types, etc. for rich text v2 would be independent

PR #436, opened 2023-05-04
fix: inlines disappearing on text edit

joshuawootonn:


packages/runtime/src/slate/BlockPlugin/index.ts
```diff
@@ -1,15 +1,24 @@
...
+  wrapInline,
+  unwrapInline,
 }
 
 export function withBlock(editor: Editor) {
   const { normalizeNode } = editor
+
+  editor.isInline = entry => {
+    return ElementUtils.isInline(entry)
+  }
```
joshuawootonn, line 20: This is the main line that was missing. Not having it caused our inlines to be normalized as slate thought they were block level.

packages/runtime/src/slate/ListPlugin/__test__/unwrapList.test.tsx
```diff
@@ -143,4 +143,111 @@ describe('Unwrap List', () => {
     expect(editor.children).toEqual(result.children)
     expect(editor.selection).toEqual(result.selection)
   })
+
+  it('WHEN unwrapList on single node THEN it is removed from list', () => {
```
joshuawootonn, line 147: These were net new tests from cosmos.

packages/runtime/src/slate/BlockPlugin/__tests__/clearBlockKeyForDevice.test.tsx
```diff
@@ -1,6 +1,6 @@
+/** @jsxRuntime classic */
```
joshuawootonn, line 1: Adding this comment prevents us from needing to use a tsconfig. Since we are already specifying the jsx pragma I like this approach in place of tsconfig.

PR #437, opened 2023-05-04
Unstable versioning endpoints for pages and swatches

migueloller:
Also resolves PRD-1812 and PRD-1815

### DIFF
--- a/packages/runtime/src/next/api-handler/handlers/manifest.ts
+++ b/packages/runtime/src/next/api-handler/handlers/manifest.ts
@@ -8,6 +8,7 @@ export type Manifest = {
   clientSideNavigation: boolean
   elementFromPoint: boolean
   customBreakpoints: boolean
+  unstable_siteVersions: boolean
 }
 
 type ManifestError = { message: string }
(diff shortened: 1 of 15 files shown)


packages/runtime/src/next/api-handler/handlers/manifest.ts
```diff
@@ -30,5 +31,6 @@ export default async function handler(
     clientSideNavigation: true,
     elementFromPoint: false,
     customBreakpoints: true,
+    unstable_siteVersions: true,
```
migueloller, line 34: Make this an argument for the API handler

packages/runtime/src/next/api-handler/index.ts
```diff
@@ -65,14 +67,14 @@ export function MakeswiftApiHandler(
       )
     }
 
-    const client = new Makeswift(apiKey, { apiOrigin })
+    const client = new Makeswift(apiKey, { apiOrigin, unstable_previewData: req.previewData })
```
joshuawootonn, line 70: Just to make sure I am understanding currently:

1. proxy passes `unstable_siteVersion: MakeswiftSiteVersion.Working` as previewData 
2. API handler passes that along to the MakeswiftClient 
3. This add headers to how MakeswiftClient fetches data from our API, which results in preview data instead of production data.

joshuawootonn (approved):
LGTM!

packages/runtime/src/next/api-handler/index.ts
migueloller: Yup!

PR #444, opened 2023-05-08
InlineModePlugin for `unstable_RichTextV2`

joshuawootonn:


.changeset/small-ducks-heal.md
```diff
@@ -0,0 +1,7 @@
+---
+'@makeswift/runtime': patch
+---
+
+Add `mode` option to `unstable_RichTextV2` control.
+
+Setting the mode of `RichTextV2` to `RichTextMode.Inline` locks down output to only include inline HTML elements. The use case we have in mind is allowing you to double-click edit button text, while protecting you from creating hydration mismatch errors.
```
joshuawootonn, line 7: Is  there are type of special indenting that I need to do changesets to format this correctly?

packages/runtime/src/runtimes/react/controls/rich-text-v2/rich-text-v2.tsx
```diff
@@ -130,17 +105,34 @@ export function useRichTextV2(data: RichTextV2ControlData, control: RichTextV2Co
...
       }
+
+      plugins.forEach(plugin => plugin?.onKeyDown?.(e))
     },
     [control, editor],
   )
 
+  const handleKeyUp = useCallback(
+    (e: KeyboardEvent) => {
+      if (editMode === BuilderEditMode.CONTENT) e.preventDefault()
```
joshuawootonn, line 124: Repro'd this in codesandbox and it is the default behavior for the events to propagate out to elements outside of content editable.

migueloller: Event propagation and default behavior are separate things (i.e., `event.stopPropagation()` and `event.preventDefault()`) so I'm not understanding what you mean in this comment

migueloller (commented):
Looking good!

PR #449, opened 2023-05-11
feat: block plugin v2

joshuawootonn:


packages/runtime/src/runtimes/react/controls/rich-text-v2/EditableTextV2/editable-text-v2.tsx
```diff
@@ -77,7 +79,57 @@ export function EditableTextV2({ text, control }: Props) {
   }, [editor, control])
 
   const renderElement = useCallback(({ attributes, children, element }: RenderElementProps) => {
+    const blockStyles = [useStyle({ margin: 0 })]
```
joshuawootonn, line 82: This is temporary. When I work on composable render functionality this would be moved into the block plugin

joshuawootonn:
@migueloller I really am just looking for an review on the last commit. 

The first two commits 1. Fix the bug I mentioned in standup yesterday for `unstable_RichTextV2` and 2. move the slate types to a new location.

joshuawootonn:
https://github.com/makeswift/makeswift/assets/20950876/12a1704a-6e3e-493c-9900-57d02234cd82

migueloller:
> CleanShot.2023-05-11.at.16.53.44.mp4

Amazing! Will review soon

packages/runtime/src/builder/serialization/control-serialization.ts
```diff
@@ -866,6 +870,52 @@ function deserializeRichTextControl(
   return { ...serializedControl, options: deserializedOptions }
 }
 
+type RichTextV2PluginControlDefinition<T extends ControlDefinition> = {
+  definition: T
+}
+
+type RichTextV2Plugin = {
+  controls?: RichTextV2PluginControlDefinition<ControlDefinition>[]
```
migueloller, line 878: I think we should change this to just `control` or `type` and if you want to compose multiple controls, you can use `Shape`. What do you think?

packages/runtime/src/builder/serialization/control-serialization.ts
```diff
@@ -866,6 +870,52 @@ function deserializeRichTextControl(
...
+}
+
+type RichTextV2Config = {
+  plugins?: RichTextV2Plugin[]
+  mode?: RichTextV2Mode
+}
```
migueloller, line 884: Should this type and the other 2 above this be in a separate file? They don't seem to be related to serialization specifically

packages/runtime/src/runtimes/react/controls/rich-text-v2/EditableTextV2/editable-text-v2.tsx
migueloller: Cool. Do you want to add a TODO comment?

packages/runtime/src/runtimes/react/controls/rich-text-v2/EditableTextV2/editable-text-v2.tsx
```diff
@@ -77,7 +79,57 @@ export function EditableTextV2({ text, control }: Props) {
...
+
     switch (element.type) {
+      case BlockType.Text:
+        return (
+          <span {...attributes} className={cx(...blockStyles)}>
```
migueloller, line 87: Should we be composing `attributes.className` here? I imagine once you make the text align control which will compose rendering that might become important?

packages/runtime/src/slate/BlockPlugin/index.ts
```diff
@@ -32,3 +36,68 @@ export function withBlock(editor: Editor) {
 
   return editor
 }
+
+export const BlockPlugin = createRichTextV2Plugin({
+  withPlugin: withBlock,
+  controls: [
+    createRichTextV2PluginControl({
```
migueloller, line 43: It seems weird that there's a special function to create a control specifically for the rich text v2 plugin. It should be able to take control definitions directly, without having to wrap them with anything.

Also, as I mentioned in another comment, I think we should just take a single control and use `Shape` if you want multiple

migueloller (changes requested):

packages/runtime/src/builder/serialization/control-serialization.ts
joshuawootonn: From my scan of the file, it seemed like there were a lot of partial types that were the result of serialization written inline above the serialization functions. I guess I'm confused about where else I would put these partial types.

Would it be better if I defined them here, but based them on the originals using the `Omit` keyword to remove anything that won't be passed to the builder?

migueloller: Ah, it threw me off because it didn’t have “serializable” on the name and I thought these were the control types. We should be explicit with naming to avoid confusion

packages/runtime/src/builder/serialization/control-serialization.ts
joshuawootonn: Let's say you create a plugin with `Shape` (or same principle goes for `List`)
```
export const BlockPlugin = createRichTextV2Plugin({
  withPlugin: withBlock,
  controls: Shape({
    type: {
      foo: {
        definition: Select({
          label: 'Block',
          options: [
            {
              value: BlockType.Heading1,
              label: 'Heading1',
            },
            /* ... */
            {
              value: BlockType.Paragraph,
              label: 'Paragraph',
            },
          ],
          defaultValue: BlockType.Paragraph,
      },
      onChange: () => {},
      getValue: () => {}
    bar: // Same thing
```

What do the keys (`foo` and `bar`) map to? There are no props or schema that we are changing. 

The reason I designed it as I did it is that these are stateless ways of mutating whatever is currently selected.  

We do need them to have a way of specifying multiple controls because a plugin isn't 1on1 with a control.

joshuawootonn: After some more thought. Our current plugins are only going to use one control per plugin. I might just move down to 
```suggestion
  controls?: RichTextV2PluginControlDefinition<ControlDefinition>
```
because I am really struggling with how `List` and `Shape` would compose `getValue` / `onChange`

migueloller: What you have above is not how the Shape control’s API works. You just have a value and on change per plugin, not per control. In fact, what you said in your second comment is what I meant. One control per plugin is fine, of you need a group of values you can use List or Shape

packages/runtime/src/runtimes/react/controls/rich-text-v2/EditableTextV2/editable-text-v2.tsx
joshuawootonn: Currently  `attributes` doesn't contain a classname because there is no way to compose the rendering just yet. 

CleanShot 2023-05-12 at 16 39 08@2x

It might have it when I am done with textAlign though! TBH I can't totally visualize it yet.

PR #452, opened 2023-05-16
chore: add testing as part of CI pipeline

chanceaclark:
## What/why?

- Adds gh-action to automate test runs on PRs and merges to main
- Adds `.nvmrc` file to define which version on node to use.

Note: didn't include linting in this PR as there's some prettier issues that I want to fix in another PR.


migueloller (approved):
Looks good to me!

Got a question, how come you went with Node.js v16 instead of v18?

chanceaclark:
@migueloller the release workflow was using [v16](https://github.com/makeswift/makeswift/blob/main/.github/workflows/release.yml#L23) so just aligned it with that. We can bump to 18 if we want 🤔

migueloller:
> @migueloller the release workflow was using [v16](https://github.com/makeswift/makeswift/blob/main/.github/workflows/release.yml#L23) so just aligned it with that. We can bump to 18 if we want 🤔

Ahh, got it. We should probably upgrade to Node.js v18 for all workflows. That can be handled in a separate PR, though

chanceaclark:
Sounds good!

Just pushed a rebase and quick line in `turbo.json` to ensure the build has been run before. I had a feeling it wouldn't run the first time.

migueloller:
Ok, cool! It looks like there's still some issues with the test. I'll push up a fix for that...

PR #469, opened 2023-05-23
Unstable localization without merging without backend

fikrikarim:


packages/runtime/src/next/client.ts
```diff
@@ -362,9 +362,9 @@ export class Makeswift {
 
   private async getPageSnapshotByPageId(
     pageId: string,
-    { preview = false }: { preview?: boolean } = {},
+    { preview = false, locale = '' }: { preview?: boolean; locale?: string } = {},
```
migueloller, line 365: Why default it to `''`, which is an invalid locale?

packages/runtime/src/state/modules/locales.ts
```diff
@@ -0,0 +1,56 @@
+import { Action, ActionTypes } from '../actions'
+
+export type Locale = string
```
migueloller, line 3: I would like us to strongly type this

migueloller (changes requested):
Pretty straightforward

packages/runtime/src/next/client.ts
```diff
@@ -362,9 +362,9 @@ export class Makeswift {
...
   ): Promise<MakeswiftPageSnapshot | null> {
-    const searchParams = new URLSearchParams({ preview: String(preview) })
+    const searchParams = new URLSearchParams({ preview: String(preview), locale })
```
migueloller, line 367: The `locale` param is just being ignored by the API right now, right?

packages/runtime/src/state/modules/locales.ts
migueloller: Hmm... I thought that ISO standards were comprehensive and didn't change. Maybe we don't use the unicode spec and instead use the ISO standards? ISO-639 and ISO-3166 was it?

fikrikarim: Yes if we just want to have `${language_code}-${country_code}` then it'll be easier to manage. But still, like the ISO-3166 is [regularly updated](https://www.iso.org/iso-3166-country-codes.html:
> ISO 3166 is regularly updated to reflect changes in country names and subdivisions. These changes are done by the ISO 3166 Maintenance Agency (ISO 3166/MA).

I was pointing to the Unicode specification because that's what the Next.js docs is using.

migueloller: Ok. Let's just do arbitrary string values, then

packages/runtime/src/state/modules/locales.ts
```diff
@@ -0,0 +1,56 @@
...
+  locales: string[]
+  defaultLocale: string
+}
+
+export function parseLocalesInput(input: LocalesInput): State {
+  return {
+    ...input,
+    locale: null,
+  }
+}
```
migueloller, line 56: There doesn't seem to be any "parsing" going on here. Maybe we just want this logic to be part of `getInitialState`?

migueloller (approved):

PR #470, opened 2023-05-23
spike: style control v2

joshuawootonn:


packages/runtime/src/runtimes/react/controls/style-v2.ts
```diff
@@ -0,0 +1,72 @@
...
+          findBreakpointOverride(breakpoints, styleControlData, breakpoint.id)?.value,
+        ),
+      }
+    }, {}),
+  }
+}
+
+export type StyleV2ControlFormattedValue = CSSObject | string
+
+export type ResolveStyleV2ControlFormattedValue<T extends ControlDefinition> =
```
joshuawootonn, line 44: This `ResolveXXXXXXXX` is something I saw in the Image control for Image format, but the `ControlValue` of the image control doesn't seem to be typed based on which format you use in `register-components.ts`, I remember bringing this up when I was working on www.makeswift.com. I might have enough context to understand why.

Why does `useImageControlValue` return `ImageControlValue` instead of `ResolveImageControlValue`

packages/runtime/src/runtimes/react/controls/style-v2.ts
```diff
@@ -0,0 +1,69 @@
...
+      ? string | undefined
+      : T['config']['format'] extends typeof StyleV2Format.Object
+      ? CSSObject | undefined
+      : never
+    : never
+
+export function useFormattedStyleV2<T extends StyleV2ControlDefinition>(
+  styleControlData: StyleV2ControlData,
+  controlDefinition: T,
+): StyleV2ControlFormattedValue {
```
joshuawootonn, line 55: When I try to use `ResolveStyleV2ControlFormattedValue` all the returns in the function as "Unassignable to the generic type"

packages/runtime/src/controls/style-v2.ts
```diff
@@ -0,0 +1,79 @@
...
+ * * A SO user implementing a tail recursive type: https://stackoverflow.com/questions/72370456/tail-recursion-elimination-on-conditional-types-doesnt-work
+ *
+ *
+ * This is what the type should be:
+ * ```
+ * export type StyleV2ControlData<T extends StyleV2ControlDefinition = StyleV2ControlDefinition> =
+ *   ResponsiveValue<ControlDefinitionData<T['config']['type']>>
+ * ```
+ */
+export type StyleV2ControlData = ResponsiveValue<any>
```
joshuawootonn, line 56: Decided on any typing the control value here @migueloller. I think we are actually reaching a limitation of the conditional type for `ControlDefinitionData`

There might be a way to rewrite `ControlDefinitionData` so it is "tail recursive." I feel out of my league though so I am going to any type this for now.

packages/runtime/src/builder/serialization/controls/style-v2.ts
```diff
@@ -0,0 +1,24 @@
...
+      config,
+    },
+    transferables,
+  ] as [Serialize<StyleV2ControlDefinition>, Transferable[]]
+}
+
+export function deserializeStyleV2Control(
+  control: Serialize<StyleV2ControlDefinition>,
+): Deserialize<Serialize<StyleV2ControlDefinition>> {
+  // TODO: `as unknown` shouldn't be required here
```
joshuawootonn, line 22: I don't understand how the function serialization is working ATM. 

I didn't think we could serialize functions, but there are types (`SerializedFunction` etc) for serialized functions. 

Just admitting this because I don't really know where to start in fixing this `as unknown`.

packages/runtime/src/runtimes/react/controls/style-v2.ts
migueloller: > Why does useImageControlValue return ImageControlValue instead of ResolveImageControlValue

Seems to be a mistake

packages/runtime/src/runtimes/react/controls/style-v2.ts
migueloller: I don't understand what you're trying to say here

packages/runtime/src/controls/style-v2.ts
migueloller: We might want to put the `any` where the limitation is happening, i.e., the inference of the data instead of there in the underlying data definition, which will be used in various places

packages/runtime/src/builder/serialization/controls/style-v2.ts
migueloller: We serialize functions by serializing arguments and return values and sending them via `postMessage`. This means that serialized functions become `async`. An example of this is the `getItemLabel` function in the `List` control which we use to render a label in the builder panel

packages/runtime/src/builder/serialization/controls/style-v2.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export function serializeStyleV2Control(
+  control: StyleV2ControlDefinition,
+): [Serialize<StyleV2ControlDefinition>, Transferable[]] {
+  const { getStyle, ...config } = control.config
+
+  const transferables: Transferable[] = []
+  return [
+    {
+      ...control,
```
migueloller, line 12: You need to recursively serialize the control here, otherwise this will throw an error for controls that have values that can't be serialized by `postMessage`

packages/runtime/src/builder/serialization/controls/style-v2.ts
migueloller, line 22: What you'll need to do here is compose the deserialize function for other controls. Take a look at how other composition controls implement deserialization

packages/runtime/src/controls/style-v2.ts
```diff
@@ -0,0 +1,79 @@
...
+
+export const StyleV2Format = {
+  Class: 'makeswift::controls::style-v2::format::class',
+  Object: 'makeswift::controls::style-v2::format::object',
+} as const
+
+export type StyleV2Format = typeof StyleV2Format[keyof typeof StyleV2Format]
+
+export type StyleV2CSSObject = CSSObject
+export const useUnstableStyleV2ClassName = useStyle
```
migueloller, line 19: Can we keep "unstable" as a prefix always: i.e., `unstable_useStyleV2ClassName` so that we can easily find it across the codebase?

migueloller (changes requested):

packages/runtime/src/controls/style-v2.ts
joshuawootonn: Opting to keep it here. The error happens on typings for List at _random_ locations that pass the threshold of what TS supports. I think it's better to have the any at one locations with the comment explaining it's presence.

packages/runtime/src/runtimes/react/controls/style-v2.ts
joshuawootonn: I can ignore this for now since I have to cast the value anyways.

--- 


Further explanation: 

Similar to the Image control I have two types. The one that is prefixed with `Resolve` is generic and should be useable with `useFormattedStyleV2`, but I am struggling to do so. 


CleanShot 2023-05-24 at 10 15 21@2x

packages/runtime/src/controls/style-v2.ts
migueloller: What I'm suggesting would be just at 1 location, `StyleV2ControlData<any>` in the conditional type `ControlData`. That way in this file when we implement functions such as `copy`, `merge`, etc. those can be properly typed

migueloller:
@joshuawootonn, something else to do that I realized we haven't properly implemented in our other controls is introspection. Don't have to do it in this PR since it's still `unstable_`, but if you decide not to, then let's make an issue to track it

packages/runtime/src/controls/style-v2.ts
joshuawootonn: I wish I had responded to this last week. I did try this and I remember getting the same error. There were a lot of roadblocks to this work, so I ended up just skipping this one and proceeding. 

When I ultimately work on this I plan to deal with this problem

packages/runtime/src/runtimes/react/controls/style-v2.ts
joshuawootonn: This type was ultimately removed so I am resolving this conversation

PR #471, opened 2023-05-26
chore: add madge to check for circular deps

chanceaclark:
## What/why?

Adds `madge` so we can check and monitory for circular dependency issues. Currently, the task should be **non-blocking** but providing it in CI so we can keep track of it.

ping @migueloller

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,3 +37,7 @@ jobs:
 
       - name: Test packages
         run: yarn run test
+
+      - name: Circular dependency check
+        continue-on-error: true
+        run: yarn run madge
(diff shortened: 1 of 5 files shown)


migueloller (approved):
Looks good!

migueloller:
I'm trying to look at the CI run and read the output but it doesn't seem to be very useful? Maybe I need to run this locally to make sense of it? Also, it seems to be picking up some build outputs, which is weird given you're specifying `./src`

chanceaclark:
> I'm trying to look at the CI run and read the output but it doesn't seem to be very useful? Maybe I need to run this locally to make sense of it? Also, it seems to be picking up some build outputs, which is weird given you're specifying `./src`

It's really strange... It has it's own dependency tree and nothing from `src/` was including anything:
Screenshot 2023-05-26 at 08 17 45

I went ahead and excluded `dist` files for now as I don't have a lot of time to figure out the underlying issue.

migueloller (approved):

PR #473, opened 2023-05-26
fix: change runtime to use ComponentIcon const

arvinpoddar:
This breaking change uses an ComponentIcon const instead of a string literal union in the ComponentMeta. With this change, icons are no longer specified with a size attached. All component registrations have been updated to reflect this update.

### DIFF
--- a/.changeset/nine-windows-join.md
+++ b/.changeset/nine-windows-join.md
@@ -1,0 +1,17 @@
+---
+"@makeswift/runtime": minor
+---
+
+BREAKING: When registering component icons, use the `ComponentIcon` enum (available under `@makeswift/runtime`) instead of the original string values. Below is a table of the deprecated string values and their new enum equivalent:
+
+| Removed           | Use Instead (enum)          |
+|-------------------|-----------------------------|
+| `'Carousel40'`    | `ComponentIcon.Carousel`    |
+| `'Code40'`        | `ComponentIcon.Code`        |
+| `'Countdown40'`   | `ComponentIcon.Countdown`   |
+| `'Cube40'`        | `ComponentIcon.Cube`        |
+| `'Divider40'`     | `ComponentIcon.Divider`     |
+| `'Form40'`        | `ComponentIcon.Form`        |
+| `'Navigation40'`  | `ComponentIcon.Navigation`  |
+| `'SocialLinks40'` | `ComponentIcon.SocialLinks` |
+| `'Video40'`       | `ComponentIcon.Video`       |
--- a/packages/runtime/src/state/modules/components-meta.ts
+++ b/packages/runtime/src/state/modules/components-meta.ts
@@ -1,15 +1,18 @@
 import { Action, ActionTypes } from '../actions'
 
-export type ComponentIcon =
-  | 'Carousel40'
-  | 'Code40'
-  | 'Countdown40'
-  | 'Cube40'
-  | 'Divider40'
-  | 'Form40'
-  | 'Navigation40'
-  | 'SocialLinks40'
-  | 'Video40'
+export const ComponentIcon = {
+  Carousel: 'carousel',
+  Code: 'code',
+  Countdown: 'countdown',
+  Cube: 'cube',
+  Divider: 'divider',
+  Form: 'form',
+  Navigation: 'navigation',
+  SocialLinks: 'social-links',
+  Video: 'video',
+} as const
+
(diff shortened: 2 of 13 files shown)


.changeset/nine-windows-join.md
```diff
@@ -0,0 +1,17 @@
+---
+"@makeswift/runtime": patch
```
migueloller, line 2: The changeset should be for a minor version, not patch

.changeset/nine-windows-join.md
```diff
@@ -0,0 +1,17 @@
...
+---
+
+BREAKING: When registering component icons, use the `ComponentIcon` enum (available under `@makeswift/runtime/state/components-meta`) instead of the original string values. Below is a table of the deprecated string values and their new enum equivalent:
```
migueloller, line 5: We should not export internals like the `component-meta` this way. Instead the export should be available from `@makeswift/runtime` directly:

```ts
import { ComponentIcon } from '@makeswift/runtime'
```

.changeset/nine-windows-join.md
```diff
@@ -0,0 +1,17 @@
...
+
+| Deprecated        | Use Instead (enum)          |
```
migueloller, line 7: Should we change the wording from "Deprecated" to "Removed"?

packages/runtime/package.json
```diff
@@ -181,4 +189,4 @@
     "vite": "^2.8.6",
     "vitest": "^0.7.4"
   }
-}
+}
```
migueloller, line 192: Missing a newline at the end of the file

packages/runtime/src/state/modules/components-meta.ts
```diff
@@ -1,15 +1,16 @@
...
-  | 'Carousel40'
-  | 'Code40'
-  | 'Countdown40'
-  | 'Cube40'
-  | 'Divider40'
-  | 'Form40'
-  | 'Navigation40'
-  | 'SocialLinks40'
-  | 'Video40'
+export enum ComponentIcon {
```
migueloller, line 3: We avoid using TypeScript enums because it's not standard JavaScript. We prefer to use TypeScript in a way were all types can just be compiled away and the JavaScript isn't affected.

https://dev.to/ivanzm123/dont-use-enums-in-typescript-they-are-very-dangerous-57bh

Instead define it like this:

```ts
const ComponentIcon = {
} as const
type ComponentIcon = typeof ComponentIcon[keyof typeof ComponentIcon]
```

packages/runtime/state/components-meta.js
```diff
@@ -0,0 +1 @@
+module.exports = require("../dist/state/components-meta.cjs")
```
migueloller, line 1: No newline at the end of the file (although I'm guessing this file will be deleted)

packages/runtime/src/state/modules/components-meta.ts
```diff
@@ -1,15 +1,16 @@
...
+  Carousel = 'Carousel',
+  Code = 'Code',
+  Countdown = 'Countdown',
+  Cube = 'Cube',
+  Divider = 'Divider',
+  Form = 'Form',
+  Navigation = 'Navigation',
+  SocialLinks = 'SocialLinks',
+  Video = 'Video',
```
migueloller, line 12: Can we make these values lower-kebab-case?

migueloller (changes requested):

migueloller:
One more thing, @arvinpoddar, can you please fixup the changeset commit with the commit that introduces the changes?

packages/runtime/vite.config.ts
```diff
@@ -30,7 +30,7 @@ export default defineConfig({
         builder: path.resolve(__dirname, 'src', 'builder'),
         controls: path.resolve(__dirname, 'src', 'controls'),
         slate: path.resolve(__dirname, 'src', 'slate'),
-        ['state/breakpoints']: path.resolve(__dirname, 'src', 'state', 'modules', 'breakpoints'),
+        ['state/breakpoints']: path.resolve(__dirname, 'src', 'state', 'modules', 'breakpoints')
```
migueloller, line 33: A trailing comma was removed from there that shouldn't have been

migueloller (changes requested):
Looking good! We'll want to make sure we update the support for icons in the app before we merge this, though

migueloller (approved):

PR #474, opened 2023-05-30
feat: add new icon support to runtime

arvinpoddar:
Introduces 13 new icons to the Makeswift runtime

### DIFF
--- a/.changeset/silent-tomatoes-cheat.md
+++ b/.changeset/silent-tomatoes-cheat.md
@@ -1,0 +1,5 @@
+---
+'@makeswift/runtime': patch
+---
+
+Adds 13 new icon options to the runtime that can be used when registering components.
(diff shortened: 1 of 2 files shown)


.changeset/silent-tomatoes-cheat.md
```diff
@@ -0,0 +1,5 @@
+---
+'@makeswift/runtime': minor
```
migueloller, line 2: This one should probably just be a `patch` since there's no breaking changes

.changeset/silent-tomatoes-cheat.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Add new icon support to runtime
```
migueloller, line 5: For the changesets we probably want to use a more "human" tone instead of the tone we use for commit messages. i.e., feel free to use full sentences, punctuation, etc.

migueloller (changes requested):

PR #476, opened 2023-05-31
chore: remove builder circular dep

chanceaclark:
## What/why?

This aims to remove the following circular dependencies:
Screenshot 2023-05-31 at 12 16 30

There are two big changes:
- Moved types into it's own file to reduce the length of the `control-serialization.ts` file.
- Moved modules that were recursively calling each other into the same file.


packages/runtime/src/builder/serialization/control-serialization.ts
```diff
@@ -12,90 +11,121 @@ import {
...
+  SerializedResponsiveNumberControl,
+  SerializedResponsiveSelectControl,
+  SerializedRichTextControl,
+  SerializedShapeControl,
+  SerializedTextInputControl,
+  SerializedTextStyleControl,
+  SerializedTypeaheadControl,
+} from './types'
+
+function serializeShapeControlDefinition(
```
chanceaclark, line 93: This is one the big changes: I ended up moving a bunch of the modules that are recursively dependent into the same file. It sort of sucks that we can't have these be their own modules but this is sort of the best approach to avoid these circular deps. I'm open to recommendations on how to split these up even better, like splitting this up into `serialization` and `deserialization` modules 🙏 

cc. @migueloller @joshuawootonn 

Here is also a [SO article](https://stackoverflow.com/questions/63414806/how-to-avoid-circular-dependency-between-modules-when-working-with-recursive-fun) that I found useful here.

packages/runtime/src/builder/serialization/types.ts
```diff
@@ -0,0 +1,517 @@
+import {
```
chanceaclark, line 1: `control-serialization.ts` was getting pretty long, so I moved the types into it's own file. This is the other big change.

packages/runtime/src/builder/serialization/index.ts
```diff
@@ -1,3 +1,3 @@
-export * from './control-serialization'
-export * from './function-serialization'
+export { serializeControls, deserializeControls } from './control-serialization'
 export type { Serialize, Deserialize } from './controls/types'
+export type { SerializedControl } from './types'
```
chanceaclark, line 3: I went ahead and only exported what was needed, but @migueloller let me know if I am missing an export that is being used in the builder.

migueloller:
@chanceaclark, should removing _all_ circular dependencies be a non-goal here? I feel like for certain structures, like mutual recursion as mentioned in the SO article you shared, it's natural to have circular dependencies. As long as the modules just have function declarations that depend on each other, there shouldn't be a problem. I think that the big issue that we had was that instead of having the serialization domain depend on each other, the dependencies were through control modules via index files that re-exported everything. This meant that things that we wouldn't want to have circular dependencies ended up having them. Think about serialization specifically, here's how I think it could work:

- Types are leaves and while we might want to colocate type files with the control folders, we still want to import types directly (e.g., `import { SomeType } from './path/to/module/type'` instead of `import { SomeType } from './path/to/module'`).
  - An alternative to colocation is to have a "types" module that is grouped by domain (e.g., `import { SomeType } from './types'` where the `types` folder has a file or folder for each control with the relevant types.
- Serialization and deserialization rely on mutual recursion, so there _will_ be circular dependencies. These circular dependencies, though, should be just between the serialization and deserialization functions and _not_ the entire modules.
  - Just like types, we could use colocation where `some-control/serialization.ts` has the logic and we import directly from that file (not from `some-control/index.ts`) or have a "serialization" module where we have a file/folder per control
- Serialization will depend on types, but since types are leaves, there's no circular dependencies
- Serialization will depend on other serialization functions, which might in turn depend on the "root" serialization function, but that's ok, it's a small local cycle that is unavoidable due to the recursive nature of the function
  - We _could_ put everything in a single file and avoid circular file dependencies, but this really doesn't buy us anything other than removing the circular dependency, but I think that should be a non-goal. I would rather have the files be well-organized rather than having _huge_ files that grow as we add more controls and inevitably have to include legacy controls

Looking at the screenshot you shared above, that circular dependency doesn't seem problematic since it's "local" and simply there because files are split-up for organization purposes.

What are your thoughts on this?

PR #478, opened 2023-06-01
chore: remove circular deps for MakeswiftApiHandler

chanceaclark:
## What/why?

Removes the circular dependencies for the `MakeswiftApiHandler` route and also starts to add `use client` directives where needed.

_**Note: The modules using `use client` still won't work since rollup strips these directives on build.**_

## Next steps
We now got it to a point where we can start to build a new module for supporting NextJS route handlers. We need to determine what the public API will need to look like before we can continue down that path.

## Testing/proof
Got the app directory working up to the point of needing a new interface for route handlers:
Screenshot 2023-06-01 at 08 48 06

### DIFF
--- a/packages/runtime/src/api/react.tsx
+++ b/packages/runtime/src/api/react.tsx
@@ -1,3 +1,5 @@
+'use client'
+
 import { createContext, ReactNode, useContext } from 'react'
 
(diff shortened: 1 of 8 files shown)


migueloller (approved):
Looks good!

PR #484, opened 2023-06-09
Locale

fikrikarim:


packages/runtime/src/next/client.ts
```diff
@@ -404,14 +404,20 @@ export class Makeswift {
 
   async getPageSnapshot(
     pathname: string,
-    { preview: previewOverride = false }: { preview?: boolean } = {},
+    { preview: previewOverride = false, locale }: { preview?: boolean; locale?: string } = {},
```
migueloller, line 407: Let's make this `unstable_locale` until we're ready for a stable release of our localization features

packages/runtime/types/locale.ts
migueloller, line 1: Let's use [Zod enums](https://zod.dev/?id=zod-enums), they were designed for this

packages/runtime/types/locale.ts
```diff
@@ -0,0 +1,471 @@
...
+    return value as Locale
+  }
+
+  ctx.addIssue({
+    code: z.ZodIssueCode.custom,
+    message: `Locale is not supported: '${value}'. Make sure the locale is in the format of 'language', e.g. 'en', or 'language-country', e.g. 'en-US'.`,
+  })
+
+  return z.NEVER
+})
```
migueloller, line 24: Why not have the transform return an `Intl.Locale`? It seems weird to cast to a special `Locale` type when we can use the native `Intl.Locale`

migueloller (changes requested):

packages/runtime/src/next/client.ts
fikrikarim: True. Updated.

packages/runtime/types/locale.ts
fikrikarim: I was thinking of using the Zod enums, but could it handle combination type for `${LanguageCode}-${CountryCode}`?

packages/runtime/types/locale.ts
fikrikarim: I was thinking about that too. I remember that we could just use `Intl.Locale` everywhere. I decided to just use string because that's what we use in the `State.Locale`. Do you also want to store `Intl.Locale` in the state?

packages/runtime/types/locale.ts
migueloller: Can you be more specific? I don't understand what you mean

packages/runtime/types/locale.ts
migueloller: We need to keep our domains separate. Whether we use a `string` or whatever else in the state shouldn't necessarily drive what's in the public API. But yeah, I don't see why we can't just keep an `Intl.Locale` in state unless we rely on it being serializable. Regardless, it would be very easy to keep the serialized string representation in state and then have a selector that transforms it into a `Intl.Locale` whenever read from state

fikrikarim: I can see that we can use `Intl.Locale` internally when reading from the state.

The custom `Locale` type here is used mainly to type the input in the `ReactRuntime`:
```
export const runtime = new ReactRuntime({
  unstable_i18n: {
    locales: ["es", "en-US", "es-SE", "id-ID"],
    defaultLocale: "en-US",
  },
});
```

packages/runtime/types/locale.ts
migueloller: I meant for the JavaScript sets you're creating. You can make a Zod enum and then use `parse` to validate their types instead of casting. As for the locale input, I think what you're doing with making a new `Intl.Locale` and then checking that the language and region codes match the allowed values works well

packages/runtime/types/locale.ts
migueloller: That type has nothing to do with `localeSchema`, though. It's just defined as a combination of the language code and country code enums, which can be used at the input of the API and should still be independent of what's stored internally.

The values allowed as inputs can follow the locale string type while internally we can always use `Intl.Locale`

fikrikarim: I updated the custom locale string to be just used when parsing the input from the user, and I made it so that we use `Intl.Locale` on the selector.

We still need to use `string` on the actions, because we're going to send it through the `postMessage`.

fikrikarim: Yes, yes. I called it `LocaleInput` though. I can change it if needed.

migueloller: Let's change it to `LocaleString` since it's not necessarily just an input (e.g., it can be used as an output in certain cases)

packages/runtime/types/locale.ts
```diff
@@ -0,0 +1,475 @@
+import { z } from 'zod'
+
+export const localeInputSchema = z.string().transform((value, ctx) => {
+  const { language, region } = new Intl.Locale(value)
+
+  if (
+    languageCodeSchema.safeParse(language as LanguageCode).success &&
```
migueloller, line 7: Why cast to `LanguageCode` here?

packages/runtime/types/locale.ts
```diff
@@ -0,0 +1,475 @@
...
+    region == null &&
+    value === language
+  ) {
+    return value as LocaleInput
+  }
+
+  if (
+    languageCodeSchema.safeParse(language as LanguageCode).success &&
+    countryCodeSchema.safeParse(region as CountryCode).success &&
```
migueloller, line 16: Why cast `LanguageCode` and `CountryCode`?

packages/runtime/types/locale.ts
```diff
@@ -0,0 +1,475 @@
...
+
+export const localeInputSchema = z.string().transform((value, ctx) => {
+  const { language, region } = new Intl.Locale(value)
+
+  if (
+    languageCodeSchema.safeParse(language as LanguageCode).success &&
+    region == null &&
+    value === language
+  ) {
+    return value as LocaleInput
```
migueloller, line 11: This casting shouldn't be necessary if we're using Zod and the parsed result

packages/runtime/src/state/modules/locales.ts
```diff
@@ -1,13 +1,11 @@
...
-export type Locales = Locale[]
+import { LocaleInput, localeInputSchema } from '../../../types/locale'
 
 export type State = {
-  locales: Locales
-  locale: Locale | null
-  defaultLocale: Locale | null
+  locales: string[]
+  locale: string | null
+  defaultLocale: string | null
```
migueloller, line 8: Why aren't we constraining the state to be `LocaleString`?

packages/runtime/src/state/actions.ts
migueloller, line 1: For the locale action payloads, why are we using `string` instead of `LocaleString`?

migueloller (changes requested):

packages/runtime/types/locale.ts
fikrikarim: Updated

packages/runtime/src/state/actions.ts
migueloller: Got it. I guess the action creators could take `Intl.Locale` and then call `toString()` and parse them so the builder doesn't have to. Not a big deal to leave them as `string`, though. Let's please not add another entry point. I'm trying to make progress with @chanceaclark on RSC support and we're in entry point hell right now

packages/runtime/types/locale.ts
migueloller: But since `LocaleString` is just ``${CountryCode}-${LanguageCode}` | `${CountryCode}`` wouldn't that be inferred by TypeScript?

packages/runtime/src/state/react-builder-preview.ts
```diff
@@ -500,9 +501,11 @@ export function messageChannelMiddleware(): Middleware<Dispatch, State, Dispatch
       const breakpoints = ReactPage.getBreakpoints(state)
       messageChannel.port1.postMessage(setBreakpoints(breakpoints))
 
-      const locales = ReactPage.getLocales(state).locales
-      const defaultLocale = ReactPage.getLocales(state).defaultLocale
-      const routerLocale = Router.locale
+      const locales = ReactPage.getLocales(state).map(locale =>
+        localeStringSchema.parse(locale.toString()),
+      )
+      const defaultLocale = localeStringSchema.parse(ReactPage.getDefaultLocale(state)?.toString())
```
migueloller, line 507: Is the idea that this will just throw an error if `getDefaultLocale` returns a nullish value?

migueloller (dismissed):

packages/runtime/types/locale.ts
fikrikarim: Cool. Finally I could make it working by using
```ts
return `${languageResult.data}-${countryCodeResult.data}` as const
```

packages/runtime/src/state/actions.ts
fikrikarim: Hahah, I see. Great. Making the action creators take `Intl.Locale` and parsing the locale there works very well. Updated.

packages/runtime/src/state/react-builder-preview.ts
fikrikarim: Yeah, that was bad. Updated so we parse in the action creator instead.

PR #485, opened 2023-06-09
Update introspection for composable controls

joshuawootonn:
* Create generic introspection functions for List and Shape
* Add pageId introspection for Link Control 

josh/prd-2056
josh/prd-2058

### DIFF
--- a/packages/runtime/src/next/client.ts
+++ b/packages/runtime/src/next/client.ts
@@ -35,17 +35,10 @@ import {
   TypographyQueryVariables,
 } from '../api/graphql/generated/types'
 import { CacheData } from '../api/react'
-import { ShapeControlData, ShapeControlType } from '../controls/shape'
-import { ListControlData, ListControlType } from '../controls/list'
-import {
-  ListValue,
-  ShapeValue,
-  Types,
-  Descriptor as PropControllerDescriptor,
-} from '../prop-controllers/descriptors'
+import { Descriptor as PropControllerDescriptor } from '../prop-controllers/descriptors'
 import {
   getElementChildren,
-  getElementSwatchIds,
+  getSwatchIds,
   getFileIds,
   getPageIds,
   getTableIds,
@@ -277,7 +270,7 @@ export class Makeswift {
         props: ElementData['props'],
       ) {
         Object.entries(elementDescriptors).forEach(([propName, descriptor]) => {
-          getElementSwatchIds(descriptor, props[propName]).forEach(swatchId => {
+          getSwatchIds(descriptor, props[propName]).forEach(swatchId => {
             swatchIds.add(swatchId)
           })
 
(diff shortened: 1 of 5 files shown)


packages/runtime/src/prop-controllers/descriptors.ts
```diff
@@ -814,6 +824,18 @@ export function Shape<T extends Record<string, PanelDescriptor>>(
...
+  value: ShapeValue | undefined,
+  func: (definition: Descriptor, data: Data) => T[],
+): T[] {
+  if (value == null) return []
+
+  return Object.entries(descriptor.options.type).flatMap(([key, definition]) =>
+    func(definition, value[key]),
+  )
+}
+
```
joshuawootonn, line 838: I was not sure where to place these functions. I am open to feedback.

packages/runtime/src/next/client.ts
```diff
@@ -298,48 +291,6 @@ export class Makeswift {
               remaining.push(child)
             }
           })
-
-          if (descriptor.type === ShapeControlType) {
-            const prop = props[propName] as ShapeControlData
-
-            if (prop == null) return
-
-            getResourcesFromElementDescriptors(descriptor.config.type, prop)
```
joshuawootonn, line 307: This feels wrong. Shouldn't we be looping through the keys and calling `getResources....` on `prop[key]`?

joshuawootonn:
@migueloller I requested feedback on this draft. The pattern is ready for review, but I still need to propagate the changes to stylev2 and richtextv2.

joshuawootonn:
On second thought I would rather just ship this and add the richtextv2 and stylev2 stuff as part of the release milestone. 

https://linear.app/makeswift/issue/PRD-2207/add-composable-introspection-for-richtextv2-and-stylev2

joshuawootonn:
Hey, @migueloller! Can you review this? I want to make sure we are on the same page about how introspections should be handled / I want to make sure that I should run introspections on the old list and shape controls.

migueloller:
On it!

packages/runtime/src/next/client.ts
migueloller: It's _very_ confusing, but it's not wrong. What's happening is that `getElementDescriptors` expects an object of propName -> descriptor and a props object. Since a Shape control keeps a map from shape key -> descriptor then it works out. Another way to think about it is that the "control" for element's props is the shape of the props

migueloller (commented):
@joshuawootonn, this looks good. I'll admit, though, that I wasn't expecting an `introspectListData` and `introspectShapeData` function and then calling that on each control. Instead, I thought we'd have a `getSwatchIds`, `getPageIds`, etc. function for the List and Shape controls.

While having just the one function makes it so there's less to type, it does make the abstraction a bit leaky by not having all controls just have a `getSwatchIds`, etc. type of function. Another way to think about this is: imagine that we had a Control interface that each control had to implement. This interface might look something like this:

```ts
interface Control<T extends Data> {
  getSwatchIds(definition: this, data: T): string[]

  // etc.
}
```

You can think of the control modules (files) as basically implementing this interface.

You could keep `introspectListData`, etc. to avoid duplicating that logic, but instead of exporting it, you could use it internally when implementing all the introspection functions.

Thoughts?

joshuawootonn:
@migueloller I ultimately decided to merge this as is. 

There is some TS work with getting this working nicely with minimal repetition, and I am going to handle that in: https://linear.app/makeswift/issue/PRD-2207/add-composable-introspection-for-richtextv2-and-stylev2

packages/runtime/src/next/client.ts
joshuawootonn: Got it. That makes sense. It's the same reason we call `getResourcesFromElementDescriptors(elementDescriptors, element.props)` and internally it iterates the props internally.

migueloller: Yeah, exactly. Again, it's very confusing since the function name has _element_ in it even though it operates on arbitrary data, but that's fine. It's going to the farm now...

joshuawootonn:
Here is what I am thinking based on our convo @migueloller.  Methods in each of these composable controls like this one for `getElementChildren` and `Shape`

```ts
import { getElementChildren as getPropControllersElementChildren } from '../prop-controllers/introspection'

/* ... */

export function getElementChildren(
  definition: ShapeControlDefinition,
  value: ShapeControlData | undefined,
): Element[] {
  return value
    ? Object.entries(definition.config.type).flatMap(([key, definition]) =>
        getPropControllerElementChildren(definition, value[key]),
      )
    : []
}
```

migueloller:
Looks good!

PR #490, opened 2023-06-15
feat: create Typography control / plugin

joshuawootonn:


packages/runtime/src/components/builtin/Text/components/Leaf/leaf.tsx
```diff
@@ -1,182 +1,7 @@
 import { RenderLeafProps } from 'slate-react'
```
joshuawootonn, line 1: This was all moved to `typography.ts` (keep scrolling)

PR #491, opened 2023-06-15
chore: add initial @makeswift/next package

chanceaclark:
## What/why?

Adds a new package `@makeswift/next` with basic build process with `tsup`. This will enable us to preserve `use client` directives.

### DIFF
--- a/packages/next/package.json
+++ b/packages/next/package.json
@@ -1,0 +1,36 @@
+{
+  "name": "@makeswift/next",
+  "version": "0.0.0",
+  "license": "MIT",
+  "repository": {
+    "url": "makeswift/makeswift",
+    "directory": "packages/next"
+  },
+  "sideEffects": false,
+  "main": "src/index.ts",
+  "files": [
+    "dist"
+  ],
+  "publishConfig": {
+    "access": "public",
+    "main": "dist/index.js",
+    "module": "dist/index.mjs",
+    "typings": "dist/index.d.ts"
+  },
+  "scripts": {
+    "build": "tsup"
+  },
+  "peerDependencies": {
+    "next": "^13.4.5",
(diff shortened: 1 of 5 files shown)


packages/next/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "access": "public",
+    "main": "dist/index.js",
+    "module": "dist/index.mjs",
+    "typings": "dist/index.d.ts"
+  },
+  "scripts": {
+    "build": "tsup"
+  },
+  "peerDependencies": {
+    "next": "^13.4.5"
```
migueloller, line 24: We should add `react` and `react-dom` as peer (and dev) dependencies as well

chanceaclark: Good point. I was going to add them when we needed to use the dependencies in code, but I guess this would be good to add now.

chanceaclark: I also added corresponding `@types/**` 🙏

packages/next/tsup.config.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export default defineConfig(
+  (options: Options): Options => ({
+    treeshake: false,
+    splitting: false,
+    entry: ['src/index.ts'],
+    format: ['esm', 'cjs'],
+    dts: true,
+    clean: true,
+    external: ['react'],
```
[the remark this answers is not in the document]
migueloller: Yeah, it will depend on how tsup/esbuild treat the externals option

packages/next/package.json
migueloller: Yeah, it came to mind since Next.js itself has those as peer dependencies. Are `@types/*` usually peer deps as well? We might want to make those optional peer deps

PR #496, opened 2023-06-20
Revert "chore: fix import issue with package version"

joshuawootonn:
This reverts commit 472fa4ccd884639c575c58cd313ef448161d9bcf.

### DIFF
--- a/packages/runtime/src/next/api-handler/handlers/manifest.ts
+++ b/packages/runtime/src/next/api-handler/handlers/manifest.ts
@@ -1,5 +1,5 @@
 import { NextApiRequest, NextApiResponse } from 'next'
-import pkg from '../../../../package.json'
+import { version } from '../../../../package.json'
 
 export type Manifest = {
   version: string
@@ -25,7 +25,7 @@ export default async function handler(
   }
 
   return res.json({
-    version: pkg.version,
+    version,
     previewMode: true,
     interactionMode: true,
     clientSideNavigation: true,


joshuawootonn:
Hey, @chanceaclark! @fikrikarim and I were going to drop a version when we noticed that main is broken. 

When the runtime is linked to the app we get this error: 
CleanShot 2023-06-20 at 14 41 59@2x

and the revert in this PR fixes said error. 

Do you have the ability to reproduce the issue / does the revert also fix it for you?

joshuawootonn:
Ok! While in the process of gathering more info, the issue stopped happening. 

I just pinged @fikrikarim to see if he is still experiencing it. This might have been a fluke. I'll keep you posted

joshuawootonn:
Whoops spoke too soon. In the above "maybe this is a fluke" comment, I thought I was on main, but I was on the revert commit from this PR.

chanceaclark:
@joshuawootonn I've been trying to dive into the _why_ it's happening and it's some pretty strange behavior. Give me a little to dive into it further.

chanceaclark:
Alright, so I think I found the culprit. When importing the `package.json` it imports the values as globally hoisted variables:
```js
const name = "@makeswift/runtime";
const version = "0.9.4";
const license = "MIT";
const main = "dist/main.cjs";
const module = "dist/main.es";
const types = "dist/types/src";
const files = [
  // ...
];
const scripts = [
  // ...
]
```

I believe the issues is the globally hoisted `files` + `scripts` constants which looking at the diff of the built, there where a few lines that were changed to account for this. **I think we can revert this for now** because it's really only an issue when you use the module directly, instead of build files.

chanceaclark (approved):

fikrikarim (approved):
Thanks for looking into this, Josh, and Chance!

joshuawootonn:
Hey @chanceaclark, what tool did you use to look at the diff of the build outputs?

chanceaclark:
@joshuawootonn just diffchecker.com with each build.

joshuawootonn:
Trueeee. There are only a few files so doing it that is reasonable. I was just curious if there was a CLI I wasn't aware of or something!

PR #497, opened 2023-06-21
chore: convert runtime build from vite to tsup

chanceaclark:
## What/why?
Converts the build over to `tsup` from `vite` in order to preserve `'use client';` directives. This work is aiming to support React Server Components (RSCs) in a future version of `@makeswift/runtime`.

## Testing

_Tested with the `makeswift/examples/bigcommerce` project_

https://github.com/makeswift/makeswift/assets/10539418/2349ba9e-5350-4fb8-9fae-6952f244cbe1

To test it out yourself, you'll need to `yarn link` this package and add a yarn `resolutions` field for your `react` version of choice (otherwise you'll get a Rule of Hooks issue).


migueloller (changes requested):
The changes look good but I found some issues when running things locally:
- The `vite.test.config.ts` file has some TypeScript errors, perhaps from missing Vite dependencies?
- We're getting a lot of errors from using `import.meta`, I think we should extract those tests into `*.test.ts` files next to the source
- DTS seems to be taking a very long time (16s while the rest of the build is just 0.3s). Is there a way to leverage incremental TypeScript compilation?
- I'm getting errors when attempting to use this in a vanilla Makeswift setup and `yarn link`
- Jump to definition in VS Code seems to be going to d.ts files instead of the source code, which it wasn't doing before

chanceaclark:
> * The `vite.test.config.ts` file has some TypeScript errors, perhaps from missing Vite dependencies?

I'll push a fix for this, simple module namespace issue.

---

> * We're getting a lot of errors from using `import.meta`, I think we should extract those tests into `*.test.ts` files next to the source

Those are just warnings and are not critical to the built files. I think we can tackle this in another PR to remove those warnings.

---

> * DTS seems to be taking a very long time (16s while the rest of the build is just 0.3s). Is there a way to leverage incremental TypeScript compilation?

Great idea, I'll see what I can throw on there.

---

> * I'm getting errors when attempting to use this in a vanilla Makeswift setup and `yarn link`

It was a little footnote at the end of the PR descriptions, but make so you do this as I ran into the same issues:
>To test it out yourself, you'll need to yarn link this package and add a yarn resolutions field for your react version of choice (otherwise you'll get a Rule of Hooks issue).

---

> * Jump to definition in VS Code seems to be going to d.ts files instead of the source code, which it wasn't doing before

Interesting, I'll check it out, but from what I know I think that's normal with other libraries. We can chat about this offline though.

PR #498, opened 2023-06-22
feat: add upgrade path from `RichText` to `RichTextV2`

joshuawootonn:
related: https://gitlab.com/makeswift/cosmos/-/merge_requests/1897

### DIFF
--- a/.changeset/metal-pets-call.md
+++ b/.changeset/metal-pets-call.md
@@ -1,0 +1,7 @@
+---
+'@makeswift/runtime': patch
+---
+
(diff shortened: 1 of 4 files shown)


joshuawootonn:
Little demo here. You can imagine me toggling the richtext version is equivalent to what they will do when they upgrade through a breaking version and the export is changed in the runtime.


https://github.com/makeswift/makeswift/assets/20950876/3ce09d23-72a8-4b58-8e95-2a715154ffdd

PR #499, opened 2023-06-23
feat: add fonts config to the ReactRuntime

chanceaclark:
## What/why?

Adds registering fonts as apart of the `ReactRuntime` instead of the `MakeswiftApiHandler`. This is some prep work to get fonts out of the API handlers since it should be the concern of the runtime, not the API.

Screenshot 2023-06-23 at 14 07 15

### DIFF
--- a/packages/runtime/src/next/api-handler/index.ts
+++ b/packages/runtime/src/next/api-handler/index.ts
@@ -15,6 +15,7 @@ export type { Manifest, Font }
 type MakeswiftApiHandlerConfig = {
   appOrigin?: string
   apiOrigin?: string
+  /** @deprecated Use the `fonts` option in the `ReactRuntime` constructor instead. */
   getFonts?: GetFonts
   unstable_siteVersions?: boolean
 }
--- a/.changeset/tasty-bikes-warn.md
+++ b/.changeset/tasty-bikes-warn.md
@@ -1,0 +1,8 @@
+---
+'@makeswift/runtime': minor
+---
+
+Adds setting the fonts in the `ReactRuntime` rather than `MakeswiftApiHandler`.
+
+`getFonts` within the `MakeswiftApiHandler` is now considered deprecated and will
+be removed in a future version of `@makeswift/runtime`.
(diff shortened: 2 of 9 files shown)


migueloller (changes requested):
This looks good. There's a couple of things that come to mind:

- Since we're deprecating an API, we might want this to go out in a `minor` version instead of a `patch`
- We're going to need to add a flag to the manifest so the builder knows whether to request fonts via `postMessage` or HTTP
- Also, we'll need the interface that the builder will be using to request fonts. In the current implementation, which seems to mirror breakpoints, we send the values at load time but this makes things more complicated and it's not a pattern we want to follow long-term
- We might want to separate the fonts domain logic from the state. I know that this isn't what we currently do with the breakpoints, but it's one of the things we've learned as we've had to deal with circular dependencies and modularization, etc. (this isn't necessary to do right now, but it's something that's top of mind right now so I thought it might be worth sharing)
- These changes will also require updates to our docs. I'm happy to give you access to our docs repo so you can make the changes yourself @chanceaclark

Of all the stuff I mention above, the more important one is the manifest flag. The rest is mostly about doing things in a "better" way than how we did them when we implemented the breakpoints API, what I'm assuming you took as inspiration for this API. If we implement this like we did breakpoints, what you have here is solid.

chanceaclark:
Added the manifest flag and updated to a `minor` ♻️ 🙏

packages/runtime/src/next/api-handler/index.ts
```diff
@@ -15,6 +15,7 @@ export type { Manifest, Font }
 type MakeswiftApiHandlerConfig = {
   appOrigin?: string
   apiOrigin?: string
+  /** @deprecated, use the `fonts` in the ReactRuntime config */
```
migueloller, line 18: There should be no comma after `@deprecated`. Let's change the message to this:

```ts
/** @deprecated Use the `fonts` option in the `ReactRuntime` constructor instead. */
```

migueloller (changes requested):
Looks good! Just have a minor comment

packages/runtime/src/next/api-handler/index.ts
chanceaclark: Done!

migueloller (approved):
This is looking good! I might wait until we can implement the builder side before we merge this if that's ok

PR #503, opened 2023-06-26
feat: add localized page meta and SEO

fikrikarim:


packages/runtime/src/components/page/Page.tsx
```diff
@@ -100,9 +100,12 @@ export function Page({ document: page }: Props): JSX.Element {
...
   const site = useCachedSite(isInBuilder ? page.site.id : null)
 
+  const baseLocalizedPage = page.localizedPages.find(({ parentId }) => parentId == null)
   const favicon = page.meta.favicon ?? defaultFavicon
-  const { title, description, keywords, socialImage } = page.meta
-  const { canonicalUrl, isIndexingBlocked } = page.seo
+  const meta = baseLocalizedPage?.meta ?? page.meta
+  const seo = baseLocalizedPage?.seo ?? page.seo
+  const { title, description, keywords, socialImage } = meta
+  const { canonicalUrl, isIndexingBlocked } = seo
```
migueloller, line 108: Shouldn't we be deeply merging values here? i.e.:

```ts
const title = baseLocalizedPage?.meta?.title ?? page.meta.title
// ...etc
```

fikrikarim: Absolutely true. Thanks for noticing that! Updated.

PR #519, opened 2023-07-05
Call versioned endpoints if using versioning

alexhwoods:


migueloller (approved):
LGTM 👍🏻 

Are these endpoints live at api.makeswift.com?

alexhwoods:
> LGTM 👍🏻
> 
> Are these endpoints live at api.makeswift.com?

They're on their way now

PR #525, opened 2023-07-11
Stabilize  RichTextV2

joshuawootonn:


packages/runtime/src/controls/rich-text-v2/rich-text-v2.ts
```diff
@@ -175,6 +175,7 @@ export class RichTextV2Control<
           this.editor.selection = null
           this.editor.children = this.defaultValue
           this.editor.onChange()
+          this.onLocalUserChange()
```
joshuawootonn, line 204: This onChange in 177 doesn't have operations so it normally wouldn't be committed back to the builder. This overrides that for this situation. 

This message is received after the user clicks reset in the `Content` panel label.

packages/runtime/src/runtimes/react/controls/rich-text-v2/EditableTextV2/useRemoteChanges.tsx
```diff
@@ -1,8 +1,8 @@
 import { useEffect } from 'react'
 import { Editor } from 'slate'
-import { useIsInBuilder } from '../../../../../react'
 import { RichTextV2ControlData, richTextV2DataToDescendents } from '../../../../../controls'
 import { LocalChange } from '../../../../../slate'
+import { useIsInBuilder } from '../../..'
```
joshuawootonn, line 5: Vite was giving me a warning and this fixed it.

packages/runtime/src/components/builtin/Text/ReadOnlyText.tsx
```diff
@@ -122,12 +122,6 @@ export function BlockElement({ descendant }: BlockProps) {
   })
 
   switch (descendant.type) {
-    case BlockType.Text:
```
joshuawootonn, line 125: This type was a bad idea. Instead both modes will default to `Default` type. This way when 
* they use `Inline` mode
* create some data
* change the mode to `Block`

there doesn't need to be a migration of the data. It's just the same default type that is rendered differently based on mode.

packages/runtime/src/controls/rich-text-v2/rich-text-v2.ts
```diff
@@ -68,12 +68,15 @@ export function createRichTextV2Plugin<T extends ControlDefinition>({
...
-  plugins?: RichTextV2Plugin[]
+export type RichTextV2Config = {
   mode?: RichTextV2Mode
+  plugins?: RichTextV2Plugin[]
+  defaultValue?: Descendant[]
 }
 
-type ExternalRichTextV2Config = Omit<RichTextV2Config, 'plugins'>
+type ExternalRichTextV2Config = Omit<RichTextV2Config, 'plugins' | 'defaultValue'> & {
+  unstable_defaultValue?: Descendant[]
```
joshuawootonn, line 78: I am exposing the `unstable_defaultValue` so that this component can be used in the `Text` component

packages/runtime/src/runtimes/react/controls/rich-text-v2/EditableTextV2/editable-text-v2.tsx
```diff
@@ -131,9 +132,18 @@ export function EditableTextV2({ text, definition, control }: Props) {
...
-    control?.setDefaultValue(defaultText)
-  }, [control, editor, defaultText])
+    control?.setDefaultValue(defaultValue)
+
+    /**
+     * When initialValue is set to the default value we need to trigger an local change so that the sidebar updates and so the data is saved
+     */
+    if (initialValue === defaultValue) {
+      control?.onLocalUserChange()
+    }
```
joshuawootonn, line 145: The initialValue in the `<Editor />` is set without an onchange being fired so this makes sure the value is sent to the builder / the prop data is set.

packages/runtime/src/runtimes/react/controls/rich-text-v2/EditableTextV2/editable-text-v2.tsx
```diff
@@ -152,6 +152,8 @@ export function EditableTextV2({ text, definition, control }: Props) {
 
   const handleKeyDown = useCallback(
     (e: KeyboardEvent) => {
+      if (isHotkey('mod+shift+z', e)) return control?.redo()
+      if (isHotkey('mod+z', e)) return control?.undo()
```
joshuawootonn, line 156: Missed adding these back in a rebase.

packages/runtime/src/controls/rich-text-v2/introspection.test.ts
```diff
@@ -57,44 +56,17 @@ const introspectionFixture: RichTextV2ControlData = {
...
+ * Running into too many circular dependencies problems. Skipping for now.
+ */
+describe.skip('GIVEN introspecting RichTextV2', () => {
   test('WHEN getSwatchIds THEN correct swatches are returned', () => {
-    expect(
-      getSwatchIds(
-        {
-          type: RichTextV2ControlType,
-          config: {
-            plugins: [TypographyPlugin()],
```
joshuawootonn, line 67: the `plugins` config option doesn't work because it is inferred from mode in the `RichText()` function, but using the `RichText()` function caused circular dependencies.

.changeset/five-parents-provide.md
```diff
@@ -0,0 +1,7 @@
+---
+'@makeswift/runtime': minor
+---
+
+BREAKING: Prior to this version the `Text` component and `RichText` control used `white-space-collapse: preserve` within app.makeswift.com and `white-space-collapse: collapse` in your live page.
+Our goal is to always exactly match what you see in makeswift with what you see in the live page.
```
migueloller, line 6: Capitalize "makeswift"

.changeset/metal-colts-love.md
```diff
@@ -0,0 +1,12 @@
+---
+'@makeswift/runtime': minor
+---
+
+BREAKING: Upgrade `Richtext` control to have with a new architecture that enables `Inline` mode and future rich text upgrades.
```
migueloller, line 5: "to have with a new architecture" is grammatically incorrect

.changeset/metal-colts-love.md
```diff
@@ -0,0 +1,12 @@
...
+
+This is the first time we have altered the data structure of a component, and we wanted you to be able to migrate the data and see the diff yourself incase the migration didn't work.
+
+When you select a `Text` component or component with the `RichText` control, you will be prompted in the sidebar to upgrade. If the migration doesn't work, simple `cmd/ctrl + z`.
+
+Details on `Inline` are in the [documentation for `RichText`](https://www.makeswift.com/docs/controls/rich-text)
+
```
migueloller, line 12: Unnecessary newline?

packages/runtime/src/components/builtin/Text/TextV2.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { RichTextV2ControlValue } from '../../../runtimes/react/controls/rich-text-v2'
+
+type Props = {
+  id?: ElementIDValue
+  text?: RichTextV2ControlValue
+  width?: string
+  margin?: string
+}
+
+function Text({ id, text, width, margin }: Props) {
```
migueloller, line 12: I think we should rename the `text` prop to `children`. Also, instead of taking `width` and `margin`, we should take `className` and use our `Style` control.

But perhaps more importantly, which would invalidate the above, how come we're creating a new component? I would've expect we just use the new control with the existing component

packages/runtime/src/components/builtin/Text/register.ts
migueloller, line 1: It seems the file extension was changed from `ts` to `tsx`. Was this necessary? The file doesn't seem to be using JSX

packages/runtime/src/controls/rich-text-v2/introspection.test.ts
migueloller: Are these issues only affecting tests or are they also affecting the API?

packages/runtime/src/runtimes/react/controls/rich-text-v2/EditableTextV2/editable-text-v2.tsx
```diff
@@ -52,8 +61,16 @@ type Props = {
...
     const plugins = [
-      ...(definition?.config?.plugins ?? []),
-      ...(definition?.config?.mode === RichTextV2Mode.Inline ? [InlineModePlugin] : []),
+      /**
+       * TODO: we are manually referencing our default plugins for each mode here because
+       * Referencing the real LinkPlugin causes a circular dependency.
+       * When circular dependencies calm down we should update the plugin definition to use real plugins,
+       * and just use the plugins that are defined by our config.
+       */
+      // ...(definition?.config?.plugins ?? []),
```
migueloller, line 70: What's the problem with the circular dependencies? Are you trying to avoid them or are they resulting in a build time or runtime error?

migueloller (changes requested):
This looks pretty good. My main worry here are all the workarounds for the circular dependencies. I would like to avoid those if possible. Are the circular dependencies breaking something or are you just trying to avoid them?

packages/runtime/src/components/builtin/Text/TextV2.tsx
joshuawootonn: > instead of taking width and margin, we should take className and use our Style control.
I can't do this without creating a migration from Style prop controller to Style control. 

> how come we're creating a new component?
The old component took rich text data / used the RichText prop controller. This new component needs to expect a react node.

packages/runtime/src/controls/rich-text-v2/introspection.test.ts
joshuawootonn: Only tests.

packages/runtime/src/runtimes/react/controls/rich-text-v2/EditableTextV2/editable-text-v2.tsx
joshuawootonn: I am trying to avoid them, and they are resulting in a build time error.

Referencing the slate plugins in `/slate` from my control creates an error because the `Link` plugin references `useElementId`.

packages/runtime/src/components/builtin/Text/TextV2.tsx
migueloller: Ok, I think we should either update the existing component to handle the new control or create a brand new component. If we keep the existing component, it's a matter of updating the registration to use the new control and then the implementation of the component so that it treats the `text` prop as a `ReactNode`. If we do a new component, I think it should have a separate type and given that it's a new component I would apply the latest patterns and avoid legacy ones (i.e., use Style control and better prop names). Also, if we do a new component we'd need to hide the other component and also make sure that the text shortcut is using the new component, which will be a bit more involved than just updating the existing component.

So, I would suggest you just update the existing component. And it really isn't a different component since it has the same type when its registered and the same props, etc.

packages/runtime/src/runtimes/react/controls/rich-text-v2/EditableTextV2/editable-text-v2.tsx
migueloller: Ok. Can you remove the circular dependencies?

packages/runtime/src/components/builtin/Text/TextV2.tsx
joshuawootonn: I was hesitant to delete the old code for the Text component, so I "upgraded" it by creating this component and changing the registration of the `Text` component. 

I can instead actually update the `Text.tsx` file so that it's all in version history

packages/runtime/src/runtimes/react/controls/rich-text-v2/EditableTextV2/editable-text-v2.tsx
joshuawootonn: Maybe :) 

I spent some time on it yesterday, and I don't know what removing them entails. The thing that is triggering the circular dependency is referencing `runtimes/react/index.ts` which is importing a lot of things.

PR #557, opened 2023-08-08
Feature Flags

alexhwoods:


alexhwoods:
1. This separates feature flags from `previewData`
2. This defaults the site version to `Live`, which can be overridden with `Working` via `previewData`.

packages/runtime/types/features.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export type MakeswiftFeatures = {
+  [K in keyof MakeswiftFeaturesConfig]-?: Exclude<MakeswiftFeaturesConfig[K], undefined>
+}
+
+export const featuresSchema = z
+  .object({
+    versioning: z.boolean().default(false),
+  })
+  .default({})
```
alexhwoods, line 15: The default is in case `undefined` is passed in

packages/runtime/src/next/preview-mode.tsx
```diff
@@ -12,12 +12,12 @@ const makeswiftPreviewDataSchema = z.object({
...
 export type MakeswiftPreviewData = z.infer<typeof makeswiftPreviewDataSchema>
 
-export function getMakeswiftSiteVersion(previewData: PreviewData): MakeswiftSiteVersion | null {
+export function getMakeswiftSiteVersion(previewData: PreviewData): MakeswiftSiteVersion {
   const result = makeswiftPreviewDataSchema.safeParse(previewData)
 
   if (result.success) return result.data.unstable_siteVersion
 
-  return null
+  return MakeswiftSiteVersion.Live
```
migueloller, line 20: I'm not sure I'd want to place this here—it combines picking a default site version with extracting one from the preview data. The purpose of this function was extracting it from the preview data and defaulting it like this here makes it impossible to know, given the returned value of this function, whether it was explicitly set to live or if it defaulted to that

packages/runtime/src/next/client.ts
```diff
@@ -134,13 +136,15 @@ export class Makeswift {
...
-  private siteVersion: MakeswiftSiteVersion | null
+  private features: MakeswiftFeatures
+  private siteVersion: MakeswiftSiteVersion
 
   constructor(
     apiKey: string,
     {
       apiOrigin = 'https://api.makeswift.com',
       runtime = ReactRuntime,
+      features,
```
migueloller, line 147: I don't think I want to nest our features under a `features` field, I would rather have top-level configuration options

packages/runtime/types/features.ts
```diff
@@ -0,0 +1,15 @@
+import { z } from 'zod'
+
+export type MakeswiftFeaturesConfig = {
+  versioning?: boolean
+}
+
+export type MakeswiftFeatures = {
+  [K in keyof MakeswiftFeaturesConfig]-?: Exclude<MakeswiftFeaturesConfig[K], undefined>
+}
```
migueloller, line 9: When using Zod, I think we should always infer the types from the schema to avoid a mistake where our type doesn't match Zod's TypeScript types

packages/runtime/src/next/api-handler/index.ts
```diff
@@ -36,7 +37,7 @@ export function MakeswiftApiHandler(
     appOrigin = 'https://app.makeswift.com',
     apiOrigin = 'https://api.makeswift.com',
     getFonts,
-    unstable_siteVersions = false,
+    features,
```
migueloller, line 40: See my other comment on not wanting to have these nested under a `features` object

migueloller (changes requested):

packages/runtime/src/next/preview-mode.tsx
alexhwoods: Agree w that

packages/runtime/src/next/api-handler/handlers/manifest.ts
```diff
@@ -32,7 +33,8 @@ export default async function handler(
     clientSideNavigation: true,
     elementFromPoint: false,
     customBreakpoints: true,
-    unstable_siteVersions,
+    versioning,
+    unstable_siteVersions: versioning,
```
migueloller, line 37: Can we get rid of this field? I'm guessing the plan is to get rid of this field as soon as we update our app so that it reads `versioning` instead?

packages/runtime/src/next/client.ts
```diff
@@ -134,13 +135,15 @@ export class Makeswift {
...
-  private siteVersion: MakeswiftSiteVersion | null
+  private versioning: boolean
+  private siteVersion: MakeswiftSiteVersion
 
   constructor(
     apiKey: string,
     {
       apiOrigin = 'https://api.makeswift.com',
       runtime = ReactRuntime,
+      versioning = false,
```
migueloller, line 146: I remember that the reason why I thought it would be weird to pass both `versioning` and `unstable_previewData` here is that it allows for an invalid state. If `versioning` is `false` but `unstable_previewData` has a version, like `'Working'` or `'Live'`, then its inconsistent. `unstable_previewData` should by itself be enough to know, no? If it's not provided, i.e., `undefined` then `isUsingVersioning` should be `false`. If it's `null` then it's a live page so `isUsingVersioning` is `true` and `siteVersion` is `'Live'`, otherwise it's whatever version is defined by `unstable_previewData`.

I know this is contrary to what we had talked about, but I'm realizing it again after seeing `versioning` side by side with `unstable_previewData`.

Is the plan to remove the `unstable_` APIs? If so, are we not removing `unstable_previewData`? If we do remove `unstable_previewData`, I think we can come up with an API that is a single field that is enough for us to determine both `isUsingVersioning` and `siteVersion`

migueloller (changes requested):
I think we need to have one more conversation on `unstable_previewData` and the constructor parameters of the client. For the manifest we're just renaming `unstable_siteVersions` to `versioning` which is simple enough, but for the client it's a bit more nuanced.

Also, one other thought came to mind. Is there a reason we're adding this as an option, vs this just being enabled by default in the next minor version of the runtime? This isn't something that we want people to opt-out of—it's new behavior. So it feels like a new version of the runtime should be enough of a "flag"

alexhwoods:
some testing

https://github.com/makeswift/makeswift/assets/7907782/52d8cbd6-2c7a-4f24-be44-0c42b9a463b8

packages/runtime/src/next/index.tsx
```diff
@@ -95,7 +95,6 @@ export async function getStaticProps(
 
   const makeswift = new Makeswift(getApiKey(), {
     apiOrigin: getApiOrigin(),
-    unstable_previewData: ctx.previewData,
```
alexhwoods, line 98: Is removing this function the right call? 

This is the deprecated API right?

migueloller: Yeah, we'll want to replace if with the new API, though. We should probably deprecate these functions...

migueloller: I think what you have here is the proper change, though! It means that if you're using these legacy functions you're not using versioning, which I'm good with.

migueloller (dismissed):

migueloller (approved):

PR #566, opened 2023-08-17
feat: add external file support in images prop controller

arvinpoddar:


packages/runtime/src/prop-controllers/copy/images.ts
```diff
@@ -18,7 +20,13 @@ export function copy(
       ...imagesPanelItem,
       props: {
         ...imagesPanelItem.props,
-        file: context.replacementContext.fileIds.get(file) ?? file,
+        file: match([descriptor, file as ImageValue])
```
migueloller, line 23: How come we're having to use a type assertion for `file` here?

migueloller: Ah... I see that we're using an `any` above. Could we perhaps change the `any` type instead of using a type assertion here?

packages/runtime/src/prop-controllers/descriptors.ts
```diff
@@ -442,14 +442,37 @@ export function Image(options: ImageOptions = {}): ImageDescriptor {
   return { type: Types.Image, version: 1, options }
 }
 
-export type ImagesValue = { key: string; props: { link?: Link; file?: string; altText?: string } }[]
+export type ImagesValueV0Item = {
+  key: string;
+  props: {
+    link?: Link;
+    file?: string;
```
migueloller, line 449: I would suggest we make this `ImageValueV0` instead of `string`

packages/runtime/src/prop-controllers/index.ts
```diff
@@ -65,3 +69,4 @@ export type {
 export { RichTextPropControllerMessageType, TableFormFieldsMessageType } from './instances'
 export * as Introspection from './introspection'
 export { copy as imageCopy } from './copy/image'
+export {copy as imagesCopy} from './copy/images'
```
migueloller, line 72: Looks like there's some spacing inconsistencies here—Prettier should automatically fix this

packages/runtime/src/prop-controllers/introspection.ts
```diff
@@ -303,7 +303,14 @@ export function getFileIds<T extends Data>(
...
     case Types.Images: {
       const value = prop as ImagesValue
-      return value?.flatMap(item => (item.props.file == null ? [] : [item.props.file])) ?? []
+      return value?.flatMap(
+        item => match(item.props.file)
+          .with(P.string, f => [f])
+          .with({type: 'makeswift-file', version: 1}, f => [f.id])
+          .with({type: 'external-file', version: 1}, () => [])
+          .with(P.nullish, () => [])
+          .exhaustive()
```
migueloller, line 312: Should we do `.otherwise(() => [])` here?

migueloller (changes requested):
@arvinpoddar, looking good! I have a couple of suggestions.

Also, a question: was the reason we didn't have to update the Carousel component that it uses the `Image` component? If so, this is a good reason for us to couple the controls and share the types and things are intended to be used together and they "just work"

arvinpoddar:
>  Also, a question: was the reason we didn't have to update the Carousel component that it uses the Image component? If so, this is a good reason for us to couple the controls and share the types and things are intended to be used together and they "just work"

Yep, because the file types match, we didn't have to update the carousel at all.

packages/runtime/src/prop-controllers/index.ts
arvinpoddar: Done

packages/runtime/src/prop-controllers/introspection.ts
migueloller: It was more a question than a suggestion but since you did it I'm assuming the answer was "yes"?

migueloller (approved):

packages/runtime/src/prop-controllers/introspection.ts
arvinpoddar: Ah my bad, the answer is that it shouldn't make a difference because it really should be exhaustive. I prefer using `.otherwise` because it allows for continuity of behavior in the strange case something does go wrong, however.

PR #568, opened 2023-08-18
feat: add external file support in backgrounds prop control

arvinpoddar:


migueloller (approved):
Looks good! Everything go well during testing?

arvinpoddar:
> Looks good! Everything go well during testing?

Yep, didn't run into anything. Tested copying, backwards compatibility, and of course normal usage. One thing: if you do upgrade your runtime, use the builder, then downgrade your runtime, you'll get errors when attempting to run the site again.

migueloller:
> > Looks good! Everything go well during testing?
> 
> @migueloller Yep, didn't run into anything. Tested copying, backwards compatibility, and of course normal usage. One thing: if you do upgrade your runtime, use the builder, then downgrade your runtime, you'll get errors when attempting to run the site again.

Are the errors recoverable? Could we perhaps handle those cases more gracefully by checking the mismatch between descriptor and data?

arvinpoddar:
> > > Looks good! Everything go well during testing?
> > 
> > 
> > @migueloller Yep, didn't run into anything. Tested copying, backwards compatibility, and of course normal usage. One thing: if you do upgrade your runtime, use the builder, then downgrade your runtime, you'll get errors when attempting to run the site again.
> 
> Are the errors recoverable? Could we perhaps handle those cases more gracefully by checking the mismatch between descriptor and data?

Here's the error: 
```
{
  errors: [
    {
      name: 'GraphQLError',
      message: 'Variable "$fileIds" got invalid value null at "fileIds[0]"; Expected non-nullable type "ID!" not to be null.',
      locations: [Array],
      nodes: [Array],
      source: [Object],
      positions: [Array],
      extensions: [Object]
    }
  ]
}
- error node_modules/@makeswift/runtime/dist/index.es.js (244:12) @ GraphQLClient.request
- error Error: GraphQL response contains errors, check the console.
    at GraphQLClient.request (webpack-internal:///./node_modules/@makeswift/runtime/dist/index.es.js:355:19)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async Makeswift.getIntrospectedResources (webpack-internal:///./node_modules/@makeswift/runtime/dist/index.es.js:986:24)
    at async Makeswift.introspect (webpack-internal:///./node_modules/@makeswift/runtime/dist/index.es.js:1060:64)
    at async Makeswift.getPageSnapshot (webpack-internal:///./node_modules/@makeswift/runtime/dist/index.es.js:1166:54)
    at async getStaticProps (webpack-internal:///./pages/[[...path]].tsx:39:22) {
  digest: undefined
}
null
```

migueloller:
> > > > Looks good! Everything go well during testing?
> > > 
> > > 
> > > @migueloller Yep, didn't run into anything. Tested copying, backwards compatibility, and of course normal usage. One thing: if you do upgrade your runtime, use the builder, then downgrade your runtime, you'll get errors when attempting to run the site again.
> > 
> > 
> > Are the errors recoverable? Could we perhaps handle those cases more gracefully by checking the mismatch between descriptor and data?
> 
> Here's the error:
> 
> ```
> {
>   errors: [
>     {
>       name: 'GraphQLError',
>       message: 'Variable "$fileIds" got invalid value null at "fileIds[0]"; Expected non-nullable type "ID!" not to be null.',
>       locations: [Array],
>       nodes: [Array],
>       source: [Object],
>       positions: [Array],
>       extensions: [Object]
>     }
>   ]
> }
> - error node_modules/@makeswift/runtime/dist/index.es.js (244:12) @ GraphQLClient.request
> - error Error: GraphQL response contains errors, check the console.
>     at GraphQLClient.request (webpack-internal:///./node_modules/@makeswift/runtime/dist/index.es.js:355:19)
>     at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
>     at async Makeswift.getIntrospectedResources (webpack-internal:///./node_modules/@makeswift/runtime/dist/index.es.js:986:24)
>     at async Makeswift.introspect (webpack-internal:///./node_modules/@makeswift/runtime/dist/index.es.js:1060:64)
>     at async Makeswift.getPageSnapshot (webpack-internal:///./node_modules/@makeswift/runtime/dist/index.es.js:1166:54)
>     at async getStaticProps (webpack-internal:///./pages/[[...path]].tsx:39:22) {
>   digest: undefined
> }
> null
> ```

Ahhh, I see. So this error can't be avoided if old versions of the runtime encounter that data unless we somehow make it so the data doesn't encounter it.

Our general policy is that if somebody runs into a data error they can fix it by upgrading the runtime. If they can't upgrade due to a breaking change we can ship a patch version on their current major

arvinpoddar:
I decided to try a different approach that supports this edge case. Even if you do upgrade the runtime, use the builder, and rollback, you'll still be able to load your site in the builder - not all the new data will be present though. I think this provides a less frustrating experience for users.

packages/runtime/src/prop-controllers/copy/backgrounds.ts
```diff
@@ -44,21 +46,39 @@ export function copy(
               })),
             },
           }
-
-        case 'image':
+        })
+        .with(
+          [
+            { version: 1 },
+            { type: 'image-v1', version: 1, payload: { image: { type: 'makeswift-file' } } },
```
joshuawootonn, line 53: My understanding of this is that copies for external files will include a reference to the same external file.

joshuawootonn (approved):
LGTM

packages/runtime/src/prop-controllers/copy/backgrounds.ts
arvinpoddar: Exactly

PR #569, opened 2023-08-18
wip

joshuawootonn:


packages/runtime/src/controls/rich-text-v2/translatableData.ts
```diff
@@ -0,0 +1,53 @@
...
+      return `<a>${children}</a>`
+    case InlineType.Link:
+      return `<sup>${children}</sup>`
+    case InlineType.Link:
+      return `<sub>${children}</sub>`
+    case InlineType.Link:
+      return `<code>${children}</code>`
+    default:
+      return children
+  }
```
joshuawootonn, line 39: This isn't mergeable / and we will have to finish the convos from slack, but it will give us a way to test a real BC page

PR #571, opened 2023-08-22
feat: add external file support in image control

arvinpoddar:


packages/runtime/src/controls/image.ts
```diff
@@ -16,12 +33,13 @@ type ImageControlConfig = { label?: string; format?: ImageControlValueFormat }
...
   type: typeof ImageControlType
   config: T
+  version?: 1
 }
 
 export function Image<T extends ImageControlConfig>(
   config: T = {} as T,
 ): ImageControlDefinition<T> {
-  return { type: ImageControlType, config }
+  return { type: ImageControlType, config, version: 1 }
```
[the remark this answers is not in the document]
migueloller: The config is what the user provides to configure the control. We don't want the version to be configuration—so no, only in the definition

packages/runtime/src/controls/image.ts
```diff
@@ -1,6 +1,23 @@
...
+  type: 'makeswift-file'
+  id: string
+}
+
+type ExternalFileImageData = {
+  type: 'external-file'
+  url: string
+  width?: number | null
+  height?: number | null
+}
```
migueloller, line 16: Didn't we have versions in these at some point?

migueloller (changes requested):
LGTM!

packages/runtime/src/controls/image.ts
arvinpoddar: We did for the Image prop controller. I can add them in here, but they won't be used as of now.

migueloller: Well, the point would be so that when we have future versions that also use an object we can discriminate between those and this ones. It's not essential but it would be nice for us to have that... eventually, even having all the types in there with the type of the control might allow us to do introspection without even having the schema in the backend 🤔

migueloller (approved):

PR #581, opened 2023-09-06
Create `/merge-translatable-data`

joshuawootonn:


packages/runtime/src/slate/BlockPlugin/index.tsx
```diff
@@ -100,10 +101,16 @@ export function onKeyDown(e: KeyboardEvent, editor: Editor) {
...
+    Editor.withoutNormalizing(editor, () => {
+      if (!editor.selection) return
+
+      Transforms.splitNodes(editor, {
+        at: editor.selection,
+        always: true,
+        match: node => ElementUtils.isListItem(node),
+      })
+
+      Transforms.setNodes(editor, { key: uuid() }, { at: editor.selection })
```
joshuawootonn, line 113: Basically, anytime we mutate the tree now we have to manually update the key field.  

In future commits I add a normalization that adds a key for you, but it doesn't replace existing keys so.

joshuawootonn:
I really just want to get this stuff through before I work on other merge operations.

packages/runtime/src/slate/InlinePlugin/__test__/onChange.test.tsx
```diff
@@ -75,7 +85,7 @@ describe('GIVEN InlinePlugin.onChange', () => {
         </Link>
       </Paragraph>,
     )
-    const next = EditorV2(
+    const result = EditorV2(
```
joshuawootonn, line 88: changed this working to match all the other tests

packages/runtime/src/slate/BlockPlugin/index.tsx
migueloller: By "mutate the tree" do you mean any change? This is something I had mentioned in a call we had last week and I thought that changing text of a node _didn't_ change its key

joshuawootonn: "mutate the tree" here means splitting nodes, altering properties on them, etc. Changing the text will not update keys in this PR.

packages/runtime/src/slate/test-helpers/index.ts
```diff
@@ -3,3 +3,4 @@ export * from './slate-test-helper'
 export * from './editor-v1'
 export * from './editor-v2-inline'
 export * from './editor-v2'
+export * from './withoutKeys'
```
migueloller, line 6: Missing newline at the end of the file

packages/runtime/src/state/react-page.ts
```diff
@@ -367,6 +374,52 @@ export function getElementTreeTranslatableData(
...
+
+      for (const propName of Object.keys(descriptors)) {
+        const descriptor = descriptors[propName]
+
+        mergedProps[propName] = mergeTranslatableData(
+          descriptor,
+          `${node.key}:${propName}`,
+          node.props[propName],
+          context,
+        )
```
migueloller, line 414: In here is where I think we should be doing `mergeTranslatableData(descriptor, prop, translatedData, context)`

migueloller (changes requested):
- I'm a bit confused with the changeset because it mentions we're only adding support for `TextInput` but then there's a bunch of changes to rich text. Did we miss a changeset for those changes?
- I think that the naming is off in a bunch of places—it should be "merge translated data" as opposed to "merge translatable data", no? I guess they're technically the same data structure so the name it's not a big deal but it reads a bit weird
- I think we need to sync up on what the semantics of the rich text keys are for because it seems that we're re-generating these keys for every single change but we want them to be somewhat stable, like element keys in our element tree—the key doesn't change whenever a prop changes

packages/runtime/src/slate/BlockPlugin/index.tsx
migueloller: Got it! Thanks for clarifying 👍🏻

migueloller: Altering properties on the nodes shouldn't change the key, though, right? Changing the text is equivalent to changing a property. For example, changing an inline or paragraph's color shouldn't generate a new key

packages/runtime/src/state/react-page.ts
joshuawootonn: `mergeTranslatedData` data is normalized with keys that include data referenced by element and propname
`copy` data is _normalized_ into a replacement context that includes data referenced by nested within the data 

This difference means that as I traverse through the composable controls I have to keep track of the key to reference nested data.

I am rethinking the call signature now though. I think it makes more sense to have the key as the last param.

joshuawootonn: I am happy to report that I am wrong.  Details to come.

packages/runtime/src/controls/control.ts
```diff
@@ -195,3 +199,29 @@ export function getTranslatableData(definition: Descriptor | ControlDefinition,
...
+): Data {
+  switch (definition.type) {
+    case Types.Grid:
+      const value = data as GridValue
+
+      if (value == null) return undefined
+      return {
+        ...value,
+        elements: value.elements.map(element => context.mergeTranslatableData(element)),
+      }
```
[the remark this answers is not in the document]
joshuawootonn: This is moved to the `descriptors.ts` file now

joshuawootonn:
Merging to move forward with things. Feel free to add comments as needed and I can address them in the future.

packages/runtime/src/controls/control.ts
```diff
@@ -195,3 +214,58 @@ export function getTranslatableData(definition: Descriptor | ControlDefinition,
...
+  context: MergeTranslatableDataContext,
+): Data {
+  switch (definition.type) {
+    case Types.TextInput:
+    case Types.TextArea:
+    case TextInputControlType:
+    case TextAreaControlType:
+      if (translatedData == null) return data
+
+      return translatedData
```
migueloller, line 231: For multi-line switch blocks, I recommend using a block (`{}`) to make it easer to parse visually

migueloller (commented):
This looks great!

PR #585, opened 2023-09-13
feat: update basic-typescript example

apledger:


examples/basic-typescript/.env
```diff
@@ -1 +1 @@
-MAKESWIFT_SITE_API_KEY=
+MAKESWIFT_SITE_API_KEY=a86621c2-5154-4c39-84f7-bbe2d3f9b48d
```
migueloller, line 1: Hmm, I think you just leaked the API key for this site

examples/basic-typescript/pages/_app.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+const inter = Inter({
+  subsets: ['latin'],
+  variable: '--font-inter',
+})
+
+function MyApp({ Component, pageProps }: AppProps) {
+  useEffect(() => {
+    document.body.classList.add(inter.variable)
+  }, [])
```
migueloller, line 17: Why do we need to add the class to the body here? Won't this cause FOUC?

examples/basic-typescript/pnpm-lock.yaml
migueloller, line 1: Have you verified that changing from Yarn to pnpm will work with our CLI?

examples/basic-typescript/tsconfig.json
```diff
@@ -13,7 +13,14 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "jsx": "preserve",
-    "incremental": true
+    "incremental": true,
+    "baseUrl": ".",
+    "paths": {
+      "@/components/*": ["components/*"],
+      "@/generated/*": ["generated/*"],
```
migueloller, line 20: We don't seem to have a `generated` folder in this repo. I'm guessing this was copy/pasted from another repo?

migueloller (commented):
Looks good 👍🏻

Most important thing is to make sure that things still work well with the CLI and Vercel deploy button. Have you tested those out yet?

examples/basic-typescript/pages/_app.tsx
apledger: This is just for modals and anything that is portaled out of the `<main>` element.

examples/basic-typescript/pnpm-lock.yaml
apledger: What's the best way to test this?

examples/basic-typescript/tsconfig.json
apledger: Will update.

apledger: Actually this is where icons are generated from the included `icons` command.

examples/basic-typescript/pages/_app.tsx
migueloller: Got it. Can this not be added to the `body` in `_document`, then? It's how I see `create-next-app` do it for the App Router, it puts it in the root layout's `body` element.

migueloller: Or maybe `next/font` can't be used in `_document`... They don't seem to do it with `create-next-app` when using the pages router

examples/basic-typescript/pnpm-lock.yaml
migueloller: I vaguely remember @alexhwoods setting up some environment variables to be able to use local versions of the app as well as draft versions of changes by providing a full GitHub URL for example names. @alexhwoods, any pointers here?

examples/basic-typescript/tsconfig.json
migueloller: Ah, got it, I missed that. Is that `icons` script being used in this example?

migueloller: We don't seem to have any icons under `public/icons` so maybe we just want to clean that stuff up?

examples/basic-typescript/.env
```diff
@@ -1 +1 @@
-MAKESWIFT_SITE_API_KEY=
+MAKESWIFT_SITE_API_KEY=
```
migueloller, line 1: Missing a newline at the end of the file

joshuawootonn (approved):

PR #591, opened 2023-09-27
feat: use localized pathname on LinkControl

fikrikarim:
This PR focuses on getting the localized pathname.

For path-based localization, we don't have to worry about the `/es` prefix, because it's already handled by NextLink. For example, if I'm on `localhost/es/localization-1`, and I have a Button component set to Localization 2 page, then the link would be `localhost/es/localization-2`. But if I'm on `foo.es/localization-1`, then the link would be `foo.es/localization-2`.

There are several decisions that we need to make when implementing this:

1. Should we modify the existing `pagePathnamesById` query, or should we make a new query?
2. Should we reuse the existing `PagePathnameSlice` fragment, or should we make a new fragment?

For localized global elements, we have a new fragment and we have a new query because we thought that we would need both the base global element and the localized global element to be able to merge it in the runtime. But this is no longer the case because we decided to merge the element tree in our API instead.

Because of that, and because we only need the localized pathname and not the base pathname, I decided to reuse the existing query and fragment.

3. If we decided to reuse the existing `pagePathnamesById` query, should we return the `pageId`, or should we return the `localizedPageId` on the `PagePathnameSlice.id` field?

On the localized global element, we made it so that you query with `globalElement` and `locale`, and the API will return the `localizedGlobalElementId` as the `id`. This is working well, but we need to maintain a mapping of `globalElementId` to `localizedGlobalElementId`.

We could do a similar thing with this `PagePathnameSlice`. We query with `pageId` and `locale`, and the API will return `localizedPageId` as the `id`.

But in this MR, right now, I return the `pageId` instead. This is working well for now, and we don't need the mapping. I also think that `PagePathnameSlice` is not necessarily the `Page`. But this might bite us in the future if the builder also fetches `pagePathnamesById` and the cache normalization replaces the localized pathname with the non-localized one because the `id` of the `PagePathnameSlice` is the same. Right now the builder are not using `pagePathnameById` and `pagePathnamesById`.

With that being said, I still like this one because we abstract away the fact that there is a localized pathname on the runtime, and only the API knows about it. The runtime just query with `pageId` and `locale`, and use the `pathname` from the result.

Thoughts @migueloller?

### DIFF
--- a/packages/runtime/src/next/api-handler/index.ts
+++ b/packages/runtime/src/next/api-handler/index.ts
@@ -132,7 +132,9 @@ export function MakeswiftApiHandler(
     }
 
     if ((m = matches<{ id: string }>('/page-pathname-slices/:id'))) {
-      return client.getPagePathnameSlice(m.params.id).then(handleResource)
+      const locale = typeof req.query.locale === 'string' ? req.query.locale : undefined
+
+      return client.getPagePathnameSlice(m.params.id, { locale }).then(handleResource)
     }
 
     if ((m = matches<{ id: string }>('/tables/:id'))) {
--- a/packages/runtime/src/next/client.ts
+++ b/packages/runtime/src/next/client.ts
@@ -1,4 +1,5 @@
 import { PreviewData } from 'next'
+import { z } from 'zod'
 import {
   APIResourceType,
   File,
@@ -10,19 +11,12 @@ import {
   Typography,
 } from '../api'
 import { GraphQLClient } from '../api/graphql/client'
-import {
-  FileQuery,
-  IntrospectedResourcesQuery,
-  PagePathnamesByIdQuery,
-  TableQuery,
-} from '../api/graphql/documents'
+import { FileQuery, IntrospectedResourcesQuery, TableQuery } from '../api/graphql/documents'
 import {
   FileQueryResult,
   FileQueryVariables,
   IntrospectedResourcesQueryResult,
   IntrospectedResourcesQueryVariables,
-  PagePathnamesByIdQueryResult,
-  PagePathnamesByIdQueryVariables,
   TableQueryResult,
   TableQueryVariables,
 } from '../api/graphql/generated/types'
@@ -128,6 +122,17 @@ export type Sitemap = {
     href: string
   }[]
 }[]
+
+const pagePathnameSlicesAPISchema = z.array(
+  z
+    .object({
+      id: z.string(),
+      basePageId: z.string(),
+      pathname: z.string(),
+      __typename: z.literal('PagePathnameSlice'),
+    })
+    .nullable(),
+)
 
 export class Makeswift {
   private apiKey: string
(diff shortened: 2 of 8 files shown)


packages/runtime/src/next/api-handler/index.ts
```diff
@@ -131,8 +131,8 @@ export function MakeswiftApiHandler(
         )
     }
 
-    if ((m = matches<{ id: string }>('/page-pathname-slices/:id'))) {
-      return client.getPagePathnameSlice(m.params.id).then(handleResource)
+    if ((m = matches<{ id: string; locale?: string }>('/page-pathname-slices/:id/:locale?'))) {
```
migueloller, line 134: This is a confusing URL pattern because `locale` is not a sub resource of a page pathname slice. For REST APIs parameters like these are usually URL search parameters

migueloller (commented):
> Should we modify the existing pagePathnamesById query, or should we make a new query?

I think that we should continue to move things to the REST host API and stop using the builder GraphQL API.

> Should we reuse the existing PagePathnameSlice fragment, or should we make a new fragment?

We should not modify existing fragments and should move away from using those ASAP. Instead, whenever a resource changes in the builder we should trigger a revalidation in the host. We could also continue to send the data from the builder to optimistically update it but that would be an optimization on top of revalidation, IMO.

> Because of that, and because we only need the localized pathname and not the base pathname, I decided to reuse the existing query and fragment.

We _should not_ be modifying or using these fragments and should instead be working to get rid of them. They've caused some pretty nasty bugs in the past and aren't well aligned with the architecture we want for the host/builder interface.

> If we decided to reuse the existing pagePathnamesById query, should we return the pageId, or should we return the localizedPageId on the PagePathnameSlice.id field?

I think that we should never return data for one resource but provide an identifier for a separate resource as that would cause a lot of confusion and potential for bugs. If we're returning a localized resource and it comes with an ID I think that it should be the ID of that resource. If it's the ID of the non-localized version then the field should communicate that explicitly.

fikrikarim:
@migueloller Updated the PR to use the new REST endpoint.

packages/runtime/src/next/client.ts
```diff
@@ -563,13 +555,50 @@ export class Makeswift {
...
+      pathname,
+      __typename: 'PagePathnameSlice',
+    }))
+  }
+
+  async getPagePathnameSlice(
+    pageId: string,
+    { preview = false, locale }: { preview?: boolean; locale?: string } = {},
+  ): Promise<PagePathnameSlice | null> {
+    const res = await this.getPagePathnameSlices([pageId], { preview, locale })
```
migueloller, line 599: For every other resource we've been using the non-bulk endpoint—how come we're doing this one differently?

migueloller, line 599: nit: this is not a response object—it's page pathname slices

migueloller (dismissed):

packages/runtime/src/next/client.ts
fikrikarim: I was thinking about this too. Why can't we just have the bulk endpoint, and use that even when we need a single resource? Why do we need the non-bulk endpoint?

fikrikarim: Updated

PR #592, opened 2023-09-27
feat: add getPage to Makeswift client

fikrikarim:


packages/runtime/src/next/client.ts
```diff
@@ -129,6 +129,12 @@ export type Sitemap = {
   }[]
 }[]
 
+type GetPageResult = {
+  path: string
+  locale: string
+  localized: { path: string; locale: string }[]
```
fikrikarim, line 135: I'm not 100% sure what to name this field, it could be `localized`, `alternate`, or something else.

packages/runtime/src/next/client.ts
```diff
@@ -202,6 +208,32 @@ export class Makeswift {
     return json
   }
 
+  async getPage(
+    pathname: string,
+    { locale: localeInput }: { locale?: string } = {},
+  ): Promise<GetPageResult | null> {
+    const siteVersion = MakeswiftSiteVersion.Live
```
fikrikarim, line 215: Should we only enable this for Live version, or should we also enable this for the working version?

packages/runtime/src/next/client.ts
```diff
@@ -202,6 +208,32 @@ export class Makeswift {
...
+
+    const searchParams = new URLSearchParams()
+    if (localeInput) {
+      searchParams.set('locale', localeInput)
+    }
+
+    const response = await this.fetch(
+      `v1/pages/${encodeURIComponent(pathname)}/meta?${searchParams.toString()}`,
```
fikrikarim, line 223: For `getPageSnapshot`, we're using `v3/pages/:pathnameOrId/document`. I don't know what to call the subpath here. It's kinda like `meta`, but I'm afraid it'll confuse with the page meta and SEO. Any idea?

fikrikarim:
This is the draft to discuss `getPage` method on the client. This should be enough to generate the hreflang tag, but I'm not 100% sure about the naming and the structure, and I added some comments there. Let me know if you have any thoughts on those @migueloller.

packages/runtime/src/next/client.ts
migueloller: I lean towards `alternate`

Also, can we please rename the field `pathname` instead of `path`? URL paths include search params but for Makeswift pages what identifies them is their pathname

packages/runtime/src/next/client.ts
migueloller: I think that it should work for both. What are we supposed to render when we're looking at the working version of the site otherwise?

packages/runtime/src/next/client.ts
```diff
@@ -129,6 +129,12 @@ export type Sitemap = {
   }[]
 }[]
 
+type GetPageResult = {
+  path: string
```
migueloller, line 133: Can we rename to `pathname`?

packages/runtime/src/next/client.ts
```diff
@@ -129,6 +129,12 @@ export type Sitemap = {
...
+  locale: string
```
migueloller, line 134: Can we make this `Intl.Locale`?

packages/runtime/src/next/client.ts
```diff
@@ -202,6 +208,32 @@ export class Makeswift {
...
+
+    const response = await this.fetch(
+      `v1/pages/${encodeURIComponent(pathname)}/meta?${searchParams.toString()}`,
+      { headers: { 'Makeswift-Site-Version': siteVersion } },
+    )
+
+    if (!response.ok) {
+      console.error('Failed to get page', await response.json())
+      throw new Error(`Failed to get page with error: "${response.statusText}"`)
+    }
```
migueloller, line 230: Can we design it so that our API always returns JSON such that we can always do `reponse.json` and then based on the status code we can get strong type safety for the possible errors and have good error messages here?

packages/runtime/src/next/client.ts
```diff
@@ -202,6 +208,32 @@ export class Makeswift {
...
+
+    const json: GetPageResult = await response.json()
```
migueloller, line 232: What do you think about using Zod to validate and transform the response?

fikrikarim:
@migueloller Thanks for the review, I'll revise the MR based on that. Do you agree about what we return on the `getPage`? It's the bare minimum data needed to generate hreflang. I'm thinking of adding `id`, but I remember you said that we might want to do that.
```ts
type GetPageResult = {
  pathname: string
  locale: string
  alternate: { pathname: string; locale: string }[]
}
```

migueloller:
> @migueloller Thanks for the review, I'll revise the MR based on that. Do you agree about what we return on the `getPage`? It's the bare minimum data needed to generate hreflang. I'm thinking of adding `id`, but I remember you said that we might not want to do that.
> 
> ```ts
> type GetPageResult = {
>   pathname: string
>   locale: string
>   alternate: { pathname: string; locale: string }[]
> }
> ```

Let's start without `id` at first. We can always add it if it becomes necessary.

I would also like to discuss the tradeoffs between having `locale` be `string` vs `Intl.Locale`

fikrikarim:
> Let's start without `id` at first. We can always add it if it becomes necessary.
> 
> I would also like to discuss the tradeoffs between having `locale` be `string` vs `Intl.Locale`

Got it. I'm leaning towards the `locale` be `string`, so you can pass it directly from the `ctx.locale`. If we need to validate it to make sure you're passing a valid locale, we could do it inside the method.

migueloller:
> > Let's start without `id` at first. We can always add it if it becomes necessary.
> > I would also like to discuss the tradeoffs between having `locale` be `string` vs `Intl.Locale`
> 
> Got it. I'm leaning towards the `locale` be `string`, so you can pass it directly from the `ctx.locale`. If we need to validate it to make sure you're passing a valid locale, we could do it inside the method.

I was thinking about the result, not the argument. But yes, your point above is a good reason to keep the argument as a string. And keeping the result the same as an argument is a reason to keep the result as a string as well

packages/runtime/src/next/client.ts
fikrikarim: Updated

packages/runtime/src/next/client.ts
fikrikarim: We agreed to make this `string`

packages/runtime/src/next/client.ts
fikrikarim: Updated to `alternate` and `pathname`

packages/runtime/src/next/client.ts
fikrikarim: Sure updated

packages/runtime/src/next/client.ts
fikrikarim: Updated

packages/runtime/src/next/client.ts
fikrikarim: Updated

packages/runtime/src/next/client.ts
fikrikarim: It might be harder to do that just for this API, because the error could also come from `ThrottlerByApiKeyGuard`, `SiteIdByApiKeyPipe`, or `ParseSiteVersionPipe`. We need to also refactor those if we want our API to always return JSON right?

migueloller: In those scenarios, won't we know the exact status codes? Also, maybe we can update those guards and pipes so that the error is JSON as well?

packages/runtime/src/next/client.ts
```diff
@@ -129,6 +130,21 @@ export type Sitemap = {
...
+    pathname: z.string(),
+    locale: z.string(),
+    alternate: z.array(
+      z.object({
+        pathname: z.string(),
+        locale: z.string(),
+      }),
+    ),
+  })
+  .nullable()
```
migueloller, line 144: The result will never be null if the status code is 200

migueloller (dismissed):

packages/runtime/src/next/client.ts
fikrikarim: True. Updated.

packages/runtime/src/next/client.ts
fikrikarim: Yeah, but I don't think I'm planning on updating those guards and pipes in this MR. I added a linear issue for this.

PR #594, opened 2023-09-29
Image priority

apledger:


migueloller (commented):
Changes to the Image component look good, but I'm not sure I understand how the changes for background image would work.

Where's the priority coming from?

packages/runtime/src/prop-controllers/descriptors.ts
```diff
@@ -205,6 +205,7 @@ type BackgroundImageV0 = {
   repeat?: BackgroundImageRepeat
   opacity?: number
   parallax?: number
+  priority?: boolean
```
migueloller, line 208: Where would this value be coming from? Are we also updating the backgrounds panel to add a priority?

packages/runtime/src/prop-controllers/descriptors.ts
```diff
@@ -215,6 +216,7 @@ type BackgroundImageV1 = {
...
```
migueloller, line 219: See my other comment on adding `priority` to `BackgroundImageV0`

migueloller (changes requested):
Also, don't forget to add a changeset @apledger

packages/runtime/src/prop-controllers/descriptors.ts
apledger: Yes. I figured we'd merge this first to expect the priority panel before we add it to the UI.

packages/runtime/src/prop-controllers/descriptors.ts
apledger: Same response.

migueloller (approved):

PR #602, opened 2023-10-10
feat: migrate repo to pnpm

joshuawootonn:


.github/workflows/ci.yml
```diff
@@ -22,22 +22,25 @@ jobs:
...
       - name: Set up Node.js environment
         uses: actions/setup-node@v3
         with:
           node-version: 18
-          cache: yarn
+          cache: pnpm
 
       - name: Install dependencies
-        run: yarn install
+        run: pnpm install --ignore-scripts
```
joshuawootonn, line 36: When I don't include this `pnpm install` runs the `prepare` command in the `makeswift` package. This just double builds
CleanShot 2023-10-10 at 15 47 56@2x


Not sure why yarn didn't/doesn't do this.

.github/workflows/ci.yml
```diff
@@ -22,22 +22,25 @@ jobs:
         uses: actions/checkout@v3
         with:
           fetch-depth: 2
-
+      - name: Set up pnpm 8.x
+        uses: pnpm/action-setup@v2
+        with:
+          version: 8
```
joshuawootonn, line 28: pnpm doesn't come pre installed: https://github.com/actions/setup-node/issues/530#issuecomment-1169317505

packages/makeswift/package.json
```diff
@@ -43,6 +43,7 @@
     "open": "^8.4.0"
   },
   "devDependencies": {
+    "@babel/types": "^7.23.0",
```
joshuawootonn, line 46: I used `pnpm import` to copy the lockfile. I am not sure why I had to add this to calm down some type errors that came from using pnpm.

packages/runtime/src/runtimes/react/find-dom-node.tsx
```diff
@@ -23,7 +23,6 @@ function suppressWarningAndFindDomNode(
     return error.apply(console, args)
   }
 
-  // @ts-expect-error @types/react-dom has a different @types/react than the one installed.
```
joshuawootonn, line 26: CleanShot 2023-10-10 at 15 50 44@2x

joshuawootonn:
Here is the previous CI time if you would like to compare: https://github.com/makeswift/makeswift/actions/runs/6436753507/job/17480718871

migueloller (approved):
Looks great! Thanks for doing this ❤️

packages/makeswift/package.json
migueloller: Yarn makes it possible to depend on transitive dependencies that aren't in package.json so it might've just gone unnoticed. It's one of the benefits of pnpm that it doesn't allow you to depend on something like that implicitly.

PR #603, opened 2023-10-10
feat: expose `defaultValue` to `RichText` control

joshuawootonn:
josh/prd-3406

### DIFF
--- a/packages/runtime/src/controls/rich-text-v2/rich-text-v2.ts
+++ b/packages/runtime/src/controls/rich-text-v2/rich-text-v2.ts
@@ -40,7 +40,7 @@ export type RichTextV2Config = {
   defaultValue?: string
 }
 
-type ExternalRichTextV2Config = Omit<RichTextV2Config, 'plugins' | 'defaultValue'>
+type ExternalRichTextV2Config = Omit<RichTextV2Config, 'plugins'>
 
 export type RichTextV2ControlDefinition<T extends RichTextV2Config = RichTextV2Config> = {
   type: typeof RichTextV2ControlType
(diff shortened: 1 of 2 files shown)


joshuawootonn:
@apledger said that this was something that would be helpful on these buildouts.

Thought it would also be good for me to check that the publishing CI still works :)

packages/runtime/src/controls/rich-text-v2/rich-text-v2.ts
```diff
@@ -54,7 +54,10 @@ export function RichText<T extends ExternalRichTextV2Config>(
     type: RichTextV2ControlType,
     config: {
       mode: config.mode,
-      defaultValue: config?.mode === RichTextV2Mode.Inline ? 'Edit this text' : ipsum(3),
+      defaultValue:
+        config?.defaultValue ?? config?.mode === RichTextV2Mode.Inline
+          ? 'Edit this text'
+          : ipsum(3),
```
migueloller, line 60: This won't work as you expect it to

CleanShot 2023-10-10 at 18 41 55

migueloller (changes requested):

packages/runtime/src/controls/rich-text-v2/rich-text-v2.ts
joshuawootonn: How is that possible?

Based on this order of operations chart: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_precedence#table

Wouldn't it resolve as 

```
'default_value' ?? 'one' === 'one' ? 'oh_no' : 'hi'
'default_value' ?? 'oh_no' 
'default_value'
```

joshuawootonn: because `'default_value'` is never null or undefined.

joshuawootonn: Ok https://stackoverflow.com/a/67575538/8239711

migueloller: The higher the number in the chart, the higher the precedence so the nullish coalesce would resolve _first_ but in your comment above you resolve it _last_

PR #622, opened 2023-11-10
feat: add normalization that prevents nested non list root blocks

joshuawootonn:


migueloller (approved):
This looks good. Are you not worried about any unintended consequences from unwrapping non-list root nodes? Perhaps losing some properties that were in the node that's being removed?

joshuawootonn:
I am not to worried. 

I think ideally it would coalesce the possible textAlign types, but since this is the first time this is being reported I am not going to spend the time.

PR #632, opened 2023-12-14
fix: update example runtimes

arvinpoddar:
Resolve mismatch between builder runtime version and example versions, which was causing incompatibility issues with builder sites that were created with images.

### DIFF
--- a/examples/basic-typescript/package.json
+++ b/examples/basic-typescript/package.json
@@ -8,7 +8,7 @@
     "icons": "npx @svgr/cli --typescript --replace-attr-values \"#000=currentColor\" --out-dir generated/icons -- public/icons"
   },
   "dependencies": {
-    "@makeswift/runtime": "0.11.4",
+    "@makeswift/runtime": "0.11.19",
     "@radix-ui/react-accordion": "^1.1.2",
     "@radix-ui/react-tabs": "^1.0.4",
     "clsx": "^2.0.0",
--- a/examples/basic-typescript/pnpm-lock.yaml
+++ b/examples/basic-typescript/pnpm-lock.yaml
@@ -6,8 +6,8 @@ settings:
 
 dependencies:
   '@makeswift/runtime':
-    specifier: 0.11.4
-    version: 0.11.4(next@13.2.1)(react-dom@18.2.0)(react@18.2.0)
+    specifier: 0.11.19
+    version: 0.11.19(next@13.2.1)(react-dom@18.2.0)(react@18.2.0)
   '@radix-ui/react-accordion':
     specifier: ^1.1.2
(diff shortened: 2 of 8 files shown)


migueloller (changes requested):
Looks good, but how come we're doing `0.11.4` when the latest version is `0.11.19`?

arvinpoddar:
> Looks good, but how come we're doing `0.11.4` when the latest version is `0.11.19`?

Consistency - since Alan just bumped to 0.11.4 in the typescript example, I wanted all examples to be consistent with that. I could just bump everything (except bigcommerce) to be 0.11.19. Any preferences?

arvinpoddar:
> > > Looks good, but how come we're doing `0.11.4` when the latest version is `0.11.19`?
> > 
> > 
> > Consistency - since Alan just bumped to 0.11.4 in the typescript example, I wanted all examples to be consistent with that. I could just bump everything (except bigcommerce) to be 0.11.19. Any preferences?
> 
> Let's bump them all to `0.11.19`. We should have all examples be up to date with the latest version of the runtime.
> 
> Also, how come we're no updating the BigCommerce example? Is that build failing as a result of the update? Given the error I'd assume that the example is broken independent of the runtime version being used.

Sounds good. Will do. For the bigcommerce issue, you are correct - the issue is irrelevant to runtime and hasn't built properly in a while. Per my discussion with @joshuawootonn in Slack, the advice was to not upgrade bigcommerce and open a ticket for the build issues. Do you think we should move forward with the upgrade anyways?

migueloller:
> > > > Looks good, but how come we're doing `0.11.4` when the latest version is `0.11.19`?
> > > 
> > > 
> > > Consistency - since Alan just bumped to 0.11.4 in the typescript example, I wanted all examples to be consistent with that. I could just bump everything (except bigcommerce) to be 0.11.19. Any preferences?
> > 
> > 
> > Let's bump them all to `0.11.19`. We should have all examples be up to date with the latest version of the runtime.
> > Also, how come we're no updating the BigCommerce example? Is that build failing as a result of the update? Given the error I'd assume that the example is broken independent of the runtime version being used.
> 
> Sounds good, will update everything. For the bigcommerce issue, you are correct - the issue is irrelevant to runtime and hasn't built properly in a while. Per my discussion with @joshuawootonn in Slack, the advice was to not upgrade bigcommerce and open a ticket for the build issues. Do you think we should move forward with the upgrade anyways?

I still think we should upgrade the runtime because if somebody uses the example but provides their own store they might get things working but if the runtime is outdated they could run into other issues.

We should definitely create an issue in Linear to track the problem with our default store, though.

arvinpoddar:
@migueloller I've updated all package runtimes and the associated lockfiles

migueloller (approved):
Great! Thanks @arvinpoddar

PR #633, opened 2023-12-18
feat: add type to color control

chanceaclark:
Adds the `type` field to the color control.

Fixes PRD-4072

### DIFF
--- a/packages/runtime/src/controls/color.ts
+++ b/packages/runtime/src/controls/color.ts
@@ -1,7 +1,22 @@
-import { CopyContext } from '../state/react-page'
-import { ColorData } from './types'
+import { match } from 'ts-pattern'
+import type { CopyContext } from '../state/react-page'
+import type { ColorData } from './types'
 
-export type ColorControlData = ColorData
+// TODO: We don't know where this is going to live yet, so exporting it
+// globally for now and reexporting it as a namespaced export below.
+export const ControlDataTypeKey = '@@makeswift/type'
+
+export const ColorControlDataTypeKey = ControlDataTypeKey
+
+export const ColorControlDataTypeValueV1 = 'color::v1'
+
+export type ColorControlDataV0 = ColorData
+
+export type ColorControlDataV1 = ColorControlDataV0 & {
+  [ControlDataTypeKey]: typeof ColorControlDataTypeValueV1
+}
+
+export type ColorControlData = ColorControlDataV0 | ColorControlDataV1
 
 export const ColorControlType = 'makeswift::controls::color'
 
@@ -15,20 +30,26 @@ type ColorControlConfig = {
 export type ColorControlDefinition<C extends ColorControlConfig = ColorControlConfig> = {
   type: typeof ColorControlType
   config: C
+  version?: 1
 }
 
 export function Color<C extends ColorControlConfig>(config = {} as C): ColorControlDefinition<C> {
-  return { type: ColorControlType, config }
+  return { type: ColorControlType, config, version: 1 }
 }
 
 export function copyColorData(
-  value: ColorData | undefined,
+  value: ColorControlData | undefined,
   context: CopyContext,
-): ColorData | undefined {
+): ColorControlData | undefined {
   if (value == null) return value
 
-  return {
-    ...value,
-    swatchId: context.replacementContext.swatchIds.get(value.swatchId) ?? value.swatchId,
-  }
+  return match(value)
+    .with({ [ControlDataTypeKey]: ColorControlDataTypeValueV1 }, val => ({
+      ...val,
+      swatchId: context.replacementContext.swatchIds.get(val.swatchId) ?? val.swatchId,
+    }))
+    .otherwise(val => ({
+      ...val,
+      swatchId: context.replacementContext.swatchIds.get(val.swatchId) ?? val.swatchId,
+    }))
 }
--- a/.changeset/chilled-llamas-dream.md
+++ b/.changeset/chilled-llamas-dream.md
@@ -1,0 +1,5 @@
+---
+'@makeswift/runtime': patch
+---
+
+Add type and version to the Color control


packages/runtime/src/controls/style.ts
```diff
@@ -28,7 +28,7 @@ type BorderStyle = 'dotted' | 'dashed' | 'solid'
 export type BorderSideShorthandPropertyData = {
   width: number | null | undefined
   style: BorderStyle
-  color?: ColorData | null
+  color?: ColorControlData | null
```
migueloller, line 31: So this is an interesting change. We want to be able to evolve the Style control independent of the Color control and while we were previous sharing the `ColorData` type, I'm not sure we should continue doing that. This `color` property doesn't come from the Color control, so it would be weird for it to have things specific to the Color control, like its data type.

I think that if this control was somehow using or composing the Color control that would be a different story. But in this case it wasn't, it was just sharing the same TypeScript type and I don't think it should have.

Something else to consider is that it seems `ColorData` was not defining the type for any specific control, like the Color control, but instead was a shared type for data used by multiple controls, like this one in addition to the Color control. We could continue to share `ColorData` and that could be used by both `ColorControlData` and this type and then certain utilities like `copyColorData` could be shared.

For example:

```ts
type ColorControlDataV0 = ColorData

type ColorControlDataV1 = { type: typeof ColorControlDataTypeV1; swatchId: string; alpha: number }

type ColorControlData = ColorControlDataV0 | ColorControlDataV1
```

```ts
color?: ColorData | null // this stays as `ColorData` instead of `ColorControlData`
```

This would result in no changes to the Style control while continuing to share the underlying color shape. Thoughts?

migueloller (approved):
LGTM 👍🏻 

I'm guessing there wasn't a need to change any of the consumers of `ColorControlData` since we're just adding a field so it isn't a breaking change.

migueloller (approved):

migueloller:
@chanceaclark, there's something else I noticed and that's that the new control type isn't in the configuration, which means it's impossible for the builder to know whether v0 or v1 of the control is being used, meaning that the builder can't possibly go from no data (i.e., `undefined`) to a valid v1 value since it has no way of knowing if it's v0 or v1

packages/runtime/src/controls/style.ts
chanceaclark: I like but idea, but I want to see if we can _extend_ that thought a little further. If we intend to have some sort of base types, e.g. `color`, instead of using that base type directly in our interfaces, we extend that base type (create a new type) to use within the implementation?

e.g.

```ts
interface ColorData {
  swatchId: string;
  alpha: number;
}

// ....

interface ColorControlDataV0 extends ColorData {}

interface ColorControlDataV1 extends ColorData {
 type: typeof ColorControlDataTypeV1;
}
```

This way the types being consumed by the component can evolve more independently from the base type, while still sharing the underlying representation of it.

chanceaclark:
@migueloller I might be missing something here like validation in the builder but can't you infer that it's v0 if `type: undefined` and v1 if `type: "makeswift-color::v1"`? I saw the config had a `version` key but was unsure why that was used.

migueloller:
> @migueloller I might be missing something here like validation in the builder but can't you infer that it's v0 if `type: undefined` and v1 if `type: "makeswift-color::v1"`? I saw the config had a `version` key but was unsure why that was used.

How would the builder know what version to use if the prop has never been set?

packages/runtime/src/controls/style.ts
migueloller: @chanceaclark, I can’t seem to find a comment that I received an email for, but yeah, having interfaces and extending would work well. We could also use types and intersections if we’d rather use types over interfaces.

chanceaclark: I ended up removing the comment after reading your comment for the third time. It was pretty much close in nature. The only difference was types vs. interfaces, which when intersecting type assignments you get the same type rather than a new type. e.g.
Screenshot 2024-01-02 at 15 24 11

chanceaclark:
>How would the builder know what version to use if the prop has never been set?

Ah, I see. We set it in the builder when consuming this new version.

packages/runtime/src/controls/color.ts
```diff
@@ -15,20 +22,20 @@ type ColorControlConfig = {
 export type ColorControlDefinition<C extends ColorControlConfig = ColorControlConfig> = {
   type: typeof ColorControlType
   config: C
+  version?: 1,
```
migueloller, line 25: If we're planning on keeping backwards-compatible versions of the control definition we might want to have different versions for the entire definition.

migueloller (changes requested):

packages/runtime/src/controls/color.ts
chanceaclark: As in, keep them separate. e.g. `ColorControlDefinitionV0` vs. `ColorControlDefinitionV1`?

migueloller: Yup

packages/runtime/src/controls/style.ts
```diff
@@ -385,5 +385,132 @@ if (import.meta.vitest) {
       // Assert
       expect(result).toMatchObject(expected)
     })
+
+    test('colors are replaced - v1', () => {
```
migueloller, line 389: There should be no need for this test as the Style control should be using `ColorData`, not control specific types which is versioned.

Basically, this file shouldn't change since we're only updating the Color control, not the Style control

packages/runtime/src/controls/style.ts
migueloller, line 31: This should still be `ColorData`

migueloller (changes requested):

migueloller (approved):
LGTM 👍🏻

PR #643, opened 2024-01-05
feat: add type field to Checkbox control

chanceaclark:
Adds the `type` field to the Checkbox control.

### DIFF
--- a/packages/runtime/src/controls/control-data-type-key.ts
+++ b/packages/runtime/src/controls/control-data-type-key.ts
@@ -1,0 +1,1 @@
+export const ControlDataTypeKey = '@@makeswift/type'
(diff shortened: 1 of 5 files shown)


packages/runtime/src/controls/control-data-type-key.ts
```diff
@@ -0,0 +1 @@
+export const ControlDataTypeKey = '@@makeswift/type'
```
chanceaclark, line 1: Would love some input here, what the name of this file should be or where it should live.

migueloller: I think it's fine here for now. We can find a better place for it later.

migueloller (approved):

PR #646, opened 2024-01-08
feat: add type field to TextInput control

chanceaclark:
Adds the type and version field for the Number control.


migueloller (approved):
LGTM 👍🏻

PR #648, opened 2024-01-11
feat: enable versioning by default

fikrikarim:
Thoughts about this MR:

1. Should we make make `siteVersion` parameter optional on public methods `getPage()` and `getPages()`?
2. Should we make the `siteVersion` parameter an object parameter like this?
```ts
async getGlobalElement(id, { siteVersion })
```
3. I think for "internal methods" like `getSwatch`, we need to make the `siteVersion` required, because otherwise we will forget to pass the `siteVersion` when we use it internally. But for external API, I'm making those optional for now, not sure which one is better here.
4. On that note, it's so weird to have a Makeswift client, but we expose our "internal methods" like `getSwatch`. In the future, it might be better to separate Makeswift client that users will use, and a Makeswift client that we use internally.

### DIFF
--- a/packages/runtime/src/next/index.tsx
+++ b/packages/runtime/src/next/index.tsx
@@ -98,7 +98,7 @@ export async function getStaticProps(
   })
   const path = '/' + (ctx.params?.path ?? []).join('/')
   const snapshot = await makeswift.getPageSnapshot(path, {
-    preview: ctx.previewData?.makeswift === true,
+    siteVersion: Makeswift.getSiteVersion(ctx.previewData),
   })
 
   if (snapshot == null) return { notFound: true, revalidate: REVALIDATE_SECONDS }
@@ -115,7 +115,9 @@ export async function getServerSideProps(
     apiOrigin: getApiOrigin(),
   })
   const path = '/' + (ctx.params?.path ?? []).join('/')
-  const snapshot = await makeswift.getPageSnapshot(path, { preview: true })
+  const snapshot = await makeswift.getPageSnapshot(path, {
+    siteVersion: Makeswift.getSiteVersion(ctx.previewData),
+  })
 
   if (snapshot == null) return { notFound: true }
 
--- a/packages/runtime/src/next/index.tsx
+++ b/packages/runtime/src/next/index.tsx
@@ -17,7 +17,7 @@ import {
   GetStaticPropsResult,
 } from 'next'
 import { Makeswift } from './client'
-import { MakeswiftPreviewData } from './preview-mode'
+import { MakeswiftPreviewData, MakeswiftSiteVersion } from './preview-mode'
 
 function getApiOrigin(): string {
   const apiOriginString = process['env'].MAKESWIFT_API_HOST ?? 'https://api.makeswift.com'
@@ -116,7 +116,7 @@ export async function getServerSideProps(
   })
   const path = '/' + (ctx.params?.path ?? []).join('/')
   const snapshot = await makeswift.getPageSnapshot(path, {
-    siteVersion: Makeswift.getSiteVersion(ctx.previewData),
+    siteVersion: MakeswiftSiteVersion.Working,
   })
 
   if (snapshot == null) return { notFound: true }
--- a/.changeset/sixty-islands-exist.md
+++ b/.changeset/sixty-islands-exist.md
@@ -13,7 +13,7 @@ Upgrade guide from version `0.12.x` to `0.13.x`:
    b. Add the new `siteVersion` parameter.
 
    ```diff
-     export async function getStaticProps(ctx){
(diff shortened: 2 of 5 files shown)


alexhwoods:
It seems I'm still able to pass the `siteVersions` parameter to the API handler config, although it isn't having any effect.

 Could we make this a type error?

```import { MakeswiftApiHandler } from '@makeswift/runtime/next'

export default MakeswiftApiHandler(process.env.MAKESWIFT_SITE_API_KEY!, {
  appOrigin: process.env.MAKESWIFT_APP_ORIGIN,
  apiOrigin: process.env.MAKESWIFT_API_ORIGIN,
  siteVersions: false,
})
```

fikrikarim:
> It seems I'm still able to pass the `siteVersions` parameter to the API handler config, although it isn't having any effect.
> 
> Could we make this a type error?
> 
> ```
> 
> export default MakeswiftApiHandler(process.env.MAKESWIFT_SITE_API_KEY!, {
>   appOrigin: process.env.MAKESWIFT_APP_ORIGIN,
>   apiOrigin: process.env.MAKESWIFT_API_ORIGIN,
>   siteVersions: false,
> })
> ```

Oh yeah thanks for catching that. I removed the type from manifest but I forgot to remove it from the API handler. Thanks!

alexhwoods:
> > It seems I'm still able to pass the `siteVersions` parameter to the API handler config, although it isn't having any effect.
> > Could we make this a type error?
> > ```
> > 
> > export default MakeswiftApiHandler(process.env.MAKESWIFT_SITE_API_KEY!, {
> >   appOrigin: process.env.MAKESWIFT_APP_ORIGIN,
> >   apiOrigin: process.env.MAKESWIFT_API_ORIGIN,
> >   siteVersions: false,
> > })
> > ```
> 
> Oh yeah thanks for catching that. I removed the type from manifest but I forgot to remove it from the API handler. Thanks!

alright, confirmed fixed!
CleanShot 2024-01-11 at 11 50 02

alexhwoods (approved):

packages/runtime/src/next/index.tsx
```diff
@@ -115,7 +115,9 @@ export async function getServerSideProps(
     apiOrigin: getApiOrigin(),
   })
   const path = '/' + (ctx.params?.path ?? []).join('/')
-  const snapshot = await makeswift.getPageSnapshot(path, { preview: true })
+  const snapshot = await makeswift.getPageSnapshot(path, {
+    siteVersion: Makeswift.getSiteVersion(ctx.previewData),
+  })
```
migueloller, line 120: This needs to be hardcoded to `MakeswiftSiteVersion.Working` to preserve its behavior.

migueloller (changes requested):
Looks good. Just have 1 request to fix the behavior of `getServerSideProps`

migueloller (approved):
LGTM!

migueloller:
> Thoughts about this MR:
> 
> 1. Should we make make `siteVersion` parameter optional on public methods `getPage()` and `getPages()`?

We could make it optional but it feels better to make them required so that it's harder to make a mistake

> 2. Should we make the `siteVersion` parameter an object parameter like this?
> 
> ```ts
> async getGlobalElement(id, { siteVersion })
> ```

I would prefer named parameter like this, yes, but it seems that they ended up being positional parameters.

> 3. I think for "internal methods" like `getSwatch`, we need to make the `siteVersion` required, because otherwise we will forget to pass the `siteVersion` when we use it internally. But for external API, I'm making those optional for now, not sure which one is better here.

Yes! For the same reasons we can make the external API required.

> 4. On that note, it's so weird to have a Makeswift client, but we expose our "internal methods" like `getSwatch`. In the future, it might be better to separate Makeswift client that users will use, and a Makeswift client that we use internally.

I think in the future we're wanting to expose all these things as part of our public API. But in the mean time, yeah, might make sense to have separate private and public clients.

migueloller:
> It seems I'm still able to pass the `siteVersions` parameter to the API handler config, although it isn't having any effect.
> 
> Could we make this a type error?
> 
> ```
> 
> export default MakeswiftApiHandler(process.env.MAKESWIFT_SITE_API_KEY!, {
>   appOrigin: process.env.MAKESWIFT_APP_ORIGIN,
>   apiOrigin: process.env.MAKESWIFT_API_ORIGIN,
>   siteVersions: false,
> })
> ```

We could do that by doing `siteVersions?: never`

fikrikarim:
> I would prefer named parameter like this, yes, but it seems that they ended up being positional parameters.

During my first iteration, I actually used a named parameter. But I found it weird to mix a  required named parameter together with a required positional parameter like this:
```
async getGlobalElement(id, { siteVersion })
```

It's a different story if the named parameter is an optional parameter like `connection`. But we can always change this later.

PR #654, opened 2024-01-16
fix: update BC example to use new BC integration

joshuawootonn:


joshuawootonn:
@migueloller There is a new BigCommerce integration that we are moving to that isn't 500'ing

joshuawootonn:
I updated the 1pass, and updated the vercel env vars

PR #660, opened 2024-01-26
RFC: Jest vs Playwright for testing runtime change

joshuawootonn:
Hey @migueloller and @chanceaclark I wanted to get your feedback on this testing project for the runtime.

I tried to just use jest, but I opted for Playwright in later commits. (Comments on why are in the commit)

Before I delete the Jest stuff I wanted to get feedback that you are both cool with moving in the playwright direction.


test/__tests__/hello-world.test.tsx
```diff
@@ -0,0 +1,24 @@
...
+  it('renders a heading containing the right text', () => {
+    // No good way to render layout here without manually rendering,
+    // and updating the container option to not be a `body` tag
+    const appScreen = render(<HelloWorldApp />)
+    const appHeading = appScreen.getByRole('heading', { level: 1 })
+    // Not working as expected with tailwind styles. Styles are default.
+    // const styles = window.getComputedStyles()
+
+    // Kinda weird I have to do this.
+    appScreen.unmount()
```
joshuawootonn, line 16: Testing library `render` calls reuse the same context, and render the component in isolation. The comments in this file are my main hesitations with using Jest.

joshuawootonn:
If we are cool with using Playwright, then my plan is to name change from `spec` to `test` and remove the jest config specific stuff. I'll do that in a commit before I merge. Approving, just means you approve the direction.

test/__specs__/hello-world.spec.ts
```diff
@@ -0,0 +1,15 @@
...
+test('Hello World test', async ({ page }) => {
+  await page.goto('/hello-world-app')
+  let heading = await page.locator('h1')
+  const appFontWeight = await heading.evaluate((element) => {
+    return window.getComputedStyle(element).fontWeight
+  })
+  await page.goto('/hello-world-page')
+  heading = await page.locator('h1')
+  const pageFontWeight = await heading.evaluate((element) => {
+    return window.getComputedStyle(element).fontWeight
```
joshuawootonn, line 12: Nice thing about playwright is you are getting the entire page -  just like a browser. So `getComputedStyles` is real world and contains all the properties the element has on the page. 

In my jest test, this wasn't the case. 

I think `getComputedStyles` or something similar is how I would write the server style registry test.

test/pages/index.tsx
```diff
@@ -0,0 +1,3 @@
+export default function Home() {
```
joshuawootonn, line 3: Playwright pings your base URL to know that your webServer is running. So I created this page.

joshuawootonn:
Currently, adding playwright to our CI. That's why it's failing.

test/__specs__/hello-world.spec.ts
chanceaclark: Yeah, jest uses `jsdom` as the virtual DOM, which has some [unimplemented APIs](https://github.com/jsdom/jsdom?tab=readme-ov-file#unimplemented-parts-of-the-web-platform). I think using playwright to use more of the DOM features is the right way to go here (and not just about these types of features, but for actually testing browser interactions).

joshuawootonn:
Ok CI is passing now. Kinda interesting, since I didn't make the test directory part of packages I had to `pnpm install --ignore-workspace` to run the test command there.

@migueloller I got the impression from our conversation / the nextjs repo that we should have the test dir seperate.

At this point I don't really see what we gain from doing that. If we do keep them seperate though, I'll probably break out the github action so that they can run in parallel.

.github/workflows/ci.yml
```diff
@@ -44,3 +44,22 @@ jobs:
       - name: Circular dependency check
         continue-on-error: true
         run: pnpm run madge
+
+      - name: Install /test dependencies
+        run: pnpm install --ignore-workspace
+        working-directory: ./test
+
+      - name: Install Playwright Browsers
+        run: pnpm exec playwright install --with-deps
```
chanceaclark, line 53: Something to be aware about, if we end up using snapshots to test diffs, running playwright in the github runner vs. locally will actually cause snapshots to be different. The reason being is that this Github runner is using `linux` and local machine will be using `macos` unless you configure the Github running to use the `macos` runner. The snapshot diffs will be different in that case, so maybe we can write tests to avoid snapshotting.

joshuawootonn: Hmm. Interesting.

What do you mean by snapshotting? I usually think of snapshot testing as diffing some sort of media( json, image, etc). 

I am thinking of just diffing attributes in the DOM, like CSS outputs. I think this should be consistent between os.

chanceaclark: https://playwright.dev/docs/test-snapshots

It's basically visual regression testing, which again, I don't think we'll use here.

migueloller: Why would the snapshots be different in different operating systems? I would expect it to be possible to have consistent snapshots regardless of the OS.

migueloller:
> Ok CI is passing now. Kinda interesting, since I didn't make the test directory part of packages I had to `pnpm install --ignore-workspace` to run the test command there.

What do you mean by "didn't make the test directory part of packages"? Do you mean that you didn't make the Next.js app a pnpm workspace? If we want to install dependencies then we need to add the package to the workspaces so that `pnpm install` will install dependencies. Maybe that's the problem you were running into?

> @migueloller I got the impression from our conversation / the nextjs repo that we should have the test dir seperate.
>
> At this point I don't really see what we gain from doing that. If we do keep them seperate though, I'll probably break out the github action so that they can run in parallel.

I think I'm missing some context here. The reason I suggested we create a test folder was so that we could organize our tests and keep them separate from source code and examples.

pnpm-lock.yaml
migueloller, line 1: I'm curious how we got a `pnpm-lock.yaml` getting generated outside of `test` without having added `test` to the pnpm workspaces

test/__tests__/hello-world.test.tsx
migueloller: I think that the reason things aren't working as you would want them to is that we're using the wrong tool for the job here. `@testing-library/react` is designed to test React components in isolation but based on the comments it seems you were expecting things _outside_ of the React component, like Tailwind CSS files, to be considered.

I think that we need to define the intent of what we're trying to do. If we're trying to write tests from the perspective of the end-user, visiting a browser, then Playwright can give us that. But if we're looking to write unit tests where we can see how different props result different component behavior, then `@testing-library/react` is very good for that.

I do agree it's weird we have to do `appScree.unmount()`. It makes sense since `render` seems to be using shared global state but it would be nice if there was a way to set up a new `appScreen` on each test and have that be isolated. Perhaps that's an option?

test/__specs__/hello-world.spec.ts
migueloller: > In my jest test, this wasn't the case.

I think there's a misunderstanding here about what Jest does out of the box. Maybe we should discuss what our testing goals are so we can decide what tool we want to use?

test/pnpm-lock.yaml
migueloller, line 1: This file is what I was expecting to see originally given we didn't add the package to the pnpm workspaces.

.github/workflows/ci.yml
migueloller, line 1: I do think it would be a good idea to explire having separate workflows instead of putting everything in `ci.yml`. Maybe now's not the time but we should at least explore it.

.github/workflows/ci.yml
chanceaclark: Snapshot files have the os as part of their name that they compare against:
Screenshot 2024-01-26 at 14 35 06

You'll have to spin up a docker container to update your snapshots.

pnpm-lock.yaml
joshuawootonn: When I ran pnpm install in the `/test` directory it updated the lock file of the surrounding workspace.

Maybe I fix this by making the test folder it's own workspace.

joshuawootonn:
> What do you mean by "didn't make the test directory part of packages"? Do you mean that you didn't make the Next.js app a pnpm workspace? If we want to install dependencies then we need to add the package to the workspaces so that pnpm install will install dependencies. Maybe that's the problem you were running into?

> Do you mean that you didn't make the Next.js app a pnpm workspace?
I meant that I didn't add the `/test` directory as a package in the existing workspace.

I think maybe the solution here is to just to make the test directory it's own sub workspace

test/__tests__/hello-world.test.tsx
joshuawootonn: Talked about in person. Think this is resolved

.github/workflows/ci.yml
joshuawootonn: This is good to know in case we go down that route.

joshuawootonn:
Demo:

https://github.com/makeswift/makeswift/assets/20950876/b90f946e-82c5-4a67-b01e-0da7e081f14b

PR #661, opened 2024-01-26
Overhaul build in preparation for Next.js App Router support

migueloller:
This PR refactors the build setup for `@makeswift/runtime` so that it can be used with Next.js App Router. Specifically, it accomplishes the following:
- Transpiles files and preserves the directory structure under `src`
- Removes SVG files in favor of React components
- Replaces Vite with tsup

Transpiling files instead of bundling makes it easier for `'use client'` directives to be consumed by Next.js in the way they were written in `src`. Furthermore, it makes debugging from `node_modules` a lot easier. Removing SVG files was necessary to leverage esbuild since transforming JS files to rewrite imports was more trouble than it was worth. Finally, replacing Vite with tsup was done to replace Rollup, which Vite uses, with esbuild, which tsup uses. This with the intent of having a speedy build, avoiding bundling, and also maintaining `'use client'` directives.

There were also various other changes made to how distributed files are setup, replacing JS files with directories with package.json files.

These changes, in addition to the build changes, were inspired by https://github.com/clerk/javascript/tree/main/packages/nextjs


packages/runtime/package.json
```diff
@@ -2,113 +2,91 @@
...
     }
   },
+  "types": "./dist/types/index.d.ts",
   "scripts": {
-    "dev": "concurrently -k 'tsc --watch --preserveWatchOutput' 'vite build --watch'",
+    "dev": "concurrently -k 'tsc --watch --preserveWatchOutput' 'tsup --watch'",
     "clean": "rm -rf dist",
-    "build": "pnpm run clean && tsc && vite build",
-    "preview": "vite preview",
+    "build": "pnpm run clean && tsc && tsup",
```
chanceaclark, line 85: What do you think about adding the `clean` option to tsup, omitting `rm -rf dist` and swapping `tsup` and `tsc`? e.g.
```
"build": "tsup && tsc"
```

chanceaclark (dismissed):
Small comment but looks good on my end.

packages/runtime/package.json
migueloller: I played around with doing stuff like that as well as driving `tsc` from the `tsup` config or using `dts`. Ended up going with this because it was equivalent to what we had before with Vite but we could also do it that way too. It's worth noting, though, that running `tsup` would blast away `tsc`'s output which might be annoying. I'm not sure how often one would run `tsup` by itself instead of running `build`, though. `tsc` is sometimes slow so perhaps quite a bit? 🤷🏻‍♂️

chanceaclark: I would assume if you are going to rebuild, you'll most likely need to compile the types too, unless you are only modifying internals and not external facing APIs 🤷‍♂️  We could also speed up `tsc` using `incremental`?

chanceaclark (approved):

examples/bigcommerce/README.md
```diff
@@ -74,18 +74,7 @@ Note: the env for this example corresponds to the [Vercel Commerce BigCommerce e
 - `BIGCOMMERCE_ACCESS_TOKEN` requires an [API account](https://support.bigcommerce.com/s/article/Store-API-Accounts?language=en_US)
 
   - This token needs modify rights on Cart, Checkout, and storefront API Tokens, and read rights on Products and Content.
-<<<<<<< HEAD
```
joshuawootonn, line 77: oof

PR #662, opened 2024-01-29
chore: rearrange react runtime code

chanceaclark:
This code just changes where some of the react runtime code lives. All the exports and internals should be the same.


packages/runtime/src/runtimes/react/controls/rich-text-v2/EditableTextV2/editable-text-v2.tsx
```diff
@@ -189,7 +189,7 @@ export function EditableTextV2({ text, definition, control }: Props) {
 
   const handleClick = useCallback(
     (e: MouseEvent) => {
-      if (editMode === BuilderEditMode.CONTENT){
+      if (editMode === BuilderEditMode.CONTENT) {
```
chanceaclark, line 192: Ran prettier on the `src/runtimes/react` folder and caught some formatting issues here and there

packages/runtime/src/runtimes/react/hooks/use-breakpoints.ts
chanceaclark, line 1: Not sure what we wanted to do naming convention wise around these hooks. I'm down if we wanted to do camelCase here, just was following the same convention of a bunch of other files surrounding these.

packages/runtime/src/runtimes/react/index.ts
```diff
@@ -0,0 +1,15 @@
...
+export { PageProvider } from './components/PageProvider'
+
+export { StoreContext, useStore } from './hooks/use-store'
+export { useBreakpoints } from './hooks/use-breakpoints'
+export { useBuilderEditMode } from './hooks/use-builder-edit-mode'
+export { useElementId } from './hooks/use-element-id'
+export { useIsInBuilder } from './hooks/use-is-in-builder'
+export { useIsPreview } from './hooks/use-is-preview'
+export { usePageId } from './hooks/use-page-id'
+export { useDocumentKey } from './hooks/use-document-key'
```
chanceaclark, line 15: I doubled checked, but these should be all the same exports as before.

packages/runtime/src/runtimes/react/utils/suppress-ref-warning.ts
```diff
@@ -0,0 +1,23 @@
+const originalError = console.error
+let patched = false
+
+/**
+ * @see https://github.com/facebook/react/blob/a2505792ed17fd4d7ddc69561053c3ac90899491/packages/react-reconciler/src/ReactFiberBeginWork.new.js#L1814-L1890
+ */
+export function suppressRefWarning(ownerName: string): void {
```
chanceaclark, line 7: I thought about co-locating this with its usage as it's only being used once but wanted to get consensus around it.

packages/runtime/src/runtimes/react/hooks/use-breakpoints.ts
migueloller: I prefer kebab-case for filenames

packages/runtime/src/runtimes/react/utils/suppress-ref-warning.ts
migueloller: It's funny but this is actually not doing anything right now because where we pass the ref changed when we introduced `<RenderHook>`. The plan is to fix this once we refactor that out. As for whether this is colocated or not, don't have a strong preference there.

packages/runtime/src/runtimes/react/index.ts
migueloller: We're missing `useSelector` but every other export seems to be here

migueloller (changes requested):
Just had a minor comment on exports but otherwise LGTM 👍🏻

packages/runtime/src/runtimes/react/index.ts
chanceaclark: Appreciate that catch! Should be good now.

packages/runtime/src/runtimes/react/utils/suppress-ref-warning.ts
chanceaclark: I'll leave it then!

migueloller (approved):

PR #663, opened 2024-01-29
Fix issue that caused a client-side exception to be thrown in Safari and Firefox

migueloller:
Unkowingly we were shipping server-only code to the browser by having dependencies from `MakeswiftApiHandler` exported at `@makeswift/runtime/next`. This became evident when customers started reporting client-side exceptions on their sites in Safari and Firefox. We were able to track down the issue to a change that was recently made to `follow-redirects`, which is a transitive dependency of `http-proxy` which `@makeswift/runtime` uses for the Preview Mode proxy handler. More info on their issue tracker: https://github.com/follow-redirects/follow-redirects/issues/237

To address the problem this PR introduces a new entry point, `@makeswift/runtime/next/server`, which exports server-only code.

The PR also includes some fixes that were necessary after the latest refactor from Vite to tsup. Specifically, there were issues with circular dependencies and exports for the Next.js plugin were broken. It also includes a QOL change to the tsconfig.


packages/runtime/tsconfig.json
```diff
@@ -19,8 +19,7 @@
     "rootDir": "src",
     "declarationDir": "dist/types",
     "emitDeclarationOnly": true,
-    "jsx": "react-jsx",
-    "incremental": true
```
chanceaclark, line 23: 🍹 Totally forgot that incremental is more useful for apps out-of-the-box rather than libraries, which require a bit more "fine-tuning". We _should_ look into this more, but we will need to think about cleaning that file whenever we need to removed along with potentially caching it in CI. We accounted for that in [BigDesign](https://github.com/bigcommerce/big-design/blob/fa36fb635610ef891a86491524aa15a8f6fe3772/packages/big-design-icons/package.json#L24) as we had a slow icon build process.

packages/runtime/package.json
```diff
@@ -52,10 +52,14 @@
...
       "types": "./dist/types/next/index.d.ts"
     },
+    "./next/server": {
+      "require": "./dist/cjs/next/server.js",
+      "module": "./dist/esm/next/server.js",
+      "types": "./dist/types/next/server.d.ts"
+    },
     "./next/plugin": {
-      "require": "./node_modules/@makeswift/next-plugin/index.js",
-      "import": "./node_modules/@makeswift/next-plugin/index.js",
```
chanceaclark, line 57: Are we not exposing a ES module anymore?

migueloller: Not for the Next.js plugin. It's a CommonJS module right now and if we want to expose an ES module we'll need to do some work so that `@makewift/next-plugin` exports an ES module as well.

chanceaclark (dismissed):

PR #670, opened 2024-02-06
Josh/eng 4288

joshuawootonn:


joshuawootonn:
Our CLI assumed that `create-next-app` was installed. I don't think this is a good assumption. So I am adding `@latest`
CleanShot 2024-02-06 at 16 35 35@2x

packages/makeswift/src/create-next-app.ts
```diff
@@ -14,7 +14,7 @@ export function createNextApp({
   usePnpm: boolean
 }): void {
   const url = coerceExampleToUrl(example)
-  const npxArgs = ['--yes', 'create-next-app', '--example', url, dir]
+  const npxArgs = ['--yes', 'create-next-app@latest', '--example', url, dir]
```
joshuawootonn, line 21: https://github.com/makeswift/makeswift/pull/670#issuecomment-1930791374

migueloller (dismissed):
Looks great!

PR #675, opened 2024-02-07
Make `Page` a React client component

migueloller:
This pull request introduces a couple of changes to make it possible to use the Makeswift `Page` component in Next.js' App Router like so:

```tsx
import { client } from '@/makeswift/client'

type ParsedUrlQuery = { path?: string[] }

export async function generateStaticParams() {
  const pages = await client.getPages()

  return pages.map((page) => ({
    path: page.path.split('/').filter((segment) => segment !== ''),
  }))
}

export default async function Page({ params }: { params: ParsedUrlQuery }) {
  const snapshot = await client.getPageSnapshot(path, { siteVersion: getSiteVersion() })

  if (snapshot == null) return notFound()

  return <MakeswiftPage snapshot={snapshot} />
}
``` 

`getSiteVersion` isn't implemented just yet but it should now be possible to use `Page`, aliased above as `MakeswiftPage` as shown above. To accomplish this, it was necessary to remove the `runtime` prop from the `Page` component and move it to a separate provider that can be used in the root layout as so:

```tsx
import { MakeswiftProvider } from './makeswift-provider'

export default function RootLayout({
  children,
}: Readonly<{
  children: React.ReactNode
}>) {
  return (
    <html lang="en">
      <body>
        <MakeswiftProvider>{children}</MakeswiftProvider>
      </body>
    </html>
  )
}
```

And here is `makeswift-provider.tsx`:

```tsx
'use client'

import {
  ReactRuntimeProvider,
  RootStyleRegistry,
} from '@makeswift/runtime/next'

export function MakeswiftProvider({ children }: { children: React.ReactNode }) {
  return (
    <ReactRuntimeProvider runtime={runtime}>
      <RootStyleRegistry>{children}</RootStyleRegistry>
    </ReactRuntimeProvider>
  )
}
```

For Pages Router integrations, the provider can be placed in the Custom App:

```tsx
import { ReactRuntimeProvider } from '@makeswift/runtime/next'
import type { AppProps } from 'next/app'

export default function App({ Component, pageProps }: AppProps) {
  return (
    <ReactRuntimeProvider runtime={runtime}>
      <Component {...pageProps} />
    </ReactRuntimeProvider>
  )
}
```

As part of this change I also had to shift a couple of files around so that we could split off client-only imports, like those used for React context.


chanceaclark (approved):

joshuawootonn (approved):
Looks great

arvinpoddar (approved):

arvinpoddar:
General clarification: so this change was made because we must only pass serializable data as props (for introspection purposes)? How does this fit into the broader goal of app router support?

migueloller:
> General clarification: so this change was made because we must only pass serializable data as props (for introspection purposes)? How does this fit into the broader goal of app router support?

Passing serializable data as props isn't an introspection requirement but a requirement of React when a server component is rendering a client-component. If our `Page` client component required a `runtime` prop that wasn't serializable it wouldn't be possible to render it from a server component in the App Router as I show above in the PR description.

PR #676, opened 2024-02-08
chore: update changsets so it properly ignores packages

joshuawootonn:
CleanShot 2024-02-08 at 11 32 49@2x

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -7,5 +7,5 @@
   "access": "public",
   "baseBranch": "main",
   "updateInternalDependencies": "patch",
-  "ignore": ["*nextjs-app-router*", "*nextjs-pages-router*"]
+  "ignore": ["nextjs-app-router", "nextjs-pages-router", "tsconfig"]
 }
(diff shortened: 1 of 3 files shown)


.changeset/config.json
```diff
@@ -7,5 +7,5 @@
   "access": "public",
   "baseBranch": "main",
   "updateInternalDependencies": "patch",
-  "ignore": ["*nextjs-app-router*", "*nextjs-pages-router*"]
```
joshuawootonn, line 10: Not sure how the heck this is working for them.

CleanShot 2024-02-08 at 11 37 00@2x

In the source code they are doing ^ and not using micromatch glob expressions like they are supposed to, and 
CleanShot 2024-02-08 at 11 38 35@2x

I am removing the * since they aren't helping anyways

joshuawootonn: Regardless in the version we were on `2.22.0` the `ignore` array wasn't being checked for the `add` command. That was the root issue. It was added in `2.25.1`

migueloller (approved):

PR #677, opened 2024-02-08
Add local dev workflow for CLI

joshuawootonn:


migueloller (approved):
Thanks @joshuawootonn!

Some suggestions:
- To avoid having to explain differences between your directory structure and others' you could have all directories be relative to the git repository root (i.e., `./packages/maekswift` where `.` is the same directory where the `.git` dir is). You could also do it relative to where this README file is.
- If you use `mktemp` you can have a temporary folder that the OS will delete later. Maybe this isn't desirable, though.
```sh
cd $(mktemp -d)
```

chanceaclark (dismissed):

joshuawootonn:
I like the `cd $(mktemp -d)` idea. And I made some other modifications (before this push) so that the notice at the top isn't really necessary. 

Going to add `cd $(mktemp -d)` and remove the notice. Thanks for the reviews.

PR #679, opened 2024-02-13
Josh/eng 4383

joshuawootonn:


apps/nextjs-app-router/src/localization.ts
migueloller, line 1: Are the extra newlines at the beginning and end of this file intentional?

apps/nextjs-pages-router/next.config.mjs
```diff
@@ -5,6 +5,10 @@ const withMakeswift = createWithMakeswift()
 /** @type {import('next').NextConfig} */
 const nextConfig = {
   reactStrictMode: true,
+  i18n: {
+    locales: ["en-US", "en-FR", "fr-FR"],
```
migueloller, line 9: Maybe it's just me but I don't want us to have `en-FR` in these examples. It's a weird quirk that BigCommerce uses but I wouldn't expect to be a common thing for people to do, specially because these locales aren't even meant to be public (i.e., they're only used "internally" and no visitor to the site should be able to see these locales).

apps/nextjs-pages-router/src/pages/[[...path]].tsx
```diff
@@ -12,6 +12,7 @@ import {
 
 import { client } from '@/makeswift/client'
 import '@/makeswift/components'
+import { runtime } from '@/makeswift/runtime'
```
Comment (migueloller) [inline apps/nextjs-pages-router/src/pages/[[...path]].tsx:15]: This looks like an unused import.

apps/nextjs-app-router/src/middleware.ts
```diff
@@ -0,0 +1,33 @@
+import { NextRequest, NextResponse } from 'next/server'
+import { match } from '@formatjs/intl-localematcher'
+import Negotiator from 'negotiator'
+import { defaultLocale, locales } from '@/localization'
+
+function getLocale(request: NextRequest) {
```
migueloller, line 6: Can we add a return type?

apps/nextjs-app-router/src/middleware.ts
```diff
@@ -0,0 +1,33 @@
...
+  const acceptLanguageHeader = request.headers.get('accept-language')
+  let languages = new Negotiator({
+    headers: { 'accept-language': acceptLanguageHeader ?? undefined },
+  }).languages()
+
+  return match(languages, locales, defaultLocale)
+}
+
+export function middleware(request: NextRequest) {
```
migueloller, line 15: Can we add a return type?

apps/nextjs-app-router/src/middleware.ts
```diff
@@ -0,0 +1,33 @@
+import { NextRequest, NextResponse } from 'next/server'
+import { match } from '@formatjs/intl-localematcher'
+import Negotiator from 'negotiator'
+import { defaultLocale, locales } from '@/localization'
+
+function getLocale(request: NextRequest) {
+  const acceptLanguageHeader = request.headers.get('accept-language')
+  let languages = new Negotiator({
+    headers: { 'accept-language': acceptLanguageHeader ?? undefined },
```
migueloller, line 9: Why do we need to coalesce `undefined` to `undefined`? Can't we do

```ts
new Negotiator({ 'Accept-Language': request.headers.get('Accept-Language') }).languages()
```

Or even cleaner

```ts
new Negotiator(request.headers).languages()
```

?

apps/nextjs-app-router/src/middleware.ts
```diff
@@ -0,0 +1,33 @@
+import { NextRequest, NextResponse } from 'next/server'
+import { match } from '@formatjs/intl-localematcher'
+import Negotiator from 'negotiator'
+import { defaultLocale, locales } from '@/localization'
+
+function getLocale(request: NextRequest) {
+  const acceptLanguageHeader = request.headers.get('accept-language')
+  let languages = new Negotiator({
```
migueloller, line 8: nit: we use `const` everywhere else

apps/nextjs-app-router/src/middleware.ts
migueloller, line 6: Should we drop this function in `src/localization` or is the logic middleware-specific?

migueloller (dismissed):

apps/nextjs-app-router/src/localization.ts
```diff
@@ -0,0 +1,19 @@
...
+export let locales: Locale[] = ['en-US', 'en-FR', 'fr-FR']
+
+export function getLocale(request: NextRequest): Locale {
+  const languages = new Negotiator({
+    headers: Object.fromEntries(request.headers.entries()),
+  }).languages()
+
+  return (
+    locales.find(
+      (locale) => locale === match(languages, locales, defaultLocale),
```
joshuawootonn, line 16: Match just returns a string so I have to validate it

PR #681, opened 2024-02-15
feat: remove client-side-routing code

chanceaclark:
## What

Removes client-side routing from the state as it's being tightly coupled to Next.js routing. We want to implement client side routing, but need to think of a better way to integrate it into state.

### DIFF
--- a/packages/runtime/src/index.ts
+++ b/packages/runtime/src/index.ts
@@ -12,7 +12,6 @@ export {
   scrollDocumentElement,
   unregisterDocument,
   setBuilderEditMode,
-  changePathname,
   builderPointerMove,
   setBreakpoints,
   setLocale,
--- a/packages/runtime/src/state/actions.ts
+++ b/packages/runtime/src/state/actions.ts
@@ -72,10 +72,6 @@ export const ActionTypes = {
   MAKESWIFT_CONNECTION_INIT: 'MAKESWIFT_CONNECTION_INIT',
   MAKESWIFT_CONNECTION_CHECK: 'MAKESWIFT_CONNECTION_CHECK',
 
-  CHANGE_PATHNAME: 'CHANGE_PATHNAME',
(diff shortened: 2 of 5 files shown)


packages/runtime/src/index.ts
```diff
@@ -12,7 +12,6 @@ export {
   scrollDocumentElement,
   unregisterDocument,
   setBuilderEditMode,
-  changePathname,
```
migueloller, line 15: This function is being used in the builder. Is the intent to remove the call-site as well when we upgrade the builder to this version of the runtime?

packages/runtime/src/state/actions.ts
```diff
@@ -72,10 +72,6 @@ export const ActionTypes = {
   MAKESWIFT_CONNECTION_INIT: 'MAKESWIFT_CONNECTION_INIT',
   MAKESWIFT_CONNECTION_CHECK: 'MAKESWIFT_CONNECTION_CHECK',
 
-  CHANGE_PATHNAME: 'CHANGE_PATHNAME',
-  CHANGE_PATHNAME_START: 'CHANGE_PATHNAME_START',
-  CHANGE_PATHNAME_COMPLETE: 'CHANGE_PATHNAME_COMPLETE',
```
migueloller, line 77: Unfortunately, these are also used in the builder (the entire `ActionTypes` object is 😢), so is the plan to remove that code from the builder?

migueloller (approved):

packages/runtime/src/index.ts
chanceaclark: Yeah, I have a branch ready to throw up in the builder to remove this.

migueloller: Ok, cool. I think we should ship that first, then.

chanceaclark: Submitted an MR for that.

migueloller (approved):

PR #684, opened 2024-02-16
feat: add draft mode support for app router

arvinpoddar:
Introduces support to set draft mode and pass arbitrary data for app router apps. Includes draft mode compatibility with pages router apps. Also includes `env` variables for pages router app initialization for better local development experience.


apps/nextjs-pages-router/src/pages/[[...path]].tsx
```diff
@@ -36,8 +36,13 @@ export async function getStaticProps(
   ctx: GetStaticPropsContext<ParsedUrlQuery>,
 ): Promise<GetStaticPropsResult<Props>> {
   const path = '/' + (ctx.params?.path ?? []).join('/')
+
+  const siteVersion = Makeswift.getSiteVersion(
+    ctx.draftMode ? ctx.previewData : null,
+  )
```
Comment (migueloller) [inline apps/nextjs-pages-router/src/pages/[[...path]].tsx:42]: How come this API changed?

packages/runtime/src/next/draft-mode/app-router-draft-mode-script.tsx
migueloller, line 1: I think we should avoid using "App Router" in our component names

packages/runtime/src/next/draft-mode/app-router-draft-mode-script.tsx
```diff
@@ -0,0 +1,14 @@
+import { DraftModeScript } from './draft-mode-script'
+import { draftMode } from 'next/headers'
+
+type Props = {
+  appOrigin?: string
+}
+
+/**
+ * Creates a draft mode script - for use within server components only.
```
migueloller, line 9: nit: this doesn't "create" a draft mode script—the component _is_ the script

packages/runtime/src/next/api-handler/handlers/proxy-preview-mode.ts
migueloller, line 1: Looks like we renamed all the parameters but the function is still named `proxyPreviewMode` instead of `proxyDraftMode`.

packages/runtime/src/next/api-handler/handlers/proxy-preview-mode.ts
```diff
@@ -50,29 +54,31 @@ async function proxyPreviewModeRouteHandler(
...
-      siteVersion: MakeswiftSiteVersion.Working,
-    }
-
-    proxyRequest.cookies.set('x-makeswift-draft-mode-data', JSON.stringify(previewData))
+    proxyRequest.cookies.set(MAKESWIFT_DRAFT_MODE_DATA_COOKIE, JSON.stringify(draftModeData))
   }
-  draftMode().disable()
+
+  // TODO: Delete Draft Mode cookie.
+  cookies().delete('__prerender_bypass')
```
migueloller, line 81: Is this out best option?

Also, should we remove the TODO?

packages/runtime/src/next/client.ts
```diff
@@ -152,7 +152,8 @@ export class Makeswift {
   private graphqlClient: GraphQLClient
   private runtime: ReactRuntime
 
-  static getSiteVersion(previewData: PreviewData): MakeswiftSiteVersion {
+  static getSiteVersion(previewData: PreviewData | null): MakeswiftSiteVersion {
+    if (previewData == null) return MakeswiftSiteVersion.Live
```
migueloller, line 156: Should this be `DraftData`?

packages/runtime/src/next/get-site-version.ts
```diff
@@ -1,15 +1,5 @@
...
-
-  if (isEnabled) {
-    const cookie = cookies().get('x-makeswift-draft-mode-data')
-    const draftModeData = cookie == null ? null : JSON.parse(cookie.value)
-
-    return getMakeswiftSiteVersion(draftModeData) ?? MakeswiftSiteVersion.Live
-  }
-
-  return MakeswiftSiteVersion.Live
+export const getSiteVersion = () => {
```
migueloller, line 3: nit: use a function declaration for functions

packages/runtime/src/next/draft-mode/index.tsx
migueloller, line 1: This file seems to be a server-only import. I think we're going to need to start organizing our files differently so that this is clearer.

packages/runtime/src/next/get-site-version.ts
migueloller, line 1: This file is also a server-only file and that's not clear from looking at it. Should also probably go into a directory for server-only modules.

packages/runtime/src/next/preview-mode.tsx
```diff
@@ -25,6 +25,9 @@ type Props = {
   appOrigin?: string
 }
 
+/**
+ * @deprecated: Use `DraftModeScript` instead.
+ */
```
migueloller, line 30: So this is deprecated, but it simply doesn't work anymore, right? Because we're not looking at `x-makeswift-preview-mode` anymore?

I think that if we're going to keep this around we might as well continue using Preview Mode for Pages Router and introduce Draft Mode just for App Router.

packages/runtime/src/next/draft-mode/index.tsx
arvinpoddar: This file actually wasn't server-only, it could be used in either a client or server context. With the changes now, draft mode script will only be a server component.

apps/nextjs-pages-router/src/pages/[[...path]].tsx
```diff
@@ -36,6 +36,7 @@ export async function getStaticProps(
   ctx: GetStaticPropsContext<ParsedUrlQuery>,
 ): Promise<GetStaticPropsResult<Props>> {
   const path = '/' + (ctx.params?.path ?? []).join('/')
+
```
Comment (migueloller) [inline apps/nextjs-pages-router/src/pages/[[...path]].tsx:39]: Random extra line?

packages/makeswift-next-plugin/index.js
```diff
@@ -49,6 +49,30 @@ module.exports =
...
+            has: [
+              {
+                type: 'query',
+                key: 'x-makeswift-draft-mode',
+                value: '(?<secret>.+)',
+              },
+            ],
+            source: '/:path(.*)',
+            destination: '/api/makeswift/proxy-draft-mode',
+            locale: false,
```
migueloller, line 62: The `locale` property shouldn't be needed as this is intended for App Router

packages/runtime/src/next/api-handler/handlers/proxy-preview-mode.ts
```diff
@@ -174,6 +128,7 @@ async function proxyPreviewModeApiRouteHandler(
     makeswift: true,
     siteVersion: MakeswiftSiteVersion.Working,
   }
+
```
migueloller, line 131: Random extra line?

packages/runtime/src/next/draft-mode/draft-mode-script.tsx
```diff
@@ -0,0 +1,97 @@
...
+      <script
+        id="makeswift-draft-mode"
+        type="module"
+        dangerouslySetInnerHTML={{ __html: draftModeScript }}
+      />
+      <script
+        id="makeswift-connection-check"
+        type="module"
+        dangerouslySetInnerHTML={{ __html: makeswiftConnectionCheckScript }}
+      />
```
migueloller, line 94: It's not a big deal, but should we be sharing this connection script?

packages/runtime/src/next/api-handler/handlers/proxy-preview-mode.ts
```diff
@@ -40,60 +39,15 @@ export default async function proxyPreviewMode(
   ...args: ProxyPreviewModeHandlerArgs
 ): Promise<NextResponse<ProxyPreviewModeResponse> | void> {
   return match(args)
-    .with(routeHandlerPattern, args => proxyPreviewModeRouteHandler(...args))
+    .with(routeHandlerPattern, () => {
+      throw new Error(
+        'Cannot request proxy preview mode endpoint from an API handler registered in `app`. Move your Makeswift API handler to the `pages/api` directory',
+      )
```
migueloller, line 45: Is this error shown to the user in a useful way? Should we instead return an error HTTP response?

packages/runtime/src/next/api-handler/handlers/proxy-preview-mode.ts
```diff
@@ -2,10 +2,9 @@ import { CookieSerializeOptions, serialize } from 'cookie'
 import { createProxyServer } from 'http-proxy'
 import { NextApiRequest, NextApiResponse } from 'next'
 import { parse } from 'set-cookie-parser'
-import { MakeswiftPreviewData, MakeswiftSiteVersion } from '../../preview-mode'
 import { NextRequest, NextResponse } from 'next/server'
 import { P, match } from 'ts-pattern'
-import { cookies, draftMode } from 'next/headers'
+import { MakeswiftPreviewData, MakeswiftSiteVersion } from '../../preview-mode'
```
migueloller, line 7: Unnecessary import move?

packages/runtime/src/next/draft-mode/index.tsx
```diff
@@ -0,0 +1,32 @@
+import { z } from 'zod'
+import { cookies, draftMode } from 'next/headers'
+export const MAKESWIFT_DRAFT_MODE_DATA_COOKIE = 'x-makeswift-draft-data'
+
+const makeswiftSiteVersionSchema = z.enum(['Live', 'Working'])
+export const MakeswiftSiteVersion = makeswiftSiteVersionSchema.Enum
+export type MakeswiftSiteVersion = z.infer<typeof makeswiftSiteVersionSchema>
```
migueloller, line 7: Didn't we already have an enum for the site version for Preview Mode?

packages/runtime/src/next/draft-mode/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+export const makeswiftDraftDataSchema = z.object({
+  makeswift: z.literal(true),
+  siteVersion: makeswiftSiteVersionSchema,
+})
```
migueloller, line 12: The draft data could just be the site version, by the way, since there is no chance of conflict with other Next.js stuff and also we control this cookie.

migueloller (approved):

packages/runtime/src/next/api-handler/handlers/proxy-preview-mode.ts
migueloller: Is `.disable()` our best option? Specifically, are we worried about sending back a `Set-Cookie` header to the client that removes `__prerender_bypass`? Do we have other alternatives?

migueloller: Had in-person discussion: no better alternatives that we know of.

migueloller:
Missing a changeset!

packages/runtime/src/next/draft-mode/draft-mode-script.tsx
arvinpoddar: Going to keep these paired together for now. I think this can be one of those things we tear apart when we more closely examine the new role of the injected script.

packages/runtime/src/next/api-handler/handlers/proxy-preview-mode.ts
arvinpoddar: Shifted towards returning HTTP errors, as well as reflecting the error in the next console.

migueloller (approved):

PR #685, opened 2024-02-16
refactor: move locale switching logic out of state

chanceaclark:
https://github.com/makeswift/makeswift/assets/10539418/68577ce9-7dc2-47f2-adbc-7bff68457e50

### DIFF
--- a/packages/runtime/src/state/react-builder-preview.ts
+++ b/packages/runtime/src/state/react-builder-preview.ts
@@ -45,7 +45,6 @@ import {
   changePathnameComplete,
   elementFromPointChange,
   setBreakpoints,
-  setLocale,
 } from './actions'
 import { ActionTypes } from './actions'
 import { createPropController } from '../prop-controllers/instances'
@@ -496,11 +495,6 @@ export function messageChannelMiddleware(
       const breakpoints = ReactPage.getBreakpoints(state)
       messageChannel.port1.postMessage(setBreakpoints(breakpoints))
 
-      const routerLocale = Router.locale
-      if (routerLocale != null) {
-        messageChannel.port1.postMessage(setLocale(new Intl.Locale(routerLocale)))
-      }
-
       Router.events.on('routeChangeStart', () => {
         messageChannel.port1.postMessage(changePathnameStart())
       })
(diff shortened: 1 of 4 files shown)


migueloller (commented):
I think there's some changes missing here. I'm not seeing how we're moving existing behavior to a new place. Before, when the `SET_LOCALE` action was fired, we'd call `Router.replace`. Now, we're dispatching a message to the builder. Also, now, when Next.js locale changes, we're dispatching an action, which it seems we weren't doing before, or if we were now we're doing twice (perhaps this would explain the weird double refresh you were seeing before)?

Maybe you also intended to remove lines 500-502 from `react-builder-preview.ts`?

```ts
      if (routerLocale != null) {
        messageChannel.port1.postMessage(setLocale(new Intl.Locale(routerLocale)))
      }
```

packages/runtime/src/state/react-builder-preview.ts
```diff
@@ -550,14 +551,6 @@ export function messageChannelMiddleware(
             window.getSelection()?.removeAllRanges()
             break
 
-          case ActionTypes.SET_LOCALE: {
-            const { pathname: currentPathname, query } = Router
-            const pathname = (action.payload.pathname ?? currentPathname).replace(/^\//, '/')
-
-            Router.replace({ pathname, query }, undefined, { locale: action.payload.locale })
-            break
-          }
```
migueloller, line 559: Now that I'm looking at this closely, I don't think this code path was every being called 🤔

joshuawootonn (approved):

migueloller (approved):
LGTM 👍🏻 

Don't forget the changeset!

chanceaclark:
Need a ♻️ for the changeset

migueloller:
Why are our deployments failing? 🤔

chanceaclark:
> Why are our deployments failing? 🤔

I'm not apart of the Vercel team, so I can't see why the `pages` router version is failing, but the app router version is failing due to the `next/router` import, which I want to tackle separately from this PR.

migueloller:
> > Why are our deployments failing? 🤔
> 
> I'm not apart of the Vercel team, so I can't see why the `pages` router version is failing, but the app router version is failing due to the `next/router` import, which I want to tackle separately from this PR.

Ok. I can see why it's failing and it's a weird error. It sucks that our CI isn't green but we can move on...

migueloller (approved):

PR #686, opened 2024-02-21
fix: app router will now render 404 and fix `useRouterLocaleSync` for app router

chanceaclark:
- This moves the `RootLayout` up a level, right under the `app/` directory in order for 404 pages to work correctly.
- Also, wraps the `useRouter` call in a try/catch in order for it to work within the App Directory.


packages/runtime/src/components/hooks/useRouterLocaleSync.ts
```diff
@@ -1,16 +1,27 @@
 import { useRouter } from 'next/router'
-
+import { P, match } from 'ts-pattern'
 import { useEffect } from 'react'
+
 import { setLocale } from '../../state/actions'
 import { useDispatch } from '../../runtimes/react/hooks/use-dispatch'
 
+function getRouter() {
```
migueloller, line 8: Should we still name this `use*` since it's technically still a React hook and we want to make sure we follow the rules of hooks?

packages/runtime/src/components/hooks/useRouterLocaleSync.ts
migueloller, line 1: This is an interesting solution. @arvinpoddar and I were discussing potential ways of sharing APIs for Pages and App Router and this isn't something I hadn't considered. Is there prior art for this? Maybe something you've seen in some OSS repo?

apps/nextjs-app-router/src/app/layout.tsx
migueloller, line 1: Nice catch. I would've thought that if there was no root layout it would use the one within `[lang]` but I guess this only applies to route groups (e.g., `(dasboard)` dir) and not dynamic routes...

migueloller (approved):

packages/runtime/src/components/hooks/useRouterLocaleSync.ts
chanceaclark: Ah sure, one sec.

packages/runtime/src/components/hooks/useRouterLocaleSync.ts
migueloller: I like it 👍🏻 

I can't think of a litmus test for whether we're on App Router or Pages Router for a component like we were able to do for `MakeswiftApiHandler`.

packages/runtime/src/components/hooks/useRouterLocaleSync.ts
```diff
@@ -1,16 +1,27 @@
...
 import { setLocale } from '../../state/actions'
 import { useDispatch } from '../../runtimes/react/hooks/use-dispatch'
 
+function useRouter() {
+  try {
+    const router = usePagesRouter()
+
+    return router
+  } catch (e) {
+    return;
```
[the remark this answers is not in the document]
migueloller: And of course, this isn't explicit, we're implicitly returning `undefined`, but whatever calls this might have to handle `undefined`. Again, this is a major nit, so feel free to ignore.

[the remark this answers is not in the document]
chanceaclark: We could easily go into a debate here about this 😅 I'm going to leave it for now and if it needs to change, we can update it in another PR.

apps/nextjs-app-router/src/app/layout.tsx
joshuawootonn: CleanShot 2024-02-21 at 16 13 23@2x

Nice catch indeed. Here is why I had originally placed it within `[lang]`. Might be a good opportunity to PR the docs.

migueloller: Yeah, this seems either like a bug or bad documentation. I'm guessing the appropriate way right now would be to have a root layout under `[lang]` as well as a root layout under `app` for top level errors, like `/404`?

chanceaclark: I was looking at [this code in `next`](https://github.com/vercel/next.js/blob/canary/packages/next/src/build/webpack/loaders/next-app-loader.ts#L359-L372) and I'm not sure they account for the root layout being in a dynamic segment 🤔

migueloller: So perhaps misleading docs? Like @joshuawootonn said, sounds like a good opportunity for a PR or GitHub issue.

chanceaclark: I also saw this issue, but it's not entirely what we are running into: https://github.com/vercel/next.js/issues/54911

chanceaclark: I was playing around with a blank slate and I was able to get it to work like it should, but this seems totally like a bug in Next.js

There is a `Error` component that you can use, but you need to wrap it in a client component in order for it to work:
https://github.com/chanceaclark/isr-app-router/commit/29787386aa79af5368a1445761c9d0de7c28ac31#diff-d1e6f43ca212fc901dcd88a7838bfab00d0c11a387888c6ff067c97274c3547bR1-R5

PR #687, opened 2024-02-21
feat: metadata/fonts/title in head for app router

arvinpoddar:
Introduces `PageHead` component, which renders title/meta/link data for pages in both app and pages router. Currently only supports snippet rendering for pages router. For pages router, uses Next/Head to render content directly in document head. For app router, uses React canary extensions to DOM elements, which are automatically rendered in the `head` tag.

### DIFF
--- a/packages/runtime/src/components/hooks/usePageSnippets.ts
+++ b/packages/runtime/src/components/hooks/usePageSnippets.ts
@@ -1,0 +1,84 @@
+import { useEffect, useMemo, useState } from 'react'
+import { useIsInBuilder } from '../../react'
+import { MakeswiftPageDocument } from '../../next'
+import deepEqual from '../../utils/deepEqual'
+import { useMakeswiftHostApiClient } from '../../next/context/makeswift-host-api-client'
+import { useSyncExternalStore } from 'use-sync-external-store/shim'
+import { Page as PageType } from '../../api'
+
+const SnippetLocation = {
+  Body: 'BODY',
+  Head: 'HEAD',
+} as const
+
+export type SnippetLocation = typeof SnippetLocation[keyof typeof SnippetLocation]
+
+export type Snippet = {
+  builderEnabled: boolean
+  cleanup: string | null
+  code: string
+  id: string
+  liveEnabled: boolean
+  location: SnippetLocation
+}
+
+const filterUsedSnippetProperties = ({
+  code,
+  builderEnabled,
+  liveEnabled,
+  location,
+  cleanup,
+}: Snippet) => ({
+  code,
+  builderEnabled,
+  liveEnabled,
+  location,
+  cleanup,
+})
+
+export function usePageSnippets({ page }: { page: MakeswiftPageDocument }) {
+  const isInBuilder = useIsInBuilder()
+  const [snippets, setSnippets] = useState(page.snippets)
+
+  // We're using cached results here for page snippets so that anytime the user
+  // changes the snippets or fonts on the builder, the change would be reflected
+  // here. See this PR for discussions and things we can do to improve it in the
+  // future: https://github.com/makeswift/makeswift/pull/77
+  const cachedPage = useCachedPage(isInBuilder ? page.id : null)
+  useEffect(() => {
+    if (cachedPage == null) return
+
+    const oldSnippets = snippets.map(filterUsedSnippetProperties)
+    const newSnippets = cachedPage.snippets.map(filterUsedSnippetProperties)
+
+    if (deepEqual(newSnippets, oldSnippets)) return
+
+    setSnippets(cachedPage.snippets)
+  }, [cachedPage])
+
+  const filteredSnippets = useMemo(
+    () => snippets.filter(snippet => (isInBuilder ? snippet.builderEnabled : snippet.liveEnabled)),
+    [snippets, isInBuilder],
+  )
+
+  const headSnippets = useMemo(
+    () => filteredSnippets.filter(snippet => snippet.location === SnippetLocation.Head),
+    [filteredSnippets],
+  )
+
+  const bodySnippets = useMemo(
+    () => filteredSnippets.filter(snippet => snippet.location === SnippetLocation.Body),
+    [filteredSnippets],
+  )
+
+  return { headSnippets, bodySnippets }
+}
+
+function useCachedPage(pageId: string | null): PageType | null {
+  const client = useMakeswiftHostApiClient()
+  const getSnapshot = () => (pageId == null ? null : client.readPage(pageId))
+
+  const page = useSyncExternalStore(client.subscribe, getSnapshot, getSnapshot)
+
+  return page
+}
--- a/packages/runtime/src/components/page/Page.tsx
+++ b/packages/runtime/src/components/page/Page.tsx
@@ -1,268 +1,34 @@
 'use client'
 
-import { ReactElement, Children, createElement, useMemo, useEffect, useRef, useState } from 'react'
-import { useSyncExternalStore } from 'use-sync-external-store/shim'
-import parse from 'html-react-parser'
-import Head from 'next/head'
-
 import { BodySnippet } from './BodySnippet'
 import { DocumentReference } from '../../runtimes/react'
 import { createDocumentReference } from '../../state/react-page'
-import { useMakeswiftHostApiClient } from '../../next/context/makeswift-host-api-client'
-import { useIsInBuilder } from '../../react'
-import deepEqual from '../../utils/deepEqual'
 import { MakeswiftPageDocument } from '../../next'
-import { Page as PageType, Site } from '../../api'
 import { useRouterLocaleSync } from '../hooks/useRouterLocaleSync'
-
-const SnippetLocation = {
-  Body: 'BODY',
-  Head: 'HEAD',
-} as const
-
-type SnippetLocation = typeof SnippetLocation[keyof typeof SnippetLocation]
-
-type Snippet = {
-  builderEnabled: boolean
-  cleanup: string | null
-  code: string
-  id: string
-  liveEnabled: boolean
-  location: SnippetLocation
-}
-
-const defaultFavicon = {
-  mimetype: 'image/png',
-  publicUrl:
-    'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAACIAAAAiCAYAAAA6RwvCAAAACXBIWXMAABcRAAAXEQHKJvM/AAABjElEQVRYhc2XzU3EMBCFB8TddAAXn6EE6GCpgNABZ1/IXnymBOgAOmA7YM8+ABVsXEHQQFaKQryeN3Yk3ilKJtEnv/nLUd/3pFG0riGi88yrnQn+UfJ5FUi0riWiB2H4nQn+KRd0DFP8agXEfkqCYJBoHdtxIQxfm+DfFgEhoith3NYE30o/qgGR2BJB+xY7kdYEL8oNFUi0jiFMJuxVWrJqEMFxsyUNCsE6AeNztvBp7aJ143vXksoRnwhYtmNdSoIQa6RlO9YXEWW7KgoCleOgxgTf1QZBT+RZ2lXFING6UxCCq+ceeUE8fYdknY599v9sJvzGBP+yCEgC7GPmETc0OJ+0awAlkhe2pAbIXAeFZ8xe2g2Nk3c3ub0xwWt6zY9qbmiqGVMbZK21ZC/YmhlbeBMTzZNDQqcvDb1kM1x32iqZSt1HaqukfKvq34BAOTLsrH+ETNmUkKHHA+428RgeclPVWozeSyAI2EdWB34jtqXNTAySOY3i/KgFIlqOa4GkFmBegorzg4joG07he/M7zl6jAAAAAElFTkSuQmCC',
-}
-
-// Taken from https://github.com/facebook/react/blob/14bac6193a334eda42e727336e8967419f08f5df/packages/react-dom/src/server/ReactPartialRenderer.js#L208
-const VALID_TAG_REGEX = /^[a-zA-Z][a-zA-Z:_.\-\d]*$/
-
-const VALID_HEAD_ELEMENT_TYPES = [
-  'title',
-  'base',
-  'link',
-  'style',
-  'meta',
-  'script',
-  'noscript',
-  'template',
-]
-
-function snippetToElement(snippet: Pick<Snippet, 'id' | 'code'>): (string | ReactElement)[] {
-  return Children.map(parse(snippet.code), element => {
-    if (typeof element === 'string') return element
-
-    if (!VALID_TAG_REGEX.test(element.type)) return null
-
-    const key = element.key ? `${snippet.id}:${element.key}` : snippet.id
-
-    return createElement(element.type, { ...element.props, key })
-  })
-}
-
-const filterUsedSnippetProperties = ({
-  code,
-  builderEnabled,
-  liveEnabled,
-  location,
-  cleanup,
-}: Snippet) => ({
-  code,
-  builderEnabled,
-  liveEnabled,
-  location,
-  cleanup,
-})
+import { usePageSnippets } from '../hooks/usePageSnippets'
+import { PageHead } from './PageHead'
 
 type Props = {
   document: MakeswiftPageDocument
 }
 
 export function Page({ document: page }: Props): JSX.Element {
-  const isInBuilder = useIsInBuilder()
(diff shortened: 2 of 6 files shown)


packages/runtime/src/components/page/PageHead.tsx
```diff
@@ -0,0 +1,188 @@
...
+    <>
+      <style>
+        {`
+      html {
+        font-family: sans-serif;
+      }
+      div#__next {
+        overflow: hidden;
+      }
+    `}
```
migueloller, line 101: nit: bad indentation

packages/runtime/src/components/page/PageHead.tsx
```diff
@@ -0,0 +1,188 @@
...
+
+      {headSnippets.map(snippetToElement).map(children =>
+        Children.map(children, child => {
+          if (typeof child === 'string') return child
+
+          if (VALID_HEAD_ELEMENT_TYPES.includes(child.type as string)) return child
+
+          return null
+        }),
+      )}
```
migueloller, line 155: I'm worried that the [nesting](https://nextjs.org/docs/pages/api-reference/components/head#use-minimal-nesting) here will be a problem. Have we tested for this?

packages/runtime/src/components/page/Page.tsx
```diff
@@ -1,268 +1,35 @@
...
-
-    const oldSnippets = snippets.map(filterUsedSnippetProperties)
-    const newSnippets = cachedPage.snippets.map(filterUsedSnippetProperties)
-
-    if (deepEqual(newSnippets, oldSnippets)) return
-
-    setSnippets(cachedPage.snippets)
-  }, [cachedPage])
-  const site = useCachedSite(isInBuilder ? page.site.id : null)
+  const { bodySnippets } = usePageSnippets({ page })
```
migueloller, line 16: We seem to be using `usePageSnippets` twice. Once in this component and another time in `PageHead`. Are we worried that this will result in two subscriptions and unnecessary re-renders or potential performance problems?

packages/runtime/src/components/hooks/usePageSnippets.ts
```diff
@@ -0,0 +1,106 @@
...
+
+      try {
+        cleanUp()
+      } catch {
+        // Ignore errors from user input.
+      }
+    })
+
+    previousHeadSnippets.current = headSnippets
+  }, [headSnippets])
```
migueloller, line 89: How will this interact with App Router and `useServerInsertedHTML`? I'm guessing that now that HTML will come from the server and React won't do much to track it, but if we're now running this effect as `headSnippets` changes on client-side navigation will the cleanup work well? Have we tested what happens when there's different snippets in the head and we client-side navigate between those pages?

packages/runtime/src/components/page/Page.tsx
```diff
@@ -1,268 +1,35 @@
...
-          Children.map(children, child => {
-            if (typeof child === 'string') return child
-
-            if (VALID_HEAD_ELEMENT_TYPES.includes(child.type as string)) return child
-
-            return null
-          }),
-        )}
-      </Head>
+      {/* Can also pass children to be added to page <head> here */}
```
migueloller, line 25: I'm not sure we should add this to the API. `useServerInsertedHTML` probably doesn't do anything useful if `children` changes and so having an API that doesn't work as expected might just make things more confusing. Given how differently `next/head` and `useServerInsertedHTML` behave we should probably not try to build some sort of useful abstraction for them as much as just hiding them behind `PageHead` and making sure the behavior works well for the exact use case `PageHead` tries to accomplish. There's already some mismatch there with cleaning up head scripts on client-side navigations that we need to potentially address.

migueloller (changes requested):

packages/runtime/src/components/page/PageHead.tsx
arvinpoddar: Confirming that nesting is not an issue for snippets in rendered in the next/head

arvinpoddar:
The recent changes reflect the conversations from above and the finding that metadata from `useServerInsertedHTML` does not update on any client-side navigations (CSN). The changes to this PR address this by updating the meta content on CSN for app router pages by portaling the meta data into the document head (in addition to adding cloned script tags to the head to enable their execution). This way, the correct metadata is set on hard navigation and CSN.

packages/runtime/src/components/page/PageHead.tsx
```diff
@@ -0,0 +1,294 @@
...
+  const [shouldRenderPortal, setShouldRenderPortal] = useState(false)
+
+  // Purge the initial head elements that were server inserted. Then,
+  // we can portal in the same head elements so they will now be tracked on
+  // client side navigation by React. Will not occur in pages router.
+  useEffect(() => {
+    if (isPagesRouter) return
+    document.head.querySelectorAll('[data-makeswift-meta]').forEach(el => el.remove())
+    setShouldRenderPortal(true)
+  }, [isPagesRouter])
```
migueloller, line 234: So my understanding of what's going on here is that we mark the server-inserted HTML elements with the `data-*` attribute and then we remove those _once_ and moving forward the portal takes care of client-side navigation. Is that accurate?

packages/runtime/src/components/page/PageHead.tsx
```diff
@@ -0,0 +1,294 @@
...
+
+          script.onload = () => resolve()
+          script.onerror = () => resolve()
+
+          node.parentNode?.appendChild(script)
+          appendedScripts.push(script)
+
+          if (!script.hasAttribute('src')) resolve()
+        })
+      })
```
migueloller, line 262: These promises seem to be swallowed and unused. I don't think the top-level `.catch` will work. For it to work this would be to need a `Promise.all` on a `scripts.current.map` where we map all script re-executions into promises. Furthermore, we would want `script.onerror` to be `reject`, not `resolve`.

packages/runtime/src/components/page/PageHead.tsx
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+    executeScriptsInOrder().catch(error => {
+      // Ignore errors from user-provided code
+      console.error(error)
+    })
+
+    return () => {
+      appendedScripts.forEach(script => script.remove())
+    }
```
migueloller, line 272: Doesn't React via the portal take care of managing the DOM? If we manually remove them I'd expect React to run into errors where the elements it tries to manage/remove don't exist anymore.

packages/runtime/src/components/page/PageHead.tsx
```diff
@@ -0,0 +1,294 @@
...
+          data-makeswift-meta="true"
+        />
+      )}
+
+      {headSnippets.map(snippetToElement).map(children =>
+        Children.map(children, child => {
+          if (typeof child === 'string') return child
+
+          if ((child.type as string) === 'script') {
+            return cloneElement(child, { ref: saveScriptRef })
```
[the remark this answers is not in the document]
migueloller: After reading the rest of this file I would recommend a different approach here. Instead of using `cloneElement` and adding this ref, I would leverage `findDOMNode` to get all DOM nodes that are put in a portal for CSN. From there, you can traverse the DOM and execute any scripts as necessary.

migueloller (changes requested):
@arvinpoddar, thanks for the changes! I've left some comments on the new implementation. Some of them questions so that I can improve my understanding of how you're solving the problem, some are suggestions on how we can simplify, and others are some bugs/gotchas that I caught.

---
On a separate note, after looking a bit into some of the new APIs that React has in its canary releases—which Next.js App Router uses—I think I have an idea for how we can simplify this work and ease a lot of the pain you're probably going through with having to manually manage this stuff.

It would involve having separate behavior for Pages Router and App Router where for Pages Router we continue to use `next/head` without changing anything and for App Router we leverage the new React canary behavior instead of using `useServerInsertedHTML`.

If this isn't clear, please let me know and we can discuss in more detail!

packages/runtime/src/components/page/PageHead.tsx
```diff
@@ -0,0 +1,294 @@
...
+        const cleanUp = new Function(snippetToCleanUp.cleanup)
+        cleanUp()
+      } catch (err) {
+        console.error('Error attempting to parse snippet cleanup', err)
+        // Ignore errors from user input.
+      }
+    })
+
+    previousHeadSnippets.current = headSnippets
+  }, [headSnippets])
```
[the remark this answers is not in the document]
arvinpoddar: We previously talked about this, we moved this behavior outside of the hook since this hook is really only for retrieving head snippets and body snippets (we call this hook in two places). We decided that registering this effect in the hook would unnecessarily it where it doesn't apply. Instead, we add this hook directly to the page head, where it should be executed.

packages/runtime/src/components/page/PageHead.tsx
arvinpoddar: Bingo

packages/runtime/src/components/page/PageHead.tsx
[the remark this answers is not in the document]
arvinpoddar: Based on my understanding of the docs, `findDOMNode` is deprecated and only for use with class components: https://react.dev/reference/react-dom/findDOMNode#finddomnode. Wouldn't it be worse to rely on this?

packages/runtime/src/components/page/PageHead.tsx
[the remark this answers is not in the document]
migueloller: Yeah, that's right... it's very confusing to see these things separated.

packages/runtime/src/components/page/PageHead.tsx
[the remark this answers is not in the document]
migueloller: We rely on `findDOMNode` in the runtime to get DOM nodes from user-registered components since they don't always use `forwardRef`. It's been deprecated for a while but hasn't been removed because it's the only way to do what it does.

I guess `cloneElement` plus a ref isn't that bad when compared to `findDOMNode`.

packages/runtime/src/components/page/PageHead.tsx
```diff
@@ -0,0 +1,294 @@
...
+      {favicon && (
+        <link
+          rel="icon"
+          type={favicon.mimetype}
+          href={favicon.publicUrl}
+          data-makeswift-meta="true"
+        />
+      )}
+
+      {canonicalUrl && <link rel="canonical" href={canonicalUrl} data-makeswift-meta="true" />}
```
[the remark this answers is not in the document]
migueloller: And that's because they're inserted using `useServerInsertedHTML`, right? If we end up going with my idea then this won't be needed since we wouldn't be using `useServerInsertedHTML`.

arvinpoddar:
Cutting scope here, using React canary extensions to DOM elements to achieve base-level app router compatibility. **Currently does not support/render any snippets for app router**. Some caveats:

- The link/style tags need a `precedence` prop specified to be rendered in the head. With this comes the behavior that the tag may be left behind by React in the case of a prop update. In our case, this manifests with fonts - in the builder, if you update the fonts loaded by the site, the old fonts tag remains in the head. Of course, it disappears on refresh in the builder.
- Meta tags with an `itemProp` appear in the body. This is an expected behavior per the React docs. We may consider removing them.

packages/runtime/src/next/components/head-tags.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+export function PageLink({ precedence, ...props }: PageLinkProps) {
+  if (useIsPagesRouter()) {
+    return (
+      <Head>
+        <link {...props} />
+      </Head>
+    )
+  }
+  //@ts-expect-error
```
migueloller, line 41: nit: we should add a space between `//` and `@`.

Also, can we add a note of what the expected error is so that this is more "deletable"? For example:

```ts
// @ts-expect-error: the current version of `@types/react-dom` doesn't include types for `precedence` since it's a React Canary feature. See https://react.dev/reference/react-dom/components/link. The functionality is still supported because Next.js App Router uses a React Canary.
```

That's a pretty verbose comment and perhaps we can do something shorter but it does specify why the error happens and also why we're ok with it happening (i.e., Next.js does support it).

packages/runtime/src/next/components/head-tags.tsx
```diff
@@ -0,0 +1,66 @@
...
+      </Head>
+    )
+  }
+  return <meta {...props} />
+}
+
+// https://react.dev/reference/react-dom/components/link#props
+type ReactLinkProps = {
+  precedence?: 'reset' | 'low' | 'medium' | 'high'
+}
```
migueloller, line 30: If we're going to define the types, we might as well augment the React DOM module in this file.

packages/runtime/src/next/components/head-tags.tsx
```diff
@@ -0,0 +1,66 @@
+import { ComponentPropsWithoutRef } from 'react'
+import { useIsPagesRouter } from '../../components/hooks/useIsPagesRouter'
```
migueloller, line 2: We probably want this hook to be under `src/next` as well. Also, as far as naming conventions, we should use `kebab-case.ts`

packages/runtime/src/next/components/head-tags.tsx
```diff
@@ -0,0 +1,66 @@
...
+import Head from 'next/head'
+
+export function PageTitle({ children, ...props }: ComponentPropsWithoutRef<'title'>) {
```
migueloller, line 5: Is there a reason we're explicitly destructuring `children`?

packages/runtime/src/next/components/head-tags.tsx
```diff
@@ -0,0 +1,66 @@
...
+  if (useIsPagesRouter()) {
+    return (
+      <Head>
+        <title {...props}>{children}</title>
+      </Head>
+    )
+  }
+  return <title {...props}>{children}</title>
```
migueloller, line 13: Note that for `<title>` in React Canary, `children` must be a string. This means that `ComponentPropsWithoutRef<'title'>` might not be the appropriate type.

packages/runtime/src/next/components/head-tags.tsx
```diff
@@ -0,0 +1,66 @@
...
+  precedence?: 'reset' | 'low' | 'medium' | 'high'
+  href?: string
+}
+type PageStyleProps = ComponentPropsWithoutRef<'style'> & ReactStyleProps
+
+export function PageStyle({ children, precedence, href, ...props }: PageStyleProps) {
+  if (useIsPagesRouter()) {
+    return (
+      <Head>
+        <style {...props}>{children}</style>
```
migueloller, line 56: Is the `href` prop not a regular prop that we'd want to pass here too?

migueloller (approved):
Looking good! ❤️

packages/runtime/src/next/components/head-tags.tsx
arvinpoddar: href is not a standard attribute for `style`: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/style

packages/runtime/src/next/components/head-tags.tsx
arvinpoddar: Style preference, in our builder we tend to separate children and pass it explicitly. I understand the use cases there are a bit different since we usually wrap the children in other elements, but I prefer the explicitness.

packages/runtime/src/next/components/head-tags.tsx
migueloller: Ah, got it. Thanks for clarifying.

packages/runtime/src/slate/types.ts
```diff
@@ -264,3 +269,21 @@ declare module 'slate' {
...
+
+  export interface PageStyleProps extends ComponentPropsWithoutRef<'style'> {
+    precedence?: 'reset' | 'low' | 'medium' | 'high'
+    href?: string
+  }
+
+  export interface PageTitleProps extends Omit<ComponentPropsWithoutRef<'title'>, 'children'> {
+    children?: string
+  }
+}
```
migueloller, line 289: These shouldn't be in the types file for Slate, which is the library we use for rich text. Instead, we should place it in a file for React DOM or in the one file we need this types. Also, I don't think we should export prop types and instead should use augmentation to add support for the new props in the elements themselves.

migueloller (commented):
The point of module augmentation is to avoid having to use `@ts-expect-error`, not to put our own prop type definitions in there.

packages/runtime/src/next/components/head-tags.tsx
arvinpoddar: Decided not to do this to avoid exposing these props to _all_ `link` and `style` elements.

migueloller (approved):
Looks good!

PR #689, opened 2024-02-22
feat: update http-proxy to forward x- headers

joshuawootonn:


joshuawootonn:
`http-proxy` doesn't forward `x-` headers by default, and nextjs doesn't gracefully handle `undefined` (currently)

I am added the [`xfwd`](https://github.com/http-party/node-http-proxy/blob/HEAD/lib/http-proxy.js#L26-L42) to turn on `x-` headers forwarding for the proxy. 

details on the bug: https://github.com/vercel/next.js/issues/61133

You can see that our app no longer blows up for pages router in the builder here: https://app.makeswift.com/workspaces/V29ya3NwYWNlOjcxMTJlYTg3LTBhNmYtNGZhOS1hNDc2LTA1NjhjMWYyYzE5MA==/sites/U2l0ZToxY2I5Y2M5OC1lYzg4LTQ2NmUtYjdjNi03YTRhNDc2NTEwN2Q=/pages/UGFnZTo5ODEyOTZlMy0xY2MyLTQ3YWUtYWFlMi00NTcwOTIyNjU2MGI=

This site in Makeswift is using the preview deployment above.

chanceaclark (dismissed):

migueloller (dismissed):

arvinpoddar (dismissed):

migueloller:
@joshuawootonn, this looks good to me. We've also got some custom code that looks at `x-forwarded-proto` and `x-forwarded-ssl` to determine what protocol we should use. Do these changes relate in some way to this? Do we need to change or remove this code?

joshuawootonn:
> Do these changes relate in some way to this? Do we need to change or remove this code?

My understanding is: 

(1) user sends request -> (2) this request hits api route -> (3) the proxy sends the request without headers to route -> (4) route throws because the invalid value of the header.

The error is at 4 because of the invalid value. At 3 the proxy still has access to the headers and their valid values. **_So no change is needed to the proxy_** and how it accesses it's headers

joshuawootonn:
@migueloller do you think this warrants a changeset?

migueloller:
> @migueloller do you think this warrants a changeset?

Yes! Every change we make should have a changeset. Maybe in the future we can extract commit info for the changelog so we can simply write good commit messages and don't also have to write a changeset.

PR #691, opened 2024-02-26
fix: fix bug in translation of detached typography

joshuawootonn:
Prior to translation we normalize all our `typography` richtext data to the highest ancestor for the simplest possible markup. To do this we go through children determining what parts of `typography` they share. Any `typography` data that is shared is pushed up the tree, while unique `typography` data is left behind.

We do an `and` and an `xor` to determine what is shared and unique. In the case of detached `typography` the logic was incorrect. It was pushing data up the tree wtihout removing it from the original location. This left our normalization in a infinite loop.

My hunch is I added this special case as an optimization, so I am just removing it.

### DIFF
--- a/.changeset/red-mirrors-agree.md
+++ b/.changeset/red-mirrors-agree.md
@@ -1,0 +1,5 @@
+---
+'@makeswift/runtime': patch
+---
+
+Fix a bug in translating `Text` components containing detached typography.
--- a/packages/runtime/src/slate/TypographyPlugin/__tests__/normalization.test.tsx
+++ b/packages/runtime/src/slate/TypographyPlugin/__tests__/normalization.test.tsx
@@ -128,7 +128,6 @@ describe('GIVEN normalizing typography', () => {
 
       expect(editor.children).toEqual(result.children)
     })
-
     it('WHEN the only non matching node is an empty text THEN it is ignored', () => {
       const editor = EditorV2(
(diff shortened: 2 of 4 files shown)


joshuawootonn:
This is the branch I'll be shipping this with: `0.14-stable`

This was the branch naming pattern I had previously created patches with.

migueloller:
> This is the branch I'll be shipping this with: `0.14-stable`
> 
> This was the branch naming pattern I had previously created patches with.

Sorry, I'm a bit confused here. Does the branch have some sort of important meaning? When doing the release there should be a git tag from changesets that points to the commit so the original branch for the PR can be deleted and we'll still have a reference to the commit that the release as for.

joshuawootonn:
> > This is the branch I'll be shipping this with: `0.14-stable`
> > This was the branch naming pattern I had previously created patches with.
> 
> Sorry, I'm a bit confused here. Does the branch have some sort of important meaning? When doing the release there should be a git tag from changesets that points to the commit so the original branch for the PR can be deleted and we'll still have a reference to the commit that the release as for.

True. I think I had the idea from other conversations that we needed to maintain a branch for manually patched versions, but you are right.

joshuawootonn:
Going to cut a release for this change

joshuawootonn:
Not sure how this test is failing.

PR #695, opened 2024-02-29
feat: add app router head snippet support

arvinpoddar:


packages/runtime/src/components/page/HeadSnippet.tsx
```diff
@@ -0,0 +1,122 @@
...
+  return Children.map(parse(snippet.code), element => {
+    if (typeof element === 'string') return element
+
+    if (!VALID_HEAD_ELEMENT_TYPES.includes(element.type as string)) return null
+
+    const key = element.key ? `${snippet.id}:${element.key}` : snippet.id
+
+    return createElement(element.type, {
+      ...element.props,
+      key,
```
arvinpoddar, line 73: Without this key, `next/head` will only insert one snippet into the head. If there are multiple snippets, only the first will get rendered. It's a very strange behavior.

packages/runtime/src/components/page/HeadSnippet.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+export function HeadSnippet({ snippet }: Props) {
+  const isPagesRouter = useIsPagesRouter()
+
+  useEffect(() => {
+    if (isPagesRouter) {
+      return () => {
+        const snippetElements = getSnippetElementsFromDOM(snippet)
+        if (snippetElements.length > 0) return
+        cleanUpSnippet(snippet)
```
arvinpoddar, line 20: Problem: `next/head` will dedupe scripts rendered in the head. Even though the page mounts/unmounts on client-side navigations, the script will not actually execute twice. It just stays in the head. However, the page mounting/unmounting means that we perform the cleanup, but the snippet never reruns. Checking the presence of the snippet prevents this.

packages/runtime/src/components/page/HeadSnippet.tsx
```diff
@@ -0,0 +1,122 @@
...
+      }
+    }
+
+    const snippetElements = getSnippetElementsFromDOM(snippet)
+    const isAlreadyInTheDOM = snippetElements.length > 0
+
+    if (isAlreadyInTheDOM) return
```
arvinpoddar, line 27: Offline discussion: choosing the no-cleanup policy for server inserted scripts because this is basically what is happening with `next/head` already. Cleanups are still performed for snippets that are DOM managed - this brings the snippet execution between Pages Router / App Router 99% in sync.

packages/runtime/src/components/page/HeadSnippet.tsx
migueloller: Makes sense. But then we still have the issue of duplicate keys. When is `element.key` not defined? Does it always fall back to `snippet.id`?

migueloller (approved):
👍🏻

packages/runtime/src/components/page/HeadSnippet.tsx
arvinpoddar: Yes, it will always fall back on `snippet.id`. In fact, this key is what `next/head` uses to dedupe script executions, so it is something we want to be one-to-one with a snippet.

migueloller: The potential issue, though, is a snippet with multiple elements using the same key for all elements

PR #696, opened 2024-03-01
Add X and Slack icons to SocialLinks

migueloller:
- Adds a new version to `SocialLinks` prop controller
- Adds "X" and "Slack" options to `SocialLinks` prop controller
- Updates Facebook logo in `SocialLinks` builtin component
- Makes `SocialLinks` component a bit less flimsy by not throwing if the icon is not found


arvinpoddar (approved):
LGTM

PR #698, opened 2024-03-04
chore: remove xfwd patch from bigcommerce-catalyst

chanceaclark:
Fixes the 400 Bad Request issue we were having within the builder. For some reason the `x-forwarded-proto` was causing the proxy to crash. I am still unsure why that is the case, but I tracked it down to that.

Screenshot 2024-03-05 at 11 01 56


migueloller (approved):
LGTM 👍🏻 

Since we've got this set up in later versions of the runtime we might have to figure out what's going on there later.

PR #700, opened 2024-03-06
Andrew/components

andrewreifman:
I created some new components to use for the example template, but I did modify the UI carousel component a little so I could use it for a static carousel component. I think there is a little degradation in the loading performance, but I'm hoping it would be ok to get this merged ASAP so I can build out the example template. Then we can go back in and either revert what I did in the carousel, or fix the issue.


migueloller (approved):
Looks good!

PR #703, opened 2024-03-07
fix: hydrate App Router link with locales

arvinpoddar:
Hydrates Link control and Button component with localized page path, if applicable. Supports both App Router and Pages Router. Also resolves an existing bug where App Router API handler requests are unauthorized due to not reading the secret from the rewritten request header.

Demo: 

https://github.com/makeswift/makeswift/assets/8345645/b505cdb7-6f20-43a5-b93a-993664a89ec8

https://github.com/makeswift/makeswift/assets/8345645/15f0ee3c-dc44-4ae2-a51b-422e3257c104

### DIFF
--- a/.changeset/sweet-monkeys-reply.md
+++ b/.changeset/sweet-monkeys-reply.md
@@ -1,0 +1,5 @@
+---
+'@makeswift/runtime': patch
+---
+
+Button component and Link control now hydrate page links with the locale, if present. Brings automatic link localization to App Router, while still supporting Pages Router.
--- a/.changeset/good-bulldogs-beg.md
+++ b/.changeset/good-bulldogs-beg.md
@@ -1,0 +1,5 @@
+---
+'@makeswift/runtime': patch
+---
+
+Resolves issue where rewritten host API requests are unauthorized due to not checking the request header for the secret.
--- a/packages/runtime/src/api/resource-types.ts
+++ b/packages/runtime/src/api/resource-types.ts
@@ -1,0 +1,10 @@
+import { z } from 'zod'
+
+const pagePathnameSliceSchema = z.object({
+  id: z.string(),
+  pathname: z.string(),
+  localizedPathname: z.string().optional().nullable(),
+  __typename: z.literal('PagePathnameSlice'),
+})
+
+export type PagePathnameSlice = z.infer<typeof pagePathnameSliceSchema>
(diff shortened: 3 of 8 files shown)


migueloller:
An interesting observation is that sometimes the host might prepend the default locale (e.g., `en-US`) to the URL instead of stripping it off by default. In this case we probably would want to prefix the links with the default locale. For us to be able to do this we'd need an additional option in our site locale settings where the user could specify whether to strip the default locale or not. This is not a feature we have right now, but perhaps we should support it! I made a Linear issue for it: https://linear.app/makeswift/issue/ENG-4736/support-configuring-stripping-default-locale

.changeset/sweet-monkeys-reply.md
```diff
@@ -0,0 +1,7 @@
+---
+'@makeswift/runtime': patch
+---
+
+Button component and Link control now hydrate page links with the locale, if present. Brings automatic link localization to App Router, while still supporting Pages Router.
+
+Also resolves issue where rewritten host API requests are unauthorized due to not checking the request header for the secret.
```
migueloller, line 7: Not a big deal but we should try and include changesets in the commit that introduce the change so that when they show up in the change log the commit shows the appropriate diff.

migueloller (dismissed):
This looks great!

I've got one question/concern, though. When the builder sends a page pathname slice that it has in its cache, since it won't include the `localizedPathname` field, will it cause any issues? It seems that we might need to update the builder so that the objects sent include this field as well.

.changeset/sweet-monkeys-reply.md
arvinpoddar: Ah thanks, I thought changesets were set on a PR granularity, not per commit. I can adjust this.

arvinpoddar:
> This looks great!
> 
> I've got one question/concern, though. When the builder sends a page pathname slice that it has in its cache, since it won't include the `localizedPathname` field, will it cause any issues? It seems that we might need to update the builder so that the objects sent include this field as well.

Per my understanding, the runtime cache uses the resource ID as the cache key, not the fields. Since the slices have a different ID due to using the localized page ID, there shouldn't be any stale cache issues.

migueloller:
> > This looks great!
> > I've got one question/concern, though. When the builder sends a page pathname slice that it has in its cache, since it won't include the `localizedPathname` field, will it cause any issues? It seems that we might need to update the builder so that the objects sent include this field as well.
> 
> Per my understanding, the runtime cache uses the resource ID as the cache key, not the fields. Since the slices have a different ID due to using the localized page ID, there shouldn't be any stale cache issues.

My comment was more about the cache fills/revalidations the builder does when changing a resource but we discussed this offline and reached an agreement on how we'd move forward.

migueloller (approved):

PR #705, opened 2024-03-13
fix(runtime): wraps the RuntimeProvider component in suspense boundary

chanceaclark:
Wraps the `RuntimeProvider` component in a `Suspense` boundary as it uses `React.lazy`. Not wrapping the component would cause a hydration mismatch between the server and client.

### DIFF
--- a/.changeset/yellow-turtles-check.md
+++ b/.changeset/yellow-turtles-check.md
@@ -1,0 +1,5 @@
+---
+'@makeswift/runtime': patch
+---
+
+Wraps the `RuntimeProvider` component in a `Suspense` boundary as it uses `React.lazy`. Not wrapping the component would cause a hydration mismatch between the server and client.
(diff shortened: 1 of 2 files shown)


.changeset/yellow-turtles-check.md
```diff
@@ -0,0 +1,5 @@
+---
+'@makeswift/runtime': patch
+---
+
+Wraps the Page component in a `Suspense` boundary for component registration with `React.lazy`. There were occurrences in which using a registered component with `React.lazy` would cause a hydration mismatch between the server and client.
```
migueloller, line 5: The problem was less with registered components and more with the fact that `RuntimeProvider` itself was using `React.lazy` or am I misunderstanding the issue here?

We already wrap all registered components with `<Suspense>` which is why I'm asking.

migueloller (approved):
LGTM 👍🏻

.changeset/yellow-turtles-check.md
chanceaclark: Ah, yeah I was misunderstanding it here.

migueloller (approved):

PR #706, opened 2024-03-20
feat: throw runtime error when passing `runtime` to `Page`

joshuawootonn:
CleanShot 2024-03-20 at 11 48 24@2x

Once documentation is public, I can update this message with a link to it.

### DIFF
--- a/packages/runtime/src/next/components/page.tsx
+++ b/packages/runtime/src/next/components/page.tsx
@@ -11,7 +11,13 @@ export type PageProps = {
   snapshot: MakeswiftPageSnapshot
 }
 
-export const Page = memo(({ snapshot }: PageProps) => {
+export const Page = memo(({ snapshot, ...props }: PageProps) => {
+  if ('runtime' in props) {
+    throw new Error(
+      `The \`runtime\` prop is no longer supported in the \`@makeswift/runtime\` \`Page\` component as of \`0.15.0\`.
+See our docs for more information on what's changed and instructions to migrate: https://docs.makeswift.com/migrations/0.15.0`,
+    )
+  }
   const client = useMemo(
     () =>
       new MakeswiftHostApiClient({
--- a/.changeset/gentle-planets-doubt.md
+++ b/.changeset/gentle-planets-doubt.md
@@ -1,0 +1,6 @@
+---
+'@makeswift/runtime': patch
+---
+
+Add console warning when `runtime` prop is passed to the `Page` component. 
+`runtime` should now be passed to the `ReactRuntimeProvider` instead of to `Page`. 


packages/runtime/src/next/components/page.tsx
```diff
@@ -11,7 +11,13 @@ export type PageProps = {
   snapshot: MakeswiftPageSnapshot
 }
 
-export const Page = memo(({ snapshot }: PageProps) => {
+export const Page = memo(({ snapshot, ...props }: PageProps) => {
+  if ('runtime' in props) {
+    const message =
+      'The `runtime` prop is no longer supported in the @makeswift/runtime `Page` component as of `0.15.0`.'
+    console.error(message)
+    throw new Error(message)
```
chanceaclark, line 19: Do we need both? Can we just throw the error and have it logged?

joshuawootonn: No we don't need both. Updated!

I was looking at the Nextjs repo for how they did similar exceptions and they did both, I just c/p'd from there. But throwing an error logs the message anyway, so there is no reason to `console.error`

joshuawootonn: Trueeeee. I think warning them makes more sense here. Now that you mention it!

migueloller: I originally suggested we throw an error because passing a prop like that is most likely due to a user that didn't migrate or hasn't migrated properly and a warning might be missed, specially in the other sea of warnings that we have. An error would make it a lot more obvious something is wrong. That being said, if the warning is prominent and hard to miss, it does feel better to do a warning than an error.

migueloller (changes requested):
I think we need a more descriptive message possibly indicating what they can do to address the problem. Think about how helpful Rust's errors are, we need to do that or better.

We're missing a changeset!

packages/runtime/src/next/components/page.tsx
```diff
@@ -11,7 +11,12 @@ export type PageProps = {
   snapshot: MakeswiftPageSnapshot
 }
 
-export const Page = memo(({ snapshot }: PageProps) => {
+export const Page = memo(({ snapshot, ...props }: PageProps) => {
+  if ('runtime' in props) {
+    throw new Error(
+      'The `runtime` prop is no longer supported in the @makeswift/runtime `Page` component as of `0.15.0`.',
```
migueloller, line 17: nit: `@makeswift/runtime` in backticks.

packages/runtime/src/next/components/page.tsx
chanceaclark: Just make the warning prominent and I think we should be :gucci:

joshuawootonn:
CleanShot 2024-03-20 at 13 19 53@2x

CleanShot 2024-03-20 at 13 20 06@2x

Here is what the new warning message looks like ^

joshuawootonn:
Here is what the warning looks like now: 
CleanShot 2024-03-20 at 13 29 59@2x

I thought about talking about `ReactRuntimeProvider` but I don't really think that content belongs in an error message, and since the solution is different for each router I think it makes most sense to add a link to the docs when they are published.

chanceaclark (approved):

migueloller:
> I thought about talking about ReactRuntimeProvider but I don't really think that content belongs in an error message, and since the solution is different for each router I think it makes most sense to add a link to the docs when they are published.

I agree with not getting into a ton of details here. Perhaps we can link to our docs and the upgrade guide for 0.15.0 and that can have more details on the change and why it happened and how to address it? Just removing the prop without understanding the implications might cause more problems, and those would be subtle and wouldn't put the user in a position to succeed.

migueloller:
@joshuawootonn, is this logged only in the browser or is it also logged in the terminal when running `next dev`?

joshuawootonn:
> @joshuawootonn, is this logged only in the browser or is it also logged in the terminal when running `next dev`?

It's logged in both

joshuawootonn:
@migueloller how do you feel about me shipping this and creating a followup ticket to update it when the docs are completed?

migueloller:
> @migueloller how do you feel about me shipping this and creating a followup ticket to update it when the docs are completed?

What are we trying to ship? If it's not the error message we want then it doesn't seem worth it? Also, I think it's ok for us to use a link that won't be live until later since there's kind of a circular dependency here with us wanting this to go out with `0.15.0` but us wanting to have the docs done before we ship `0.15.0`.

migueloller:
@joshuawootonn, here's what I would say in the warning:

```
Warning: The \`runtime\` prop is no longer supported in the \`@makeswift/runtime\` \`Page\` component as of \`0.15.0\`. See our docs for more information on what's changed: https://docs.makeswift.com/migrations/0.15.0
```

We could also link to the GitHub release instead of the docs.

Also, I'm still worried that if we don't do an error this won't help users as much. An error would be so much more helpful since it would prevent the user from getting into a bad state.

joshuawootonn:
Good point the migration URL isn't random. Changed the message and threw an error.
CleanShot 2024-03-22 at 09 29 59@2x

PR #707, opened 2024-03-20
refactor(runtime): remove http-proxy from proxy-preview-mode

chanceaclark:
Using the `DraftModeScript` in an `edge` runtime environment, webpack was trying to eagerly load in `http-proxy` which caused the `http` module to not be found. The `http` module is a `nodejs` only runtime, so it will always cause issues within the `edge` runtime.

This PR refactors the `/proxy-preview-mode` endpoint to use native API's to proxy the request to the builder.

https://github.com/makeswift/makeswift/assets/10539418/13179f9d-8331-4f79-b538-e0f6e078093e


packages/runtime/src/next/api-handler/handlers/proxy-preview-mode.ts
```diff
@@ -85,71 +91,63 @@ async function proxyPreviewModeApiRouteHandler(
...
-  return await new Promise<void>((resolve, reject) =>
-    previewModeProxy.web(req, res, { target, headers: { cookie }, secure }, err => {
-      if (err) reject(err)
-      else resolve()
-    }),
-  )
+  const response = await fetch(url, {
+    referrer: req.headers.referer,
+    headers: {
+      ...headers as Record<string, string>,
```
migueloller, line 130: How come we need to cast it here? Shouldn't we be transforming the headers from Node.js headers to fetch headers?

packages/runtime/src/next/api-handler/handlers/proxy-preview-mode.ts
```diff
@@ -85,71 +91,63 @@ async function proxyPreviewModeApiRouteHandler(
...
+      'Set-Cookie': cookie,
```
migueloller, line 131: `Set-Cookie` only makes sense in response headers, not request headers. Did you mean to use the `Cookie` header?

packages/runtime/src/next/api-handler/handlers/proxy-preview-mode.ts
```diff
@@ -85,71 +91,63 @@ async function proxyPreviewModeApiRouteHandler(
...
+    },
+  });
+
+  res.removeHeader('Set-Cookie')
```
migueloller, line 135: How come we're removing the `Set-Cookie` header from the response?

migueloller: Oh, I see now that this was to remove the Preview Mode cookie. We should keep this colocated with the `res.setPreviewData` call.

packages/runtime/src/next/api-handler/handlers/proxy-preview-mode.ts
```diff
@@ -85,71 +91,63 @@ async function proxyPreviewModeApiRouteHandler(
...
 
-  const proto = isForwardedProtoHttps || isForwardedSSL ? 'https' : 'http'
-  let target = `${proto}://${host}`
-
-  // If the user generates a locally-trusted CA for their SSL cert, it's likely that Node.js won't
-  // trust this CA unless they used the `NODE_EXTRA_CA_CERTS` option
-  // (see https://stackoverflow.com/a/68135600). To provide a better developer experience, instead
-  // of requiring the user to provide the CA to Node.js, we don't enforce SSL during local
-  // development.
-  const secure = process.env['NODE_ENV'] === 'production'
```
migueloller, line 132: Did we check that this is still being handled properly?

packages/runtime/src/next/api-handler/handlers/proxy-preview-mode.ts
```diff
@@ -85,71 +91,63 @@ async function proxyPreviewModeApiRouteHandler(
     } catch {}
   }
 
-  previewModeProxy.once('proxyReq', proxyReq => {
-    proxyReq.removeHeader('X-Makeswift-Preview-Mode')
-
-    // The following headers are Next.js-specific and are removed to prevent Next.js from
-    // short-circuiting requests, breaking routing.
-    proxyReq.removeHeader('X-Invoke-Path')
-    proxyReq.removeHeader('X-Invoke-Query')
```
migueloller, line 94: We seem to be losing some context here as to why these headers were being removed. As we move things around we should try and maintain this information.

packages/runtime/src/next/api-handler/handlers/proxy-preview-mode.ts
```diff
@@ -60,7 +64,8 @@ async function proxyPreviewModeApiRouteHandler(
   res: NextApiResponse<ProxyPreviewModeResponse>,
   { apiKey }: { apiKey: string },
 ): Promise<void> {
-  const previewModeProxy = createProxyServer({ xfwd: true })
```
migueloller, line 63: Are we adding `X-Forwarded-*` headers?

packages/runtime/src/next/api-handler/handlers/proxy-preview-mode.ts
```diff
@@ -85,71 +91,63 @@ async function proxyPreviewModeApiRouteHandler(
...
+    'X-Invoke-Path': xInvokePath,
+    'X-Invoke-Query': xInvokeQuery,
+     ...headers
+  } = req.headers
 
-    proxyReq.path = url.pathname + url.search
-  })
-
-  if (req.query.secret !== apiKey) return res.status(401).send('Unauthorized')
+  const isForwardedProtoHttps = match(req.headers['X-Forwarded-Proto'])
```
migueloller, line 101: Doesn't `red.headers` keep headers in lowercase? If we want to avoid case-sensitivity we might want to use `getHeader`? Otherwise maybe we want to keep the behavior we had before?

packages/runtime/src/next/api-handler/handlers/proxy-preview-mode.ts
```diff
@@ -85,71 +91,63 @@ async function proxyPreviewModeApiRouteHandler(
...
+    .with(P.string, x => x.split(','))
+    .with(P.array(P.string), x => x)
+    .otherwise(() => [])
+    .includes('https')
 
-  const host = req.headers.host
-  const originalCookies = req.headers.cookie
+  const isForwardedSSL = match(req.headers['X-Forwarded-SSL'])
```
migueloller, line 107: See my comment above about `req.headers` case sensitivity

packages/runtime/src/next/api-handler/handlers/proxy-preview-mode.ts
```diff
@@ -85,71 +91,63 @@ async function proxyPreviewModeApiRouteHandler(
...
-
-    const url = new URL(proxyReq.path, 'http://n')
-
-    url.searchParams.delete('x-makeswift-preview-mode')
+  const {
+    'X-Makeswift-Preview-Mode': xMakeswiftPreviewMode,
+    'X-Invoke-Path': xInvokePath,
+    'X-Invoke-Query': xInvokeQuery,
+     ...headers
+  } = req.headers
```
migueloller, line 99: Maybe to make things clearer we can make a `proxyHeaders` object as an instance of `Headers` and a `proxyUrl` as an instance of `URL` and then use that to apply modifications like removing the `X-Makeswift-Preview-Mode` search param and header so it's clear where we're doing intended business logic vs avoiding Next.js-specific gotchas?

migueloller (changes requested):
We need a changeset!

packages/runtime/src/next/api-handler/handlers/proxy-preview-mode.ts
chanceaclark: The type has an index signature of `string | string[] | undefined` and I was just trying to bypass TS here. I'll update this.

packages/runtime/src/next/api-handler/handlers/proxy-preview-mode.ts
chanceaclark: Ah good catch.

packages/runtime/src/next/api-handler/handlers/proxy-preview-mode.ts
chanceaclark: Pretty sure we don't need to now that we are just fetching the page directly. I could be wrong though 🤷‍♂️

packages/runtime/src/next/api-handler/handlers/proxy-preview-mode.ts
chanceaclark: > Key-value pairs of header names and values. Header names are lower-cased.

Ahh yeah, it'll fix that to use lowercase, which I think we would prefer?

migueloller: Yeah, let's just keep the code as it was before.

packages/runtime/src/next/api-handler/handlers/proxy-preview-mode.ts
migueloller: I think you're right since this is more of a configuration for `http-proxy` 👍🏻

packages/runtime/src/next/api-handler/handlers/proxy-preview-mode.ts
chanceaclark: Confirmed we are
Screenshot 2024-03-21 at 17 42 31

chanceaclark:
I have a branch for most of the fixes, just waiting on some help to figure out why the query param is not being removed after the proxy request: https://github.com/makeswift/makeswift/compare/test-proxy-preview-mode

packages/runtime/src/next/api-handler/handlers/proxy-preview-mode.ts
```diff
@@ -25,6 +24,39 @@ type ProxyPreviewModeHandlerArgs =
 const routeHandlerPattern = [P.instanceOf(Request), P.any, P.any] as const
 const apiRoutePattern = [P.any, P.any, P.any] as const
 
+const previewData: MakeswiftPreviewData = {
+  makeswift: true,
+  siteVersion: MakeswiftSiteVersion.Working,
+}
```
migueloller, line 30: By the way, this is eventually going to be dynamic based on request data which is why it was next to the call to `res.setPreviewData`.

packages/runtime/src/next/api-handler/handlers/proxy-preview-mode.ts
```diff
@@ -25,6 +24,39 @@ type ProxyPreviewModeHandlerArgs =
...
+
+const mapRequestHeadersToHeaders = (requestHeaders: NextApiRequest['headers']) => {
```
migueloller, line 32: nit: Can we please add return types? Also, can we use function declarations?

packages/runtime/src/next/api-handler/handlers/proxy-preview-mode.ts
```diff
@@ -144,12 +139,35 @@ async function proxyPreviewModeApiRouteHandler(
...
-    previewModeProxy.web(req, res, { target, headers: { cookie }, secure }, err => {
-      if (err) reject(err)
-      else resolve()
-    }),
-  )
+  const response = await fetch(proxyUrl, {
+    referrer: req.headers.referer,
+    headers: {
+      ...proxyHeaders,
+      cookie,
```
migueloller, line 149: How come we're not using the `Headers` API to add the Cookie header?

packages/runtime/src/next/api-handler/handlers/proxy-preview-mode.ts
```diff
@@ -25,6 +24,39 @@ type ProxyPreviewModeHandlerArgs =
...
+  makeswift: true,
+  siteVersion: MakeswiftSiteVersion.Working,
+}
+
+const mapRequestHeadersToHeaders = (requestHeaders: NextApiRequest['headers']) => {
+  const headers = new Headers()
+
+  Object.entries(requestHeaders).forEach(([key, value]) => {
+    match(value)
+      .with(P.array(P.string), value => headers.set(key, value.join(',')))
```
migueloller, line 37: In this case, we'd want to do `headers.append` which is the API for headers with multiple values.

packages/runtime/src/next/api-handler/handlers/proxy-preview-mode.ts
```diff
@@ -144,12 +139,35 @@ async function proxyPreviewModeApiRouteHandler(
     .map(cookie => serialize(cookie.name, cookie.value, cookie as CookieSerializeOptions))
     .join(';')
 
+  const originalCookies = req.headers.cookie
```
migueloller, line 142: Since we've created the proxy headers here it might be easier to use the `Headers` APIs to get the cookie, set the new one, etc.?

migueloller (changes requested):
Can we add a changeset?

packages/runtime/src/next/api-handler/handlers/proxy-preview-mode.ts
chanceaclark: Ahh that's why, I'll push that back down and add a note.

packages/runtime/package.json
```diff
@@ -118,15 +118,11 @@
...
     "@popmotion/popcorn": "^0.4.4",
-    "@types/cookie": "^0.5.1",
-    "@types/http-proxy": "^1.17.9",
     "@types/is-hotkey": "^0.1.7",
-    "@types/set-cookie-parser": "^2.4.2",
     "@types/use-sync-external-store": "^0.0.3",
     "@types/uuid": "^9.0.1",
     "@use-gesture/react": "^10.2.24",
     "color": "^3.1.3",
-    "cookie": "^0.5.0",
```
chanceaclark, line 129: I was also able to remove this and `set-cookie-parser` :feelsgood:

migueloller: Nice!

packages/runtime/src/next/api-handler/handlers/proxy-preview-mode.ts
```diff
@@ -85,71 +112,53 @@ async function proxyPreviewModeApiRouteHandler(
...
-  // (see https://stackoverflow.com/a/68135600). To provide a better developer experience, instead
-  // of requiring the user to provide the CA to Node.js, we don't enforce SSL during local
-  // development.
-  const secure = process.env['NODE_ENV'] === 'production'
+  proxyUrl.searchParams.delete('x-makeswift-preview-mode')
+  proxyHeaders.delete('x-makeswift-preview-mode')
+  // The following headers are Next.js-specific and are removed to prevent Next.js from
+  // short-circuiting requests, breaking routing.
+  proxyHeaders.delete('x-invoke-path')
+  proxyHeaders.delete('x-invoke-query')
```
migueloller, line 123: Are we worried about issues with case sensitivity? We were doing a different casing before.

packages/runtime/src/next/api-handler/handlers/proxy-preview-mode.ts
```diff
@@ -85,71 +112,53 @@ async function proxyPreviewModeApiRouteHandler(
...
 
   if (!Array.isArray(setCookie)) return res.status(500).send('Internal Server Error')
 
-  const additionalCookies = parse(setCookie)
-    .map(cookie => serialize(cookie.name, cookie.value, cookie as CookieSerializeOptions))
-    .join(';')
+  setCookie.forEach((cookie) => proxyHeaders.append('cookie', cookie))
+
+  const response = await fetch(proxyUrl, {
+    referrer: req.headers.referer,
```
migueloller, line 139: How come it's necessary to specify the referrer here?

packages/runtime/src/next/api-handler/handlers/proxy-preview-mode.ts
```diff
@@ -85,71 +112,53 @@ async function proxyPreviewModeApiRouteHandler(
...
+  const arrayBuffer = await response.arrayBuffer()
 
-  return await new Promise<void>((resolve, reject) =>
-    previewModeProxy.web(req, res, { target, headers: { cookie }, secure }, err => {
-      if (err) reject(err)
-      else resolve()
-    }),
-  )
+  res.write(arrayBuffer)
+  res.end()
```
migueloller, line 163: I think we can consolidate these two lines to `res.send(arrayBuffer)`

migueloller (approved):
LGTM 👍🏻

packages/runtime/src/next/api-handler/handlers/proxy-preview-mode.ts
chanceaclark: According to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Headers#:~:text=case%2Dinsensitive%20byte%20sequence) it's case-insensitive if using the `Headers` constructor. That, along with the lower-cased `req.headers` I think we should be fine.

packages/runtime/src/next/api-handler/handlers/proxy-preview-mode.ts
chanceaclark: I guess it doesn't really matter in preview mode. I was thinking it will send `about:client` as the referrer, but we also send it in `headers` so it's actually duplicative.

packages/runtime/src/next/api-handler/handlers/proxy-preview-mode.ts
chanceaclark: It doesn't actually work. It treats the `arrayBuffer` as an `application/json` mime type.

migueloller: Oh, interesting. Well, what you have is perfectly good.

packages/runtime/src/next/api-handler/handlers/proxy-preview-mode.ts
migueloller: Yeah, I thought the headers would take care of that.

joshuawootonn (approved):

PR #708, opened 2024-03-21
feat(bigcommerce-catalyst): add ProductCarousel component

chanceaclark:
https://github.com/makeswift/makeswift/assets/10539418/f76906d5-8ab4-4680-8499-25370240dd33

### Loading states

https://github.com/makeswift/makeswift/assets/10539418/930413f2-3c58-47e2-ae5c-01cb86fae282

### DIFF
--- a/examples/bigcommerce-catalyst/components/product-carousel/product-carousel.makeswift.ts
+++ b/examples/bigcommerce-catalyst/components/product-carousel/product-carousel.makeswift.ts
@@ -1,0 +1,49 @@
+import dynamic from 'next/dynamic';
+
+import { Combobox, TextInput } from '@makeswift/runtime/controls';
+import { forwardNextDynamicRef } from '@makeswift/runtime/next';
+
+import { runtime } from '~/lib/makeswift/runtime';
+
+type Category = {
+  entityId: number;
+  name: string;
+  path: string;
+}
+
+const fetchCategories = async (): Promise<Category[]> => {
+  const response = await fetch('/api/categories')
+
+  const { categories }: { categories: Category[] } = await response.json()
+
+  return categories
+}
+
+export const props = {
+  title: TextInput({ label: 'Title', defaultValue: 'Products', selectAll: true }),
+  categoryId: Combobox({
+    label: "Category",
+    async getOptions(query) {
+      const categories = await fetchCategories();
+
+      return categories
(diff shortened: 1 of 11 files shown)


examples/bigcommerce-catalyst/components/product-carousel/product-carousel.makeswift.ts
```diff
@@ -0,0 +1,46 @@
...
+export const props = {
+  title: TextInput({ label: 'Title', defaultValue: 'Products', selectAll: true }),
+  products: Combobox({
+    label: "Category",
+    async getOptions(query) {
+      const { categories }: { categories: Category[] } = await fetch('/api/categories').then((res) => res.json())
+
+      const mappings = await Promise.all(categories
+        .map(async (category) => {
+          const { products } = await fetch(`/api/quick-search?categoryId=${category.entityId}&imageHeight=500&imageWidth=500&limit=10`).then((res) => res.json())
```
chanceaclark, line 23: BC has some limited GraphQL endpoints so this was the monstrosity I had to come up with unfortunately.

chanceaclark (commented):
Most of this was copy/pasta from the other `ProductCardCarousel` and `ProductCard` components. This was mostly because the `ProductCard` component relied on Server Actions, which don't work in the pages directory.

chanceaclark:
@andrewreifman tagging you just to make sure the custom `No category selected` design is good from your end.

examples/bigcommerce-catalyst/components/product-carousel/product-carousel.makeswift.ts
```diff
@@ -0,0 +1,44 @@
...
+
+          return {
+            id: category.entityId.toString(),
+            label: category.name,
+            value: products,
```
migueloller, line 28: Instead of storing the products in the data like this I thought we were going to pass the category ID to the component and then fetch products from the component dynamically. Otherwise these products will be set when the prop is set and won't change it the catalog data changes. And as for making it work well with SSR we'd do an SSR pre-pass.

andrewreifman:
> @andrewreifman tagging you just to make sure the custom `No category selected` design is good from your end.

Looks good to me

andrewreifman (dismissed):

migueloller (approved):

PR #709, opened 2024-03-25
Add version type to Background, Border, BorderRadius, and Checkbox prop controllers

fikrikarim:
This is needed for server-side introspection.

### DIFF
--- a/packages/runtime/src/components/builtin/Box/Box.tsx
+++ b/packages/runtime/src/components/builtin/Box/Box.tsx
@@ -23,6 +23,7 @@ import {
   GapXValue,
   CheckboxValue,
   BackgroundsValue,
+  getCheckboxValue,
 } from '../../../prop-controllers/descriptors'
 import { BoxModelHandle, parse, createBox } from '../../../box-model'
 import BackgroundsContainer from '../../shared/BackgroundsContainer'
@@ -202,7 +203,7 @@ const Box = forwardRef(function Box(
             </GridItem>
           ))
         ) : (
-          <Placeholder hide={hidePlaceholder} />
+          <Placeholder hide={getCheckboxValue(hidePlaceholder) ?? undefined} />
         )}
       </div>
     </BackgroundsContainer>
--- a/packages/runtime/src/components/builtin/Image/Image.tsx
+++ b/packages/runtime/src/components/builtin/Image/Image.tsx
@@ -5,12 +5,14 @@ import NextImage from 'next/image'
 import type NextLegacyImageType from 'next/legacy/image'
 
 import {
+  CheckboxValue,
   ElementIDValue,
   ImageValue,
   LinkValue,
   ResponsiveOpacityValue,
   TextInputValue,
   WidthValue,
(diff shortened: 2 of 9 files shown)


fikrikarim:
@migueloller Two questions:

1. One main difference with the control is that we don't have separate `data` and `value` for prop-controllers. For example, for the `CheckboxControl`, we have `CheckboxControlData` and `CheckboxControlValue`. While on the `CheckboxPropController` we only have the `CheckboxValue` and not `CheckboxData`. I was thinking of introducing the `data` type, but I'm afraid it'll complicate things because we don't have a centralized place to turn the `data` into the value`.

2. I saw that we're importing prop-controller values from the `control-serialization`. As we're changing those, this shouldn't make a problem right? It's weird that we have old prop-controllers but not the new one.

image

migueloller:
> ### ⚠️ No Changeset found
> Latest commit: [14c7adf](https://github.com/makeswift/makeswift/commit/14c7adf68116418fc403ffe485a31bea929d3811)
> 
> Merging this PR will not cause a version bump for any packages. If these changes should not result in a new version, you're good to go. **If these changes should result in a version bump, you need to add a changeset.**
> 
> This PR includes no changesets
> [Click here to learn what changesets are, and how to add one](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md).
> 
> [Click here if you're a maintainer who wants to add a changeset to this PR](https://github.com/makeswift/makeswift/new/fikri/prop-version-1?filename=.changeset/sixty-rockets-remember.md&value=---%0A%22%40makeswift%2Fruntime%22%3A%20patch%0A---%0A%0AAdd%20version%20type%20to%20Background%2C%20Border%2C%20BorderRadius%2C%20and%20Checkbox%20prop%20controllers%0A)

@fikrikarim, where's the changesets?

migueloller:
> 1. One main difference with the control is that we don't have separate `data` and `value` for prop-controllers. For example, for the `CheckboxControl`, we have `CheckboxControlData` and `CheckboxControlValue`. While on the `CheckboxPropController` we only have the `CheckboxValue` and not `CheckboxData`. I was thinking of introducing the `data` type, but I'm afraid it'll complicate things because we don't have a centralized place to turn the `data` into the value`.

We do have a centralized place to handle these things, the `PropsValue` component.

> 2. I saw that we're importing prop-controller values from the `control-serialization`. As we're changing those, this shouldn't make a problem right? It's weird that we have old prop-controllers but not the new one.
> 
> image

I'm not sure I understand this question. We still need to make sure the control and prop controller definitions are serializable. New controls are all serializable except for a couple like composable controls and advanced ones like Combobox which have functions in the definition. Older controls have stuff like `hidden` so most of them need to be serialized. This is related to definition serialization, not data serialization, though. So since all we should be doing is adding or bumping the version number in the definition, that shouldn't affect serialization. Which, by the way, I'm not seeing that in this MR, adding a version to the control so that the panels/overlays in the builder know to write in the latest data format.

fikrikarim:
Closing in favor of #711

PR #710, opened 2024-03-26
Add data types to Shadows, ResponsiveLength, Number, and ResponsiveColor

migueloller:
This PR adds data types to the `Shadows`, `ResponsiveLength`, `Number`, and `ResponsiveColor` prop controllers.

Some notes:
- It was necessary to keep types for the older variant of the descriptor for consumers of this type to be able to support older versions of the runtime (i.e., panels have to be able to write in the format of older descriptors thus we need to preserve older data types as well as older descriptor types).
- I finally figured out a pattern for naming: prop controller data uses the name `<name>PropControllerData`, prop controller hydrated values (i.e., what a component receives) uses the name `<name>PropControllerValue`, shared data (e.g., color data, shadows data, rich text data, etc.) uses the name `<name>Data` (e.g., `ColorData`), and finally, shared data that has been hydrated (i.e., a color _value_ with hue, saturation, lightness) uses the name `<name>Value` (e.g., `ColorValue`).
- I started moving various things to `@makeswift/prop-controllers` so that we start having things centralized there and sharing it there.
- I explicitly exported the named exports from `@makeswift/prop-controllers` to avoid accidental leaking of internals by being explicit about the public API.
- I added Zod so that we can parse prop controller data and handle data corruption and other validation issues.
- I created some factory functions to create prop controller data given a definition and shared data. This allows the panels to be decoupled from the data versions.
- There are a couple of TODOs around where to put the resolve types for prop controllers and manually constructing preset data that I think can be addressed in a future PR.


migueloller:
@fikrikarim, this is ready for review. I have some more plans for this PR, but those will come in as additional commits:
- Tests for functionality in `src/shadows.ts`
- A GitHub action for canary releases

I'll get to those after I set up the MR in GitLab for the panel.

packages/prop-controllers/src/number.ts
```diff
@@ -0,0 +1,94 @@
...
+}
+
+export function createNumberPropControllerDataFromNumber(
+  value: number,
+  definition?: NumberDescriptor,
+): NumberPropControllerData {
+  return match(definition)
+    .with(
+      { version: 1 },
+      P.nullish,
```
fikrikarim, line 86: What is this `P.nullish` for?

packages/runtime/src/components/builtin/Navigation/register.ts
```diff
@@ -30,13 +37,20 @@ export function registerComponent(runtime: ReactRuntime) {
...
-            {
-              deviceId: getBaseBreakpoint(runtime.getBreakpoints()).id,
-              value: { value: 100, unit: 'px' },
-            },
-          ],
+        logoWidth: ResponsiveLength(props => ({
+          // TODO(miguel): We're manually constructing the data here but should be using a factory
+          // function instead. This is because the factory function currently expects a definition
+          // but we don't have one to pass here. Perhaps we should make the factory function not
+          // require the definition and use the latest version when a definition isn't provided.
```
fikrikarim, line 44: Isn't it a safe assumption that we can always use the latest data version? Because if you have the v1 definition it means the runtime could handle the v1 version.

fikrikarim (changes requested):

fikrikarim (approved):

packages/prop-controllers/src/number.ts
fikrikarim: Ok I understand this is for handling when definition is nullish

packages/prop-controllers/src/number.ts
```diff
@@ -0,0 +1,94 @@
...
+      { [ControlDataTypeKey]: NumberPropControllerDataV1Type },
+      (v1) => v1.value,
+    )
+    .otherwise((v0) => v0)
+}
+
+export function createNumberPropControllerDataFromNumber(
+  value: number,
+  definition?: NumberDescriptor,
+): NumberPropControllerData {
```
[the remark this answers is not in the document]
migueloller: Great question. One of the things I realized is that for presets we might want to be able to create the latest version of the prop controller data without having a definition. For example, `createNumberPropControllerDataFromNumber(5)`.

These APIs would be a lot cleaner if it was something like `NumberPropControllerData.fromNumber(5)` or something like that... Maybe I'll explore a cleaner API like this.

packages/runtime/src/components/builtin/Navigation/register.ts
migueloller: Yes, it should be safe to do so.

PR #711, opened 2024-03-26
feat: add a new prop-controllers package and migrate LinkPropController

fikrikarim:


.changeset/spicy-dolls-talk.md
migueloller, line 1: I think we might want to have separate changesets for each package.

packages/prop-controllers/.gitignore
migueloller, line 1: Don't we already have a gitignore file at the repo's root that takes care of all these files?

packages/prop-controllers/README.md
```diff
@@ -0,0 +1 @@
+This package contains deprecated prop-controllers and is intended for internal use only.
```
migueloller, line 1: Let's add a title to this README and maybe some more information as to why it's needed?

packages/prop-controllers/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "@makeswift/prop-controllers",
+  "version": "0.0.1",
+  "license": "MIT",
+  "type": "module",
+  "files": [
+    "dist"
+  ],
+  "main": "./dist/index.cjs",
+  "types": "./dist/types/index.d.ts",
```
migueloller, line 10: Does this give us the type setup we want for IDEs? For example, does jump to definition work? Do we need to add source maps?

packages/prop-controllers/tsconfig.json
migueloller, line 1: How did we decide on these values?

packages/runtime/src/components/builtin/Box/Box.tsx
```diff
@@ -22,17 +22,17 @@ import {
...
 import BackgroundsContainer from '../../shared/BackgroundsContainer'
 import { useResponsiveStyle } from '../../utils/responsive-style'
 import { GridItem } from '../../shared/grid-item'
 import { useStyle } from '../../../runtimes/react/use-style'
+import { BackgroundsPropControllerData } from '@makeswift/prop-controllers'
 
 type Props = {
   id?: ElementIDValue
-  backgrounds?: BackgroundsValue
+  backgrounds?: BackgroundsPropControllerData
```
migueloller, line 35: This shouldn't be tied to the prop controller data but instead the provided value.

packages/runtime/src/components/builtin/Button/Button.tsx
```diff
@@ -27,6 +26,7 @@ import {
   useResponsiveWidth,
 } from '../../utils/responsive-style'
 import { cx } from '@emotion/css'
+import { LinkValue } from '@makeswift/prop-controllers'
```
migueloller, line 29: We should decouple the value from the data, or is this the type for the value?

packages/runtime/src/components/builtin/Carousel/Carousel.tsx
```diff
@@ -32,6 +31,7 @@ import { cx } from '@emotion/css'
 import { useResponsiveStyle } from '../../utils/responsive-style'
 import { getBaseBreakpoint } from '../../../state/modules/breakpoints'
 import { useBreakpoints } from '../../../runtimes/react'
+import { ImagesValue } from '@makeswift/prop-controllers'
```
migueloller, line 34: Is this type prop controller data or a shared value we compute from the data? These components should mostly be decoupled from controls.

packages/runtime/src/components/builtin/Root/Root.tsx
```diff
@@ -16,7 +12,7 @@ import { useStyle } from '../../../runtimes/react/use-style'
 
 type Props = {
   children?: GridValue
-  backgrounds?: BackgroundsValue
+  backgrounds?: BackgroundsPropControllerData
```
migueloller, line 15: This should be decoupled from the prop controller data

packages/runtime/src/prop-controllers/backgrounds.ts
migueloller, line 1: Why is there an empty file here?

migueloller (commented):
This looks good but there's _a lot_ going on here. I've got a couple of questions:
- Can we try to tackle just 1 prop controller to reduce the amount of surface area we're going over there so that we can focus on one thing at a time? Maybe we start with just the Image prop controller since that one's pretty simple?
- Are we trying to simply move existing code around or are we also trying to clean things up a bit as we do this? I think we should try to clean things up a bit to continue to make it easier to maintain this code and continue to simplify the user of it.
- We need to figure out the distinction between data and provided values and where that stuff goes. This is going to apply to prop controllers and well as controls so we might as well do that now.

packages/runtime/src/components/builtin/Button/Button.tsx
fikrikarim: I only copy paste the `LinkValue` and haven't updated the type yet.

packages/runtime/src/components/builtin/Box/Box.tsx
fikrikarim: I discarded the changes related to the Background. Let's focus on Link for now.

packages/runtime/src/components/builtin/Carousel/Carousel.tsx
fikrikarim: I discarded the changes related to the Images. Let's focus on Link for now.

packages/runtime/src/prop-controllers/backgrounds.ts
fikrikarim: Thank you. Deleted.

packages/runtime/src/components/builtin/Root/Root.tsx
fikrikarim: I discarded the changes related to the Background. Let's focus on Link for now.

fikrikarim (commented):
> Can we try to tackle just 1 prop controller to reduce the amount of surface area we're going over there so that we can focus on one thing at a time? Maybe we start with just the Image prop controller since that one's pretty simple?

Sure. Image depends on Link. I discarded the changes for Background and Image. Let's focus on Link for now.

> Are we trying to simply move existing code around or are we also trying to clean things up a bit as we do this? I think we should try to clean things up a bit to continue to make it easier to maintain this code and continue to simplify the user of it.

Sure. Let me know if there are better way to clean things up.

> We need to figure out the distinction between data and provided values and where that stuff goes. This is going to apply to prop controllers and well as controls so we might as well do that now.

Yes. I agree. I plan to have a separate commit for this, so we can differentiate the commit that just move code and the commit that actually change the code.

packages/prop-controllers/src/link.ts
```diff
@@ -32,21 +33,45 @@ export type Link =
...
-  preset?: LinkValue
+  preset?: LinkPropControllerData
   label?: string
   defaultValue?: Link
   options?: { value: Link['type']; label: string }[]
   hidden?: boolean
 }>
 
-export type LinkDescriptor<_T = LinkValue> = {
+export type LinkDescriptor<_T = LinkPropControllerValue> = {
```
fikrikarim, line 57: @migueloller this should use the value and not the data right?

fikrikarim (commented):
@migueloller I added a new commit that adds the v1 data type.

packages/prop-controllers/package.json
migueloller: Let's make sure to try it from an external package like Leo. The easiest way is using the `file://` protocol from pnpn for a dependency

packages/prop-controllers/tsconfig.json
migueloller: I'm still figuring out myself what the best minimal options are. There's a blog post I recently shared on Slack that went over some that maybe we can use as inspo: https://www.totaltypescript.com/tsconfig-cheat-sheet

packages/runtime/src/prop-controllers/descriptors.ts
```diff
@@ -554,26 +530,6 @@ export function Images(options: ImagesOptions = {}): ImagesDescriptor {
...
-
-export type LinkDescriptor<_T = LinkValue> = { type: typeof Types.Link; options: LinkOptions }
-
-/**
- * @deprecated Imports from `@makeswift/runtime/prop-controllers` are deprecated. Use
- * `@makeswift/runtime/controls` instead.
- */
-export function Link(options: LinkOptions = {}): LinkDescriptor {
-  return { type: Types.Link, options }
-}
```
migueloller, line 575: These exports, shouldn't we be re-exporting the equivalent bindings from `@makeswift/prop-controllers` so that we're not breaking things?

packages/prop-controllers/src/link.ts
migueloller: I think that the generics we use here are the descriptor data. Mapping it to a value is done separately.

packages/runtime/src/components/builtin/Button/Button.tsx
```diff
@@ -27,12 +26,13 @@ import {
...
 } from '../../utils/responsive-style'
 import { cx } from '@emotion/css'
+import { LinkPropControllerValue } from '@makeswift/prop-controllers'
 
 type BaseProps<T extends ElementType> = {
   as?: T
   id?: ElementIDValue
   children?: ReactNode
-  link?: LinkValue
+  link?: LinkPropControllerValue
```
migueloller, line 35: We should export a generic `Link` type which is used by multiple prop controllers.

packages/prop-controllers/package.json
fikrikarim: Didn't know there was a `file://` protocol. I tested it and it's working.

https://github.com/makeswift/makeswift/assets/13066728/c2d00706-15fe-4333-b6f1-afd9b2acf27c

packages/prop-controllers/tsconfig.json
fikrikarim: That's a good reference. I'll say it's good enough for now. 

We can iterate on this together we iterate on the runtime tsconfig.

packages/prop-controllers/src/link.ts
fikrikarim: Ok updated to `LinkPropControllerData`

packages/prop-controllers/src/link.ts
```diff
@@ -0,0 +1,174 @@
...
+  switch (link.type) {
+    case 'OPEN_PAGE':
+      return link.payload.pageId == null ? [] : [link.payload.pageId]
+
+    default:
+      return []
+  }
+}
+
+export function copyLinkPropControllerData(
```
fikrikarim, line 113: @migueloller We also need to move the copy function to the `@makeswift/prop-controllers` package right?

migueloller: Yes, since it's an introspection function.

packages/runtime/src/prop-controllers/copy.ts
```diff
@@ -24,7 +24,7 @@ export function copy(descriptor: Descriptor, value: any, context: CopyContext) {
     case 'NavigationLinks':
       return navigationLinksCopy(value, context)
     case 'Link':
```
migueloller, line 26: Maybe we should use the enum here?

packages/runtime/src/controls/rich-text/introspection.ts
```diff
@@ -84,7 +84,9 @@ export function getRichTextPageIds(value: RichTextControlData) {
     switch (inline.type) {
       case 'link': {
         const nodePageIds = inline.nodes?.flatMap(getNodePageIds) ?? []
-        const dataPageIds = inline.data ? getLinkDataPageIds(inline.data as LinkValue) : []
+        const dataPageIds = inline.data
+          ? getLinkPropControllerPageIds(inline.data as LinkPropControllerData)
```
migueloller, line 88: We should be casting this to `Link` instead, no? Since there won't be multiple versions and data types for the prop controller here?

packages/prop-controllers/src/link.ts
migueloller, line 1: It would be great to have some tests for the functions defined in this file.

migueloller (dismissed):
Looking great! Got a couple of comments on some minor stuff but this looks like it's basically good to go.

packages/prop-controllers/src/link.ts
fikrikarim: Added

packages/prop-controllers/src/link/link.test.ts
```diff
@@ -0,0 +1,295 @@
...
+        test: 'unknown',
+      }
+
+      // Act
+      // @ts-expect-error: invalid data
+      const result = getLinkPropControllerValue(data)
+
+      // Assert
+      expect(result).toEqual(data)
+    })
```
migueloller, line 66: This is an interesting test case. One could argue that this should be a parsing error.

fikrikarim: That's true. I think the problem is that right now we don't have a way to gracefully handle parse errors in the `PropsValue`

migueloller: Yeah, maybe as we evolve this we can do a better job of handling bad data in these functions. I'll see what I can do when working with the Shadows prop controller

PR #713, opened 2024-03-26
feat: migrate basic typescript to app router

joshuawootonn:


examples/basic-typescript/package.json
```diff
@@ -8,18 +8,18 @@
...
-    "next": "13.2.1",
+    "next": "14.1.4",
     "react": "18.2.0",
     "react-dom": "18.2.0"
   },
   "devDependencies": {
     "@trivago/prettier-plugin-sort-imports": "^4.2.0",
     "@types/node": "17.0.37",
-    "@types/react": "^18.2.0",
+    "@types/react": "18.2.60",
```
joshuawootonn, line 22: https://linear.app/makeswift/issue/ENG-4902/figure-out-why-latest-version-of-typesreact-breaks-lazy-component

I will fill this in at some point.

The Bigcommerce catalyst example isn't seeing this problem because it has a lock file.

examples/basic-typescript/tsconfig.json
```diff
@@ -1,7 +1,11 @@
 {
   "compilerOptions": {
     "target": "es6",
-    "lib": ["dom", "dom.iterable", "esnext"],
+    "lib": [
```
joshuawootonn, line 4: auto changed by next upgrade

examples/basic-typescript/app/[[...path]]/page.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { getSiteVersion } from '@makeswift/runtime/next/server'
+
+import { client } from '@/lib/makeswift/client'
+
+type ParsedUrlQuery = { path?: string[] }
+
+export async function generateStaticParams() {
+  const pages = await client.getPages()
+
+  return pages.flatMap(page => ({
```
Comment (migueloller) [inline examples/basic-typescript/app/[[...path]]/page.tsx:13]: Why are we using `flatMap` here?

examples/basic-typescript/lib/makeswift/provider.tsx
migueloller, line 1: nit: I think this file shouldn't exist and instead an `app/provider.tsx` client component should be used were all third-party lib providers are used in a glorious provider Christmas tree 🎄

migueloller (dismissed):

examples/basic-typescript/app/[[...path]]/page.tsx
Comment (joshuawootonn) [reply examples/basic-typescript/app/[[...path]]/page.tsx:13]: Good question. I actually used the preview of the docs to do this migration. This was a c/p from there. surprised it worked. Glad you caught this 🙌🏼 

Fixing the docs

examples/basic-typescript/lib/makeswift/provider.tsx
joshuawootonn: Think I'm going to leave this for now, since v0 of these migration docs will point to having this file / location. I'm down for this suggestion though.

This is making me question though.. could we just expose a client component called `MakeswiftProviders` that could just be placed in the root layout? Why didn't we do that?

PR #728, opened 2024-04-22
fix(runtime): pages are not editable when running in dev w/ NextJS 14.2

agurtovoy:
Decouple `MessageChannel` creation from the middleware creation to make store initalization compatible with React's [strict mode](https://react.dev/reference/react/StrictMode). 

Next.js has React's strict mode [enabled by default](https://nextjs.org/docs/pages/api-reference/next-config-js/reactStrictMode) in dev; when running w/ Next.js 14.2 and [App Router](https://nextjs.org/docs/app) in dev, creation of the channel in the middleware was resulting in a secondary "ghost" store instance that was handling some of the dispatch methods, leading to a de-synchronized runtime state.

### DIFF
--- a/.changeset/chilly-peaches-wonder.md
+++ b/.changeset/chilly-peaches-wonder.md
@@ -1,0 +1,5 @@
+---
+'@makeswift/runtime': minor
+---
+
+Next.js 14.2 compatibility fix: decouple `PreviewProvider`'s message channel setup from the middleware creation to make store initialization compatible with React's strict mode.
--- a/packages/runtime/src/state/react-builder-preview.ts
+++ b/packages/runtime/src/state/react-builder-preview.ts
@@ -2,11 +2,13 @@ import {
   applyMiddleware,
   combineReducers,
   createStore,
+  compose,
   Dispatch as ReduxDispatch,
   Middleware,
   MiddlewareAPI,
   PreloadedState,
   Store as ReduxStore,
+  StoreEnhancer,
 } from 'redux'
 import thunk, { ThunkAction, ThunkDispatch } from 'redux-thunk'
 
@@ -29,19 +31,21 @@ import {
   Action,
   changeDocumentElementSize,
   changeElementBoxModels,
+  elementFromPointChange,
+  handleWheel,
+  handlePointerMove,
   messageBuilderPropController,
   registerBuilderComponent,
+  registerComponent,
   registerMeasurable,
   registerPropControllers,
   registerPropControllersHandle,
+  setBreakpoints,
+  setIsInBuilder,
   unregisterBuilderComponent,
+  unregisterComponent,
   unregisterMeasurable,
   unregisterPropControllers,
-  setIsInBuilder,
-  handleWheel,
-  handlePointerMove,
-  elementFromPointChange,
-  setBreakpoints,
 } from './actions'
 import { ActionTypes } from './actions'
 import { createPropController } from '../prop-controllers/instances'
@@ -402,15 +406,47 @@ function startPollingElementFromPoint(): ThunkAction<() => void, State, unknown,
   }
 }
 
-export function initialize(): ThunkAction<() => void, State, unknown, Action> {
-  return dispatch => {
+function registerBuilderComponents(): ThunkAction<() => void, State, unknown, Action> {
+  return (dispatch, getState) => {
+    const state = getState()
+    const componentsMeta = getComponentsMeta(state)
+
+    componentsMeta.forEach((meta, type) => {
+      const propControllerDescriptors = getComponentPropControllerDescriptors(state, type)
+      if (propControllerDescriptors != null) {
+        dispatch(registerComponent(type, meta, propControllerDescriptors))
+      }
+    })
+
+    return () => {
+      componentsMeta.forEach((_, componentType) => {
+        dispatch(unregisterComponent(componentType))
+      })
+    }
+  }
+}
+
+export interface IMessageChannel {
+  postMessage(message: any, transferables?: Transferable[]): void
+  dispatchBuffered(): void
+}
+
+export function initialize(
+  channel: IMessageChannel,
+): ThunkAction<() => void, State, unknown, Action> {
+  return (dispatch, getState) => {
     const stopMeasuringElements = dispatch(startMeasuringElements())
     const stopMeasuringDocumentElement = dispatch(startMeasuringDocumentElement())
     const stopHandlingFocusEvent = dispatch(startHandlingFocusEvents())
     const unlockDocumentScroll = dispatch(lockDocumentScroll())
     const stopHandlingPointerMoveEvent = dispatch(startHandlingPointerMoveEvent())
     const stopPollingElementFromPoint = dispatch(startPollingElementFromPoint())
+    const unregisterBuilderComponents = dispatch(registerBuilderComponents())
+
+    const breakpoints = ReactPage.getBreakpoints(getState())
+    dispatch(setBreakpoints(breakpoints))
     dispatch(setIsInBuilder(true))
+    channel.dispatchBuffered()
 
     return () => {
       stopMeasuringElements()
@@ -419,6 +455,7 @@ export function initialize(): ThunkAction<() => void, State, unknown, Action> {
       unlockDocumentScroll()
       stopHandlingPointerMoveEvent()
       stopPollingElementFromPoint()
+      unregisterBuilderComponents()
       dispatch(setIsInBuilder(false))
     }
   }
@@ -457,41 +494,13 @@ function measureBoxModelsMiddleware(): Middleware<Dispatch, State, Dispatch> {
 
 export function messageChannelMiddleware(
   client: MakeswiftHostApiClient,
+  channel: IMessageChannel,
 ): Middleware<Dispatch, State, Dispatch> {
-  return ({ dispatch, getState }: MiddlewareAPI<Dispatch, State>) =>
+  return ({ dispatch }: MiddlewareAPI<Dispatch, State>) =>
     (next: ReduxDispatch<Action>) => {
+      if (typeof window === 'undefined') return () => {}
+
       let cleanUp = () => {}
-
-      if (typeof window === 'undefined') return cleanUp
-
-      const messageChannel = new window.MessageChannel()
-
-      window.parent.postMessage(messageChannel.port2, '*', [messageChannel.port2])
-
-      messageChannel.port1.onmessage = (event: MessageEvent<Action>) => dispatch(event.data)
-
-      const state = getState()
-      const registeredComponentsMeta = getComponentsMeta(state)
-
-      registeredComponentsMeta.forEach((componentMeta, componentType) => {
-        const propControllerDescriptors = getComponentPropControllerDescriptors(
-          state,
-          componentType,
-        )
-
-        if (propControllerDescriptors != null) {
-          const [serializedControls, transferables] = serializeControls(propControllerDescriptors)
-
-          messageChannel.port1.postMessage(
-            registerBuilderComponent(componentType, componentMeta, serializedControls),
-            transferables,
-          )
-        }
-      })
-
-      const breakpoints = ReactPage.getBreakpoints(state)
-      messageChannel.port1.postMessage(setBreakpoints(breakpoints))
-
       return (action: Action): Action => {
         switch (action.type) {
           case ActionTypes.CHANGE_ELEMENT_BOX_MODELS:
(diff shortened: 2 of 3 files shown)


packages/runtime/src/state/react-builder-preview.ts
```diff
@@ -455,22 +455,23 @@ function measureBoxModelsMiddleware(): Middleware<Dispatch, State, Dispatch> {
...
       let cleanUp = () => {}
-
-      if (typeof window === 'undefined') return cleanUp
-
-      const messageChannel = new window.MessageChannel()
-
-      window.parent.postMessage(messageChannel.port2, '*', [messageChannel.port2])
-
-      messageChannel.port1.onmessage = (event: MessageEvent<Action>) => dispatch(event.data)
+      if (typeof window === 'undefined' || !messageChannel) return cleanUp
```
migueloller, line 470: It's interesting we've got a `cleanUp` function we're returning but we're really not returning any clean up further down this function. It just looks like bad naming since a Redux middleware should be returning a dispatch function here, not a clean up function.

migueloller (commented):
@agurtovoy, we use [changesets](https://github.com/changesets/changesets) to manage releases. It takes care of generating change logs as well as updating package versions, GitHub releases, git tags, canary releases, etc.

packages/runtime/src/state/react-builder-preview.ts
agurtovoy: I noticed that, too, will rename.

migueloller (approved):
This seems reasonable! Again, nice find. Imm wondering if this will also address a memory leak we’ve been having due to not handling message channels from function serialization.

Instead of actions we could also perform the side effects in a thunk. I realized as I was reading this that that would be more friendly to Redux since doing a side effect on an action is against Redux’s assumptions. And this applies regardless of whether we do a mixin or not (I incorrectly couple these two when we discussed this earlier). The crux here is wether we perform the side effect and cleanup as a response to a Redux action or within the thunk itself. If done in an action, certain things, like time travel in Redux DevTools won’t work as expected.

Don’t forget to add a changeset!

packages/runtime/src/state/react-builder-preview.ts
```diff
@@ -402,15 +404,50 @@ function startPollingElementFromPoint(): ThunkAction<() => void, State, unknown,
...
 }
 
-export function initialize(): ThunkAction<() => void, State, unknown, Action> {
-  return dispatch => {
+export interface IMessageChannel {
+  postMessage(message: any, transferables?: Transferable[]): void
+  dispatchBuffered(): void
+}
+
+function registerBuilderComponents(state: State, messageChannel: IMessageChannel) {
```
agurtovoy, line 412: Factor out builder component registration into its own function, make it a part of the initialization sequence (see `initialize` below).

packages/runtime/src/state/react-builder-preview.ts
```diff
@@ -402,15 +404,50 @@ function startPollingElementFromPoint(): ThunkAction<() => void, State, unknown,
...
     const stopHandlingFocusEvent = dispatch(startHandlingFocusEvents())
     const unlockDocumentScroll = dispatch(lockDocumentScroll())
     const stopHandlingPointerMoveEvent = dispatch(startHandlingPointerMoveEvent())
     const stopPollingElementFromPoint = dispatch(startPollingElementFromPoint())
+    const unregisterBuilderComponents = registerBuilderComponents(state, channel)
+
+    const breakpoints = ReactPage.getBreakpoints(state)
+    dispatch(setBreakpoints(breakpoints))
     dispatch(setIsInBuilder(true))
+    channel.dispatchBuffered()
```
agurtovoy, line 450: Dispatch channel's buffered messages, if any, as a part of the initialization sequence after the connection has been established.

packages/runtime/src/state/react-builder-preview.ts
```diff
@@ -457,41 +495,13 @@ function measureBoxModelsMiddleware(): Middleware<Dispatch, State, Dispatch> {
...
     (next: ReduxDispatch<Action>) => {
-      let cleanUp = () => {}
-
-      if (typeof window === 'undefined') return cleanUp
-
-      const messageChannel = new window.MessageChannel()
-
-      window.parent.postMessage(messageChannel.port2, '*', [messageChannel.port2])
-
-      messageChannel.port1.onmessage = (event: MessageEvent<Action>) => dispatch(event.data)
```
agurtovoy, line 471: Decoupled channel creation from the middleware creation, see `MessageChannel` below.

packages/runtime/src/state/react-builder-preview.ts
```diff
@@ -457,41 +495,13 @@ function measureBoxModelsMiddleware(): Middleware<Dispatch, State, Dispatch> {
...
-
-        if (propControllerDescriptors != null) {
-          const [serializedControls, transferables] = serializeControls(propControllerDescriptors)
-
-          messageChannel.port1.postMessage(
-            registerBuilderComponent(componentType, componentMeta, serializedControls),
-            transferables,
-          )
-        }
-      })
```
agurtovoy, line 490: Moved to its own function, see `registerBuilderComponents` above.

packages/runtime/src/state/react-builder-preview.ts
```diff
@@ -457,41 +495,13 @@ function measureBoxModelsMiddleware(): Middleware<Dispatch, State, Dispatch> {
...
-
-      const breakpoints = ReactPage.getBreakpoints(state)
-      messageChannel.port1.postMessage(setBreakpoints(breakpoints))
```
agurtovoy, line 493: Moved to the initialization sequence, see the `initialize` function above.

packages/runtime/src/state/react-builder-preview.ts
```diff
@@ -648,7 +661,70 @@ function makeswiftApiClientSyncMiddleware(
...
-export type Store = ReduxStore<State, Action> & { dispatch: Dispatch }
+class MessageChannel {
+  private channel: MessagePort | null = null
+  private bufferedMessages: [Action, Transferable[]?][] = []
+
+  public postMessage(message: any, transferables?: Transferable[]) {
+    if (this.channel) {
+      this.channel.postMessage(message, transferables ?? [])
+    } else {
+      this.bufferedMessages.push([message, transferables])
```
agurtovoy, line 672: Buffer messages posted before the underlying channel has been created.

.changeset/chilly-peaches-wonder.md
```diff
@@ -0,0 +1,5 @@
+---
+'@makeswift/runtime': minor
+---
+
+Next.js 14.2 compatibility fixes
```
migueloller, line 5: Should we outline here what changed? 🤔

packages/runtime/src/state/react-builder-preview.ts
```diff
@@ -648,7 +661,70 @@ function makeswiftApiClientSyncMiddleware(
...
+
+    // connect channel to the parent window, see
+    // https://developer.mozilla.org/en-US/docs/Web/API/Channel_Messaging_API
+    window.parent.postMessage(channel.port2, '*', [channel.port2])
+
+    this.channel = channel.port1
+  }
+
+  public dispatchBuffered() {
+    console.assert(this.channel != null, 'channel is not setup')
```
migueloller, line 688: This is cool. Didn't know about this API.

packages/runtime/src/state/react-builder-preview.ts
```diff
@@ -648,7 +661,70 @@ function makeswiftApiClientSyncMiddleware(
...
+
+    this.bufferedMessages.forEach(([message, transferables]) => {
+      this.postMessage(message, transferables ?? [])
+    })
```
migueloller, line 692: Are there any potential issues here with messages pushing more messages resulting in a potential loop. Does `forEach` take a snapshot of the array or should be first get a clone of the buffered messages before we loop?

packages/runtime/src/state/react-builder-preview.ts
```diff
@@ -503,22 +513,23 @@ export function messageChannelMiddleware(
           case ActionTypes.HANDLE_POINTER_MOVE:
           case ActionTypes.ELEMENT_FROM_POINT_CHANGE:
           case ActionTypes.SET_LOCALE:
-            messageChannel.port1.postMessage(action)
+          case ActionTypes.SET_BREAKPOINTS:
```
migueloller, line 516: Was this case (SET_BREAKPOINTS) just missing before?

packages/runtime/src/state/react-builder-preview.ts
```diff
@@ -402,15 +404,50 @@ function startPollingElementFromPoint(): ThunkAction<() => void, State, unknown,
...
+): ThunkAction<() => void, State, unknown, Action> {
+  return (dispatch, getState) => {
+    const state = getState()
     const stopMeasuringElements = dispatch(startMeasuringElements())
     const stopMeasuringDocumentElement = dispatch(startMeasuringDocumentElement())
     const stopHandlingFocusEvent = dispatch(startHandlingFocusEvents())
     const unlockDocumentScroll = dispatch(lockDocumentScroll())
     const stopHandlingPointerMoveEvent = dispatch(startHandlingPointerMoveEvent())
     const stopPollingElementFromPoint = dispatch(startPollingElementFromPoint())
+    const unregisterBuilderComponents = registerBuilderComponents(state, channel)
```
migueloller, line 445: Should we make `registerBuilderComponents` a thunk to be consistent so we can do here `dispatch(registerBuilderComponents(channel))` and let it get `state` via `getState` itself?

packages/runtime/src/state/react-builder-preview.ts
migueloller: I now see that it's handling the `messageChannel.port1.postMessage(setBreakpoints(breakpoints))` that was removed above 👍🏻

migueloller (approved):
This looks great! 👍🏻 

Left some comments but they're either questions or nits. Feel free to ship without addressing them.

Thanks @agurtovoy 🙏🏻

packages/runtime/src/state/react-builder-preview.ts
agurtovoy: I considered it, but the extra indirection didn't seem to be worth it to me. I can kind of see why you went for decomposing thunks into more thunks here, though, as opposite to just going with a simpler `startMeasuringElements(dispatch)`, will do the same.

packages/runtime/src/state/react-builder-preview.ts
agurtovoy: Good point about making sure we don't loop if the precondition fails. I think this should be just `this.channel?.postMessage`.

packages/runtime/src/state/react-builder-preview.ts
```diff
@@ -402,15 +406,47 @@ function startPollingElementFromPoint(): ThunkAction<() => void, State, unknown,
...
+      }
+    })
+
+    return () => {
+      componentsMeta.forEach((_, componentType) => {
+        dispatch(unregisterComponent(componentType))
+      })
+    }
+  }
+}
```
agurtovoy, line 427: @migueloller Check this out, no channel needed here after all, we already handle `REGISTER_COMPONENT`/`UNREGISTER_COMPONENT` below, just needed to dispatch these two.

packages/runtime/src/state/react-builder-preview.ts
agurtovoy: Realized we actually don't need the channel to register/unregister here, see my other comment.

packages/runtime/src/state/react-builder-preview.ts
migueloller: Well, yes, but also what I mean is something like:

```
const bufferedMessages = this.bufferedMessages.slice()

bufferedMessages.forEach(// ...
```

This makes it so that it `this.bufferedMessages` is mutated in the loop, we don't go back in. This might be unnecessary depending on how `forEach` works. If we were using a `for` loop on the mutable array, for example, it would definitely be an issue.

packages/runtime/src/state/react-builder-preview.ts
migueloller: Yeah, I definitely asked myself a lot whether it was useful to make everything a thunk. I opted for it because it seemed like a robust abstraction of "do this thing and I don't care whether it depends on dispatch or state, I just care about the _action_". It's an easy thing to evolve, though. So I don't have a super strong opinion here and am happy to go in whatever direction you prefer.

packages/runtime/src/state/react-builder-preview.ts
migueloller: Nice!

migueloller: Was there a difference between the `registerComponent` and the `registerBuilderComponent` actions? It seems that there would be given the naming, but perhaps it's silly tech debt.

packages/runtime/src/state/react-builder-preview.ts
agurtovoy: That's how I understood you, but I felt like changing the loop body to `this.channel?.postMessage` eliminates the possibility of `bufferedMessages` being mutated. If `channel` is not null and we actually do post a message that generates a response from the builder and dispatches another message, because `channel` is not null, that message won't be buffered and will be just sent to the builder. 

We could do `.slice()` just to "future-proof" this, but I feel like that would actually make me think that this "re-buffering" scenario is possible (since we don't `slice()` everywhere as a general rule), which would be confusing. Does this make sense?

migueloller: Oh, whoops, I totally missed that you were using the channel directly 🤦🏻‍♂️  This makes total sense.

PR #729, opened 2024-04-22
Add data type to PropControllers: Date, Font, Video, Table, Margin, Padding, Width

fikrikarim:


migueloller:
I think we gotta be consistent with file structure. I was dropping `.ts` and `.test.ts` files under `src` but you seem to be making a folder per prop controller and putting things in there. I personally prefer avoiding the extra folder, but don't have a strong preference other than consistency.

packages/prop-controllers/src/margin/margin.test.ts
```diff
@@ -0,0 +1,160 @@
...
+        createMarginPropControllerDataFromResponsiveMarginData(data)
+
+      // Assert
+      expect(result).toEqual({
+        [ControlDataTypeKey]: MarginPropControllerDataV1Type,
+        value: data,
+      })
+    })
+
+    test('returns string value when definition version is not 1', () => {
```
migueloller, line 132: There's some copy pasta here 🍝

packages/prop-controllers/src/padding/padding.test.ts
```diff
@@ -0,0 +1,160 @@
...
+      }
+
+      // Act
+      // @ts-expect-error: invalid data
+      const result = getPaddingPropControllerDataResponsivePaddingData(data)
+
+      // Assert
+      expect(result).toEqual(data)
+    })
+  })
```
migueloller, line 72: Again, this test case feels super weird to me. I feel like I'd rather "break" behavior by failing fast if we encounter this case.

I guess we'd have to introduce parsing to the `getPaddingPropControllerDataResponsivePaddingData` function and perhaps we don't want to do that? But in that case, then it seems this test case isn't even necessary since we're assuming types are _correct_. If types can be wrong, then parsing might not be a bad idea, then, hence the necessity for a test case.

packages/prop-controllers/src/padding/padding.test.ts
```diff
@@ -0,0 +1,160 @@
...
+        createPaddingPropControllerDataFromResponsivePaddingData(data)
+
+      // Assert
+      expect(result).toEqual({
+        [ControlDataTypeKey]: PaddingPropControllerDataV1Type,
+        value: data,
+      })
+    })
+
+    test('returns string value when definition version is not 1', () => {
```
migueloller, line 132: Copy pasta 🍝

packages/prop-controllers/src/table/table.ts
```diff
@@ -0,0 +1,130 @@
...
+): TableData | undefined {
+  return match(data)
+    .with(
+      { [ControlDataTypeKey]: TablePropControllerDataV1Type },
+      (v1) => v1.value,
+    )
+    .otherwise((v0) => v0)
+}
+
+export function createTablePropControllerDataFromTableData(
```
migueloller, line 84: I think the name here doesn't help much. It's all table data, instead this should probably be `createTablePropControllerDataFromTableId`. Or better yet, it would be nice if this stuff was organized under a namespace or class so we could do `TablePropControllerData.fromId(tableId)`.

packages/prop-controllers/src/table/table.ts
```diff
@@ -0,0 +1,130 @@
...
+  ControlDataTypeKey,
+  CopyContext,
+  Options,
+  Types,
+} from '../prop-controllers'
+import { P, match } from 'ts-pattern'
+
+const tableDataSchema = z.string()
+
+export type TableData = z.infer<typeof tableDataSchema>
```
migueloller, line 12: Why don't we name this `TableId`?

packages/prop-controllers/src/width/width.test.ts
```diff
@@ -0,0 +1,134 @@
...
+        test: 'unknown',
+      }
+
+      // Act
+      // @ts-expect-error: invalid data
+      const result = getWidthPropControllerDataResponsiveWidthData(data)
+
+      // Assert
+      expect(result).toEqual(data)
+    })
```
migueloller, line 61: See my other comment on weird test cases like this.

packages/prop-controllers/src/width/width.ts
```diff
@@ -0,0 +1,135 @@
...
+ */
+export function Width<T extends WidthOptions>(
+  options: T & WidthOptions = {} as T,
+): WidthDescriptor<WidthPropControllerData, T> {
+  return { type: Types.Width, version: 1, options }
+}
+
+Width.Format = WidthPropControllerFormat
+
+export function getWidthPropControllerDataResponsiveWidthData(
```
migueloller, line 109: I think that some of this stuff is too hardcoded to what the specific control is. For example, in this case, this could be returning just responsive length data.

packages/runtime/src/next/components/tests/page-date-prop-controller.test.tsx
```diff
@@ -0,0 +1,123 @@
...
+} from '../../../utils/tests/element-data-test-test'
+
+describe('Page', () => {
+  test('can render DatePropController v0 data', async () => {
+    // Arrange
+    const dateDefinitionV0: DateDescriptor = {
+      type: Types.Date,
+      options: {},
+    }
+    const dateData = new Date(Date.now()).toISOString()
```
migueloller, line 31: `new Date(Date.now())`? Is this Copilot? Why not just `new Date()`?

packages/runtime/src/next/components/tests/page-date-prop-controller.test.tsx
```diff
@@ -0,0 +1,123 @@
...
+        )
+      }),
+      {
+        type: TestComponentType,
+        label: 'TestComponent',
+        props: {
+          date: DatePropController(),
+        },
+      },
+    )
```
migueloller, line 111: There's a ton of boilerplate here that looks copy pasted from other test cases in this file which makes it hard to follow these tests. Maybe we can extract them into a utility in this file?

migueloller (dismissed):
Nice work man! These e2e tests, specially the ones checking CSS styles, are super nice. Thanks for all the hard work in getting those set up.

I think this is good to go but I've got some minor thoughts and comments here and inline in the PR's code:
- I think we gotta be consistent with file structure. I was dropping `.ts` and `.test.ts` files under `src` but you seem to be making a folder per prop controller and putting things in there. I personally prefer avoiding the extra folder, but don't have a strong preference other than consistency.
- There's _a lot_ of boilerplate on some of these tests and that makes them very hard to follow. I want to encourage us to use these tests, write more, and change them over time but if they're full of boilerplate it's going to be a pain to deal write and read them.

packages/prop-controllers/src/margin/margin.test.ts
fikrikarim: Updated 🍝

packages/prop-controllers/src/padding/padding.test.ts
fikrikarim: Updated 🍝

packages/runtime/src/next/components/tests/page-date-prop-controller.test.tsx
fikrikarim: I didn't use copilot when writing this PR. This was a human mistake. I tried `Date.now().toISOString()` and it didn't work 🤦 

Updated to be:
```ts
new Date().toISOString()
```

packages/prop-controllers/src/width/width.ts
fikrikarim: Yeah, I was thinking about that too. I removed the `ResponsiveWidthData` and just used `ResponsiveLength` everywhere.

packages/prop-controllers/src/padding/padding.test.ts
fikrikarim: I agree that this test case is weird. Let's remove this one for now.

packages/prop-controllers/src/width/width.test.ts
fikrikarim: Removed

packages/prop-controllers/src/margin/margin.test.ts
```diff
@@ -0,0 +1,160 @@
...
+        test: 'unknown',
+      }
+
+      // Act
+      // @ts-expect-error: invalid data
+      const result = getMarginPropControllerDataResponsiveMarginData(data)
+
+      // Assert
+      expect(result).toEqual(data)
+    })
```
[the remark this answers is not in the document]
fikrikarim: Sure. I agree. Let's remove it for now till we have better patterns for handling invalid data.

packages/prop-controllers/src/table/table.ts
fikrikarim: Yeah, I thought about that too. Changed.

packages/prop-controllers/src/table/table.ts
fikrikarim: I agree. Updated to `TableId`.

I have been thinking about the namespace or class since some time ago too 😢

packages/runtime/src/next/components/tests/page-date-prop-controller.test.tsx
fikrikarim: I couldn't find a nice abstraction to extract them yet. For each prop-controller, we need to register different prop-controller, and different custom components because sometimes we're testing the style, while sometimes we're just testing the value.

If you have an idea of how we can extract this into a utility file please let me know. I don't think it'll be hard to migrate them later, as we only have two test cases for each prop-controller.

fikrikarim:
- Yes I noticed that you put the test files under the `src` folder. I don't have a strong opinion either. I can go either way. I made those into a folder just because we will have a lot of prop-controllers.
- I couldn't find a nice abstraction to extract them yet. For each prop-controller, we need to register different prop-controller, and different custom components because sometimes we're testing the style, while sometimes we're just testing the value.
If you have an idea of how we can extract this into a utility file please let me know. I don't think it'll be hard to migrate them later, as we only have two test cases for each prop-controller.

PR #731, opened 2024-04-24
fix(runtime): "function components cannot be given refs" on the builtin `Text` component

agurtovoy:
Bringing `Text` in line with the rest of the builtins.


migueloller (dismissed):
LGTM 👍🏻

PR #734, opened 2024-04-26
Add data type to TextArea, GapX, GapY PropControllers

fikrikarim:


packages/prop-controllers/src/data.ts
```diff
@@ -13,3 +14,15 @@ export const lengthDataSchema = z.object({
 })
 
 export type LengthData = z.infer<typeof lengthDataSchema>
+
+export const gapDataSchema = z.object({
+  value: z.number(),
+  unit: z.literal('px'),
+})
+
+export type GapData = z.infer<typeof gapDataSchema>
```
migueloller, line 23: I would name this `LengthData` and the one above `LengthPercentageData`.

For reference:
- https://developer.mozilla.org/en-US/docs/Web/CSS/length
- https://developer.mozilla.org/en-US/docs/Web/CSS/length-percentage

Basically, `type LengthPercentageData = LengthData | PercentageData`

Although our version of `<length>` doesn't support units other than `px`. We might add that eventually.

migueloller (approved):
LGTM! 👍🏻

packages/prop-controllers/src/data.ts
fikrikarim: We already have `LengthData` and it supports percentage:
image

I'll keep this as `GapData` for now.

PR #735, opened 2024-04-28
fix(runtime): resolve occasional `ERR_INVALID_ARG_TYPE` error in Node.js console output

agurtovoy:
Resolve occasional `TypeError [ERR_INVALID_ARG_TYPE]: The "chunk" argument must be of type string or an instance of Buffer or Uint8Array. Received an instance of Array Buffer` when previewing a site built using Next.js Pages router.

### DIFF
--- a/packages/runtime/src/next/api-handler/handlers/proxy-preview-mode.ts
+++ b/packages/runtime/src/next/api-handler/handlers/proxy-preview-mode.ts
@@ -158,6 +158,6 @@ async function proxyPreviewModeApiRouteHandler(
 
   const arrayBuffer = await response.arrayBuffer()
 
-  res.write(arrayBuffer)
+  res.write(new Uint8Array(arrayBuffer))
   res.end()
 }
--- a/.changeset/tiny-lies-check.md
+++ b/.changeset/tiny-lies-check.md
@@ -1,0 +1,5 @@
+---
+'@makeswift/runtime': patch
+---
+
+Resolve occasional `ERR_INVALID_ARG_TYPE` error when previewing a site built using Next.js Pages router.


agurtovoy:
Note that this only reproduced for me on the first run in a clean env. Subsequent runs in a new terminal, even after clearing the browser cache and removing `node_modules`, `.next` and `dist` dirs, do not report the error, even though the code is executing through the same path. I ended up using Github Codespaces to confirm that my fix does indeed resolve the issue.

packages/runtime/src/next/api-handler/handlers/proxy-preview-mode.ts
```diff
@@ -158,6 +158,6 @@ async function proxyPreviewModeApiRouteHandler(
 
   const arrayBuffer = await response.arrayBuffer()
 
-  res.write(arrayBuffer)
+  res.write(Buffer.from(arrayBuffer))
```
agurtovoy, line 161: Or maybe we should do `new Uint8Array` in case this ever runs in [Edge Runtime](https://nextjs.org/docs/app/api-reference/edge)?

migueloller: Yeah, good point. Is it just a matter if calling the constructor?

migueloller (approved):
This looks good to me. Given your findings I wonder if customers mainly ran into this after running the CLI but never on subsequent runs 🤔

Don’t forget the changeset!

packages/runtime/src/next/api-handler/handlers/proxy-preview-mode.ts
agurtovoy: Yes, `new Uint8Array(arrayBuffer)`. I think `Uint8Array` is just a [typed projection of `ArrayBuffer`](https://javascript.info/arraybuffer-binary-arrays). I'll switch over just in case.

fikrikarim:
Somehow the last push fixed the pipeline issue on the `makeswift-examples-bigcommerce-catalyst`:

https://vercel.com/makeswift/makeswift-examples-bigcommerce-catalyst/C3N8gQ4J4vZWHNDwxgm34XaatsqP

PR #736, opened 2024-04-29
chore: fix package build failure due to parallel execution of `prepare` scripts

agurtovoy:
Switch from `prepare` to [`prepublishOnly`](https://docs.npmjs.com/cli/v10/using-npm/scripts#prepare-and-prepublish), use `turbo` to build the packages so that dependencies are also built as needed.


package.json
```diff
@@ -11,6 +11,6 @@
     "@playwright/test": "^1.41.1",
     "@types/node": "^20.11.11",
     "prettier": "^2.7.1",
-    "turbo": "^1.4.3"
+    "turbo": "^1.13.3"
```
agurtovoy, line 14: Update to the latest stable version of `turbo` so that it can be run from anywhere in the repo, see https://github.com/vercel/turbo/issues/1412

package.json
```diff
@@ -1,5 +1,9 @@
 {
   "private": true,
+  "packageManager": "pnpm@8.15.3",
+  "engines": {
+    "pnpm": "^8.15.0"
```
agurtovoy, line 5: When `corepack` is not enabled, fallback to `npm`'s built-in "engine" enforcement.

.github/workflows/ci.yml
```diff
@@ -33,10 +33,10 @@ jobs:
           cache: pnpm
 
       - name: Install dependencies
-        run: pnpm install --ignore-scripts
+        run: pnpm install --ignore-scripts --frozen-lockfile
```
agurtovoy, line 36: Explicitly request `--frozen-lockfile`, as the [implicit default](https://pnpm.io/cli/install#--frozen-lockfile) turned out to be unreliable.

packages/runtime/package.json
```diff
@@ -116,8 +116,8 @@
     "@emotion/server": "^11.4.0",
     "@emotion/sheet": "^1.0.3",
     "@emotion/utils": "^1.0.0",
-    "@makeswift/next-plugin": "0.3.0",
-    "@makeswift/prop-controllers": "^0.1.0",
+    "@makeswift/next-plugin": "workspace:*",
+    "@makeswift/prop-controllers": "workspace:*",
```
agurtovoy, line 120: Switch to the [`workspace:` protocol](https://pnpm.io/workspaces#workspace-protocol-workspace); the non-workspace references were observed to be pulled from the registry instead of the repo in some cases, even when the local package version and the dependency version range matched exactly.

PR #743, opened 2024-05-02
feat(prop-controllers): `ResponsiveNumber`

agurtovoy:
`ResponsiveNumber` prop + `GapX` prop reworked using the new generic prop infra


package.json
```diff
@@ -14,7 +14,7 @@
     "@changesets/cli": "^2.27.1",
     "@playwright/test": "^1.41.1",
     "@types/node": "^20.11.11",
-    "prettier": "^2.7.1",
+    "prettier": "^3.2.5",
```
agurtovoy, line 17: Update `prettier` for [TS 5.0 support](https://github.com/prettier/prettier/issues/14240).

packages/prop-controllers/src/gap-x/gap-x.ts
```diff
@@ -1,102 +1,16 @@
...
-])
-
-export type GapXPropControllerData = z.infer<
-  typeof gapXPropControllerDataSchema
->
-
-export type GapXOptions = Options<{
-  preset?: ResponsiveGapData
+export const GapX = versionedPropDef<{
+  preset?: unknown
```
agurtovoy, line 6: If present in "raw" options, the `preset` prop will recieve the correct type as a part of the property definition "construction" here.

fikrikarim:
I think we need a changeset here so we can release a new version?

PR #751, opened 2024-05-07
`@makeswift/controls` package

alexhwoods:
- Do we need to set up CI for this, or can it remain unpublished until we need it in the API?
- I do intend to move all controls, but want to pause for a sanity check


packages/runtime/src/controls/index.ts
```diff
@@ -1,19 +1,15 @@
-export * from './checkbox'
```
agurtovoy, line 1: @alexhwoods I think we still want to re-export the controls from here.

fikrikarim:
Is it possible to have them not published? 

Let's say we release a new canary version of the runtime, how could it use the code from the new controls package if it's not published? We can try it, and if we don't need to publish it then it's fine.

packages/runtime/src/controls/color.ts
```diff
@@ -1,39 +1,10 @@
...
-  config: C
-  version?: 1
-}
-
-export function Color<C extends ColorControlConfig>(config = {} as C): ColorControlDefinition<C> {
-  return { type: ColorControlType, config, version: 1 }
-}
 
+// @todo: move this into controls package
 export function copyColorData(
```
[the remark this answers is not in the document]
alexhwoods: That's a good point; that's actually less complicated than i thought it was. Can look into this

migueloller:
> > Is it possible to have them not published?
> > Let's say we release a new canary version of the runtime, how could it use the code from the new controls package if it's not published? We can try it, and if we don't need to publish it then it's fine.
> 
> I'll test this, but I believe it goes into the runtime's node modules when it does install and build

We're probably going to need to publish it because when we do a canary release of the runtime to test things in the app, it should depend on a canary release of `@makeswift/prop-controllers` and `@makeswift/controls`. Whatever setup we have for `@makeswift/prop-controllers` we probably want to do the same.

packages/controls/README.md
```diff
@@ -0,0 +1,3 @@
+# @makeswift/controls
+
+This is for internal use only.
```
migueloller, line 3: This is temporary, right? I was thinking that this package would be properly documented and is effectively just re-exported via `@makeswift/runtime/controls` but we point people here for contributions, documentation, etc.

packages/runtime/src/controls/index.ts
```diff
@@ -13,7 +13,6 @@ export * from './slot'
 export * from './style'
 export * from './style-v2'
 export * from './text-area'
-export * from './text-input'
```
migueloller, line 16: We still need this re-export so that we don't break the current public API

packages/controls/src/index.ts
```diff
@@ -1,3 +1,4 @@
 export * from './common/data-type-key'
 
 export * from './text-input'
+export * from './text-area'
```
migueloller: nit: no newline at end of file, Prettier should be configured so that this is taken care of automatically.

packages/runtime/src/controls/index.ts
```diff
@@ -12,7 +12,6 @@ export * from './shape'
 export * from './slot'
 export * from './style'
 export * from './style-v2'
-export * from './text-area'
```
migueloller, line 15: Still want this export.

packages/runtime/src/controls/color.ts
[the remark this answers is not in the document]
migueloller: My opinion is we should go ahead and move them like we did for `@makeswift/prop-controllers` and we can copy/paste the type definition for now and figure out how to share it later if necessary.

packages/controls/.prettierrc
migueloller, line 1: If we're adding a Prettier config file, we should also install Prettier as a dev dependency of `@makeswift/controls`. It would also be nice to set up an `fmt` script that runs `prettier --write .`. We might want a `.prettierignore` as well to ignore build artifacts.

migueloller: We might also want to set up CI to run `prettier --check .`

packages/controls/src/color.ts
```diff
@@ -1,4 +1,3 @@
-import { match } from 'ts-pattern'
```
migueloller: Hmm, we probably want to set up ESLint to run on CI

migueloller (approved):
I think this is a good start to consolidating controls in a separate package! Left some minor comments on some improvements I think we should make.

packages/controls/src/color.ts
alexhwoods: It does, that's how I caught this

migueloller: Awesome!

PR #765, opened 2024-05-10
feat(prop-controllers): `ResponsiveSelect`

agurtovoy:


packages/prop-controllers/src/gap-x/gap-x.ts
```diff
@@ -1,16 +1,21 @@
...
-  defaultValue?: GapData
-  min?: number
-  max?: number
-  step?: number
-  hidden?: boolean
-}>()(Types.GapX, responsiveGapDataSchema, {
-  version: 1,
-  dataKey: 'prop-controllers::gap-x::v1',
-})
+export const GapX = versionedPropDef(
```
agurtovoy, line 5: Here and below, adjusting existing prop controller definitions for the revised `versionedPropDef` API, which takes options wrapped in `typeArg`.

packages/prop-controllers/src/responsive-select/responsive-select.ts
```diff
@@ -0,0 +1,49 @@
...
+  /**
+   * @deprecated Prop controllers are deprecated. Use `@makeswift/runtime/controls` instead.
+   */
+  <_T extends string, T extends _T, U extends T>(
+    options: Options<RawSelectOptions<T, U>>,
+  ): VersionedDescriptor<
+    typeof discriminator,
+    typeof Types.ResponsiveSelect,
+    Options<RawSelectOptions<T, U>>
+  >
```
agurtovoy, line 40: Typed `ResponsiveSelect` "constructor".

packages/prop-controllers/src/data.ts
```diff
@@ -63,3 +66,10 @@ export const responsiveNumberValueSchema = createResponsiveValueSchema(
 )
 
 export type ResponsiveNumberValue = z.infer<typeof responsiveNumberValueSchema>
+
+export const responsiveSelectValueSchema = createResponsiveValueSchema(
+  z.string(),
+)
```
migueloller, line 72: Since the select prop controller is generic over options, would it make sense to have a `createResponsiveSelectValueSchema` that takes in options?

packages/prop-controllers/src/gap-x/gap-x.ts
```diff
@@ -1,16 +1,21 @@
...
+  },
+  typeArg<{
+    preset?: unknown
+    label?: string
+    defaultValue?: GapData
+    min?: number
+    max?: number
+    step?: number
+    hidden?: boolean
+  }>(),
```
migueloller, line 20: This is a very interesting API. Particularly because it's a type-only construct. I've had some bad experience in TypeScript when I've had to manually pass types on generics or have had to cast types. Particularly around type safety and discrepancies between run time and compile time. I wonder if there are any gotchas with this API and if we could enforce it at run time by taking an options Zod schema.

packages/prop-controllers/src/responsive-opacity/responsive-opacity.ts
```diff
@@ -10,11 +10,12 @@ export type ResponsiveOpacityValue = z.infer<
...
 
-export const ResponsiveOpacity = versionedPropDef<{}>()(
+export const ResponsiveOpacity = versionedPropDef(
   Types.ResponsiveOpacity,
   responsiveOpacityValueSchema,
   {
     version: 1,
     dataKey: 'prop-controllers::responsive-opacity::v1',
   },
+  typeArg<{}>(),
```
migueloller, line 20: The `{}` type is effectively `any` in TypeScript. For an empty options object, should we be using `Record<string, never>`?

See https://www.totaltypescript.com/the-empty-object-type-in-typescript.

packages/prop-controllers/src/responsive-select/responsive-select.test.ts
```diff
@@ -0,0 +1,186 @@
...
+        version: 1,
+        options: {
+          options: [
+            { value: 'solid', label: 'Solid' },
+            { value: 'dashed', label: 'Dashed' },
+            { value: 'dotted', label: 'Dotted' },
+            { value: 'blended', label: 'Blended' },
+          ],
+        },
+      } satisfies Descriptor<typeof ResponsiveSelect>
```
migueloller, line 133: I'm curious about your criteria for using `satisfies`, or giving a type to the binding, or using the descriptor constructor and letting TypeScript infer it.

packages/prop-controllers/src/responsive-select/responsive-select.ts
migueloller: This is very interesting. So this allows an associated type that is generic on the "constructor" of the definition (i.e., the options)?

packages/prop-controllers/src/versioned.ts
migueloller, line 1: This is some very interesting TypeScript. I would love to go over it together so that I can understand in better detail how it all works.

migueloller (approved):
I finally got around to reviewing this one! Sorry for the delay 😞 

This is great. I've got a ton of questions and would love to go over it together at some point.

packages/prop-controllers/src/responsive-opacity/responsive-opacity.ts
agurtovoy: Good catch, it was actually `Record<string, never>` in the original code (https://github.com/makeswift/makeswift/pull/754/files#diff-1ecb7c462ca01ab9ac5cef93354d548cc023a085d175fcb3697ade52c2e831ffL183), but got mangled during transition to `versionedPropDef`.

packages/prop-controllers/src/data.ts
agurtovoy: Right now we are using schemas only to validate values in prop controllers that are created based on deserialized control definitions, and the finer-typed options do not survive control serialization.

packages/prop-controllers/src/gap-x/gap-x.ts
agurtovoy: We could certainly validate options at run-time if we want to. My understanding was that options are specified exclusively in the source code, though, and thus compile-time validation is sufficient.

packages/prop-controllers/src/responsive-select/responsive-select.test.ts
agurtovoy: I could have used a descriptor constructor here, the main reason for why the descriptor is manually spelled out here is because our tests are largely copy & pasted, and the descriptor was spelled out in the original tests that this is based on. That said, I have considered using a constructor here when I was reworking these, but we use the constructor elsewhere in these tests, and so I thought a test with an explicit descriptor literal with a `satisfies` would actually be a good thing to keep.

As far `satisfies` vs giving an explicit type to the variable, I was mostly just testing how the type is deduced. I do prefer to use `satisfies` for object literals by default.

packages/prop-controllers/src/responsive-select/responsive-select.ts
agurtovoy: Effectively this "types" the existing constructor implementation, which otherwise is the same for all descriptors. `Descriptor<typeof ResponsiveSelect>` or `OptionsType<typeof ResponsiveSelect>` will still return a non-generic, type-erased descriptor/options type, but `ResponsiveSelect(...)` constructor will return a finely-typed descriptor.

PR #769, opened 2024-05-20
fix: global element not selectable after it's blurred

fikrikarim:
This bug only happened on component that doesn't use forwardRef.

This happened because we're relying on FindDomNode when the component does not use forwardRef. Unfortunately, for ElementReference, we render the ElementData, instead of Element, which has the FindDomNode fallback.

This commit fix that issue, and we render Element instead of ElementData.

### DIFF
--- a/.changeset/curly-dragons-train.md
+++ b/.changeset/curly-dragons-train.md
@@ -1,0 +1,5 @@
+---
+"@makeswift/runtime": patch
+---
+
+Fix global component cannot be selected after it is blurred.
--- a/packages/runtime/src/runtimes/react/components/ElementReference.tsx
+++ b/packages/runtime/src/runtimes/react/components/ElementReference.tsx
@@ -8,7 +8,7 @@ import {
   ElementReference as ReactPageElementReference,
 } from '../../../state/react-page'
 import { FallbackComponent } from '../../../components/shared/FallbackComponent'
-import { ElementData } from './ElementData'
+import { Element } from './Element'
 import { Document } from './Document'
 import { DisableRegisterElement } from '../hooks/use-disable-register-element'
 
@@ -59,7 +59,8 @@ export const ElementReference = memo(
           <Document document={elementReferenceDocument} ref={ref} />
         ) : (
           <DisableRegisterElement.Provider value={true}>
-            <ElementData elementData={globalElementData} ref={ref} />
+            {/* We render Element instead of ElementData because we rely on the FindDomNode */}
+            <Element element={globalElementData} ref={ref} />
           </DisableRegisterElement.Provider>
         )}
       </DocumentCyclesContext.Provider>


migueloller (approved):
This is a nice fix but I wonder if it's going to have unintended side effects. Did you try, for example, to have two of the same global element in a page? I remember when I set this up I had to be very careful as to what each component did. Does the `Element` component just get the DOM node? Maybe this is a big enough change we want to mark it as a `minor`?

fikrikarim:
@migueloller I was thinking the same thing. The side effects of the `Element` component are the ref-handling logic and the component registration. We need the ref-handling effects, and we don't want the component registration. The component registration effect is disabled with `isRegisterElementDisabled`.

Here's the `Element.tsx` code. Anything that might cause an issue here?

```ts
export const Element = memo(
  forwardRef(function Element(
    { element }: ElementProps,
  ): JSX.Element {
    const elementKey = element.key
    const useFindDomNodeRef = useRef(true)
    const imperativeHandleRef = useRef(new ElementImperativeHandle())
    const findDomNodeCallbackRef = useCallback((current: (() => Element | Text | null) | null) => {
      if (useFindDomNodeRef.current === true) {
        imperativeHandleRef.current.callback(() => current?.() ?? null)
      }
    }, [])
    const elementCallbackRef = useCallback((current: unknown | null) => {
      useFindDomNodeRef.current = false

      imperativeHandleRef.current.callback(() => current)
    }, [])
    const isRegisterElementDisabled = useDisableRegisterElement()

    useImperativeHandle(ref, () => imperativeHandleRef.current, [])

    useEffect(() => {
      if (documentKey == null || isRegisterElementDisabled) return

      return dispatch(
        registerComponentHandleEffect(documentKey, elementKey, imperativeHandleRef.current),
      )
    }, [dispatch, documentKey, elementKey, isRegisterElementDisabled])

    useEffect(() => {
      if (documentKey == null || isRegisterElementDisabled) return

      return dispatch(mountComponentEffect(documentKey, elementKey))
    }, [dispatch, documentKey, elementKey, isRegisterElementDisabled])

    return (
      <FindDomNode ref={findDomNodeCallbackRef}>
        {isElementReference(element) ? (
          <ElementReference key={elementKey} ref={elementCallbackRef} elementReference={element} />
        ) : (
          <ElementData key={elementKey} ref={elementCallbackRef} elementData={element} />
        )}
      </FindDomNode>
    )
  }),
)
```

We're calling the `isRegisterElementDisabled` on the ElementReference:
```tsx
<DisableRegisterElement.Provider value={true}>
  <Element element={globalElementData} ref={ref} />
</DisableRegisterElement.Provider>
```

Another alternative is to copy the `FindDomNode` code to the `ElementReference`.

I'll do more testing later.

migueloller:
Nothing immediately jumps out other than element registration.

My feeling right now is that this is hitting a threshold of complexity where it should probably be re-thought and refactored. I'll leave it up to you whether you want to do that or not.

fikrikarim:
> My feeling right now is that this is hitting a threshold of complexity where it should probably be re-thought and refactored. I'll leave it up to you whether you want to do that or not.

I agree with this. I think refactoring the code should be a separate task so it doesn't block the bug fix.

PR #772, opened 2024-05-22
feat!(makeswift): remove integrate existing app

migueloller:
Our CLI would fail to integrate existing Next.js apps most of the time. Instead of leaving the sub par experience as a feature of the CLI, often resulting in users being confused, we're going to remove it in favor of manually following our docs for now. Later on we might be able to provide the user experience we desire when using the Makeswift CLI in an existing Next.js app.


alexhwoods:
https://youtu.be/7qH4qyi1-Ys

alexhwoods (approved):

PR #776, opened 2024-06-07
fix: uncaught client exception on incomplete, protocol-only links

agurtovoy:
Workaround for https://github.com/vercel/next.js/issues/66650


migueloller (approved):
Nice catch!

PR #780, opened 2024-06-12
feat: introduce v4 `getPages` endpoint

arvinpoddar:
Updates the sample applications to use the new paginated v4 `getPages` endpoint. Also adds a deprecation notice to `getSitemap`.

### DIFF
--- a/.changeset/little-ads-return.md
+++ b/.changeset/little-ads-return.md
@@ -1,0 +1,106 @@
+---
+"@makeswift/runtime": minor
+---
+
+BREAKING: The latest upgrade to the Makeswift client `getPages` method
+introduces sorting, path filtering, and pagination. This method was not
+previously paginated - in order to get all your pages, you may now use our
+`toArray` pagination helper method, which will automatically paginate through
+all results and aggregate them into an array:
+
+```tsx
+import { client } from '@/makeswift/client'
+import { MakeswiftPage } from '@makeswift/runtime/next'
+
+async function getAllPages(): Promise<MakeswiftPage[]> {
+  return await client.getPages().toArray()
+}
+```
+
+`getPages` now returns an instance of `IterablePaginationResult`, a decorated
+async iterator which includes methods `.map` and `.filter`, in addition to
+`.toArray`, mentioned above.
+
+This change also deprecates the client `getSitemap` method, with the
+recommendation that sitemaps should now be generated using data returned from
+`getPages`. Note that the deprecation of `getSitemap` now involves the host
+being responsible for the construction of page paths in the sitemap (either with
+domain or path based localization). Below is an example that uses path-based
+localization with the `next-sitemap` library:
+
+```tsx pages/sitemap.xml.tsx
+import { getServerSideSitemapLegacy } from 'next-sitemap'
+import { MakeswiftPage } from '@makeswift/runtime/next'
+import { client } from '@makeswift/client'
+
+const DOMAIN = 'https://example.com'
+const DEFAULT_PRIORITY = 0.75
+const DEFAULT_FREQUENCY = 'hourly'
+
+function pageToSitemapItem(page: MakeswiftPage) {
+  const pageUrl = new URL(page.path, DOMAIN)
+  return {
+    loc: pageUrl.href,
+    lastmod: page.createdAt,
+    changefreq: page.sitemapFrequency ?? DEFAULT_FREQUENCY,
+    priority: page.sitemapPriority ?? DEFAULT_PRIORITY,
+    alternateRefs: page.localizedVariants.map(variant => {
+      const localizedPath = `/${variant.locale}/${variant.path}`
+      const localizedPageUrl = new URL(localizedPath, DOMAIN)
+      return {
+        hreflang: variant.locale,
+        href: localizedPageUrl.href,
+      }
+    }),
+  }
+}
+
+export async function getServerSideProps(context) {
+  const sitemap = client
+    .getPages()
+    .filter(page => !page.excludedFromSearch)
+    .map(page => pageToSitemapItem(page))
+    .toArray()
+
+  return getServerSideSitemapLegacy(context, sitemap)
+}
+
+export default function Sitemap() {}
+
+```
+
+Here's another example for Next.js's App Router built-in support for sitemaps:
+
+```ts app/sitemap.ts
+
+import { MetadataRoute } from 'next'
+import { MakeswiftPage } from '@makeswift/runtime/dist/types/next'
+import { client } from '@/lib/makeswift/client'
+
+type NextSitemapItem = MetadataRoute.Sitemap[number]
+
+const DOMAIN = 'https://example.com'
+const DEFAULT_PRIORITY = 0.75
+const DEFAULT_FREQUENCY = 'hourly'
+
+function pageToSitemapEntry(page: MakeswiftPage): NextSitemapItem {
+  const pageUrl = new URL(page.path, DOMAIN)
+  return {
+    url: pageUrl.href,
+    lastModified: page.createdAt,
+    changeFrequency: page.sitemapFrequency ?? DEFAULT_FREQUENCY,
+    priority: page.sitemapPriority ?? DEFAULT_PRIORITY,
+  }
+}
+
+export default async function sitemap(): Promise<MetadataRoute.Sitemap> {
+  return client
+    .getPages()
+    .filter(p => p.path != null)
+    .map(page => pageToSitemapEntry(page))
+    .toArray()
+}
+```
+
+BREAKING: The exported `MakeswiftPage` type now includes several more data
+fields from the Makeswift page.
--- a/apps/nextjs-app-router/src/app/[lang]/[[...path]]/page.tsx
+++ b/apps/nextjs-app-router/src/app/[lang]/[[...path]]/page.tsx
@@ -3,19 +3,27 @@ import '@/makeswift/components'
 import { getSiteVersion } from '@makeswift/runtime/next/server'
 import { notFound } from 'next/navigation'
 import { Page as MakeswiftPage } from '@makeswift/runtime/next'
-import { locales } from '@/localization'
 
 type ParsedUrlQuery = { lang: string; path?: string[] }
 
 export async function generateStaticParams() {
-  const pages = await client.getPages()
 
-  return pages.flatMap((page) =>
-    locales.map((locale) => ({
-      path: page.path.split('/').filter((segment) => segment !== ''),
-      lang: locale,
+  const pages = await client.getPages().toArray()
+
+  return pages.flatMap((page) => [
+    {
+      params: {
+        path: page.path.split('/').filter((segment) => segment !== ''),
+        lang: page.locale,
+      },
+    },
+    ...page.localizedVariants.map((variant) => ({
+      params: {
+        path: variant.path.split('/').filter((segment) => segment !== ''),
+        lang: variant.locale,
+      },
     })),
-  )
+  ])
 }
 
 export default async function Page({ params }: { params: ParsedUrlQuery }) {
--- a/packages/runtime/src/next/client.ts
+++ b/packages/runtime/src/next/client.ts
@@ -39,11 +39,73 @@ import {
   Data,
 } from '../state/react-page'
 import { getMakeswiftSiteVersion, MakeswiftSiteVersion } from './preview-mode'
+import { toIterablePaginationResult } from './utils/pagination'
 
-export type MakeswiftPage = {
-  id: string
-  path: string
+const makeswiftPageResultSchema = z.object({
+  id: z.string(),
+  path: z.string(),
+  title: z.string().nullable(),
+  description: z.string().nullable(),
+  canonicalUrl: z.string().nullable(),
+  socialImageUrl: z.string().nullable(),
+  sitemapPriority: z.number().nullable(),
+  sitemapFrequency: z
+    .enum(['always', 'hourly', 'daily', 'weekly', 'monthly', 'yearly', 'never'])
+    .nullable(),
+  createdAt: z.string(),
+  updatedAt: z.string(),
+  publishedAt: z.string().nullable(),
+  isOnline: z.boolean().nullable(),
+  excludedFromSearch: z.boolean().nullable(),
+  locale: z.string(),
+  localizedVariants: z.array(
+    z.object({
+      locale: z.string(),
+      path: z.string(),
+    }),
+  ),
+})
+
+const makeswiftGetPagesResultAPISchema = z.object({
+  data: z.array(makeswiftPageResultSchema),
+  hasMore: z.boolean(),
+})
+
+const makeswiftGetPagesParamsSchema = z.object({
+  limit: z.number().optional(),
+  after: z.string().optional(),
+  sortBy: z.enum(['title', 'path', 'createdAt', 'updatedAt', 'publishedAt']).optional(),
+  sortDirection: z.enum(['asc', 'desc']).optional(),
+  includeOffline: z.boolean().optional(),
+  pathPrefix: z.string().optional(),
+  locale: z.string().optional(),
+})
+
+function getPagesQueryParams({
+  limit = 20,
+  after,
+  sortBy,
+  sortDirection,
+  includeOffline,
+  pathPrefix,
+  locale,
+}: GetPagesParams): URLSearchParams {
+  const params = new URLSearchParams()
+
+  if (limit != null) params.set('limit', limit.toString())
+  if (after != null) params.set('after', after)
+  if (sortBy != null) params.set('sortBy', sortBy)
+  if (sortDirection != null) params.set('sortDirection', sortDirection)
+  if (includeOffline != null) params.set('includeOffline', includeOffline.toString())
+  if (pathPrefix != null) params.set('pathPrefix', pathPrefix)
+  if (locale != null) params.set('locale', locale)
+
+  return params
 }
+
+type GetPagesParams = z.infer<typeof makeswiftGetPagesParamsSchema>
+export type MakeswiftPage = z.infer<typeof makeswiftPageResultSchema>
+export type MakeswiftGetPagesResult = z.infer<typeof makeswiftGetPagesResultAPISchema>
 
 export type MakeswiftPageDocument = {
   id: string
@@ -201,22 +263,31 @@ export class Makeswift {
     return response
   }
 
-  async getPages({
+  private getPagesInternal = async ({
     siteVersion = MakeswiftSiteVersion.Live,
+    ...params
   }: {
     siteVersion?: MakeswiftSiteVersion
-  } = {}): Promise<MakeswiftPage[]> {
-    const response = await this.fetch(`v3/pages`, siteVersion)
+  } & GetPagesParams = {}): Promise<MakeswiftGetPagesResult> => {
+    const queryParams = getPagesQueryParams(params)
 
+    const response = await this.fetch(`v4/pages?${queryParams.toString()}`, siteVersion)
     if (!response.ok) {
       console.error('Failed to get pages', await response.json())
       throw new Error(`Failed to get pages with error: "${response.statusText}"`)
     }
 
-    const json = await response.json()
-
-    return json
+    const result = await response.json()
+    const parsedResponse = makeswiftGetPagesResultAPISchema.safeParse(result)
+    if (!parsedResponse.success) {
+      throw new Error(
+        `Failed to parse getPages response: ${parsedResponse.error.errors.join(', ')}`,
+      )
+    }
+    return parsedResponse.data
   }
+
+  getPages = toIterablePaginationResult(this.getPagesInternal)
 
   async getPage(
     pathname: string,
@@ -636,6 +707,10 @@ export class Makeswift {
     return result.table
   }
 
+  /**
+   * @deprecated `getSitemap` is deprecated. We recommend constructing a sitemap
+   * using data from `getPages` instead.
+   */
   async getSitemap({
     limit = 50,
     after,
--- a/packages/runtime/src/next/utils/pagination.ts
+++ b/packages/runtime/src/next/utils/pagination.ts
@@ -1,0 +1,52 @@
+import { AsyncIterableExt } from './async-iterable'
+
+type PaginationResult<T> = { data: T[]; hasMore: boolean }
+type IterablePaginationResult<T> = Promise<PaginationResult<T>> & AsyncIterableExt<T>
+
+type PaginationFunction<Options, R extends { id: any }> = Options extends {
+  after?: string
+}
+  ? (options: Options) => Promise<PaginationResult<R>>
+  : never
+
+type IterableMethod<Options, R> = Partial<Options> extends Options
+  ? (options?: Options) => IterablePaginationResult<R>
+  : (options: Options) => IterablePaginationResult<R>
+
+export function toIterablePaginationResult<Options, R extends { id: string }>(
+  fn: PaginationFunction<Options, R>,
+): IterableMethod<Options, R> {
+  return (options?: Options): IterablePaginationResult<R> => {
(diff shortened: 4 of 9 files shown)


arvinpoddar:
In addition to these changes, do I need to update the `basic-typescript-example` and the others, or can I only do that after the new version has been released?

migueloller:
> In addition to these changes, do I need to update the `basic-typescript-example` and the others, or can I only do that after the new version has been released?

You'll need to wait until it's released so you can update the dependency

.changeset/little-ads-return.md
```diff
@@ -0,0 +1,33 @@
...
+import { client } from '@/makeswift/client'
+import { MakeswiftPageResult } from '@makeswift/runtime/next'
+
+async function getAllPages(): Promise<MakeswiftPageResult[]> {
+  let hasMorePages = true
+  let after: string | undefined = undefined
+  const pages: MakeswiftPageResult[] = []
+
+   do {
+    const { data, hasMore } = await client.getPages({ after })
```
agurtovoy, line 20: I think the awkwardness of this example probably calls for the async iterator protocol @migueloller has been advocating for.

.changeset/little-ads-return.md
```diff
@@ -0,0 +1,33 @@
...
+    hasMorePages = hasMore
+    after = data.at(-1)?.id
+  } while (hasMorePages)
+
+  return pages
+}
+
+This change also deprecates the client `getSitemap` method, with the
+recommendation that sitemaps should now be generated using data returned from
+`getPages`.
```
migueloller, line 31: A link to the upgrade guide or even a quick example inline showing how to migrate would be super helpful!

packages/runtime/src/next/client.ts
```diff
@@ -40,9 +40,25 @@ import {
...
+  updatedAt: string
+  publishedAt: string | null
+  isOnline: boolean | null
+  excludedFromSearch: boolean | null
+  locale: string
+  localizedVariants: {
+    locale: string
+    path: string
+  }[]
 }
```
migueloller, line 62: How do you guys feel about creating a Zod schema so that we can verify the response at runtime?

packages/runtime/src/next/client.ts
```diff
@@ -636,6 +675,10 @@ export class Makeswift {
     return result.table
   }
 
+  /**
+   * @deprecated `getSitemap` is deprecated. We recommend constructing a sitemap
+   * using data from `getPages` instead.
+   */
```
migueloller, line 681: I wonder if we should go ahead and remove this method 🤔 

I know we've talked about deprecating it but since there's feature parity with the new `getPages` and it's a new minor, it's probably ok to remove it?

packages/runtime/src/next/index.tsx
```diff
@@ -1,6 +1,11 @@
 export type { PageProps } from './components/page'
 export { Page } from './components/page'
-export type { MakeswiftPage, MakeswiftPageDocument, MakeswiftPageSnapshot, Sitemap } from './client'
```
migueloller, line 3: Seems like we're making a breaking _type_ change here. Maybe we should call that out in the changeset, upgrade guide, etc.?

migueloller (approved):
Looking good! Can't wait to see this with the new async iterable pattern ❤️

.changeset/little-ads-return.md
```diff
@@ -0,0 +1,73 @@
...
+        }
+      }
+    }
+  })
+  .toArray()
+
+  return getServerSideSitemapLegacy(context, sitemap)
+}
+
+export default function Sitemap() {}
```
agurtovoy, line 67: I think this example is broken, need to double-check. Also, I'd move all of the prop mapping minutiae into a separate local function (`pageToSitemapItem`) so that this one can be just (untested):

```typescript
export async function getServerSideProps(context) {
  const sitemap = client.getPages()
    .filter(page => !page.excludedFromSearch)
    .map(pageToSitemapItem)
    .toArray()

  return getServerSideSitemapLegacy(context, sitemap)
}
```

packages/runtime/src/next/client.ts
```diff
@@ -201,23 +230,51 @@ export class Makeswift {
...
+    } = args
+
+    const url = new URL('v4/pages', this.apiOrigin)
+    url.searchParams.set('limit', limit.toString())
+    if (after != null) url.searchParams.set('after', after)
+    if (includeOffline != null) url.searchParams.set('includeOffline', String(includeOffline))
+    if (sortBy != null) url.searchParams.set('sortBy', sortBy)
+    if (sortDirection != null) url.searchParams.set('sortDirection', sortDirection)
+    if (pathPrefix != null) url.searchParams.set('pathPrefix', pathPrefix)
+    if (locale != null) url.searchParams.set('locale', locale)
```
agurtovoy, line 263: I'd consider

```typescript
const query = getPagesQueryParams(args)
const response = await this.fetch('v4/pages?' + query, args.siteVersion)
```

where `getPagesQueryParams`  takes the args and returns `URLSearchParams`.

.changeset/little-ads-return.md
```diff
@@ -0,0 +1,70 @@
...
+
+```
+
+BREAKING: The exported `MakeswiftPage` type has been replaced with
+`MakeswiftPageResult`, which includes several more data fields from the
+Makeswift page.
```
agurtovoy, line 70: Do we need to rename it? `MakeswiftPageResult` is kind of an awkward name.

arvinpoddar: Do you have any suggestions? I wanted to be more specific than `Page` or `MakeswiftPage` to avoid collision with the component of the same/similar name that renders page snapshots.

packages/runtime/src/next/client.ts
agurtovoy: Note that I'm not suggesting factoring out `getPagesQueryParams` for reuse in any other place, but rather as a way to "tidying up" the implementation here. Basically we have 20 lines of code that are dedicated to constructing the query params, inlined and mixed in with the much more compact and higher-level code that does the work of calling the endpoint and transforming the result.

.changeset/little-ads-return.md
agurtovoy: I'd just keep it as `MakeswiftPage`, since it's an established name we have and it fits with `MakeswiftPageDocument` and `MakeswiftPageSnapshot`. I think if we want to revise our naming convention/namespacing with regards to API types vs components, we need to do it wholesale.

packages/runtime/src/next/client.ts
agurtovoy: Here are a couple of links on the topic: 
* https://medium.com/@brooknovak/single-level-of-abstraction-d7fe8aa5c813
* https://dev.to/bornfightcompany/why-you-shouldn-t-mix-levels-of-abstraction-in-your-functions-332f

apps/nextjs-app-router/src/app/[lang]/[[...path]]/page.tsx
```diff
@@ -3,19 +3,27 @@ import '@/makeswift/components'
...
-  return pages.flatMap((page) =>
-    locales.map((locale) => ({
-      path: page.path.split('/').filter((segment) => segment !== ''),
-      lang: locale,
+  const pages = await client.getPages().toArray()
+
+  return pages.flatMap((page) => [
+    {
+      params: {
+        path: page.path.split('/').filter((segment) => segment !== ''),
```
Comment (agurtovoy) [inline apps/nextjs-app-router/src/app/[lang]/[[...path]]/page.tsx:16]: What is `page.path.split('/').filter((segment) => segment !== '')` for?

packages/runtime/src/next/client.ts
migueloller: Actually, maybe Zod can take care of this for us? Is it possible for it to exclude `undefined` keys?

```ts
new URLSearchParams(serachParamsSchema.parse(params))
```

apps/nextjs-app-router/src/app/[lang]/[[...path]]/page.tsx
```diff
@@ -2,20 +2,39 @@ import { client } from '@/makeswift/client'
...
+        lang: page.locale,
+      },
+    },
+    ...page.localizedVariants.map((variant) => ({
+      params: {
+        path: variant.path.split('/').filter((segment) => segment !== ''),
+        lang: variant.locale,
+      },
```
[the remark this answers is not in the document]
Comment (migueloller) [reply apps/nextjs-app-router/src/app/[lang]/[[...path]]/page.tsx:24]: I think this is better since there could be certain variants that are offline and so the cartesian product might include 404s that we'd otherwise not want

.changeset/little-ads-return.md
migueloller: I agree on doing this wholesale at another time and keeping it as `MakeswiftPage` for now

.changeset/little-ads-return.md
```diff
@@ -0,0 +1,70 @@
...
+  return await client.getPages().toArray()
+}
+```
+
+This change also deprecates the client `getSitemap` method, with the
+recommendation that sitemaps should now be generated using data returned from
+`getPages`. Note that the deprecation of `getSitemap` now involves the host
+being responsible for the construction of page paths in the sitemap (either with
+domain or path based localization). Below is an example that uses path-based
+localization with the `next-sitemap` library:
```
migueloller, line 25: Should we show the example using App Router and its native support for sitemaps?

.changeset/little-ads-return.md
```diff
@@ -0,0 +1,70 @@
...
+    }),
+  }
+}
+
+export async function getServerSideProps(context) {
+  const sitemap = client
+    .getPages()
+    .filter(page => !page.excludedFromSearch)
+    .map(page => pageToSitemapItem(page))
+    .toArray()
```
migueloller, line 59: Does this even work? You can't use `.filter` on a promise or async iterable, right?

apps/nextjs-app-router/src/app/[lang]/[[...path]]/page.tsx
```diff
@@ -3,19 +3,27 @@ import '@/makeswift/components'
...
 type ParsedUrlQuery = { lang: string; path?: string[] }
 
 export async function generateStaticParams() {
-  const pages = await client.getPages()
 
-  return pages.flatMap((page) =>
-    locales.map((locale) => ({
-      path: page.path.split('/').filter((segment) => segment !== ''),
-      lang: locale,
+  const pages = await client.getPages().toArray()
```
Comment (migueloller) [inline apps/nextjs-app-router/src/app/[lang]/[[...path]]/page.tsx:11]: I'm ok with leaving this like this for now but I don't think we should recommend listing all pages in general. Users should be getting a subset of their pages that are important. Once we have a good example of this we should update our docs so that we're not steering users in the what I think is the wrong direction here

packages/runtime/src/next/client.ts
```diff
@@ -39,11 +39,40 @@ import {
...
+    }),
+  ),
+})
 
-export type MakeswiftPage = {
-  id: string
-  path: string
-}
+const makeswiftGetPagesResultAPISchema = z.object({
+  data: z.array(makeswiftPageResultSchema),
```
migueloller, line 70: Did we have other candidates for the name of this field other than `data`?

packages/runtime/src/next/client.ts
```diff
@@ -201,23 +230,51 @@ export class Makeswift {
...
+    const url = new URL('v4/pages', this.apiOrigin)
+    url.searchParams.set('limit', limit.toString())
+    if (after != null) url.searchParams.set('after', after)
+    if (includeOffline != null) url.searchParams.set('includeOffline', String(includeOffline))
+    if (sortBy != null) url.searchParams.set('sortBy', sortBy)
+    if (sortDirection != null) url.searchParams.set('sortDirection', sortDirection)
+    if (pathPrefix != null) url.searchParams.set('pathPrefix', pathPrefix)
+    if (locale != null) url.searchParams.set('locale', locale)
+
+    const response = await this.fetch(url.pathname + url.search, siteVersion)
```
migueloller, line 265: Why are we concatenating URL parts here? We can use `url.href` or `url.toString()`, or better yet, just pass the URL itself to `fetch`, it's supported. Or does `this.fetch` only support relative URLs as a string?

packages/runtime/src/next/client.ts
```diff
@@ -201,23 +230,51 @@ export class Makeswift {
...
     if (!response.ok) {
       console.error('Failed to get pages', await response.json())
       throw new Error(`Failed to get pages with error: "${response.statusText}"`)
     }
 
-    const json = await response.json()
-
-    return json
+    const result = await response.json()
+    return makeswiftGetPagesResultAPISchema.parse(result)
```
migueloller, line 273: I think we should be proactive about error handling here. i.e., do `safeParse` and if it doesn't parse throw an error ourselves, not whatever Zod throws

packages/runtime/src/next/client.ts
```diff
@@ -201,23 +230,51 @@ export class Makeswift {
     return response
   }
 
-  async getPages({
-    siteVersion = MakeswiftSiteVersion.Live,
-  }: {
-    siteVersion?: MakeswiftSiteVersion
-  } = {}): Promise<MakeswiftPage[]> {
-    const response = await this.fetch(`v3/pages`, siteVersion)
+  private getPagesInternal = async (
```
migueloller, line 233: How come we changed this from a method to a field with an arrow function?

packages/runtime/src/next/client.ts
```diff
@@ -201,23 +230,51 @@ export class Makeswift {
...
     }
 
-    const json = await response.json()
-
-    return json
+    const result = await response.json()
+    return makeswiftGetPagesResultAPISchema.parse(result)
   }
 
+  getPages = toIterablePaginationResult(this.getPagesInternal)
```
migueloller, line 276: Is it possible to have `getPages` be a method or do we have to settle for a class field if we're going to decorate it like this?

.changeset/little-ads-return.md
migueloller: I see now that we're implementing these methods. This is very cool—we should document this, though, as this isn't standard and users won't know and it will be weird to see it used this way without us explaining the interface we're providing.

packages/runtime/src/next/client.ts
```diff
@@ -201,23 +230,51 @@ export class Makeswift {
...
+      siteVersion?: MakeswiftSiteVersion
+      limit?: number
+      after?: string
+      sortBy?: 'title' | 'path' | 'createdAt' | 'updatedAt' | 'publishedAt'
+      sortDirection?: 'asc' | 'desc'
+      includeOffline?: boolean
+      pathPrefix?: string
+      locale?: string
+    } = {},
+  ) => {
```
migueloller, line 244: We should define function return types in addition to argument types to avoid TypeScript implicitly setting a return type

apps/nextjs-app-router/src/app/[lang]/[[...path]]/page.tsx
Comment (arvinpoddar) [reply apps/nextjs-app-router/src/app/[lang]/[[...path]]/page.tsx:16]: I assume this is to avoid empty path segments in the parsed outputs, which is important given the leading slash. When you do `"/test".split("/")`, for example, the first member of the array is an empty string. Similarly, doing `"/".split("/")` results in `["", ""]`

packages/runtime/src/next/client.ts
```diff
@@ -201,23 +230,51 @@ export class Makeswift {
...
+    const {
+      siteVersion = MakeswiftSiteVersion.Live,
+      limit = 20,
+      after,
+      includeOffline,
+      sortBy,
+      sortDirection,
+      pathPrefix,
+      locale,
+    } = args
```
migueloller, line 254: Why aren't we destructuring arguments directly in the arguments? Do we need this intermediate `args` variable?

packages/runtime/src/next/utils/pagination.ts
```diff
@@ -0,0 +1,54 @@
...
+  after?: string
+}
+  ? (options: Options) => Promise<PaginationResult<R>>
+  : never
+
+type IterableMethod<Options, R> = Partial<Options> extends Options
+  ? (options?: Options) => IterablePaginationResult<R>
+  : (options: Options) => IterablePaginationResult<R>
+
+export function toIterablePaginationResult<Options, R extends { id: any }>(
```
migueloller, line 16: This function is a bit tricky to follow but I'm assuming all the type inference works as expected? Also, is there a reason not to make `id` be of type `string` instead of `any`?

packages/runtime/src/next/utils/pagination.ts
```diff
@@ -0,0 +1,54 @@
...
+  fn: PaginationFunction<Options, R>,
+): IterableMethod<Options, R> {
+  return (options?: Options): IterablePaginationResult<R> => {
+    return (() => {
```
migueloller, line 20: Is the IIFE necessary here?

packages/runtime/src/next/client.ts
arvinpoddar: We considered `results` and `pages`. Ultimately, we went with `data` since it appears to be a standard for paginated APIs (Stripe currently uses `data` in their response: : https://docs.stripe.com/api/pagination)

.changeset/little-ads-return.md
arvinpoddar: Yep, the documentation is currently being updated to reflect this.

packages/runtime/src/next/client.ts
arvinpoddar: `this.fetch` only accepts relative URLs, since it takes on the responsibility of constructing the URL using the `apiOrigin`

packages/runtime/src/next/client.ts
agurtovoy: `results` is another common one; Stripe uses `data`: https://docs.stripe.com/api/pagination

migueloller: I've seen `list` as well. I think `data` is good 👍🏻

packages/runtime/src/next/client.ts
agurtovoy: It's possible to make it a method at the cost of giving the arguments record type a name. Does it matter if it's a method? I don't think we can make `toIterablePaginationResult` a decorator since it changes the signature.

packages/runtime/src/next/utils/pagination.ts
agurtovoy: Good question, I don't think it is anymore since we already have an outer function scope where we can store the state.

packages/runtime/src/next/client.ts
agurtovoy: This is an internal method that we are passing to `toIterablePaginationResult` wrapper. Defining it as a class field auto-binds `this`.

packages/runtime/src/next/utils/pagination.ts
arvinpoddar: Realistically, no. We wanted this helper to be flexible, so perhaps an entity could have a numerical `id`, etc.

packages/runtime/src/next/client.ts
agurtovoy: If we factor out the query construction out of this method, we wouldn't need to destructure them at all here. That said, IIRC, we originally had this argument as optional, and you can't destructure an optional arg.

.changeset/little-ads-return.md
arvinpoddar: I was planning on documenting this in our actual documentation page, as opposed to here (which is more focused on the deprecation). Do we need to do both?

packages/runtime/src/next/client.ts
migueloller: Ah, makes sense. As for destructuring from an optional param, this is what I usually do:

```ts
class Foo() {
  someMethod({ foo = 1, bar = '', baz }: { foo?: number, baz?: string, baz?: string } = {}) {
    //
  }
}
```

If not all fields were optional, then it wouldn't make sense for the parameter itself to be optional so this pattern works in all cases where you have optional named parameters.

packages/runtime/src/next/utils/pagination.ts
migueloller: In that scenario we could do `string | number` instead of `any`. Not a big deal, but `any` is a type that spreads around so I generally like to avoid it whenever possible

packages/runtime/src/next/client.ts
migueloller: Got it, I didn't notice we were relying on `this.fetch`. We could pay the price of instantiation at function call time by doing something like this:

```ts
  private async getPagesInternal() {
    //
  }

  getPages(...args) {
    return toIterablePaginationResult(this.getPagesInternal.bind(this))(...argrs)
  }
```

This makes it so the instances share all methods... but again, this is a super nit so what we have right now is perfectly ok.

.changeset/little-ads-return.md
migueloller: I think our release and our upgrade guide should be similar and/or link to each other. Also, the App Router example might be cleaner since it doesn't involve external dependencies, etc.

packages/runtime/src/next/utils/pagination.ts
arvinpoddar: Got it - then to avoid spreading I'll restrict `id` to a `string` for now.

packages/runtime/src/next/client.ts
arvinpoddar: Had an offline discussion, we're going to deprecate the method for now and remove it in a future release

.changeset/little-ads-return.md
```diff
@@ -0,0 +1,106 @@
...
+const DEFAULT_PRIORITY = 0.75
+const DEFAULT_FREQUENCY = 'hourly'
+
+function pageToSitemapEntry(page: MakeswiftPage): NextSitemapItem {
+  const pageUrl = new URL(page.path, DOMAIN)
+  return {
+    url: pageUrl.href,
+    lastModified: page.createdAt,
+    changeFrequency: page.sitemapFrequency ?? DEFAULT_FREQUENCY,
+    priority: page.sitemapPriority ?? DEFAULT_PRIORITY,
```
agurtovoy, line 92: What about the variants?

agurtovoy (approved):
Looking great, just one comment: https://github.com/makeswift/makeswift/pull/780/files#r1645196474

.changeset/little-ads-return.md
arvinpoddar: This is the funniest thing, you currently cannot generate a localized sitemap. There's an issue on Github for it: https://github.com/vercel/next.js/issues/65817. To replicate this yourself:

1. Create a new Next.js app with `npx create-next-app@latest` (accept all the default settings)
2. Create a new `app/sitemap.ts` file
3. Copy and paste their example from: https://nextjs.org/docs/app/api-reference/file-conventions/metadata/sitemap#generating-a-sitemap-using-code-js-ts. This will work when you visit `localhost:3000/sitemap.xml`
4. Now, copy and paste their localized example: https://nextjs.org/docs/app/api-reference/file-conventions/metadata/sitemap#generate-a-localized-sitemap. This will not work. Remove all the `alternates`, and it works again.

I intentionally omitted this from the example for now.

packages/runtime/src/next/index.tsx
arvinpoddar: Resolved, called out in the changeset

packages/runtime/src/next/client.ts
arvinpoddar: Resolved, moved param construction outside the client class.

.changeset/little-ads-return.md
agurtovoy: Got it, I think we should file a new issue for it since the linked one didn't follow their template and got auto-closed.

migueloller: Yeah, we can also reach out via Slack if needed.

arvinpoddar: Opened an issue: https://github.com/vercel/next.js/issues/67005

packages/runtime/src/next/client.ts
arvinpoddar: Addressed, using `safeParse` and throwing an error if not successful.

PR #781, opened 2024-06-15
runtime: unify controls interface

agurtovoy:
Corresponding `cosmos` PR: https://github.com/makeswift/cosmos/pull/212


packages/controls/src/color/color.ts
```diff
@@ -0,0 +1,118 @@
...
+        .optional(),
+      defaultValue: z.string().optional(),
+      hideAlpha: z.boolean().optional(),
+    })
+
+    type ControlData = z.infer<typeof dataSchema>
+    type Config = z.infer<typeof configSchema>
+
+    type ControlDefinition<C extends Config = Config> =
+      VersionedControlDefinition<typeof type, C, ColorData, typeof version>
```
arvinpoddar, line 50: Is there a reason for these to be defined inside the IIFE?

agurtovoy: That's the easiest way to limit the scope of these definitions. The main point of this refactoring is defining a common traits interface, and making all the type info and functions accessible through the interface rather through an ad-hoc collection of top-level types and functions.

.changeset/nasty-drinks-judge.md
```diff
@@ -0,0 +1,27 @@
...
+'@makeswift/runtime': minor
+---
+
+This change updates the runtime to use the latest version of
+`@makeswift/controls`. As part of this update, this package is no longer
+exposing internal data types and functions associated with our controls.
+
+## BREAKING:
+
+1. Registering controls with arbitrary options will now result in a TypeScript
```
agurtovoy, line 11: @arvinpoddar How about this:

1. An attempt to instantiate a control with extraneous configuration parameters will now result in a TypeScript compilation error:

    ```typescript
     import { Number } from '@makeswift/runtime/controls'
    
     const num = Number({ foo: 'bar' }) // error, `foo` is not a valid `Number` param
     ```
     Prior to this version, the extraneous parameters were silently ignored.

?

.changeset/nasty-drinks-judge.md
```diff
@@ -0,0 +1,27 @@
...
+   import { Number } from '@makeswift/runtime/controls'
+
+   const num = Number({ foo: 'bar' })
+   ```
+
+   `foo` is not a valid `Number` control option, resulting in an error. Any
+   invalid options in your component registrations must be removed for
+   successful compilation.
+
+2. When passing `options` to a `Select` control, you must provide at least one
```
agurtovoy, line 24: 2. `Select` control now requires the `options` array to have at least one entry:
    
   ```typescript
    import { Select } from '@makeswift/runtime/controls'

    const sel = Select({
      label: 'Block type',
      options: [] // error, non-empty array is required
    })
   ```

    Previously, the `options` parameter was allowed to be empty.

.changeset/nasty-drinks-judge.md
arvinpoddar: Well said, thank you

packages/controls/.prettierrc
agurtovoy, line 1: Renamed for consistency with `.prettierrc.json` at the root. Ideally the formatting settings should be shared between these two configs, but Prettier doesn't support config extension/merging, and we can't just delete this one in favor of the one at the root because the "sort imports" settings in `.prettierrc.json`  below are specific to this package.

packages/controls/.prettierrc.json
```diff
@@ -0,0 +1,31 @@
...
+  "bracketSpacing": true,
+  "jsxSingleQuote": false,
+  "printWidth": 80,
+  "quoteProps": "as-needed",
+  "semi": false,
+  "singleQuote": true,
+  "tabWidth": 2,
+  "trailingComma": "all",
+  "useTabs": false,
+  "plugins": ["@ianvs/prettier-plugin-sort-imports"],
```
agurtovoy, line 14: https://github.com/IanVS/prettier-plugin-sort-imports

packages/controls/jest.config.json
```diff
@@ -1,4 +1,5 @@
 {
+  "testRegex": "(\\./__tests__/.*|(\\.|/)(test|spec))\\.(types\\.)?[jt]sx?$",
```
agurtovoy, line 2: Adjust Jest's default test regex to pick up our type tests.

packages/controls/package.json
```diff
@@ -2,29 +2,38 @@
   "name": "@makeswift/controls",
   "version": "0.0.1",
   "license": "MIT",
-  "type": "module",
   "files": [
     "dist"
   ],
-  "main": "./dist/index.cjs",
+  "main": "./dist/cjs/index.js",
+  "module": "./dist/esm/index.js",
```
agurtovoy, line 9: Changes here and below bring `controls`' `package.json` in line with `prop-controllers`.

packages/controls/src/checkbox.ts
agurtovoy, line 1: Moved to https://github.com/makeswift/makeswift/pull/781/files#diff-e8ab449822fb3b353f4bd8da8edf44cfbe5d05fbc483b9053d68e9f49086c7fb

packages/controls/src/color.ts
agurtovoy, line 1: Moved to https://github.com/makeswift/makeswift/pull/781/files#diff-bf2e9306cf6e0d0238fcb9b8807fa29849be692e7358137d11c1f6d77cebc291

packages/controls/src/combobox.ts
agurtovoy, line 1: Moved to https://github.com/makeswift/makeswift/pull/781/files#diff-bd4cc846403737cf673af7c2845d11da3df054909e6d1af8d6708891e4aa0931

packages/controls/src/common/schema.ts
agurtovoy, line 1: Zod schemas for common data types, moved into their own file so we can namespace and use them in a more zod-like manner, e.g. `Schema.colorData` or `Schema.responsiveValue(...)` instead of `colorDataSchema` or `createResponsiveValueSchema(...)`.

packages/controls/src/common/types.ts
```diff
@@ -1,111 +1,26 @@
...
+export function isElementReference(
+  element: Element,
+): element is ElementReference {
+  return !('props' in element)
 }
 
-export type CopyContext = {
-  replacementContext: ReplacementContext
-  copyElement: (node: Element) => Element
-}
```
agurtovoy, line 111: Moved contexts to their own file: https://github.com/makeswift/makeswift/pull/781/files#diff-69a50df54dce20530e7f5f578014e4c0f4d453ea327ff76393ae3d309b29f4ea

packages/controls/src/controls/associated-types.ts
```diff
@@ -0,0 +1,31 @@
+import { ControlDefinition } from './definition'
+
+export type ControlType<D> =
+  D extends ControlDefinition<infer ControlType, any, any, any, any, any>
+    ? ControlType
+    : never
```
agurtovoy, line 6: Funny story with these ones: originally they were formulated as

```typescript

export type AssociatedTypes<D> =
  D extends ControlDefinition<
    infer ControlType,
    infer Config,
    infer DataType,
    infer ValueType,
    infer ResolvedValueType,
    infer InstanceType
  >
    ? {
        ControlType: ControlType
        Config: Config
        DataType: DataType
        ValueType: ValueType
        ResolvedValueType: ResolvedValueType
        InstanceType: InstanceType
      }
    : never

export type ControlType<D> = AssociatedTypes<D>['ControlType']
export type DataType<D> = AssociatedTypes<D>['DataType']
...
```
... which seems like a less repetitive but otherwise equivalent way to infer these types. However, it turned out that in some situations the above is NOT equivalent to the current formulation. I'll produce a sandbox reproduce for this at some point.

packages/controls/src/controls/checkbox/checkbox.test.ts
```diff
@@ -0,0 +1,54 @@
...
+    assignTest(Checkbox({ label: undefined, defaultValue: undefined }))
+  })
+
+  const invalidValues = [null, 17, 'text']
+
+  testDefinition(
+    Checkbox({ defaultValue: false, label: 'visible' }),
+    [true, false],
+    invalidValues,
+  )
```
agurtovoy, line 47: This runs a battery of predefined version data tests defined in https://github.com/makeswift/makeswift/pull/781/files#diff-a6e34394c847cf596574da34a2839d77b7e7e2b7abdb7142536dc36adac60764

packages/controls/src/controls/checkbox/checkbox.test.types.ts
```diff
@@ -0,0 +1,60 @@
+import { expectTypeOf } from 'expect-type'
```
agurtovoy, line 1: https://github.com/mmkal/expect-type

packages/controls/src/controls/definition.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  abstract get controlType(): ControlType
+
+  abstract get schema(): {
+    definition: SchemaType<unknown>
+    type: SchemaType<ControlType>
+    data: SchemaType<DataType>
+    value: SchemaType<ValueType>
+    resolvedValue: SchemaType<ResolvedValueType>
+  }
```
agurtovoy, line 43: Exposing the schemas is crucial for implementing composable controls (list and shape).

packages/controls/src/controls/definition.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  abstract createInstance(sendMessage: SendMessage<any>): InstanceType
+
+  abstract serialize(): [SerializedRecord, Transferable[]]
+
+  introspect<R>(
+    data: DataType | undefined,
+    target: IntrospectionTarget<R>,
+  ): R[] {
+    return target.introspect(data)
```
agurtovoy, line 85: Introspection methods are collapsed into a single generic method parametrized by the introspection target.

packages/controls/src/controls/definition.ts
```diff
@@ -0,0 +1,101 @@
...
+  }
+}
+
+export function serialize(
+  config: unknown,
+  rest: { type: string } & Record<string, unknown>,
+): [SerializedRecord, Transferable[]] {
```
agurtovoy, line 92: High-level serialization helper. Most controls call this to implement the `serialize` method.

packages/controls/src/controls/instance.ts
```diff
@@ -0,0 +1,30 @@
+export type ControlMessage<Payload = any> = {
+  type: string
+  payload?: Payload
+  [key: string]: any
```
agurtovoy, line 4: Not all of the control messages put their payload data under the `payload`.

packages/controls/src/controls/instance.ts
```diff
@@ -0,0 +1,30 @@
...
+  message: M,
+) => void
+
+export abstract class ControlInstance<
+  M extends ControlMessage = ControlMessage,
+> {
+  constructor(protected readonly sendMessage: SendMessage<M>) {}
+
+  abstract recv(message: M): void
+  abstract child(key: string): ControlInstance | undefined
```
agurtovoy, line 17: Added a `child(key)` method to standardize composite controls code.

packages/controls/src/controls/rich-text/__fixtures__/introspection.ts
```diff
@@ -93,47 +88,3 @@ const introspectionFixture: IndexSignatureHack<RichTextControlData> = {
     isFocused: false,
   },
 }
-
-describe('GIVEN introspecting RichText', () => {
```
agurtovoy, line 97: Split up the fixture from the test code.

packages/controls/src/controls/rich-text/dto/schema.ts
```diff
@@ -0,0 +1,39 @@
...
+  return z.any() as z.ZodType<AddIndexSignature<R>>
+}
+
+export const valueJSON = z.object({
+  object: objectType.optional(),
+  document: any<DocumentJSON>().optional(),
+  selection: any<SelectionJSON>().optional(),
+  annotations: z.record(z.string(), any<AnnotationJSON>()).optional(),
+  data: z.record(z.string(), z.any()).optional(),
+})
```
agurtovoy, line 39: Partial schema for `valueJSON`; the schema itself is needed to fix some of the Zod typing issues, there is no good reason why it's partial other than that we took a shortcut here to avoid converting about a dozen other DTO types to Zod schemas.

packages/controls/src/controls/rich-text/slate.ts
agurtovoy, line 1: Tidied up code from https://github.com/makeswift/makeswift/pull/781/files#diff-071fa0d38ad0707de7530ee7ad1a9daef4c2dbae72e807eee37d2e6a777c11aa

packages/controls/src/controls/checkbox/checkbox.ts
```diff
@@ -0,0 +1,186 @@
...
+        value === undefined
+          ? undefined
+          : {
+              ...Definition.dataSignature.v1,
+              value,
+            },
+      )
+      .with(undefined, () => value)
+      .exhaustive()
+  }
```
arvinpoddar, line 149: Might be able to tighten this up:

```typescript
  toData(value: ValueType<C>): DataType<C> {
    return match({ version: this.version, value })
      .with({ version: 1, value: P.boolean }, ({ value }) => ({
        ...Definition.dataSignature.v1,
        value,
      }))
      .with({ version: P.nullish }, ({ value }) => value)
      .with({ value: undefined }, () => undefined)
      .exhaustive()
  }
```

packages/controls/src/controls/checkbox/checkbox.ts
```diff
@@ -0,0 +1,186 @@
...
+
+  copyData(
+    data: DataType<C> | undefined,
+    _context: CopyContext,
+  ): DataType<C> | undefined {
+    return data
```
arvinpoddar, line 155: Should we be shallowing cloning this?

packages/controls/src/controls/combobox/combobox.test.ts
```diff
@@ -0,0 +1,74 @@
...
+        ],
+      }).config satisfies {
+        label?: string
+        getOptions: () => Promise<
+          { id: string; value: 'a' | 'b'; label: string }[]
+        >
+      }
+    })
+
+    test('disallows extraneous properties', () => {
```
arvinpoddar, line 34: Not every control has these tests, should I go ahead and backfill them to the other controls?

agurtovoy: All of them should:

Screenshot 2024-08-21 at 12 41 58 PM

packages/controls/src/controls/checkbox/checkbox.ts
agurtovoy: Good question, the old code did not:

copy

packages/controls/src/controls/checkbox/checkbox.ts
agurtovoy: Sure, go for it.

packages/controls/src/controls/rich-text/v1/copy.ts
agurtovoy, line 1: Tidied up code from https://github.com/makeswift/makeswift/pull/781/files#diff-b34738b7e147e454e4d555944db2a179b52ed4e31ff2bad1acc9f404a5b283a5

packages/controls/src/controls/rich-text/v1/introspection.ts
agurtovoy, line 1: Tidied up code from https://github.com/makeswift/makeswift/pull/781/files#diff-4d6e6b1aa817a930c37e67d419cb9d39e1b5c82920fef45a8e817c1f1cbd2e90

packages/controls/src/controls/rich-text/v2/copy.ts
agurtovoy, line 1: Tidied up code from https://github.com/makeswift/makeswift/pull/781/files#diff-f06dad41f3b035cfa194987d3644b48b0a50086707f9b5666128b3f65d606e6f

packages/controls/src/controls/rich-text/v2/introspection.ts
agurtovoy, line 1: Moved from https://github.com/makeswift/makeswift/pull/781/files#diff-12efb2da5b68a2a73b6a558d788b98d5062a4c222729e03a5848cc7440dd6215

packages/controls/src/controls/rich-text/v2/plugin.ts
```diff
@@ -0,0 +1,22 @@
...
+import { type DataType } from '../../associated-types'
+import { ControlDefinition } from '../../definition'
+
+export type RichTextPluginControl<
+  Def extends ControlDefinition = ControlDefinition,
+  ValueType extends Data = Data,
+> = {
+  definition: Def
+  getValue?(editor: Editor): ValueType | undefined
+  onChange?(editor: Editor, value: ValueType): void
```
agurtovoy, line 13: Made `getValue` and `onChange` optional here so we can serialized/deserialize `RichTextV2` definition w/o having to provide/generate "deserialized plugin" type -- see the corresponding serialization code here: https://github.com/makeswift/makeswift/pull/781/files#diff-43f3bf8883378456d53431e89b27f0c02e470722d8ae9e3ebcba9c6008fe32d4R147

packages/controls/src/controls/rich-text/v2/rich-text.ts
```diff
@@ -0,0 +1,166 @@
...
+type UserConfig = z.infer<typeof Definition.schema.userConfig>
+type DataV1Type = z.infer<typeof Definition.schema.dataV1>
+type DataV2Type = z.infer<typeof Definition.schema.dataV2>
+type DataType = z.infer<typeof Definition.schema.data>
+type ValueType = z.infer<typeof Definition.schema.value>
+
+abstract class Definition<
+  RuntimeNode,
+  Config extends UserConfig = UserConfig,
+  InstanceType extends ControlInstance<any> = ControlInstance<any>,
```
agurtovoy, line 26: Parametrized by `InstanceType` because the current implementation of the control instance depends on Slate's `ReactEditor` type.

packages/controls/src/controls/rich-text/v2/rich-text.ts
```diff
@@ -0,0 +1,166 @@
...
+
+  static normalizeData(data: DataType): DataV2Type {
+    return Definition.isV1Data(data)
+      ? Definition.nodesToDataV2(richTextDTOtoDAO(data))
+      : data
+  }
+
+  abstract toText(data: DataType | undefined): string
+  abstract get pluginControls(): RichTextPluginControl[]
+  abstract pluginControlAt(index: number): RichTextPluginControl | undefined
```
agurtovoy, line 159: Actual `RichTextV2` plugins depend on React, but the introspection and data copying code only needs access to plugin controls, so we're providing access to the latter through these abstract methods.

packages/controls/src/controls/select/select.test.ts
```diff
@@ -0,0 +1,68 @@
...
+        options,
+        // @ts-expect-error
+        extra: 'extra',
+      })
+    })
+  })
+
+  describe('assignability', () => {
+    // FIXME
+    // function assignTest(_def: SelectDefinition) {}
```
agurtovoy, line 47: These are currently failing, will come back to this later.

packages/controls/src/controls/select/select.ts
```diff
@@ -0,0 +1,182 @@
...
+import {
+  type DeserializedRecord,
+  type SerializedRecord,
+} from '../../serialization'
+
+import { ControlDefinition, serialize, type SchemaType } from '../definition'
+import { DefaultControlInstance, type SendMessage } from '../instance'
+
+type Option<T extends string> = { readonly value: T; readonly label: string }
+type OptionList<T extends string> = readonly [Option<T>, ...Option<T>[]]
```
agurtovoy, line 16: Made `options` a non-empty array: it makes sense to require at least one option, and it allows us to implement Zod's `unionOfLiterals` below w/o hacks.

packages/controls/src/controls/shape/shape-control.ts
agurtovoy, line 1: Extracted from https://github.com/makeswift/makeswift/pull/781/files#diff-dd10f639557e51d89c5195320460362d150316ac7a21a304e2a649edaff52ad0

packages/controls/src/controls/slot/slot-control.ts
agurtovoy, line 1: Extracted from https://github.com/makeswift/makeswift/pull/781/files#diff-cd4a7f92cff39705fbf6faabdc3d088cc5910e7476d67c3b5ce3d28c4675f197

packages/controls/src/controls/slot/slot.ts
```diff
@@ -0,0 +1,164 @@
...
+      const overrideElement = override.elements.find(
+        (e) => baseElement.type === e.type && baseElement.key === e.key,
+      )
+
+      if (overrideElement == null) return [baseElement]
+      if (overrideElement.deleted) return []
+      if (isElementReference(overrideElement)) return [overrideElement]
+      if (isElementReference(baseElement)) return [baseElement]
+
+      return context.mergeElement(baseElement, overrideElement)
```
agurtovoy, line 122: Original is here: https://github.com/makeswift/makeswift/pull/781/files#diff-cd4a7f92cff39705fbf6faabdc3d088cc5910e7476d67c3b5ce3d28c4675f197L78

packages/controls/src/controls/style-v2/style-v2-control.ts
agurtovoy, line 1: Extracted from https://github.com/makeswift/makeswift/pull/781/files#diff-0290b1196b1f88e304650c8afd9123691c1ed7cb0b882cc9ef3ed21811b03b86

packages/controls/src/controls/style/schema.ts
```diff
@@ -0,0 +1,215 @@
+import { z } from 'zod'
```
agurtovoy, line 1: Converted types from https://github.com/makeswift/makeswift/pull/781/files#diff-fe2757ce8e16ddd867978a37261e04fe23962f06fdaafd490fc0b26f226411ce and https://github.com/makeswift/makeswift/tree/main/packages/runtime/src/css to Zod schemas.

packages/controls/src/controls/style/style-control.ts
```diff
@@ -0,0 +1,26 @@
+import { type BoxModel } from '../../common'
```
agurtovoy, line 1: Extracted from https://github.com/makeswift/makeswift/pull/781/files#diff-fe2757ce8e16ddd867978a37261e04fe23962f06fdaafd490fc0b26f226411ceL201

packages/controls/src/controls/style/style.ts
```diff
@@ -0,0 +1,237 @@
...
+  }
+
+  toData(value: ValueType<C>): DataType<C> {
+    return value
+  }
+
+  copyData(
+    data: DataType<C> | undefined,
+    context: CopyContext,
+  ): DataType<C> | undefined {
```
agurtovoy, line 144: Tidied up version of https://github.com/makeswift/makeswift/pull/781/files#diff-fe2757ce8e16ddd867978a37261e04fe23962f06fdaafd490fc0b26f226411ceL244

packages/controls/src/controls/typography/schema.ts
```diff
@@ -0,0 +1,30 @@
...
+            value: z.number().nullable(),
+            unit: z.string().nullable(),
+          })
+          .nullable()
+          .optional(),
+        color: color.optional(),
+      }),
+    ),
+  })
+}
```
agurtovoy, line 30: Zod schema for https://github.com/makeswift/makeswift/pull/781/files#diff-b70dc315c7902612240b31f74d787d26f92c109d2147374ee57dbd6d528da820L3, except that this schema is for a single typography object vs an array.

packages/controls/src/controls/typography/typography.ts
```diff
@@ -0,0 +1,180 @@
...
+
+  static get schema() {
+    const type = z.literal(Definition.type)
+
+    const data = Schema.typography(
+      z
+        .object({
+          swatchId: z.string().nullable(),
+          alpha: z.number().nullable(),
+        })
```
agurtovoy, line 42: @migueloller Here's the nullable color data definition that you asked about earlier.

packages/controls/src/controls/typography/typography.ts
```diff
@@ -0,0 +1,180 @@
...
+
+    const resolvedData = Schema.typography(
+      z
+        .object({
+          swatch: z.object({
+            hue: z.number().nullable(),
+            saturation: z.number().nullable(),
+            lightness: z.number().nullable(),
+          }),
+          alpha: z.number().nullable(),
```
agurtovoy, line 54: ... and another one for resolved data.

packages/controls/src/icon-radio-group.ts
```diff
@@ -1,47 +0,0 @@
-export const unstable_IconRadioGroupIcon = {
```
agurtovoy, line 1: Moved to https://github.com/makeswift/makeswift/pull/781/files#diff-eabb08fb171c3a089a3f5d3823857c5521aa07b0957a21211f32abc73875d3b4

packages/controls/src/lib/add-index-signature.ts
```diff
@@ -0,0 +1,5 @@
+// See https://github.com/microsoft/TypeScript/issues/15300
+export type AddIndexSignature<T> =
```
agurtovoy, line 2: This is a legalized version of https://github.com/makeswift/makeswift/pull/781/files#diff-7135d2d9b277af3c39d03ead4f2e0d109d50c72190c9f8ed5ae415d410afdd32. 

The final comment from the TS team on this issue is "[interfaces should declare an explicit index signature if they intend to be indexed](https://github.com/microsoft/TypeScript/issues/15300#issuecomment-1320620897)."

packages/controls/src/controls/style-v2/style-v2-control.ts
```diff
@@ -0,0 +1,59 @@
...
+    super(sendMessage)
+    this.control = propDef.createInstance((message) =>
+      this.sendMessage({
+        type: StyleV2Control.CHILD_CONTROL_MESSAGE,
+        payload: { message },
+      }),
+    )
+  }
+
+  get propControl() {
```
migueloller, line 36: Should this be `control` instead of `propControl`?

packages/controls/src/controls/typography/typography.ts
migueloller: All the fields are nullable but the object itself doesn't seem to be nullable. Is that correct?

packages/controls/src/introspection.ts
migueloller, line 1: Can you explain why we created this module? It seems that all the default implementations return an empty array and that we don't seem to extend these to override the implementation since we seem to be constructing them and exporting the instances. Feels like a weird layer of indirection.

packages/prop-controllers/jest.config.json
```diff
@@ -1,5 +1,8 @@
 {
   "transform": {
     "^.+\\.(t|j)sx?$": "@swc/jest"
+  },
+  "moduleNameMapper": {
+    "^@makeswift/controls$": "<rootDir>/node_modules/@makeswift/controls/dist/esm/index.js"
```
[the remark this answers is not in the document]
migueloller: Yeah, this seems unnecessary

packages/prop-controllers/package.json
```diff
@@ -2,28 +2,29 @@
...
-    "dev": "concurrently -k 'tsc --watch --preserveWatchOutput' 'tsup src/index.ts --watch'",
+    "dev": "concurrently -k 'tsc --watch --preserveWatchOutput' 'tsup --watch'",
     "clean": "rm -rf dist",
-    "build": "tsc && tsup src/index.ts",
+    "build": "tsc && tsup",
     "prepublishOnly": "turbo run build",
     "test": "jest"
   },
   "dependencies": {
+    "@makeswift/controls": "workspace:*",
```
migueloller, line 19: I'm guessing we've already checked that when publishing we're using `pnpm` instead of `npm` and that it's replacing this with the appropriate value?

packages/runtime/builder/rich-text/package.json
migueloller, line 1: How come we're introducing a new entry point instead of exporting on the existing one? I'm trying to move away as much as possible from all these different entry points and eventually have separate packages.

packages/runtime/core/package.json
migueloller, line 1: See my other comment on multiple entry points

packages/runtime/package.json
```diff
@@ -170,12 +182,12 @@
     "@types/react": "^18.2.14",
     "@types/react-dom": "^18.2.6",
     "concurrently": "^5.3.0",
+    "expect-type": "^0.19.0",
     "jest": "^29.7.0",
     "jest-environment-jsdom": "^29.7.0",
     "madge": "^6.0.0",
     "msw": "^2.3.1",
     "next": "14.1.0",
-    "prettier": "^2.5.1",
```
migueloller, line 178: How come we're removing Prettier as a dependency of the runtime? Are we now enforcing this at the top-level for the entire monorepo?

packages/runtime/src/slate/index.ts
```diff
@@ -1,10 +1,11 @@
-export * from './TypographyPlugin'
+export { type NodeJSON, type ValueJSON, type InlineJSON, type MarkJSON } from '@makeswift/controls'
```
arvinpoddar, line 1: We may not need to expose these anymore, will confirm

packages/runtime/src/controls/rich-text/__tests__/dto.test.ts
```diff
@@ -1,60 +0,0 @@
...
-    )
-  })
-  test('WHEN translating typography data THEN nothing is lost', () => {
-    expect(typographyFixture).toStrictEqual(
-      richTextDAOToDTO(
-        richTextDTOtoDAO(typographyFixture),
-        richTextDTOtoSelection(typographyFixture),
-      ),
-    )
-  })
```
arvinpoddar, line 37: Do we have equivalent tests for these?

migueloller (approved):

.changeset/nasty-drinks-judge.md
```diff
@@ -0,0 +1,27 @@
...
+   option - passing an empty array will result in an error. If you are not
+   defining your options inline with the `Select` control, you may need to
+   declare the value `as const`.
```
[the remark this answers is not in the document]
agurtovoy: Yep, that's what we are doing here: 
Screenshot 2024-08-21 at 6 35 16 PM

packages/controls/src/common/schema.ts
```diff
@@ -0,0 +1,55 @@
...
+])
+
+// See https://github.com/colinhacks/zod?tab=readme-ov-file#recursive-types
+type Data = z.infer<typeof primitives> | Data[] | { [key: string]: Data }
+
+export const data: z.ZodType<Data> = z.union([
+  primitives,
+  z.array(z.lazy(() => data)),
+  z.record(z.lazy(() => data)),
+])
```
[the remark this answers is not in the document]
agurtovoy: This used to be defined as `export const dataSchema: z.ZodType<Data> = z.any()`, will revert to that 🙏

packages/controls/src/controls/color/color.ts
```diff
@@ -0,0 +1,201 @@
...
+  createInstance(sendMessage: SendMessage<any>) {
+    return new DefaultControlInstance(sendMessage)
+  }
+
+  serialize(): [SerializedRecord, Transferable[]] {
+    return serialize(this.config, {
+      type: Definition.type,
+      version: this.version,
+    })
+  }
```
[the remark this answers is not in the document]
agurtovoy: Re: `createInstance`, not straightforwardly; `createInstance` in the `ControlDefinition` can return a derived instance type, thus `ControlDefinition` cannot provide a default implementation of `createInstance` that returns `DefaultControlInstance` since this could be a lie.

Re: `serialize`, we can provide a default implementation that serializes config + type, which would cover 80% of the controls. I choose not to because I wanted to force the question of serialization on the control author (in this case, ourselves) as we were going through all of the controls. We can definitely revise this if we feel that there is no need to force this on the authors of any new controls.

packages/controls/src/introspection.ts
agurtovoy: There are two aspects to this. 

The first one is the idea that introspection for various resources can be implemented as a single generic method that takes an introspection target (type of the resource), and returns that target's "result type":
introspect method


Beside reducing the API surface of a control, this will also allow us to collapse "client" introspection code like this into a simple loop over all of the introspection targets (not done in this PR):
resource introspection

The second aspect of this was our hypothesis that if we also made the introspection target to be a visitor, i.e. give it an ability to extract the corresponding resource from data if data matches certain structure, then we can further simplify some code. The visitor part really only worked out for the `SwatchTarget`, but nevertheless, as the following couple of examples demonstrate, it's not an entirely useless abstraction:

https://github.com/makeswift/makeswift/blob/3b19965d565ba56281daa7feea92a11786b2da2c/packages/controls/src/controls/typography/typography.ts#L237
typography.ts

https://github.com/makeswift/makeswift/blob/3b19965d565ba56281daa7feea92a11786b2da2c/packages/controls/src/controls/style/style.ts#L245
style.ts

packages/runtime/package.json
agurtovoy: We already have Prettier as a dependency in the repo's `package.json`, and it's a newer version (3.2.5). Prettier 2.5.x chokes on some newer Typescript syntaxes, so we either needed to upgrade the runtime dep to the newer version, or have Prettier be picked up from the repo's root, and we opted for the latter.

Are there any caveats with this that we missed?

packages/runtime/src/controls/rich-text/__tests__/dto.test.ts
agurtovoy: We do: https://github.com/makeswift/makeswift/blob/aleksey/eng-5464-part2/packages/controls/src/controls/rich-text/dto/dto.test.ts

packages/runtime/builder/rich-text/package.json
arvinpoddar: We don't want the runtime to directly leak that it's using `Slate` for the RichText editing experience. Moreover, everything being exported from this entry point is only for use within the builder - both of these reasons are why we decided not to expose it via `@makeswift/runtime/slate` (which we hope to get rid of eventually). We could expose everything in this package via `@makeswift/runtime/builder`, but then it became an organizational question - we felt this entry point was too broad since it already exposes serialization related code.

packages/runtime/core/package.json
arvinpoddar: Similar line of thinking to the other introduction of the endpoint - originally, we were exposing these associated types via `@makeswift/runtime/controls`. This is arguably our most trafficked entry point, so we wanted to keep it as clean as possible, exposing only control definitions and the functions that create them. That's why we exposed these associated types via a new entry point.

Alternatively, we could have the builder directly depend on `@makeswift/controls`, and remove this exposure altogether. We didn't do this to avoid cases of having version mismatches between our three packages.

.changeset/spicy-pumas-suffer.md
```diff
@@ -0,0 +1,65 @@
...
+     definition, namely its type and config options
+
+   - `copyColorData` - Creates a copy of a given `ColorControlData`, given a copy
+     context.
+
+   All of these have been removed. Most of them have been replaced by
+   functionality available on the `ColorDefinition` active class. When we
+   register our components, we are creating an instance of a `ColorDefinition`
+   class. Let's see how we can use this class to achieve the functionality of
+   the removed exports above:
```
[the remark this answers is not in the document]
arvinpoddar: I think it's still relevant for internal consumption - someone attempting to introduce/update a control may be confused by how different everything looks, so this provides a good overview of how everything was consolidated into the unified interface. I can remove the reference to "exports", however.

.changeset/spicy-pumas-suffer.md
```diff
@@ -0,0 +1,65 @@
...
+
+   ```typescript
+   import { DataType } from './associated-types'
+   import { Color, ColorDefinition } from './color'
```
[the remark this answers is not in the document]
arvinpoddar: Ah, I defined these from the perspective of an internal user, not an external one (since this is meant for internal use only). That's not appropriate for a changelog, so I'll revert to what it was.

packages/runtime/src/slate/index.ts
arvinpoddar: Confirmed - not used in the builder anymore after we replaced the duplicated RichText copy functionality. Removing.

packages/controls/src/resources/types.ts
```diff
@@ -0,0 +1,7 @@
+import { z } from 'zod'
+
+import * as Schema from './schema'
+
+export type Swatch = z.infer<typeof Schema.swatch>
+export type File = z.infer<typeof Schema.file>
+export type PagePathnameSlice = z.infer<typeof Schema.pagePathnameSlice>
```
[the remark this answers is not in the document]
arvinpoddar: Looked into it, the types themselves are not used in this PR, but the underlying schemas (only swatch) are consumed when defining other schemas. To keep surface area down, I'm going to remove these types for now.

packages/runtime/src/runtimes/react/use-style.ts
```diff
@@ -22,3 +24,15 @@ export function useStyle(style: CSSObject): string {
 
   return `${cache.key}-${serialized.name}`
 }
+
+export function useStyles(cache: EmotionCache, styles: SerializedStyles[]) {
+  styles.forEach(s => registerStyles(cache, s, false))
+  useUniversalInsertionEffect(() => {
+    styles.forEach(s => insertStyles(cache, s, false))
+  })
+}
```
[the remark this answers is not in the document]
arvinpoddar: Confirmed, this isn't used anywhere at the moment. I'm going to revert this file.

packages/controls/src/controls/typography/typography.ts
arvinpoddar: Good call out, resolving.

.changeset/nasty-drinks-judge.md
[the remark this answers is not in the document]
arvinpoddar: Can you clarify the reason you brought this up? Was it because you thought we may be throwing a runtime error if the user provided an empty array? The changelog (at the time of review) may not have been super explicit over the type of error that would occur.

packages/prop-controllers/package.json
arvinpoddar: Yep, confirmed with a snapshot release:

CleanShot 2024-08-22 at 13 29 35@2x

Also confirmed this from a snapshot release installation of the runtime into an actual Next app. We verified that the dependencies on `@makeswift/controls` and `@makeswift/prop-controllers` are correctly replaced.

packages/runtime/package.json
migueloller: The main thing would be that if we're going to use Prettier across the monorepo when we should make sure that there's no individual package trying to manage code formatting and that we've got CI enforcing the top-level config as well as make sure that VS Code picks it up in normal development scenarios. We've had various issues where different developers check in different formatting based on their local configuration and I'd like that to stop happening. So this gives us an opportunity to standardize across the entire monorepo with the root Prettier configuration.

packages/controls/src/introspection.ts
migueloller: I still don't understand how it's supposed to work and its benefits but that's ok.

.changeset/nasty-drinks-judge.md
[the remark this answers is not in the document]
migueloller: I was mentioning it in case you all weren't aware that this could be enforced with TypeScript as well

packages/controls/src/controls/style-v2/style-v2-control.ts
agurtovoy: Removed in favor of already existing `child()` accessor

packages/runtime/package.json
agurtovoy: Sharing the Prettier config (vs the dependency) is a bit trickier if we want to do things like import sorting (something that we've set up in the `controls` package) through Prettier, which (often) requires package-specific settings. I think we'd have to switch to specifying settings in a JS config file so we can do something like

```js
import baseConfig from '../../prettier.config.js'

export default {
  ...baseConfig,
  "importOrder": [...]
};
```

packages/controls/src/controls/definition.ts
agurtovoy: Yes, but only because this signature mandates defining all of these schemas on all of the controls. I think we can drop `value` and `resolvedValue` from the formal `ControlDefinition` interface and leave the decision whether it's beneficial to have these two schemas or not up to an individual control (even though we never use `value` and `resolvedValue` schemas to parse runtime data, oftentimes having them does simplify corresponding type definition for controls where both data and values depend on the config).

agurtovoy: I'm going to leave these intact for now, though, as a quick attempt at removing `value` and `resolvedValue` schemas broke `ResolvedValueType` type inference pretty much everywhere, and we still have more important issues to tackle in this cycle.

packages/controls/src/controls/instance.ts
agurtovoy: Good point, updated to

```typescript
export type ControlMessage<Payload = Data> = {
  [key: string]: Data
} & {
  type: string
  payload?: Payload
}
```

arvinpoddar (approved):
What took you so long? LGTM.

packages/runtime/package.json
migueloller: If we're going to enforce one configuration across the entire monorepo I think we should only have 1 config at the top level. And even something like import order should probably be consistent as well so I don't think we should worry about having package-specific overrides.

agurtovoy: Import order for sure should be consistent at a high-level, but specific package idiosyncrasies often lead to minor tweaks in that plugin's configuration; for example, a particular directory might be so voluminous -- or semantically different from the rest -- that you want its imports to be separated by newlines from the rest. 

Either way, I think we are on the same page re: where to go from here.

PR #784, opened 2024-06-19
chore: update example apps with new `getPages` method

arvinpoddar:
Currently using the snapshot version of the runtime to test everything. Will modify this PR when the new version of the runtime is released.

### DIFF
--- a/examples/bigcommerce-catalyst/app/(default)/[...path]/page.tsx
+++ b/examples/bigcommerce-catalyst/app/(default)/[...path]/page.tsx
@@ -1,28 +1,28 @@
-import { getSiteVersion } from '@makeswift/runtime/next/server'
-import { notFound } from 'next/navigation'
-import { Page as MakeswiftPage } from '@makeswift/runtime/next'
+import { getSiteVersion } from '@makeswift/runtime/next/server';
+import { notFound } from 'next/navigation';
+import { Page as MakeswiftPage } from '@makeswift/runtime/next';
 
-import { client } from '~/lib/makeswift/client'
-import '~/lib/makeswift/components'
+import { client } from '~/lib/makeswift/client';
+import '~/lib/makeswift/components';
 
-type ParsedUrlQuery = { path?: string[] }
+type ParsedUrlQuery = { path?: string[] };
 
 export async function generateStaticParams() {
-  const pages = await client.getPages()
-
-  return pages.flatMap((page) => ({
+  return await client
+    .getPages()
+    .map((page) => ({
       path: page.path.split('/').filter((segment) => segment !== ''),
-    }),
-  )
+    }))
+    .toArray();
 }
 
 export default async function Page({ params }: { params: ParsedUrlQuery }) {
-  const path = '/' + (params?.path ?? []).join('/')
+  const path = '/' + (params?.path ?? []).join('/');
   const snapshot = await client.getPageSnapshot(path, {
     siteVersion: getSiteVersion(),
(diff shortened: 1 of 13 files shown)


pnpm-lock.yaml
migueloller, line 1: How come the root `pnpm-lock.yaml` changed? None of the examples are pnpm workspaces so I wouldn't expect a change here at all.

migueloller (approved):
Looks good! Just had a couple of questions. Also, can we verify that the Vercel deployments for all these examples deployed without issues?

arvinpoddar:
> Looks good! Just had a couple of questions. Also, can we verify that the Vercel deployments for all these examples deployed without issues?

Took a look at the preview deployments and the build logs, everything is looking good. I also cross-referenced some of these deployments with some from previous PRs, and the number of statically generated pages is the same. Feeling good!

examples/bigcommerce-catalyst/app/(default)/[...path]/page.tsx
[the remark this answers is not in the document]
Comment (arvinpoddar) [reply examples/bigcommerce-catalyst/app/(default)/[...path]/page.tsx:1]: The catalyst example has its own Prettier config that is different from the others. Their config enforces semicolons, while the others don't. I think this file managed to get by somehow - perhaps it was previously saved without formatting.

pnpm-lock.yaml
arvinpoddar: Yep, that was my mistake. Thanks for catching that.

examples/bigcommerce-catalyst/app/(default)/[...path]/page.tsx
```diff
@@ -1,28 +1,28 @@
...
+    .toArray();
 }
 
 export default async function Page({ params }: { params: ParsedUrlQuery }) {
-  const path = '/' + (params?.path ?? []).join('/')
+  const path = '/' + (params?.path ?? []).join('/');
   const snapshot = await client.getPageSnapshot(path, {
     siteVersion: getSiteVersion(),
-  })
+  });
```
Comment (agurtovoy) [inline examples/bigcommerce-catalyst/app/(default)/[...path]/page.tsx:23]: Semicolons ^^

examples/bigcommerce-catalyst/package.json
```diff
@@ -17,7 +17,7 @@
     "@bigcommerce/catalyst-client": "^0.1.1",
     "@graphql-typed-document-node/core": "^3.2.0",
     "@icons-pack/react-simple-icons": "^9.3.0",
-    "@makeswift/runtime": "^0.18.0",
+    "@makeswift/runtime": "^0.19.0",
```
agurtovoy, line 20: The gotcha with the `bigcommerce-catalyst` example is that it's the only example where we store `package-lock.json` in the repo, so we need to update that as well.

agurtovoy (approved):
Looking good, just a couple of comments

examples/bigcommerce-catalyst/app/(default)/[...path]/page.tsx
Comment (arvinpoddar) [reply examples/bigcommerce-catalyst/app/(default)/[...path]/page.tsx:23]: The catalyst example has a separate prettier config, so the auto-formatting is part of the change - the rest of the files in this example have semicolons. I believe this file somehow made it past auto-formatting.

examples/bigcommerce-catalyst/package.json
arvinpoddar: Good catch, thank you. I should have been using `npm` when upgrading the runtime instead of `pnpm`

agurtovoy (approved):

PR #789, opened 2024-06-24
feat: add tests for checkbox control

arvinpoddar:
Introduces use of snapshots for testing controls. Also creates separate test directories for prop-controllers and controls.


packages/runtime/src/next/components/tests/controls/__snapshots__/checkbox-control.test.tsx.snap
```diff
@@ -0,0 +1,603 @@
...
+  display: -webkit-box;
+  display: -webkit-flex;
+  display: -ms-flexbox;
+  display: flex;
+  -webkit-box-flex-wrap: wrap;
+  -webkit-flex-wrap: wrap;
+  -ms-flex-wrap: wrap;
+  flex-wrap: wrap;
+  width: 100%;
+}
```
arvinpoddar, line 31: I'm having some doubts about using snapshot testing given that the snapshot must be a full match. I don't want to cause people headaches any time the Page component changes and our snapshots are stale. Are we sure that the `screen` testing not sufficient? I understand that changing snapshots is a red flag, but I would argue that modifying older tests should raise the same alarm bells.

agurtovoy: Yeah, I didn't mean to suggest that we snapshot the DOM, there is just too much noise here.

packages/runtime/src/next/components/tests/controls/checkbox-control.test.tsx
```diff
@@ -0,0 +1,126 @@
...
+    // Assert
+    const renderResult = await act(async () =>
+      render(
+        <ReactRuntimeProvider runtime={runtime}>
+          <Page snapshot={snapshot} />
+        </ReactRuntimeProvider>,
+      ),
+    )
+
+    expect(renderResult).toMatchSnapshot()
```
agurtovoy, line 71: What I was thinking is that we snapshot the `snapshot` here, so we have a record of the element tree that can be examined visually: `expect(snapshot).toMatchSnapshot()`

agurtovoy (dismissed):

packages/runtime/src/next/components/tests/controls/checkbox-control.test.tsx
arvinpoddar: Got it, that makes more much less noise. Will address.

PR #791, opened 2024-06-24
fix: create stricter checkbox controls tests

arvinpoddar:
Rewrites the checkbox control tests to check exact DOM node contents rather than ternary based text inclusions. Also uses data-driven unit tests to check all boolean values.

### DIFF
--- a/packages/runtime/src/next/components/tests/controls/checkbox-control.test.tsx
+++ b/packages/runtime/src/next/components/tests/controls/checkbox-control.test.tsx
@@ -26,9 +26,9 @@ const ROOT_ID = '00000000-0000-0000-0000-000000000000'
 const ELEMENT_ID = '11111111-1111-1111-1111-111111111111'
 
 describe('Page', () => {
-  test('can render CheckboxControl v0 data', async () => {
+  test.each([true, false])('can render CheckboxControl v0 data', async bool => {
     // Arrange
-    const checkboxControlData: CheckboxControlDataV0 = true
+    const checkboxControlData: CheckboxControlDataV0 = bool
     const TestComponentType = 'TestComponent'
     const testId = 'test-id'
     const elementData: ElementData = createRootComponent(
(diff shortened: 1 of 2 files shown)


.changeset/soft-penguins-perform.md
```diff
@@ -0,0 +1,7 @@
+---
+"@makeswift/runtime": patch
+---
+
+Refactors checkbox control tests to be stricter by asserting against exact DOM
+node content matching for prop values. Also uses data-driven unit tests for more
+exhaustive coverage.
```
agurtovoy, line 7: I think we don't need a changeset if we are not changing any user-visible behavior.

packages/runtime/src/next/components/tests/controls/checkbox-control.test.tsx
```diff
@@ -74,14 +74,14 @@ describe('Page', () => {
     )
 
     expect(snapshot).toMatchSnapshot()
-    expect(screen.getByTestId(testId)).toHaveTextContent('alpha')
+    expect(screen.getByTestId(testId).textContent).toBe(String(bool))
```
agurtovoy, line 77: Don't have to change it, but I think a more generic way to assert here would be something like `expect(JSON.parse(screen.getByTestId(testId).textContent)).toBe(bool)`

agurtovoy: More generic == something that would work universally for all control types.

agurtovoy (approved):

packages/runtime/src/next/components/tests/controls/checkbox-control.test.tsx
arvinpoddar: You're right that it's more generic, but I decided to intentionally avoid doing any operations on the DOM content to prevent any kind of unintentional transformation. Obviously that won't matter here since the boolean is a trivial case, it's more of a general principle thing.

.changeset/soft-penguins-perform.md
arvinpoddar: Addressed, removing

packages/runtime/src/next/components/tests/controls/checkbox-control.test.tsx
agurtovoy: Non-trivial cases is actually why I brought this up in the first place. Ad hoc testing for a boolean is easy, but for other controls with object values we will have to parse the text content back -- which is totally fine. `JSON.parse` is the exact reverse of `stringify`, and will throw if there is anything extra in the `textContent` (other than trailing whitespace).

PR #795, opened 2024-07-03
refactor: react prop subscriptions

arvinpoddar:


packages/runtime/src/runtimes/react/controls.tsx
```diff
@@ -153,35 +155,59 @@ type PropsValueProps = {
...
-//   for (const [key, def] of Object.entries(propDefs)) {
-//     const value = useResolvedControlValue(elementData[key], def, resourceResolver)
-//     if (value !== undefined) {
-//       result[key] = value
-//     }
-//   }
-
-//   return result
-// }
+const useResolvedProps = (
```
arvinpoddar, line 158: This is the refactored hook that resolves prop values and establishes subscriptions to the underlying resources (via the unified interface). It no longer makes a looped hook call, appropriately responds to prop changes, and correctly unsubscribes when the element changes.

packages/controls/src/color/color.ts
```diff
@@ -128,17 +126,36 @@ export const Color = controlTraitsRegistry.add(
...
+    ctor.subscribeValue = (
+      value: ValueType<ControlDefinition>,
+      definition: ControlDefinition,
+      resolver: ResourceResolver,
+    ): ValueSubscription<string | undefined> => {
+      const swatchSubscription = resolver.subscribeSwatch(value?.swatchId)
+
+      if (value.swatchId !== null && swatchSubscription.readValue() == null) {
+        resolver.fetchSwatch(value.swatchId).catch(console.error)
+      }
```
arvinpoddar, line 148: Drawing attention to this: here, we make a fetch call to populate the redux store if the current value is empty. We do a similar process in `useSwatch`, with a `useEffect`. Making this fetch call gives us the expected behavior when a swatch is not in the cache: the first page render populates with the default config value, and the second render populates with the retrieved swatch.

PR #797, opened 2024-07-10
fix: swatch introspection on Style control

fikrikarim:


packages/runtime/src/controls/style.ts
```diff
@@ -219,6 +219,19 @@ export class StyleControl extends PropController<StyleControlMessage> {
...
 
+export function getStyleSwatchIds(data: StyleControlData): string[] {
+  const swatchIds: string[] = []
+
+  data?.border?.forEach(({ value }) => {
+    if (value.borderTop?.color?.swatchId) swatchIds.push(value.borderTop.color.swatchId)
+    if (value.borderRight?.color?.swatchId) swatchIds.push(value.borderRight.color.swatchId)
+    if (value.borderBottom?.color?.swatchId) swatchIds.push(value.borderBottom.color.swatchId)
+    if (value.borderLeft?.color?.swatchId) swatchIds.push(value.borderLeft.color.swatchId)
+  })
```
migueloller, line 230: Do we have a `getSwatchIds` function for border data that we can reuse here?

migueloller (dismissed):
LGTM 👍🏻

packages/runtime/src/controls/style.ts
fikrikarim: Good idea. Updated.

arvinpoddar (approved):
Looking good

PR #798, opened 2024-07-10
Rogue Canary Job

alexhwoods:


alexhwoods:
Ah, this: 
```
    if: github.ref != 'refs/heads/changeset-release/main'
```

may also, work, and is a bit cleaner..

PR #801, opened 2024-07-11
docs: add a guide for releasing a patch for an old version [ci skip]

fikrikarim:


migueloller (approved):
This is great! Can we put this in a `CONTRIBUTING.md` or maybe link to it from the `README.md`?

fikrikarim:
@migueloller I don't think it's a good idea to link it from `README.md` because the readme is shown on the GitHub repo page and the npm page. We could link it from `CONTRIBUTING.md` but we don't have that. 

I'll merge this as is. Thanks for the review!

PR #809, opened 2024-08-08
Add link command

alexhwoods:
https://github.com/user-attachments/assets/d15245c5-052d-4687-8c06-ca6f038b8d94


alexhwoods:
I've verified that the init command still works.

PR #813, opened 2024-08-09
Fix components not editable in the builder for pages router

fikrikarim:


packages/runtime/src/state/react-builder-preview.ts
```diff
@@ -464,11 +469,15 @@ export function initialize(
...
-
     const breakpoints = ReactPage.getBreakpoints(getState())
     dispatch(setBreakpoints(breakpoints))
     dispatch(setIsInBuilder(true))
+
+    // This will dispatch any buffered actions, but we don't control the order of the messages.
     channel.dispatchBuffered()
+    channel.initialize()
+    // This will dispatch any buffered actions that need to be run after the builder receive all the previous messages.
+    channel.dispatchBuffered({ lifecycle: Lifecycle.AFTER_INITIALIZED })
```
agurtovoy, line 480: I'd do `channel.initialize()` here and move `dispatchBuffered()` and `dispatchBuffered({ lifecycle: Lifecycle.AFTER_INITIALIZED })` inside of `channel.initialize()`

packages/runtime/src/state/react-builder-preview.ts
```diff
@@ -541,6 +554,11 @@ export function messageChannelMiddleware(
             channel.postMessage(action)
             break
 
+          case ActionTypes.MOUNT_COMPONENT:
+          case ActionTypes.UNMOUNT_COMPONENT:
+            channel.postMessage(action, undefined, { lifecycle: Lifecycle.AFTER_INITIALIZED })
```
agurtovoy, line 559: Once we have a state machine that defines the "correct" message order, I wonder if we can keep the `postMessage` simple and automatically dispatch buffered messages in the right order on `initialize` behind the scene.

packages/runtime/src/state/react-builder-preview.ts
```diff
@@ -685,11 +703,24 @@ function makeswiftApiClientSyncMiddleware(
...
 }
 
+type BufferedMessages = [Action, Transferable[]?][]
+
 class MessageChannel {
   private channel: MessagePort | null = null
-  private bufferedMessages: [Action, Transferable[]?][] = []
+  private bufferedMessages: BufferedMessages = []
+  private bufferedMessagesAfterInitialized: BufferedMessages = []
+  private initialized = false
```
agurtovoy, line 712: Not suggesting that we change anything here, but there is some tension here between "initialization" and "setup": are they the same, if not, why not?

packages/runtime/src/state/react-builder-preview.ts
```diff
@@ -708,14 +739,26 @@ class MessageChannel {
...
+        messages = this.bufferedMessages
+        break
+    }
+
+    messages.forEach(([message, transferables]) => {
       this.channel?.postMessage(message, transferables ?? [])
     })
 
-    this.bufferedMessages = []
+    messages = []
```
agurtovoy, line 761: This won't reset buffered messages. I'd change the `forEach` loop above to something like this (untested):

```typescript
    while (messages.length > 0) {
      const [message, transferables] = messages.shift()
      this.channel?.postMessage(message, transferables ?? [])
    }
```

agurtovoy (approved):
Generally looks good, one issue that needs to be addressed: https://github.com/makeswift/makeswift/pull/813/files#r1711623777

agurtovoy:
Fixed on the builder side instead: https://github.com/makeswift/cosmos/pull/218

fikrikarim:
Yes. Thank you for reviewing and closing the PR @agurtovoy!

I read your reviews and it's really helpful. Especially this part:

> This won't reset buffered messages, just reassign the local variable. I'd change the forEach loop above to something like this (untested):
```
    while (messages.length > 0) {
      const [message, transferables] = messages.shift()
      this.channel?.postMessage(message, transferables ?? [])
    }
```

I chuckled when I saw this because it's such a silly error 😅

PR #817, opened 2024-08-23
runtime: forward ref only if the component supports it

agurtovoy:
We still always attempt to forward to lazy components since we can't know if they accept refs without loading them.

Partially resolves https://linear.app/makeswift/issue/ENG-1609/solve-forwardref-warnings-coming-from-renderhook-components

### DIFF
--- a/packages/runtime/src/runtimes/react/utils/can-accept-ref.test.tsx
+++ b/packages/runtime/src/runtimes/react/utils/can-accept-ref.test.tsx
@@ -1,0 +1,27 @@
+import { Component, forwardRef, memo, lazy } from 'react'
+
+import { canAcceptRef } from './can-accept-ref'
+
+class Greeting extends Component {
+  render() {
+    return <h1>Hello, world!</h1>
+  }
+}
+
+describe('canAcceptRef', () => {
+  test.each([
+    [false, 'function component', () => <div />],
+    [
+      true,
+      'function component wrapped in `forwardRef',
+      forwardRef<HTMLDivElement>((_, ref) => <div ref={ref} />),
+    ],
+    [true, 'class component', Greeting],
+    // we try to pass a ref to all lazy components since we can't know if they accept refs without loading them
+    [true, 'lazy function component', lazy(async () => ({ default: () => <div /> }))],
+    [true, 'lazy class component', lazy(async () => ({ default: Greeting }))],
+  ])('returns %s for a %s', (expected, _, c) => {
+    expect(canAcceptRef(c)).toBe(expected)
+    expect(canAcceptRef(memo(c))).toBe(expected)
+  })
+})
--- a/packages/runtime/src/runtimes/react/utils/can-accept-ref.ts
+++ b/packages/runtime/src/runtimes/react/utils/can-accept-ref.ts
@@ -1,0 +1,56 @@
+import type { Component, PropsWithoutRef, RefAttributes } from 'react'
+import { type ComponentType } from '../../../state/react-page'
+
+type WrapperComponent = {
+  $$typeof: string
+  new (props: PropsWithoutRef<any> & RefAttributes<any>, context?: any): Component<any>
+}
+
+type MemoComponent = WrapperComponent & {
+  type: ComponentType
+}
+
+// See https://github.com/facebook/react/blob/main/packages/shared/ReactSymbols.js
+const REACT_MEMO_TYPE: symbol = Symbol.for('react.memo')
+const REACT_LAZY_TYPE: symbol = Symbol.for('react.lazy')
+const REACT_FORWARD_REF_TYPE: symbol = Symbol.for('react.forward_ref')
+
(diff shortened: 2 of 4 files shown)


packages/runtime/src/runtimes/react/utils/can-accept-ref.ts
```diff
@@ -0,0 +1,52 @@
...
+type MemoComponent = WrapperComponent & {
+  type: ComponentType
+}
+
+function hasTypeofSymbol(c: ComponentType, symbol: string): c is WrapperComponent {
+  // React uses `$$typeof` field on its wrapper components to identify them,
+  // see https://github.com/facebook/react/blob/main/packages/shared/ReactElementType.js
+  // and https://github.com/search?q=repo%3Afacebook%2Freact%20%24%24typeof&type=code
+  return c != null && '$$typeof' in c && c.$$typeof?.toString() === `Symbol(${symbol})`
+}
```
agurtovoy, line 18: Links:
* https://github.com/facebook/react/blob/main/packages/shared/ReactElementType.js
* https://github.com/search?q=repo%3Afacebook%2Freact%20%24%24typeof&type=code

packages/runtime/src/runtimes/react/utils/can-accept-ref.ts
```diff
@@ -0,0 +1,52 @@
...
+
+// See https://github.com/facebook/react/blob/main/packages/shared/ReactSymbols.js
+function isMemoComponent(c: ComponentType): c is MemoComponent {
+  return hasTypeofSymbol(c, 'react.memo')
```
agurtovoy, line 22: ttps://github.com/facebook/react/blob/main/packages/shared/ReactSymbols.js

packages/runtime/src/runtimes/react/utils/can-accept-ref.ts
```diff
@@ -0,0 +1,56 @@
...
+}
+
+type MemoComponent = WrapperComponent & {
+  type: ComponentType
+}
+
+// See https://github.com/facebook/react/blob/main/packages/shared/ReactSymbols.js
+const REACT_MEMO_TYPE: symbol = Symbol.for('react.memo')
+const REACT_LAZY_TYPE: symbol = Symbol.for('react.lazy')
+const REACT_FORWARD_REF_TYPE: symbol = Symbol.for('react.forward_ref')
```
agurtovoy, line 16: https://github.com/facebook/react/blob/main/packages/shared/ReactSymbols.js

packages/runtime/src/runtimes/react/utils/can-accept-ref.test.tsx
```diff
@@ -0,0 +1,27 @@
...
+      'function component wrapped in `forwardRef',
+      forwardRef<HTMLDivElement>((_, ref) => <div ref={ref} />),
+    ],
+    [true, 'class component', Greeting],
+    // we try to pass a ref to all lazy components since we can't know if they accept refs without loading them
+    [true, 'lazy function component', lazy(async () => ({ default: () => <div /> }))],
+    [true, 'lazy class component', lazy(async () => ({ default: Greeting }))],
+  ])('returns %s for a %s', (expected, _, c) => {
+    expect(canAcceptRef(c)).toBe(expected)
+    expect(canAcceptRef(memo(c))).toBe(expected)
```
agurtovoy, line 25: `memo`-wrapped components require additional logic, testing that it's there

arvinpoddar (dismissed):
1. Should we also update our control rendering test utilities to not forward the ref to the sample component, since it isn't necessary anymore?
2. Is there a way to actually test if the original error is thrown or not? I.e., not just unit tests on the `canAcceptRef` function?
3. I think this change should merit a changeset.

agurtovoy:
@arvinpoddar Good questions,

1. Yes, will do once this is blessed in general
2. Probably there is, but it's labor-intensive and I don't feel like it's worth it. 
3. Yes, will add one once this is blessed in general

packages/runtime/src/runtimes/react/utils/can-accept-ref.ts
agurtovoy: `Symbol.for` creates a [shared symbol](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol#shared_symbols_in_the_global_symbol_registry):

> Because registered symbols can be arbitrarily created anywhere, they behave almost exactly like the strings they wrap.

migueloller (approved):
Looks good 👍🏻 

We're missing a changeset.

Would you consider this a breaking change? It technically isn't but there's a chance there we're missing something and we stop forwarding a ref resulting in a change of behavior.

agurtovoy:
@migueloller I don't consider this a breaking change. We might be missing something and thus break some user code that used to work before, but that's pretty much always the case. If that happens, we either do a follow-up fix or revert this.

PR #818, opened 2024-08-26
fix: gracefully deserialize controls

arvinpoddar:
Modifies controls deserialization to still continue if an invalid config from the user is provided. This allows for the panels of successfully created controls to still appear in the builder. Addresses the case where some users are not passing valid configs during prop registration, but the issue is not surfaced due to them most likely not using TypeScript.

CleanShot 2024-08-26 at 10 17 02

### DIFF
--- a/.changeset/late-wolves-vanish.md
+++ b/.changeset/late-wolves-vanish.md
@@ -1,8 +1,9 @@
 ---
-"@makeswift/runtime": patch
+'@makeswift/runtime': patch
 ---
 
-Gracefully deserialize controls - if an invalid control definition is provided,
-a helpful console error message is shown and the remaining controls will still
-be deserialized. In the builder, the panel for the invalid control will be
-unavailable.
+- Relax `Select` options schema to allow values that can be coerced to a string.
+
+- `deserializeControls` now handles deserialization issues more gracefully:
+  the function attempts to deserialize all of the controls, reporting
+  deserialization errors through an optional error callback.
--- a/.changeset/late-wolves-vanish.md
+++ b/.changeset/late-wolves-vanish.md
@@ -1,0 +1,8 @@
+---
+"@makeswift/runtime": patch
+---
+
+Gracefully deserialize controls - if an invalid control definition is provided,
+a helpful console error message is shown and the remaining controls will still
+be deserialized. In the builder, the panel for the invalid control will be
+unavailable.
--- a/packages/controls/src/controls/select/select.test.js
+++ b/packages/controls/src/controls/select/select.test.js
@@ -1,0 +1,39 @@
+import { SelectDefinition } from './select'
+
+describe('Select', () => {
+  describe('deserialize', () => {
+    test('string values', () => {
+      expect(
+        SelectDefinition.deserialize({
+          type: SelectDefinition.type,
+          config: {
+            label: 'Block type',
+            labelOrientation: 'horizontal',
+            options: [
+              { value: 'p', label: 'Paragraph' },
+              { value: 'h1', label: 'Heading 1' },
+            ],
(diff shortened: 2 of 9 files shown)


packages/runtime/src/builder/serialization/control-serialization.ts
```diff
@@ -1044,7 +1044,18 @@ export function deserializeControls(
 ): Record<string, DeserializedControl> {
   return Object.entries(serializedControls).reduce(
     (deserializedControls, [key, serializedControl]) => {
-      return { ...deserializedControls, [key]: deserializeControl(serializedControl) }
+      try {
+        const deserializedControl = deserializeControl(serializedControl)
+        return { ...deserializedControls, [key]: deserializedControl }
+      } catch (err: unknown) {
+        if (err instanceof Error) {
+          console.error(
```
agurtovoy, line 1052: Will this still get reported to Sentry?

packages/runtime/src/builder/serialization/control-serialization.ts
```diff
@@ -1044,7 +1044,18 @@ export function deserializeControls(
 ): Record<string, DeserializedControl> {
   return Object.entries(serializedControls).reduce(
     (deserializedControls, [key, serializedControl]) => {
-      return { ...deserializedControls, [key]: deserializeControl(serializedControl) }
+      try {
+        const deserializedControl = deserializeControl(serializedControl)
+        return { ...deserializedControls, [key]: deserializedControl }
+      } catch (err: unknown) {
```
agurtovoy, line 1050: I'd write a unit test for this.

packages/runtime/src/builder/serialization/control-serialization.ts
arvinpoddar: It will not, and I'm not sure that we'd want to? I'd imagine that reporting to Sentry for this would raise a lot of noise for what are effectively user errors.

agurtovoy: I think we want it to be reported, it's unclear to me whether all or even most of these are going to be user errors, especially after we relax the schema to enable built-in coercions.

packages/runtime/src/builder/serialization/control-serialization.ts
arvinpoddar: https://github.com/makeswift/makeswift/pull/818/files#diff-5cbd6458c52f743b6f82f50f72b22da93215b86db7e90b3c9dc89a61770e8e98

agurtovoy (approved):

packages/controls/src/controls/select/select.test.js
migueloller, line 1: Shouldn't this be a TypeScript file?

.changeset/late-wolves-vanish.md
migueloller, line 1: Huh, a bit weird that we're editing an existing changeset vs having a new one 🤔

migueloller (commented):
Changes look good. Got a few comments. Also, for commit message style, we try to use [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/). It would be great if we could keep consistency on that 😄

packages/controls/src/controls/select/select.test.js
agurtovoy: These are specifically tests for vanilla JS usage.

.changeset/late-wolves-vanish.md
agurtovoy: Yes, I didn't know what I was doing.

agurtovoy:
Fixes https://linear.app/makeswift/issue/ENG-5973/zoderror-expected-string-received-number

PR #822, opened 2024-08-26
fix: gracefully handle the mismatch between `Shape` definition vs data when introspecting

agurtovoy:
`introspect` should never return `undefined`.


arvinpoddar (dismissed):

agurtovoy:
Fixes https://linear.app/makeswift/issue/ENG-5988/typeerror-cannot-read-properties-of-undefined-reading-key et al.

PR #826, opened 2024-08-30
refactor: move legacy descriptor prop rendering into its own file

agurtovoy:
Prep for [removing `<RenderHook>`](https://github.com/makeswift/makeswift/pull/814), no semantic changes

### DIFF
--- a/packages/runtime/src/runtimes/react/controls.tsx
+++ b/packages/runtime/src/runtimes/react/controls.tsx
@@ -3,26 +3,6 @@ import { useRef } from 'react'
 import * as ReactPage from '../../state/react-page'
 
 import {
-  useBoxShadow,
-  useBorder as useBorderData,
-  useLinkPropControllerData,
-  useCheckboxPropControllerData,
-  useDatePropControllerData,
-  useFontPropControllerData,
-  useVideoPropControllerData,
-  useTablePropControllerData,
-} from '../../components/hooks'
-import type { ColorValue } from '../../components/utils/types'
-import {
-  useResponsiveBorder,
-  useResponsiveBorderRadius,
-  useResponsiveMargin,
-  useResponsivePadding,
-  useResponsiveShadow,
-  useResponsiveWidth,
-} from '../../components/utils/responsive-style'
-
-import {
   CheckboxDefinition,
   NumberDefinition,
   RichTextV2Definition,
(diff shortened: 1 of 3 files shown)


agurtovoy:
"Hide whitespace" option to make reviewing a bit easier:

hide whitespace

packages/runtime/src/runtimes/react/controls.tsx
```diff
@@ -47,105 +27,20 @@ import {
...
-  return useStyle(useResponsiveMargin(value))
-}
-
-export function useBorderRadiusStyle(data: BorderRadiusPropControllerData | undefined): string {
-  const value = getBorderRadiusPropControllerDataResponsiveBorderRadiusData(data)
-
-  return useStyle(useResponsiveBorderRadius(value))
-}
-
-export function useShadowsStyle(data: ShadowsPropControllerData | undefined): string {
```
arvinpoddar, line 137: Looks like these functions are exported previously, but not in the change. Were these exports publicly exposed/consumed outside of this package?

arvinpoddar (approved):
Looking good, thanks for organizing this. Also, do we need a patch changeset?

packages/runtime/src/runtimes/react/controls.tsx
agurtovoy: Good question, forgot to comment on this, they weren't publicly exposed or used outside of the `controls.tsx` file.

agurtovoy:
@arvinpoddar I don't think we need a changeset since this is a purely internal refactoring.

migueloller:
> @arvinpoddar I don't think we need a changeset since this is a purely internal refactoring.

I think generally we want to have a changeset for any change, even if there's no public facing change. The reason is that we use changesets not just to generate the changelog but also to do new releases. This means a pure refactor would still require a changelog to cut a release. That being said, I'm also of the opinion that internal refactors (and any change for that matter) should be part of the changelog. Those kinds of changes don't require any more details than the commit messages though (in fact ideally the commit log would be the base changelog and we'd augment it with changesets for public-facing changes but alas it doesn't work that way so I opt for adding a changeset to every commit)

agurtovoy:
> The reason is that we use changesets not just to generate the changelog but also to do new releases. 

Yep, I wasn't gonna cut a release for this one.

> That being said, I'm also of the opinion that internal refactors (and any change for that matter) should be part of the changelog.

Sure, I don't think we've doing it for internal changes, at least not consistently, but can certainly do that going forward.

PR #827, opened 2024-08-31
feat: `partition`/`partitionRecord` utils

agurtovoy:
Pure addition, prep for https://github.com/makeswift/makeswift/pull/814.

### DIFF
--- a/packages/runtime/src/utils/__tests__/partition.test.ts
+++ b/packages/runtime/src/utils/__tests__/partition.test.ts
@@ -1,0 +1,75 @@
+import { expectTypeOf } from 'expect-type'
+
+import { partition, partitionRecord } from '../partition'
+
+function isNumber(value: unknown): value is number {
+  return typeof value === 'number'
+}
+
+describe('partition', () => {
+  it('splits the array into two based on a predicate', () => {
+    const array = [1, 2, 3, 4, 5]
+    const [even, odd] = partition(array, x => x % 2 === 0)
+    expect(even).toStrictEqual([2, 4])
+    expect(odd).toStrictEqual([1, 3, 5])
+  })
+
+  it('correctly handles an always-false predicate', () => {
+    const array = [1, 2, 3, 4, 5]
+    const [overTen, underTen] = partition(array, x => x > 10)
+    expect(overTen).toStrictEqual([])
+    expect(underTen).toStrictEqual(array)
+  })
+
+  it('correctly handles an always-true predicate', () => {
+    const array = [1, 2, 3, 4, 5]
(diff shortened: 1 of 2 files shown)


packages/runtime/src/utils/__tests__/partition.test.ts
```diff
@@ -0,0 +1,47 @@
...
+  return typeof value === 'number'
+}
+
+describe('partition', () => {
+  it('splits the array into two based on a predicate', () => {
+    const array = [1, 2, 3, 4, 5]
+    const [even, odd] = partition(array, x => x % 2 === 0)
+    expect(even).toStrictEqual([2, 4])
+    expect(odd).toStrictEqual([1, 3, 5])
+  })
```
arvinpoddar, line 15: I'd consider adding a test to make sure the method returns an empty partition when none of the values match the predicate. I know the function signature protects us from someone accidentally modifying it to return `null` or `undefined`, but I feel that it's a helpful addition.

packages/runtime/src/utils/partition.ts
```diff
@@ -0,0 +1,43 @@
+type Predicate<T, U extends T> = (value: T) => value is U
+
+export function partition<T, U extends T>(
+  array: readonly T[],
+  predicate: Predicate<T, U>,
+): [U[], Exclude<T, U>[]]
+
+export function partition<T>(array: readonly T[], predicate: (value: T) => boolean): [T[], T[]]
+
+export function partition(
```
arvinpoddar, line 10: May be helpful to add a JSDoc string clarifying that the first returned array is the matches, and that the results are returned in stable order.

arvinpoddar (approved):

packages/runtime/src/utils/__tests__/partition.test.ts
agurtovoy: Good idea.

packages/runtime/src/utils/partition.ts
agurtovoy: Good point, we should probably document all of the general-purpose "utils" functions.

agurtovoy: This particular convention for `partition` is a de facto standard, but docs never hurt.

migueloller (dismissed):
LGTM 👍🏻

migueloller:
Minor note, it was a bit weird to have to ignore changes from another PR. I really dislike how GitHub doesn't allow to do a review commit by commit and instead forces one to review all files at once 🙄

agurtovoy:
@migueloller My fault, I haven't rebased on `main` after merging the other one. Rebased now.

PR #829, opened 2024-09-04
refactor: move `is`, `deepEqual` and `shallowEqual` predicates to `@makeswift/controls`

agurtovoy:
No semantic changes, prep for https://github.com/makeswift/makeswift/pull/814

### DIFF
--- a/packages/controls/src/lib/predicates/deepEqual.ts
+++ b/packages/controls/src/lib/predicates/deepEqual.ts
@@ -1,0 +1,31 @@
+import { shallowEqual } from './shallowEqual'
+
+const { hasOwnProperty } = Object.prototype
+
+export const deepEqual = (a: unknown, b: unknown): boolean => {
+  if (shallowEqual(a, b)) return true
+
+  if (
+    typeof a !== 'object' ||
+    a === null ||
+    typeof b !== 'object' ||
+    b === null
+  )
+    return false
+
+  const keysA = Object.keys(a)
+  const keysB = Object.keys(b)
+
+  if (keysA.length !== keysB.length) return false
+
+  for (let i = 0; i < keysA.length; i += 1) {
+    if (
+      !hasOwnProperty.call(b, keysA[i]) ||
+      // @ts-expect-error: {}[string] is OK.
+      !deepEqual(a[keysA[i]], b[keysA[i]])
+    )
+      return false
+  }
+
+  return true
+}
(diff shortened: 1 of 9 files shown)


packages/controls/src/lib/predicates/deepEqual.ts
```diff
@@ -0,0 +1,31 @@
+import { shallowEqual } from './shallowEqual'
+
+const { hasOwnProperty } = Object.prototype
+
+export const deepEqual = (a: unknown, b: unknown): boolean => {
```
agurtovoy, line 5: Switched from default to named export, no other changes.

packages/controls/src/lib/predicates/is.ts
```diff
@@ -0,0 +1,5 @@
+export function is(x: unknown, y: unknown): boolean {
```
agurtovoy, line 1: Switched from default to named export, no other changes.

packages/controls/src/lib/predicates/shallowEqual.ts
```diff
@@ -0,0 +1,28 @@
+import { is } from './is'
+
+const { hasOwnProperty } = Object.prototype
+
+export const shallowEqual = (a: unknown, b: unknown): boolean => {
```
agurtovoy, line 5: Switched from default to named export, no other changes.

packages/runtime/src/utils/deepEqual.ts
```diff
@@ -1,27 +1,3 @@
...
-      !deepEqual(a[keysA[i]], b[keysA[i]])
-    )
-      return false
-  }
-
-  return true
-}
+import { deepEqual } from '@makeswift/controls'
 
 export default deepEqual
```
agurtovoy, line 3: Forwarding for backward compatibility.

arvinpoddar (approved):
Looks good to me

migueloller (commented):
I failed again to leave a review before this was merged. Sorry!

I wanted to say that it feels a bit weird that we're exporting utilities from `@makeswift/controls`. It seems that what we probably want is a shared library of utilities. Given this is all in a monorepo, I wonder if we could get away with a package that we don't even publish that's bundled with `@makeswift/controls` and `@makeswift/runtime` at build time using Vite.

agurtovoy:
> It seems that what we probably want is a shared library of utilities.

@migueloller For sure, and I think it should actually be a published package, since we have a number of these copied and pasted into our private monorepo as well. I'm just keeping things simple for now to stay on track.

PR #834, opened 2024-09-18
fix: `Shape` iterates through definition instead of data/value keys

agurtovoy:
This doesn't help us to gracefully handle property "swaps" along the lines of an example outlined below, but it is more conceptually correct and stops us from erring on stale/abandoned props. The current behavior I'm replacing is a regression introduced in https://github.com/makeswift/makeswift/pull/781.

"Property swap" example (to be tackled later):
1. Define a custom component taking a `Shape` prop with a key `a` mapped to `Number`
2. Add the component to the page
3. Change prop `a` to be a `List({ type: Number() })`

### DIFF
--- a/packages/controls/jest.config.json
+++ b/packages/controls/jest.config.json
@@ -2,5 +2,6 @@
   "testRegex": "(\\./__tests__/.*|(\\.|/)(test|spec))\\.(types\\.)?[jt]sx?$",
   "transform": {
     "^.+\\.(t|j)sx?$": "@swc/jest"
-  }
+  },
+  "setupFilesAfterEnv": ["<rootDir>/src/jest-setup.ts"]
 }
--- a/packages/controls/src/controls/shape/__snapshots__/shape.test.ts.snap
+++ b/packages/controls/src/controls/shape/__snapshots__/shape.test.ts.snap
@@ -19,6 +19,82 @@ ShapeDefinition {
 }
 `;
 
+exports[`Shape gracefully handles missing/extra props copyData 1`] = `
+{
+  "color": {
+    "alpha": 1,
+    "swatchId": "[swatch-id]",
+  },
+  "link": undefined,
+  "list": undefined,
+  "text": "Hello world",
+}
+`;
+
+exports[`Shape gracefully handles missing/extra props fromData 1`] = `
+{
+  "color": {
+    "alpha": 1,
+    "swatchId": "[swatch-id]",
+  },
+  "link": undefined,
+  "list": undefined,
+  "text": "Hello world",
+}
+`;
+
+exports[`Shape gracefully handles missing/extra props getTranslatableData 1`] = `
+{
+  "color": null,
+  "link": null,
+  "list": null,
+  "text": "Hello world",
+}
+`;
+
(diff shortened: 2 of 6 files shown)


packages/controls/jest.config.json
```diff
@@ -2,5 +2,6 @@
   "testRegex": "(\\./__tests__/.*|(\\.|/)(test|spec))\\.(types\\.)?[jt]sx?$",
   "transform": {
     "^.+\\.(t|j)sx?$": "@swc/jest"
-  }
+  },
+  "setupFilesAfterEnv": ["<rootDir>/src/jest-setup.ts"]
```
agurtovoy, line 6: Added a mock for `uuid`.

packages/controls/src/controls/shape/__snapshots__/shape.test.ts.snap
```diff
@@ -19,6 +19,82 @@ ShapeDefinition {
...
+      },
+    },
+  ],
+  "text": {
+    "@@makeswift/type": "text-area::v1",
+    "value": "Hello world",
+  },
+}
+`;
+
```
agurtovoy, line 97: Capture the fact that we are iterating over definition keys.

packages/controls/src/controls/shape/shape.test.ts
```diff
@@ -159,28 +162,6 @@ describe('Shape', () => {
       const fileIds = shape.introspect(shapeData, Targets.File)
       expect(fileIds).toEqual(['file-id'])
     })
-
-    test('gracefully handles missing/extra props', () => {
```
agurtovoy, line 163: Moved/expanded below.

packages/controls/src/jest-setup.ts
```diff
@@ -0,0 +1,3 @@
+jest.mock('uuid', () => ({
```
arvinpoddar, line 1: Is there a way this mock can be confined to the setup of the shape tests, rather than a global setup?

arvinpoddar (commented):
Agreed that relying on the definition keys is more robust and conceptually correct compared to reading the data keys.

packages/controls/src/jest-setup.ts
agurtovoy: We have it globally mocked in `runtime`, and IMHO this is a reasonable default. You could probably come up w/ a test use case where you want more flexibility, but it can be overridden locally then.

packages/controls/jest.config.json
migueloller: Is it for readability or repeatability or both? Or some other reason?

packages/controls/src/jest-setup.ts
migueloller: This seems like it could cause issue with tests that expect UUIDs to be different, no? Is the purpose here to make it so snapshots don't fail whenever we're using UUIDs? If so, it would be great if we could address this with some sort of seed value that makes tests deterministic

migueloller (dismissed):
Looks great! 👍🏻

packages/controls/jest.config.json
agurtovoy: `uuid` is mocked for stable results across different runs.

PR #836, opened 2024-09-19
runtime: multi-doc support, part 1: hoisting store, client

agurtovoy:
Depends on https://github.com/makeswift/makeswift/pull/839, rework of https://github.com/makeswift/makeswift/pull/830

"Hide whitespace" when viewing diffs for easier reviewing:

Hide whitespace


apps/nextjs-app-router/src/app/[lang]/[[...path]]/layout.tsx
```diff
@@ -13,10 +14,15 @@ export default function RootLayout({
   return (
     <html lang={params.lang}>
       <head>
-        <DraftModeScript />
+        <DraftModeScript appOrigin={process.env.MAKESWIFT_APP_ORIGIN} />
```
Comment (agurtovoy) [inline apps/nextjs-app-router/src/app/[lang]/[[...path]]/layout.tsx:17]: Add `appOrigin` for debugging w/ local builder, see https://makeswifthq.slack.com/archives/C014SCS4G75/p1725915902722189

apps/nextjs-app-router/src/app/[lang]/[[...path]]/layout.tsx
```diff
@@ -1,5 +1,6 @@
 import { MakeswiftProvider } from '@/makeswift/provider'
```
Comment (agurtovoy) [inline apps/nextjs-app-router/src/app/[lang]/[[...path]]/layout.tsx:1]: Moved under `app/[lang]/[...path]` to be able to retrieve page language/locale.

apps/nextjs-app-router/src/app/[lang]/[[...path]]/layout.tsx
```diff
@@ -13,10 +14,15 @@ export default function RootLayout({
...
     <html lang={params.lang}>
       <head>
-        <DraftModeScript />
+        <DraftModeScript appOrigin={process.env.MAKESWIFT_APP_ORIGIN} />
       </head>
       <body>
-        <MakeswiftProvider>{children}</MakeswiftProvider>
+        <MakeswiftProvider
+          previewMode={draftMode().isEnabled}
+          locale={params.lang}
```
Comment (agurtovoy) [inline apps/nextjs-app-router/src/app/[lang]/[[...path]]/layout.tsx:22]: Have to pass `previewMode` at this level because `draftMode()` can't be used in a client component (and `MakeswiftProvider` has to be a client component so we can pass `runtime` to it)

apps/nextjs-pages-router/src/pages/[[...path]].tsx
```diff
@@ -30,22 +30,35 @@ export async function getStaticPaths(): Promise<
...
+    locale,
   })
 
   if (snapshot == null) return { notFound: true }
 
-  return { props: { snapshot } }
+  return {
+    props: {
+      snapshot,
+      previewMode: Makeswift.getPreviewMode(previewData),
```
Comment (agurtovoy) [inline apps/nextjs-pages-router/src/pages/[[...path]].tsx:56]: Using a new method for deriving `previewMode` from `previewData`, see below.

packages/runtime/src/api/react.ts
```diff
@@ -19,21 +19,8 @@ import {
   CreateTableRecordMutationVariables,
 } from './graphql/generated/types'
 
-export type SerializedLocalizedResourcesMap = {
-  [resourceId: string]: string | null
-}
-
-type LocalizedResourcesMap = Map<string, string | null>
```
agurtovoy, line 26: Moved `LocalizedResourcesMap` to `MakeswiftApiClient`'s Redux store.

packages/runtime/src/api/react.ts
```diff
@@ -19,21 +19,8 @@ import {
...
-
 export type CacheData = MakeswiftApiClient.SerializedState
 
-export type MakeswiftClientOptions = {
-  uri: string
-  cacheData?: CacheData
-  localizedResourcesMap?: SerializedLocalizedResourcesMap
-  locale?: Intl.Locale
-}
-
```
agurtovoy, line 36: These weren't/shouldn't used externally, collapsed into the constructor definition.

packages/runtime/src/api/react.ts
```diff
@@ -57,15 +44,15 @@ export class MakeswiftHostApiClient {
   graphqlClient: GraphQLClient
   makeswiftApiClient: MakeswiftApiClient.Store
   subscribe: MakeswiftApiClient.Store['subscribe']
-  private localizedResourcesMap: LocalizedResourcesMap
-  private locale: Intl.Locale | undefined
 
-  constructor({ uri, cacheData, localizedResourcesMap = {}, locale }: MakeswiftClientOptions) {
+  constructor({ uri, cacheData, locale }: { uri: string; cacheData?: CacheData; locale?: string }) {
```
agurtovoy, line 48: Note that `locale` is a string now.

packages/runtime/src/api/react.ts
```diff
@@ -124,67 +111,46 @@ export class MakeswiftHostApiClient {
     )
   }
 
-  readLocalizedGlobalElement(globalElementId: string): LocalizedGlobalElement | null {
-    const localizedGlobalElementId = this.getLocalizedResourceId(globalElementId)
-
-    if (localizedGlobalElementId == null) return null
-
+  readLocalizedGlobalElement(
+    locale: string,
```
agurtovoy, line 115: Added an explicit `locale` parameter.

packages/runtime/src/api/react.ts
```diff
@@ -124,67 +111,46 @@ export class MakeswiftHostApiClient {
     )
   }
 
-  readLocalizedGlobalElement(globalElementId: string): LocalizedGlobalElement | null {
-    const localizedGlobalElementId = this.getLocalizedResourceId(globalElementId)
-
-    if (localizedGlobalElementId == null) return null
-
```
agurtovoy, line 131: This logic has been absorbed into `MakeswiftApiClient.getAPIResource`.

packages/runtime/src/api/react.ts
```diff
@@ -124,67 +111,46 @@ export class MakeswiftHostApiClient {
...
-    // TODO(fikri): We're checking the cache here because the cache hit will fail on fetchAPIResource.
-    // This is because in the cache we're saving the ID of the localizedGlobalElementId,
-    // but we're reading the cache using the globalElementId.
-    // This is very confusing and it can lead to more bugs. We need to refactor how we handle
-    // localized global element.
-    const cacheResult = this.readLocalizedGlobalElement(globalElementId)
-
-    if (cacheResult) return cacheResult
-
-    const result = await this.makeswiftApiClient.dispatch(
```
agurtovoy, line 161: Here as well, logic has been absorbed into `MakeswiftApiClient.fetchAPIResource`.

packages/runtime/src/api/react.ts
```diff
@@ -124,67 +111,46 @@ export class MakeswiftHostApiClient {
...
+    return await this.makeswiftApiClient.dispatch(
       MakeswiftApiClient.fetchAPIResource(
         APIResourceType.LocalizedGlobalElement,
         globalElementId,
         locale,
       ),
     )
-
-    this.setLocalizedResourceId({
```
agurtovoy, line 169: Moved to Redux.

packages/runtime/src/api/react.ts
```diff
@@ -124,67 +111,46 @@ export class MakeswiftHostApiClient {
...
-      resourceId: globalElementId,
-      localizedResourceId: result?.id ?? null,
-    })
-
-    return result
   }
 
-  readPagePathnameSlice(pageId: string): PagePathnameSlice | null {
+  readPagePathnameSlice(pageId: string, locale: string | null): PagePathnameSlice | null {
```
agurtovoy, line 139: Added an explicit `locale` parameter.

packages/runtime/src/api/react.ts
```diff
@@ -124,67 +111,46 @@ export class MakeswiftHostApiClient {
...
       APIResourceType.PagePathnameSlice,
       pageId,
+      locale,
     )
   }
 
-  async fetchPagePathnameSlice(pageId: string): Promise<PagePathnameSlice | null> {
+  async fetchPagePathnameSlice(
+    pageId: string,
+    locale: string | null,
```
agurtovoy, line 150: Added an explicit `locale` parameter.

packages/runtime/src/api/react.ts
```diff
@@ -232,18 +198,4 @@ export class MakeswiftHostApiClient {
...
-
-  setLocalizedResourceId({
-    resourceId,
-    localizedResourceId,
-  }: {
-    resourceId: string
-    localizedResourceId: string | null
-  }): void {
-    this.localizedResourcesMap.set(resourceId, localizedResourceId)
-  }
```
agurtovoy, line 248: Moved to Redux.

packages/runtime/src/components/builtin/Form/Form.tsx
```diff
@@ -36,11 +36,11 @@ import { BoxModelHandle, getBox } from '../../../box-model'
 import { PropControllersHandle } from '../../../state/modules/prop-controller-handles'
 import { DescriptorsPropControllers } from '../../../prop-controllers/instances'
 import { useTableFormFieldRefs } from '../../hooks/useTableFormFieldRefs'
-import { useMakeswiftHostApiClient } from '../../../next/context/makeswift-host-api-client'
 import { cx } from '@emotion/css'
 import { useResponsiveGridItem, useResponsiveStyle } from '../../utils/responsive-style'
 import { type ResponsiveColor } from '../../utils/types'
 
+import { useMakeswiftHostApiClient } from '../../../runtimes/react/host-api-client'
```
agurtovoy, line 43: Moved `useMakeswiftHostApiClient` from `next/context` to `runtimes/react`.

packages/runtime/src/components/hooks/usePageSnippets.ts
```diff
@@ -11,7 +11,7 @@ const SnippetLocation = {
   Head: 'HEAD',
 } as const
 
-export type SnippetLocation = typeof SnippetLocation[keyof typeof SnippetLocation]
+export type SnippetLocation = (typeof SnippetLocation)[keyof typeof SnippetLocation]
```
agurtovoy, line 14: Prettier reformat.

packages/runtime/src/components/page/Page.tsx
```diff
@@ -1,19 +1,19 @@
...
 
 type Props = {
-  pageDocument: MakeswiftPageDocument
-  documentKey: string
+  page: MakeswiftPageDocument
+  rootDocument: Document
 }
 
-export function Page({ pageDocument: page, documentKey }: Props): JSX.Element {
+export function Page({ page, rootDocument }: Props): JSX.Element {
```
agurtovoy, line 16: Take the whole root document instead of the `documentKey` here so that we don't have to re-fetch it from the store (the usage site does have the document already).

packages/runtime/src/components/page/Page.tsx
```diff
@@ -22,7 +22,7 @@ export function Page({ pageDocument: page, documentKey }: Props): JSX.Element {
     <>
       <PageHead document={page} />
 
-      <DocumentReference documentReference={createDocumentReference(documentKey)} />
+      <DocumentRoot rootDocument={rootDocument} />
```
agurtovoy, line 25: Renamed `DocumentReference` to `DocumentRoot`.

packages/runtime/src/next/client.ts
```diff
@@ -121,11 +122,7 @@ export type MakeswiftPageDocument = {
 
 export type MakeswiftPageSnapshot = {
   document: MakeswiftPageDocument
-  apiOrigin: string
   cacheData: CacheData
-  preview: boolean
-  localizedResourcesMap: SerializedLocalizedResourcesMap
-  locale: string | null
```
agurtovoy, line 128: Page snapshot is now just `document` and `cacheData`.

packages/runtime/src/next/client.ts
```diff
@@ -219,6 +216,10 @@ export class Makeswift {
     return getMakeswiftSiteVersion(previewData) ?? MakeswiftSiteVersion.Live
   }
 
+  static getPreviewMode(previewData: PreviewData): boolean {
+    return getMakeswiftSiteVersion(previewData) === MakeswiftSiteVersion.Working
+  }
```
agurtovoy, line 221: Added a method for deriving `previewMode` from `previewData`, see the Pages router example.

packages/runtime/src/next/client.ts
```diff
@@ -381,7 +382,7 @@ export class Makeswift {
     element: Element,
     siteVersion: MakeswiftSiteVersion,
     locale?: string,
-  ): Promise<{ cacheData: CacheData; localizedResourcesMap: SerializedLocalizedResourcesMap }> {
+  ): Promise<CacheData> {
```
agurtovoy, line 385: `introspect` simply returns the cache data.

packages/runtime/src/next/components/page.tsx
```diff
@@ -1,51 +1,82 @@
...
 }
 
+const isServer = typeof window === 'undefined'
+
+function useCacheData(cacheData: CacheData) {
+  const client = useMakeswiftHostApiClient()
+
+  if (isServer) {
+    client.makeswiftApiClient.dispatch(updateAPIClientCache(cacheData))
+  }
```
agurtovoy, line 26: Hack to fill the cache when rendering on the server.

packages/runtime/src/next/components/page.tsx
```diff
@@ -1,51 +1,82 @@
...
-  const client = useMemo(
-    () =>
-      new MakeswiftHostApiClient({
-        uri: new URL('graphql', snapshot.apiOrigin).href,
-        cacheData: snapshot.cacheData,
-        localizedResourcesMap: snapshot.localizedResourcesMap,
-        locale: snapshot.locale ? new Intl.Locale(snapshot.locale) : undefined,
-      }),
-    [snapshot],
-  )
```
agurtovoy, line 30: Client instance moved from `Page` to `ReactRuntimeProvider`.

packages/runtime/src/next/components/page.tsx
```diff
@@ -1,51 +1,82 @@
...
-          pageDocument={snapshot.document}
-          documentKey={rootElement.key}
-        />
-      </RuntimeProvider>
+      {/* We use a key here to reset the Snippets state in the PageMeta component */}
+      <PageComponent
+        key={snapshot.document.data.key}
+        page={snapshot.document}
+        rootDocument={rootDocument}
+      />
```
agurtovoy, line 79: Pass the whole `rootDocument` instead of just the `documentKey`, see my earlier comment.

packages/runtime/src/next/components/page.tsx
```diff
@@ -1,51 +1,82 @@
...
-    ? { key: localizedPage.elementTreeId, data: localizedPage.data }
-    : { key: snapshot.document.id, data: snapshot.document.data }
+  useCacheData(snapshot.cacheData)
 
-  const rootElements = new Map([[rootElement.key, rootElement.data]])
+  const rootDocument = useRegisterPageDocument(snapshot.document)
 
   return (
     <Suspense>
-      <RuntimeProvider client={client} rootElements={rootElements} preview={snapshot.preview}>
```
agurtovoy, line 41: `RuntimeProvider` is gone, functionality has been hoisted up to `ReactRuntimeProvider`.

packages/runtime/src/next/components/tests/controls/__snapshots__/checkbox-control.test.tsx.snap
```diff
@@ -4,8 +4,10 @@ exports[`Page Checkbox control data v0 when value is false: resolvedValue 1`] =
 
 exports[`Page Checkbox control data v0 when value is false: snapshot 1`] = `
 {
-  "apiOrigin": "https://test-api-origin.com",
-  "cacheData": {},
+  "cacheData": {
+    "apiResources": {},
+    "localizedResourcesMap": {},
+  },
```
agurtovoy, line 10: Here and below, lots of Jset snapshot updates reflecting changes to Makeswift snapshot structure.

packages/runtime/src/next/components/tests/global-element-rendering.test.tsx
```diff
@@ -224,9 +225,6 @@ describe('Page', () => {
   test('correctly renders a localized global element', async () => {
     await testGlobalElementRendering({
       locale: 'fr-FR',
-      localizedResourcesMap: {
-        [globalElementId]: localizedGlobalElementId,
-      },
```
agurtovoy, line 229: Simple presence of `localizedResourcesMap` in the cache no longer triggers global component localization, moved the cache inside of the `testGlobalElementRendering` helper.

packages/runtime/src/runtimes/react/components/DocumentReference.tsx
```diff
@@ -1,25 +0,0 @@
...
-import { useDocument } from '../hooks/use-document'
-import { FallbackComponent } from '../../../components/shared/FallbackComponent'
-import { Document } from './Document'
-
-type DocumentReferenceProps = {
-  documentReference: ReactPageDocumentReference
-}
-
-export const DocumentReference = memo(
-  forwardRef(function DocumentReference(
```
agurtovoy, line 13: Moved/renamed to `DocumentRoot`, as it takes a root document now.

packages/runtime/src/runtimes/react/components/ElementReference.tsx
```diff
@@ -22,8 +23,9 @@ export const ElementReference = memo(
     ref: Ref<ElementImperativeHandle>,
   ): JSX.Element {
     const globalElement = useGlobalElement(elementReference.value)
+    const locale = useDocumentLocale()
     // Update the logic here when we can merge element trees
-    const localizedGlobalElement = useLocalizedGlobalElement(elementReference.value)
+    const localizedGlobalElement = useLocalizedGlobalElement(locale, elementReference.value)
```
agurtovoy, line 28: Explicitly pass the current document's locale.

packages/runtime/src/runtimes/react/components/LiveProvider.tsx
```diff
@@ -1,27 +1,16 @@
...
-}
+import { useReactRuntime } from '../hooks/use-react-runtime'
+import { StoreContext } from '../hooks/use-store'
 
-export default function LiveProvider({ client, children, rootElements }: Props): JSX.Element {
+export default function LiveProvider({ children }: PropsWithChildren): JSX.Element {
   const runtime = useReactRuntime()
   const store = useMemo(
-    () => ReactPage.configureStore({ preloadedState: runtime.store.getState(), rootElements }),
-    [rootElements, runtime],
```
agurtovoy, line 19: `rootElements` are filled in through [`useRegisterDocuments` effect](https://github.com/makeswift/makeswift/pull/836/files#diff-574436562edd329137174ba80e6ecdb186f78f8e2b7e6cbc929afe8ede60e4e9R33)

packages/runtime/src/runtimes/react/components/LiveProvider.tsx
```diff
@@ -1,27 +1,16 @@
...
+    () => ReactPage.configureStore({ preloadedState: runtime.store.getState() }),
+    [runtime],
   )
 
-  return (
-    <StoreContext.Provider value={store}>
-      <MakeswiftHostApiClientProvider client={client}>{children}</MakeswiftHostApiClientProvider>
```
agurtovoy, line 24: Hoisted up to `ReactRuntimeProvider`.

packages/runtime/src/runtimes/react/components/PreviewProvider.tsx
```diff
@@ -1,55 +1,30 @@
...
   const runtime = useReactRuntime()
+  const client = useMakeswiftHostApiClient()
   const store = useMemo(
     () =>
       ReactBuilderPreview.configureStore({
         preloadedState: runtime ? runtime.store.getState() : ReactRuntime.store.getState(),
-        rootElements,
         client,
       }),
-    [client, rootElements, runtime],
```
agurtovoy, line 29: `rootElements` are filled in through [`useRegisterDocuments` effect](https://github.com/makeswift/makeswift/pull/836/files#diff-574436562edd329137174ba80e6ecdb186f78f8e2b7e6cbc929afe8ede60e4e9R33)

packages/runtime/src/runtimes/react/components/PreviewProvider.tsx
```diff
@@ -1,55 +1,30 @@
...
-      ([documentKey, rootElement]) =>
-        store.dispatch(registerDocumentEffect(ReactPage.createDocument(documentKey, rootElement))),
-    )
-
-    return () => {
-      unregisterDocuments.forEach(unregisterDocument => {
-        unregisterDocument()
-      })
-    }
-  }, [store, rootElements])
```
agurtovoy, line 48: Moved to [`useRegisterDocuments` effect](https://github.com/makeswift/makeswift/pull/836/files#diff-574436562edd329137174ba80e6ecdb186f78f8e2b7e6cbc929afe8ede60e4e9R33)

packages/runtime/src/runtimes/react/components/PreviewProvider.tsx
```diff
@@ -1,55 +1,30 @@
...
-
-  return (
-    <StoreContext.Provider value={store}>
-      <MakeswiftHostApiClientProvider client={client}>{children}</MakeswiftHostApiClientProvider>
-    </StoreContext.Provider>
```
agurtovoy, line 53: `MakeswiftHostApiClientProvider` hoisted up to `ReactRuntimeProvider`.

packages/runtime/src/runtimes/react/components/RuntimeProvider.tsx
agurtovoy, line 1: `RuntimeProvider` is gone, "reusing" this file to house the `ReactRuntimeProvider`.

packages/runtime/src/runtimes/react/hooks/makeswift-api.ts
```diff
@@ -118,38 +119,46 @@ export function useGlobalElement(globalElementId: string | null): GlobalElement
 }
 
 export function useLocalizedGlobalElement(
+  locale: string | null,
```
agurtovoy, line 122: Take an explicit `locale` parameter.

packages/runtime/src/runtimes/react/hooks/use-document-context.ts
agurtovoy, line 1: Generalized `use-document-key`.

packages/runtime/src/state/actions.ts
```diff
@@ -84,6 +85,8 @@ export const ActionTypes = {
   SET_LOCALE: 'SET_LOCALE',
 
   SET_LOCALIZED_RESOURCE_ID: 'SET_LOCALIZED_RESOURCE_ID',
+
+  UPDATE_API_CLIENT_CACHE: 'UPDATE_API_CLIENT_CACHE',
```
agurtovoy, line 89: New action to update the client cache.

packages/runtime/src/state/actions.ts
```diff
@@ -299,7 +302,13 @@ type SetLocaleAction = {
 
 type SetLocalizedResourceIdAction = {
   type: typeof ActionTypes.SET_LOCALIZED_RESOURCE_ID
-  payload: { resourceId: string; localizedResourceId: string | null }
+  // TODO: make `locale` required once we've upgraded the builder to always provide it
+  payload: { locale: string | undefined; resourceId: string; localizedResourceId: string | null }
```
agurtovoy, line 306: Extend `SET_LOCALIZED_RESOURCE_ID` action with an optional `locale` prop.

packages/runtime/src/state/makeswift-api-client.ts
```diff
@@ -1,44 +1,98 @@
...
+  type Table,
+  type LocalizedGlobalElement,
+  type APIResourceLocale,
 } from '../api'
 
-const reducer = APIResources.reducer
+const reducer = combineReducers({
+  apiResources: APIResources.reducer,
+  localizedResourcesMap: LocalizedResourcesMap.reducer,
+})
```
agurtovoy, line 30: Combine the `APIResources` and `LocalizedResourcesMap` states. We could move the `LocalizedResourcesMap` state into the `APIResources`, but this layered version felt cleaner to me.

packages/runtime/src/state/makeswift-api-client.ts
```diff
@@ -1,44 +1,98 @@
...
+      if (locale == null) {
+        console.error(`Attempt to access ${resourceType} ${resourceId} without a locale`)
+        return false
+      }
+
+      const localizedId = getLocalizedResourceId(state, locale, resourceId)
+      return (
+        localizedId != null &&
+        APIResources.getHasAPIResource(state.apiResources, resourceType, localizedId, locale)
+      )
```
agurtovoy, line 68: Global to localized element ID translation logic, moved from `MakeswiftHostApiClient`: https://github.com/makeswift/makeswift/pull/836/files#diff-e66ba2364c981fb13b774e6d7484c9ed1599cc9131c664ec58d4c64db6eb9268R114

packages/runtime/src/state/makeswift-api-client.ts
```diff
@@ -1,44 +1,98 @@
...
+    case APIResourceType.LocalizedGlobalElement:
+      if (locale == null) {
+        console.error(`Attempt to access ${resourceType} ${resourceId} without a locale`)
+        return null
+      }
+
+      const localizedId = getLocalizedResourceId(state, locale, resourceId)
+      return localizedId != null
+        ? APIResources.getAPIResource(state.apiResources, resourceType, localizedId, locale)
+        : null
```
agurtovoy, line 91: Global to localized element ID translation logic, moved from `MakeswiftHostApiClient`: https://github.com/makeswift/makeswift/pull/836/files#diff-e66ba2364c981fb13b774e6d7484c9ed1599cc9131c664ec58d4c64db6eb9268R114

packages/runtime/src/state/makeswift-api-client.ts
```diff
@@ -64,13 +118,13 @@ async function fetchJson<T>(url: string): Promise<T | null> {
 export function fetchAPIResource<T extends APIResourceType>(
   resourceType: T,
   resourceId: string,
-  locale?: Intl.Locale,
+  locale?: APIResourceLocale<T>,
```
agurtovoy, line 121: "Conditional" locale (string).

packages/runtime/src/state/makeswift-api-client.ts
```diff
@@ -95,16 +149,31 @@ export function fetchAPIResource<T extends APIResourceType>(
       case APIResourceType.LocalizedGlobalElement: {
         if (locale == null) throw new Error('Locale is required to fetch LocalizedGlobalElement')
 
+        // If `getLocalizedResourceId` returns null, it means we have tried to fetch the resource,
+        // but the resource is not available. If we haven't fetched it yet, it'll return undefined.
+        if (getLocalizedResourceId(state, locale, resourceId) === null) {
+          return null
+        }
```
agurtovoy, line 156: Pre-fetch logic, moved from `MakeswiftHostApiClient`: https://github.com/makeswift/makeswift/pull/836/files#diff-e66ba2364c981fb13b774e6d7484c9ed1599cc9131c664ec58d4c64db6eb9268R126

packages/runtime/src/state/makeswift-api-client.ts
```diff
@@ -128,6 +197,43 @@ export type Dispatch = ThunkDispatch<State, unknown, Action>
...
+    return (action: Action): Action => {
+      switch (action.type) {
+        case ActionTypes.SET_LOCALIZED_RESOURCE_ID: {
+          return next(
+            action.payload.locale != null
+              ? action
+              : { ...action, payload: { ...action.payload, locale: defaultLocale } },
+          )
+        }
+      }
```
agurtovoy, line 215: Fill in `SET_LOCALIZED_RESOURCE_ID`'s `locale` argument w/ page locale if missing.

packages/runtime/src/state/mixins/setup-teardown.ts
```diff
@@ -0,0 +1,17 @@
...
+export function withSetupTeardown(
+  setup: () => void,
+  teardown: () => void,
+): StoreEnhancer<SetupTeardownMixin> {
+  return next => (reducer, preloadedState?) => ({
+    ...next(reducer, preloadedState),
+    setup,
+    teardown,
+  })
+}
```
agurtovoy, line 17: Store enchancer extracted from the preview store.

packages/runtime/src/state/modules/__tests__/__snapshots__/api-resources.test.ts.snap
```diff
@@ -293,6 +293,120 @@ exports[`state / APIResources \`Actions.evictApiResource\` correctly removes the
 }
 `;
 
+exports[`state / APIResources \`Actions.updateAPIClientCache\` fills in the state from cache w/o overriding existing resources 1`] = `
```
agurtovoy, line 296: Snapshots for the new `APIResources` test suite.

packages/runtime/src/state/modules/__tests__/__snapshots__/localized-resources-map.test.ts.snap
```diff
@@ -0,0 +1,32 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`state / LocalizedResourcesMap \`Actions.setLocalizedResourceId\` correctly updates the store 1`] = `
+{
```
agurtovoy, line 4: Snapshots for the `LocalizedResourcesMap` tests.

packages/runtime/src/state/modules/read-only-documents.ts
```diff
@@ -20,24 +20,17 @@ export function createDocumentReference(key: string): DocumentReference {
 export type Document = {
   key: string
   rootElement: Element
+  locale: string | null
```
agurtovoy, line 23: Locale at the document level allows us to have multiple-language sections on the same page.

packages/runtime/src/utils/branded.ts
```diff
@@ -1,5 +1,5 @@
 // FIXME: copy & paste from the builder repo, we need a `lib` package
 declare const __brand: unique symbol
-type Brand<B> = { [__brand]: B }
+export type Brand<B> = { [__brand]: B }
```
agurtovoy, line 3: Can't use a branded type in a reducer that is passed to the `combineReducers` call w/o having to export this.

packages/runtime/src/state/modules/read-only-documents.ts
```diff
@@ -20,24 +20,17 @@ export function createDocumentReference(key: string): DocumentReference {
...
-  rootElements = new Map(),
-}: { rootElements?: Map<string, Element> } = {}): State {
-  const initialState = new Map()
-
-  rootElements.forEach((rootElement, documentKey) => {
-    initialState.set(documentKey, createDocument(documentKey, rootElement))
-  })
-
-  return initialState
+export function getInitialState({ documents = [] }: { documents?: Document[] } = {}): State {
```
agurtovoy, line 32: Initial state is directly seeded with docu

packages/runtime/src/state/react-builder-preview.ts
```diff
@@ -569,11 +570,6 @@ export function messageChannelMiddleware(
             window.getSelection()?.removeAllRanges()
             break
 
-          case ActionTypes.SET_LOCALIZED_RESOURCE_ID: {
-            client.setLocalizedResourceId(action.payload)
-            break
```
agurtovoy, line 574: Moved `SET_LOCALIZED_RESOURCE_ID` logic here: https://github.com/makeswift/makeswift/pull/836/files#diff-77b56149aa218322fa2db5bbccfd0e1e813309d76ea228b2790afb409dc3ead7R34

agurtovoy: `makeswiftApiClientSyncMiddleware` ensures that the action gets forwarded to the API client store.

packages/runtime/src/state/modules/read-only-documents.ts
fikrikarim, line 23: Unfortunately, this `Document` type is used in some actions as the arg and the payload, for example, the `registerDocument` and the `registerBuilderDocument`:
```ts
type RegisterDocumentAction = {
  type: typeof ActionTypes.REGISTER_DOCUMENT
  payload: { documentKey: string; document: Document }
}
```
Perhaps we need to decouple the type of the state and the actions, so that we don't accidentally change the payload of the action when we update the state. Otherwise it could cause an error because the builder is expecting a payload that's different on an old runtime.

apps/nextjs-app-router/src/makeswift/provider.tsx
```diff
@@ -1,14 +1,24 @@
...
+  locale = undefined,
+  previewMode = false,
+}: {
+  children: ReactNode
+  locale?: string
+  previewMode: boolean
+}) {
   return (
-    <ReactRuntimeProvider runtime={runtime}>
+    <ReactRuntimeProvider {...{ runtime, previewMode, locale }}>
```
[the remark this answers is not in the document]
agurtovoy: It's not about reducing typing but rather about reducing verbosity/tautology of `runtime={runtime}` etc. It's not a super common pattern out there, but I wouldn't call it very non-idiomatic. `<Component {...props} />` is pretty common, this is just a generalization of that.

packages/runtime/src/api/react.ts
agurtovoy: I didn't see a good reason to construct a `Intl.Locale` instance that we always convert to back a language string anyway. If we want to type the `locale` argument, we can brand the string type.

packages/runtime/src/next/components/page.tsx
```diff
@@ -1,51 +1,82 @@
...
+
+function useCacheData(cacheData: CacheData) {
+  const client = useMakeswiftHostApiClient()
+
+  if (isServer) {
+    client.makeswiftApiClient.dispatch(updateAPIClientCache(cacheData))
+  }
+
+  useEffect(() => {
+    client.makeswiftApiClient.dispatch(updateAPIClientCache(cacheData))
```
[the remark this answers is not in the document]
agurtovoy: Can you clarify what do you mean by "unconditionally"? It's dispatched only if `cacheData` (or the client) has changed, and we do want to update the cache in the Redux store every time it's updated from the outside.

apps/nextjs-pages-router/src/pages/[[...path]].tsx
Comment (agurtovoy) [reply apps/nextjs-pages-router/src/pages/[[...path]].tsx:56]: @migueloller The result of this method is what's passed as props to the `App` [here](https://github.com/makeswift/makeswift/pull/836/files#diff-7726b5511392f97ec99ec7c5c5727dda11bd3d5c8d01621b2e11dabb4c9cfa20R7), isn't it?

apps/nextjs-app-router/src/app/[lang]/[[...path]]/layout.tsx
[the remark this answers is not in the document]
Comment (agurtovoy) [reply apps/nextjs-app-router/src/app/[lang]/[[...path]]/layout.tsx:22]: Yes, we lifted up `MakeswiftHostApiClient` to `ReactRuntimeProvider`, so it can't use the snapshot's locale as the default anymore. It shouldn't be needing a default locale at all, but first we need to adjust the builder to always pass a locale when performing resource mutations.

See also https://github.com/makeswift/makeswift/pull/836/files#diff-701ba0ed420276bea2f613b55a022aecc9f08aca9e519242cab81445ee070c7dR206 below.

packages/runtime/src/state/actions.ts
agurtovoy: It's dispatched here: https://github.com/makeswift/makeswift/pull/836/files#diff-574436562edd329137174ba80e6ecdb186f78f8e2b7e6cbc929afe8ede60e4e9R21

apps/nextjs-app-router/src/app/[lang]/layout.tsx
Comment (agurtovoy) [inline apps/nextjs-app-router/src/app/[lang]/layout.tsx:1]: Moved under `app/[lang]` to be able to retrieve page language/locale:

next.js app router internationalization docs

apps/nextjs-pages-router/src/pages/_document.tsx
```diff
@@ -1 +1,7 @@
-export { Document as default } from '@makeswift/runtime/next'
+import { Document } from '@makeswift/runtime/next'
+
+export default class MyDocument extends Document {
+  static defaultProps = {
+    appOrigin: process.env.MAKESWIFT_APP_ORIGIN,
```
agurtovoy, line 5: Required to make the pages router work w/ a local builder.

packages/runtime/src/next/components/page.tsx
```diff
@@ -1,51 +1,68 @@
...
+  }, [cacheData, client])
+}
+
+function useRegisterPageDocument(pageDocument: MakeswiftPageDocument): ReactPage.Document {
+  const dispatch = useDispatch()
+
+  const rootDocument = pageToRootDocument(pageDocument)
+  useEffect(() => dispatch(registerDocumentsEffect([rootDocument])), [rootDocument])
+  return rootDocument
+}
```
agurtovoy, line 44: @migueloller Is this more along the lines of what you envisioned re: moving application logic out of React?

packages/runtime/src/next/document.tsx
```diff
@@ -36,7 +38,7 @@ export class Document extends NextDocument {
     return (
       <Html>
         <Head>
-          <PreviewModeScript isPreview={isPreview} />
+          <PreviewModeScript isPreview={isPreview} appOrigin={this.props.appOrigin} />
```
agurtovoy, line 41: Support running Pages router apps in local builder.

packages/runtime/src/runtimes/react/components/DocumentRoot.tsx
```diff
@@ -0,0 +1,25 @@
...
+type Props = {
+  rootDocument: Document
+}
+
+export const DocumentRoot = memo(
+  forwardRef(function DocumentRoot(
+    { rootDocument }: Props,
+    ref: Ref<ElementImperativeHandle>,
+  ): JSX.Element {
+    const document = useDocument(rootDocument.key) ?? rootDocument
```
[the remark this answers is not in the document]
agurtovoy: @migueloller We don't register anything here, but rather try to retrieve the (latest version of) document from the state by its key and fall back to the initial version if there is none (yet). 

If that's what you meant, can you clarify your comment about "keeping concerns separate"?

packages/runtime/src/state/modules/read-only-documents.ts
agurtovoy: Yes.

agurtovoy: Decoupled `Document` from document payload 🙏.

apps/nextjs-app-router/src/makeswift/provider.tsx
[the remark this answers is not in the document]
migueloller: I understand. I guess stylistically it's something I rarely see and I had to do a double-take to realize what was going on. When I said "non-idiomatic" this is more what I mean—that it's something not commonly done. It's a nit, so not a hill I will die on.

packages/runtime/src/api/react.ts
migueloller: That's fair. We were trying to use "parsed" types like `Intl.Locale` or `URL` whenever we were using a string that could be better represented with one of these platform APIs. I think that's stronger than a branded type, in particular because a branded type is only useful once you've done some sort of validation that what you've got is validated or parsed and it can be branded. `new Intl.Locale()` or `parseLocale(locale: string): BrandedLocale` somewhat achieve the same thing but it feels better to "use the platform".

I think right now the validation for locales is "does calling `Intl.Locale` throw?" and we want to validate user-provided values somehow. We might change it to something else (we've had some issues with `Intl.Locale`, like it changing `en-UK` to `en-GB`) but when we do it will have to be across the board (i.e., the builder and API as well).

packages/runtime/src/next/components/page.tsx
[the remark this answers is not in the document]
migueloller: Effects will run at least once so when this component mounts it will dispatch the `updateAPIClientCache` action (twice in `StrictMode` during development). This might result in unnecessary JavaScript execution for live pages, which could impact performance, unless the reducer is smart enough to make this effectively a no-op and we're only paying the price of dispatching the action and reducers processing it but returning the current state (thus there are no re-renders).

apps/nextjs-pages-router/src/pages/[[...path]].tsx
Comment (migueloller) [reply apps/nextjs-pages-router/src/pages/[[...path]].tsx:56]: Yes, I see this now. We can move this discussion to a separate thread re: `_app` because, while I won't go as far as calling it an anti-pattern, I consider this sort of stuff (reading props from a page on `_app`) undesirable in Next.js Pages Router.

apps/nextjs-app-router/src/app/[lang]/[[...path]]/layout.tsx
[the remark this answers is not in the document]
Comment (migueloller) [reply apps/nextjs-app-router/src/app/[lang]/[[...path]]/layout.tsx]: Great. Do we have a comment on `MakeswiftProvider`'s props as well so we remember to remove it from there in addition to removing the middleware?

apps/nextjs-pages-router/src/pages/_document.tsx
migueloller: This is interesting. How come we're setting it as a static property on `Document` vs passing it as a prop to a component or as a parameter to a Makeswift constructor

packages/runtime/src/runtimes/react/components/DocumentRoot.tsx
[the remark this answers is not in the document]
migueloller: I see. I wonder if we should have `useDocument` take the "fallback" document as an argument similar to how `useSyncExternalStore`, `useState`, `useActionState`, and other React hooks take a default value as a parameter.

As for "keep concerns separate" I thought that `useDocument` was registering the document (I see now that's impossible since it just takes a key) so I was alluding to separating registration from reading.

Where does registration happen now?

apps/nextjs-pages-router/src/pages/_document.tsx
migueloller: I'm seeing now this is using the `defaultProps` functionality from React, not something custom we're making. I think in the past we'd read directly from `process.env.MAKESWIFT_APP_ORIGIN` when destructuring the props. Or was this simply not working before?

packages/runtime/src/api/react.ts
migueloller, line 1: The main changes in this file since my last review was making the methods take named parameters I'm guessing?

packages/runtime/src/next/components/page.tsx
migueloller: Yes, for sure. In this case, though, I'd move the `pageToRootDocument` inside the effect and make the dependency `pageDocument` so that `pageToRootDocument` doesn't have to be memoized or attempt to preserve the reference equality of `pageDocument`. You could alternatively use `useMemo`.

migueloller, line 44: What do you think if we centralized the "hack" of server-side registration here? That way the hook works both in the client and the server?

packages/runtime/src/next/document.tsx
migueloller: I see now what you're doing with the document's `static defaultProps`. Seems like the only choice before was to use a custom `Document` and passing it in as a prop. I guess this allows to extend `Document` without having to re-implement the `render()` method if all you want to do is add a custom app origin?

migueloller (approved):

apps/nextjs-pages-router/src/pages/_document.tsx
agurtovoy: `Document` is a class component, defining `appOrigin` in `defaultProps` is effectively passing it as a prop. That prop is passed through to `<PreviewModeScript>` here: https://github.com/makeswift/makeswift/pull/836/files#diff-2d843b17408d4e8bbb89a142b11260df343c89c7a2c715a2f5153e78d8c26ffeR41

packages/runtime/src/api/react.ts
agurtovoy: I guess I don't buy into using "parsed" objects as a better representation of primitive values in a React app where the default method of value comparison is `Object.is`. I'd rather have a validated primitive value up until the point where we need the "parsed" type functionality, if there is such a point.

packages/runtime/src/next/components/page.tsx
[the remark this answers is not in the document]
agurtovoy: How would we dispatch it conditionally here? 

In this case, the reducer will make the dispatch of cache data identical to what's been prepopulated on the server a no-op, as it only fills in non-existent state: https://github.com/makeswift/makeswift/pull/836/files#diff-cbd5599cdef9ce1b5d413cd682e2d662f33071c2b64be862ecc43799bce436faR105

packages/runtime/src/next/document.tsx
agurtovoy: > I guess this allows to extend `Document` without having to re-implement the `render()` method if all you want to do is add a custom app origin?

Yes.

packages/runtime/src/next/components/page.tsx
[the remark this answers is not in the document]
migueloller: I would perhaps have a thunk that checks if we're in editable mode or not before dispatching so that you can dispatch the thunk unconditionally and it would conditionally dispatch the action.

As for the cache data being identical, it still seems to create a new instance of the state's objects, eventually resulting in a re-render, if I'm reading things correctly.

Perhaps the ideal approach is not to have to sync the data in the first place and have it happen at the place the data is fetch. Unfortunately I don't think this is possible given the server-client boundary present in Next.js. If we fetched the data in the server and populated the API cache there and the only way to pass that to the client is via the `snapshot` prop, we have no choice but to perform an effect on a component to then pull the cache data out of the `snapshot` and insert it into the _client's_ API cache 🤷🏻‍♂️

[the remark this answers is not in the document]
agurtovoy: > As for the cache data being identical, it still seems to create a new instance of the state's objects, eventually resulting in a re-render, if I'm reading things correctly.

It does not:

reducer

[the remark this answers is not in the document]
migueloller: Ah, my bad

PR #839, opened 2024-09-24
feat: add optional locale scope to the API resources state

agurtovoy:
This enables us to store "localized" API resources such as `PagePathnameSlice`/`LocalizedGlobalElement` across several different locales. No breaking changes.


packages/runtime/src/api/fragments.ts
```diff
@@ -0,0 +1,5 @@
+import { DocumentNode } from 'graphql'
```
agurtovoy, line 1: Lifted up `fragments` and `types` from the `graphql` dir to `api` to draw a distinction between GraphQL's generated types and runtime API types, with the latter being an extension of the former (see below).

packages/runtime/src/api/index.ts
```diff
@@ -1,2 +1,2 @@
-export * from './graphql/types'
-export * from './graphql/fragments'
+export * from './types'
+export * from './fragments'
```
agurtovoy, line 2: See my comment above.

packages/runtime/src/api/resource-types.ts
```diff
@@ -1,10 +0,0 @@
-import { z } from 'zod'
-
-const pagePathnameSliceSchema = z.object({
-  id: z.string(),
-  pathname: z.string(),
-  localizedPathname: z.string().optional().nullable(),
-  __typename: z.literal('PagePathnameSlice'),
-})
-
-export type PagePathnameSlice = z.infer<typeof pagePathnameSliceSchema>
```
agurtovoy, line 10: This was replaced by a simpler extension of the `PagePathnameSliceFragment` definition in `api/types.ts` below.

packages/runtime/src/api/types.ts
```diff
@@ -2,14 +2,19 @@ import {
...
   SnippetFragment as Snippet,
   PageFragment as Page,
   SiteFragment as Site,
-} from './generated/types'
+} from './graphql/generated/types'
+
+type PagePathnameSlice = PagePathnameSliceFragment & {
+  basePageId?: string
+  localizedPathname?: string | null
+}
```
agurtovoy, line 17: Reflect the actual type we're operating with in the runtime.

packages/runtime/src/api/types.ts
```diff
@@ -49,4 +54,4 @@ export const APIResourceType: { [R in APIResource as R['__typename']]: R['__type
   Site: 'Site',
 }
 
-export type APIResourceType = typeof APIResourceType[keyof typeof APIResourceType]
+export type APIResourceType = (typeof APIResourceType)[keyof typeof APIResourceType]
```
agurtovoy, line 57: Prettier change.

packages/runtime/src/next/components/tests/controls/__snapshots__/link-control.test.tsx.snap
```diff
@@ -260,14 +260,14 @@ exports[`Page Link control data { type: 'SEND_EMAIL', payload: [Object] }: snaps
 }
 `;
 
-exports[`Page Link control data OPEN_PAGE fetches pathname slice when not in cache: resolvedValue 1`] = `
+exports[`Page Link control data OPEN_PAGE for locale fr-FR fetches pathname slice when not in cache: resolvedValue 1`] = `
```
agurtovoy, line 263: Updated Link control rendering test to include rendering localized links.

packages/runtime/src/next/components/tests/controls/link-control.test.tsx
```diff
@@ -15,12 +15,19 @@ const Button = forwardRef(function Button({ id }: { id?: string }, _ref) {
...
   id: pageId,
-  pathname: 'fake-pathname',
+  pathname: 'test-pathname',
+}
+
+const localizedPagePathnameSlice = {
+  __typename: APIResourceType.PagePathnameSlice,
+  id: pageId,
+  basePageId: '[test-base-page-id]',
+  pathname: 'fr-FR/test-localized-pathname',
```
agurtovoy, line 30: Updated Link control rendering test to include rendering of localized links.

packages/runtime/src/next/components/tests/controls/page-control-prop-rendering.tsx
```diff
@@ -43,6 +44,7 @@ export async function testPageControlPropRendering<D extends ControlDefinition>(
   }: {
     toData?: (value: ValueType<D>) => DataType<D>
     value: ValueType<D> | undefined
+    locale?: string | null
```
agurtovoy, line 47: Add support for specifying a snapshot locale in a control rendering test.

packages/runtime/src/runtimes/react/hooks/makeswift-api.ts
```diff
@@ -6,11 +6,12 @@ import {
   File,
   GlobalElement,
   LocalizedGlobalElement,
+  PagePathnameSlice,
   Swatch,
   Table,
   Typography,
 } from '../../../api'
-import { PagePathnameSlice } from '../../../api/resource-types'
```
agurtovoy, line 13: See my comment above.

packages/runtime/src/state/actions.ts
```diff
@@ -239,12 +239,12 @@ type MessageBuilderPropControllerAction<T = PropControllerMessage> = {
 
 type ChangeAPIResourceAction = {
   type: typeof ActionTypes.CHANGE_API_RESOURCE
-  payload: { resource: APIResource }
+  payload: { resource: APIResource; locale: string | null }
```
agurtovoy, line 242: Added an optional `locale` argument to actions operating on the API resources state.

packages/runtime/src/state/actions.ts
```diff
@@ -616,12 +621,15 @@ export function messageBuilderPropController<T>(
   }
 }
 
-export function changeApiResource(resource: APIResource): ChangeAPIResourceAction {
-  return { type: ActionTypes.CHANGE_API_RESOURCE, payload: { resource } }
+export function changeApiResource(
+  resource: APIResource,
+  locale?: string | null,
```
agurtovoy, line 626: Here and below, keeping the `locale` argument optional for now for backward compatibility.

packages/runtime/src/state/modules/__tests__/__snapshots__/api-resources.test.ts.snap
```diff
@@ -0,0 +1,309 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`state / APIResources \`Actions.apiResourceFulfilled\` correctly populates the store 1`] = `
```
agurtovoy, line 3: Brand new snapshot for a brand new test, see below.

packages/runtime/src/state/modules/__tests__/api-resources.test.ts
```diff
@@ -0,0 +1,267 @@
...
+
+const site: Site = {
+  __typename: APIResourceType.Site,
+  id: '[test-site-id]',
+  googleFonts: {
+    edges: [],
+  },
+}
+
+describe('state / APIResources', () => {
```
agurtovoy, line 115: Test API resources actions/reducer.

packages/runtime/src/state/modules/api-resources.ts
```diff
@@ -1,13 +1,26 @@
...
+function compositeId(resourceId: string, locale?: string | null): CompositeResourceId {
+  return (locale != null ? `${resourceId}@${locale}` : resourceId) as CompositeResourceId
+}
+
+function parseCompositeId(compositeId: CompositeResourceId): [string, string | undefined] {
+  const [resourceId, locale] = compositeId.split('@')
+  return [resourceId, locale]
+}
+
+type State = Map<APIResourceType, Map<CompositeResourceId, APIResource | null>>
```
agurtovoy, line 17: Use "composite" resource ID (resource ID + locale) to make it possible to store "localized" API resources such as `PagePathnameSlice`/`LocalizedGlobalElement` across several different locales.

agurtovoy: This is a backward-compatible change.

arvinpoddar: My gut reaction says that this should be done with another level of Maps, where the keys are the locales or `null`, but I also assume you had a reason to favor this path. Can you fill me in?

agurtovoy: @arvinpoddar Good question, the main reason for "flattening" the `Map` hierarchy here is that the extra level of nesting makes both the state updates and the serialization/deserialization code much more verbose/harder to understand. 

The state updates part can probably be redeemed if we pull in [Immer](https://immerjs.github.io/immer/), but serialization/deserialization will still be messier w/ the extra level of the nesting.

arvinpoddar: Yeah, writing state updates with 2+ degrees of nesting is painful. I'm wondering if this causes us to lose the ability to cleanly/efficiently write certain selectors for resources, such as if my selector wants to retrieve from state using only one part of the composite ID. My imagination is failing me with examples of where it would be awful, but just something to think about. I'm not really opposed to this.

packages/runtime/src/state/modules/api-resources.ts
```diff
@@ -1,13 +1,26 @@
...
+import { Branded } from '../../utils/branded'
 
-type State = Map<APIResourceType, Map<string, APIResource | null>>
+type CompositeResourceId = Branded<string, 'CompositeResourceId'>
+
+function compositeId(resourceId: string, locale?: string | null): CompositeResourceId {
+  return (locale != null ? `${resourceId}@${locale}` : resourceId) as CompositeResourceId
+}
+
+function parseCompositeId(compositeId: CompositeResourceId): [string, string | undefined] {
```
arvinpoddar, line 12: Could the return type be `[string, string | null]`?

packages/runtime/src/state/modules/api-resources.ts
agurtovoy: We can always rewrite this if ID composition becomes an issue in the future, it's just an implementation detail. I do like how little code I had to change here comparing to the alternative.

packages/runtime/src/state/modules/api-resources.ts
agurtovoy: For consistency with `compositeId` signature? We can do that, although `locale` in `compositeId` can also be `undefined`, and the intentional conversion to `null` here might make it looks like it matter which one it is, when the semantics is the same for either.

packages/runtime/src/api/types.ts
```diff
@@ -24,29 +29,43 @@ export type {
   Site,
 }
 
+export type LocalizableAPIResource = PagePathnameSlice | LocalizedGlobalElement
```
agurtovoy, line 32: Distinguish between "localizable" and non-localizable resources.

packages/runtime/src/state/actions.ts
```diff
@@ -639,14 +647,15 @@ export function handlePointerMove(payload: {
   return { type: ActionTypes.HANDLE_POINTER_MOVE, payload }
 }
 
-export function apiResourceFulfilled(
-  resourceType: APIResourceType,
+export function apiResourceFulfilled<T extends APIResourceType>(
+  resourceType: T,
   resourceId: string,
   resource: APIResource | null,
+  locale?: APIResourceLocale<T>,
```
agurtovoy, line 654: Disallow passing a locale for non-localizable resources.

packages/runtime/src/state/modules/__tests__/api-resources.test.ts
```diff
@@ -0,0 +1,303 @@
...
+      APIResourceType.Typography,
+      APIResourceType.GlobalElement,
+      APIResourceType.Table,
+      APIResourceType.Snippet,
+      APIResourceType.Page,
+      APIResourceType.Site,
+    ])('getter for %s', resourceType => {
+      test.each(['fr-FR', null, undefined])('fails to compile with locale %s', locale => {
+        //@ts-expect-error
+        expect(APIResources.getHasAPIResource(state, resourceType, '', locale)).toBe(false)
```
agurtovoy, line 282: @migueloller Following up on your comment earlier, I disallowed passing a locale for non-localizable resources. You can still NOT pass a locale when calling a method for a localizable resource because disallowing that, while possible, breaks down function composition downstream.

packages/runtime/src/state/modules/api-resources.ts
```diff
@@ -1,13 +1,28 @@
...
+  return [resourceId, locale]
+}
+
+type State = Map<APIResourceType, Map<CompositeResourceId, APIResource | null>>
 
 export type SerializedState = {
   [key in APIResourceType]?: {
     id: string
     value: Extract<APIResource, { __typename: key }> | null
+    locale?: APIResourceLocale<key>
```
agurtovoy, line 25: Disallow specifying a locale for non-localizable resources.

packages/runtime/src/state/modules/__tests__/api-resources.test.ts
agurtovoy: Example error:

typography locale error

packages/runtime/src/api/resource-types.ts
migueloller: Are the types equivalent?

packages/runtime/src/next/components/tests/controls/link-control.test.tsx
```diff
@@ -59,10 +66,13 @@ describe('Page', () => {
     },
   )
 
-  describe('Link control data OPEN_PAGE', () => {
+  describe.each([null, 'fr-FR'])('Link control data OPEN_PAGE for locale %s', locale => {
+    const pathNameSlice = locale ? localizedPagePathnameSlice : pagePathnameSlice
```
migueloller, line 70:
```suggestion
    const pathnameSlice = locale ? localizedPagePathnameSlice : pagePathnameSlice
```

Note, if you accept the suggestion above, you'd also need to update other references to `pathNameSlice` in this file

packages/runtime/src/state/actions.ts
migueloller: Should `locale` here be an optional field? In particular, is this action ever constructed by the builder (it's unfortunate I even have to ask this question but we are reusing these actions there... ideally we'd have a `Builder` and `Host` interface that encapsulates the parts of the builder and host that talk to each other, but I digress)?

migueloller: In particular, when we handle the action payload in a reducer, the action received from the builder might not include the field at all so it wouldn't be neither `string` or `null`, it would be `undefined`

packages/runtime/src/state/modules/api-resources.ts
migueloller: I think I like @arvinpoddar's suggestion. To make the semantics be the same, the implementation could look like this:

```ts
  const [resourceId, locale = null] = compositeId.split('@')
  return [resourceId, locale]
```

packages/runtime/src/state/modules/api-resources.ts
migueloller: I would love to come up with a Map data structure that does this for us. We will be having some pretty gnarly composite keys for element trees where the key is a combination of `type`, `key`, `locale`, and `siteVersion`. Being able to do something like `map.set([type, key, locale, siteVersion], elementTree)` and transparently get good serialization/deserialization, debug visuals of keys, etc. would be great.

migueloller: Also in this case it would allow us to flatten it and avoid the extra layer with the API resource type

packages/runtime/src/state/modules/api-resources.ts
```diff
@@ -48,11 +63,12 @@ export function getSerializedState(state: State): SerializedState {
   }
 
   Array.from(state.entries()).forEach(([resourceType, resources]) => {
-    const particularResourceMap: { id: string; value: any }[] = []
+    const particularResourceMap: { id: string; value: any; locale?: any }[] = []
```
migueloller, line 66: Unintended `any` here?

packages/runtime/src/utils/branded.ts
migueloller, line 1: What if instead of introducing a branded utility we used Zod since the runtime already depends on it? https://zod.dev/?id=brand

migueloller (approved):
Looking great! I absolutely love the new tests ❤️

packages/runtime/src/state/modules/api-resources.ts
agurtovoy: Some prior art on this: https://hackernoon.com/composite-keys-a-guide-on-how-to-handle-them

packages/runtime/src/utils/branded.ts
agurtovoy: I actually considered it, but defining a branded type w/ Zod is more verbose since you first need to define a schema, which we won't use, and it felt like a wrong pattern to establish:

```typescript
const compositeResourceIdSchemae = z.string().brand<'CompositeResourceId'>()
type CompositeResourceId = z.infer<typeof compositeResourceIdSchemae>
```

vs.

```typescript
type CompositeResourceId = Branded<string, 'CompositeResourceId'>
```

packages/runtime/src/state/modules/api-resources.ts
agurtovoy: No, this is intentional. Conditional typing of `locale` in `SerializedState` (`APIResourceLocale<key>`) makes TS to reject anything but `any` here or forces us to add a type cast to `resourceMap[resourceType] = particularResourceMap` in line 75 below. I opted in for `any` as the lesser of two evils.

packages/runtime/src/state/actions.ts
agurtovoy: I have `locale` as optional in the action(s) "constructor(s)", which converts `undefined` to `null`: https://github.com/makeswift/makeswift/pull/839/files#diff-f587b032a852ddc3abcdb9915e263b0a3b74536e71fc024e97e0a13623858cb9R658

migueloller: Right, but the current builder dependency doesn't import that action creator so it will send `undefined` until the dependency is upgraded. If we parsed actions at runtime with Zod, this action would fail to parse if a host used this version of the runtime but the builder wasn't updated yet.

agurtovoy: Ah, I see, good point.

packages/runtime/src/api/resource-types.ts
agurtovoy: No, the new type is more correct, the Zod type above is actually missing the `basePageId` prop.

PR #841, opened 2024-09-25
fix: validate component type at runtime

arvinpoddar:
CleanShot 2024-09-25 at 19 03 16

### DIFF
--- a/packages/runtime/src/runtimes/react/react-runtime.ts
+++ b/packages/runtime/src/runtimes/react/react-runtime.ts
@@ -9,6 +9,15 @@ import type { ComponentType } from '../../state/react-page'
 
 import { RuntimeCore } from './runtime-core'
 
+function validateComponentType(type: string, component?: ComponentType): void {
+  const componentName = component?.name ?? 'Component'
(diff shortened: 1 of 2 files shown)


packages/runtime/src/runtimes/react/react-runtime.ts
```diff
@@ -9,6 +9,15 @@ import type { ComponentType } from '../../state/react-page'
 
 import { RuntimeCore } from './runtime-core'
 
+function validateComponentType(type: string, component?: ComponentType): void {
+  const componentName = component?.name ?? 'Component'
+  if (typeof type !== 'string') {
```
agurtovoy, line 14: Do we want to check that the string is not empty as well?

migueloller: I think we probably should disallow empty strings, yeah. A user would probably not want to do that intentionally

migueloller (dismissed):

packages/runtime/src/runtimes/react/react-runtime.ts
arvinpoddar: Addressed

PR #843, opened 2024-09-27
Add unstable getComponentSnapshot and MakeswiftComponent

fikrikarim:


packages/runtime/src/state/modules/embedded-components.ts
```diff
@@ -0,0 +1,33 @@
...
+  embeddedComponents = new Map(),
+}: { embeddedComponents?: Map<string, EmbeddedComponent> } = {}): State {
+  return embeddedComponents
+}
+
+export function getEmbeddedComponents(state: State): Map<string, EmbeddedComponent> {
+  return state
+}
+
+export function getEmbeddedComponent(state: State, key: string): EmbeddedComponent | null {
```
arvinpoddar, line 24: Could we have a more specific name for the `key` param here? I assume the `key` is the unique combination of `type`, `key`, `locale`, etc.

packages/runtime/src/utils/deterministic-uuid.ts
```diff
@@ -0,0 +1,32 @@
+import crypto from 'crypto'
+import { canonicalize } from './canonicalize'
+
+/**
+ * Generates a deterministic UUID using SHA-256.
+ * @param value - The value to generate the UUID from.
+ * @param secretNamespace - A secret key used as the HMAC namespace (optional).
+ * @returns A deterministic UUID string.
+ */
+export function deterministicUUID(value: any): string {
```
arvinpoddar, line 10: Could this be done with namespaced uuid V5?

packages/runtime/src/state/modules/embedded-components.ts
fikrikarim: Good question. I'm using the generated `uuid` as the key here. Note that this is different than the `encoded` base64 key that we're using on the builder<>orion.

What do you think we should use to make this more clear?

packages/runtime/src/utils/deterministic-uuid.ts
fikrikarim: uuidv5 is using SHA-1. Because we need to make this unique per site, we need to also seed the siteId. But we don't have access to the siteId when the returned snapshot is 404. So I'm using x-site-api-key here.

Because I'm using x-site-api-key, I'm using the more secure hash instead of SHA-1. Do you think it's overkill and SHA-1 is secure enough?

arvinpoddar: Ah, good thought. Personally, I'd be more concerned with keeping our codebase simple and robust over securing against the likelihood that someone uses a hash collision attack to crunch API key — that's a pretty computationally/time costly attack vector. That being said, this is a pretty well encapsulated utility function, so I'm not concerned either way.

packages/runtime/src/state/modules/embedded-components.ts
fikrikarim: Thank you. We severely need a naming convention. Please let me know if I'm not following that. In this case. We're registering the embedded based on the `documentKey`. So, in this case, does it make sense to use `documentKey` here?

packages/runtime/src/utils/deterministic-uuid.ts
fikrikarim: I totally agree with that. I just don't want to accidentally expose a security leak because we hash the API key. Let's keep this for now. Let me know if there's a more concise yet secure way to hash this.

fikrikarim: Oh. We could also not use the whole API key. For example, we could use the first half so we don't have to worry about the API key leaking. I'll update that.

fikrikarim: Updated to use uuid v5

packages/runtime/src/state/modules/embedded-components.ts
fikrikarim: Updated to `documentKey`

packages/runtime/src/state/actions.ts
```diff
@@ -297,6 +301,25 @@ type SetLocalizedResourceIdAction = {
...
+type RegisterBuilderEmbeddedComponentAction = {
+  type: typeof ActionTypes.REGISTER_BUILDER_EMBEDDED_COMPONENT
+  payload: {
+    documentKey: string
+    embeddedComponent: {
+      documentKey: string
+      key: string
+      type: string
+      locale: string | null
+      rootElement: Element
```
fikrikarim, line 313: Do we want to save and send the `rootElement`?

packages/runtime/src/state/actions.ts
```diff
@@ -297,6 +301,25 @@ type SetLocalizedResourceIdAction = {
   payload: { resourceId: string; localizedResourceId: string | null }
 }
 
+type RegisterBuilderEmbeddedComponentAction = {
+  type: typeof ActionTypes.REGISTER_BUILDER_EMBEDDED_COMPONENT
+  payload: {
+    documentKey: string
```
fikrikarim, line 307: drop this

packages/runtime/src/next/client.ts
```diff
@@ -569,6 +574,70 @@ export class Makeswift {
...
+      })
+
+      const emptyElementData: ElementData = {
+        key: generatedKey,
+        type: type,
+        props: {},
+      }
+
+      return {
+        document: { id: generatedKey, type, key, locale, data: emptyElementData },
```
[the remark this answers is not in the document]
fikrikarim: We could use just uuid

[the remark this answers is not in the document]
fikrikarim: try applying create op on the host on non null doc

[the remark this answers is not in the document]
fikrikarim: Let's continue to use generated key.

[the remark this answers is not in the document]
fikrikarim: Should we just use the id and insert the id?

We don't want to share the documentKey and the

PR #845, opened 2024-10-04
debug(runtime): configure Redux Devtools logging for all `runtime` stores

agurtovoy:
Logging to Redux Devtools is enabled in development only.

https://github.com/user-attachments/assets/a56f09fe-2e8c-485d-b049-14013453dca0

### DIFF
--- a/packages/runtime/package.json
+++ b/packages/runtime/package.json
@@ -132,6 +132,7 @@
     "@makeswift/next-plugin": "workspace:*",
     "@makeswift/prop-controllers": "workspace:*",
     "@popmotion/popcorn": "^0.4.4",
+    "@redux-devtools/extension": "^3.3.0",
     "@types/is-hotkey": "^0.1.7",
     "@types/use-sync-external-store": "^0.0.3",
     "@types/uuid": "^9.0.1",
--- a/packages/runtime/src/runtimes/react/components/LiveProvider.tsx
+++ b/packages/runtime/src/runtimes/react/components/LiveProvider.tsx
@@ -15,7 +15,12 @@ type Props = {
 export default function LiveProvider({ client, children, rootElements }: Props): JSX.Element {
   const runtime = useReactRuntime()
   const store = useMemo(
-    () => ReactPage.configureStore({ preloadedState: runtime.store.getState(), rootElements }),
+    () =>
+      ReactPage.configureStore({
+        name: 'Host store',
+        preloadedState: runtime.store.getState(),
+        rootElements,
+      }),
     [rootElements, runtime],
   )
 
--- a/packages/runtime/src/state/makeswift-api-client.ts
+++ b/packages/runtime/src/state/makeswift-api-client.ts
@@ -1,8 +1,11 @@
-import { applyMiddleware, createStore, Store as ReduxStore } from 'redux'
-import thunk, { ThunkAction, ThunkDispatch } from 'redux-thunk'
+import { applyMiddleware, createStore, type Store as ReduxStore } from 'redux'
+import thunk, { type ThunkAction, type ThunkDispatch } from 'redux-thunk'
+import { composeWithDevToolsDevelopmentOnly } from '@redux-devtools/extension'
+
+import { serializeState } from '../utils/serializeState'
 
 import * as APIResources from './modules/api-resources'
-import { Action, apiResourceFulfilled } from './actions'
+import { Action, ActionTypes, apiResourceFulfilled } from './actions'
 import {
   APIResource,
   APIResourceType,
@@ -129,5 +132,20 @@ export type Dispatch = ThunkDispatch<State, unknown, Action>
 export type Store = ReduxStore<State, Action> & { dispatch: Dispatch }
 
 export function configureStore({ serializedState }: { serializedState?: SerializedState }): Store {
-  return createStore(reducer, APIResources.getInitialState(serializedState), applyMiddleware(thunk))
+  const composeEnhancers = composeWithDevToolsDevelopmentOnly({
+    name: `API client store (${new Date().toISOString()})`,
+    serialize: true,
+    stateSanitizer: (state: any) => serializeState(state),
+    actionsDenylist: [
+      ActionTypes.BUILDER_POINTER_MOVE,
+      ActionTypes.HANDLE_POINTER_MOVE,
+      ActionTypes.ELEMENT_FROM_POINT_CHANGE,
+    ],
+  })
+
+  return createStore(
+    reducer,
+    APIResources.getInitialState(serializedState),
+    composeEnhancers(applyMiddleware(thunk)),
+  )
 }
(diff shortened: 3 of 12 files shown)


packages/runtime/package.json
```diff
@@ -132,6 +132,7 @@
     "@makeswift/next-plugin": "workspace:*",
     "@makeswift/prop-controllers": "workspace:*",
     "@popmotion/popcorn": "^0.4.4",
+    "@redux-devtools/extension": "^3.3.0",
```
agurtovoy, line 135: https://github.com/reduxjs/redux-devtools/tree/main/extension#13-use-redux-devtoolsextension-package-from-npm

packages/runtime/src/runtimes/react/components/LiveProvider.tsx
```diff
@@ -15,7 +15,12 @@ type Props = {
 export default function LiveProvider({ client, children, rootElements }: Props): JSX.Element {
   const runtime = useReactRuntime()
   const store = useMemo(
-    () => ReactPage.configureStore({ preloadedState: runtime.store.getState(), rootElements }),
+    () =>
+      ReactPage.configureStore({
+        name: 'Host store',
```
agurtovoy, line 20: Give each store a name so we can differentiate them more easily in the devtools UI.

packages/runtime/src/state/makeswift-api-client.ts
```diff
@@ -129,5 +132,20 @@ export type Dispatch = ThunkDispatch<State, unknown, Action>
 export type Store = ReduxStore<State, Action> & { dispatch: Dispatch }
 
 export function configureStore({ serializedState }: { serializedState?: SerializedState }): Store {
-  return createStore(reducer, APIResources.getInitialState(serializedState), applyMiddleware(thunk))
+  const composeEnhancers = composeWithDevToolsDevelopmentOnly({
+    name: `API client store (${new Date().toISOString()})`,
```
agurtovoy, line 136: Because we instantiate the API client in a React hook, multiple copies of this store will be created when running in dev / React Strict Mode; timestamp the store name so we can find the most recent one more easily.

packages/runtime/src/state/makeswift-api-client.ts
```diff
@@ -129,5 +132,20 @@ export type Dispatch = ThunkDispatch<State, unknown, Action>
...
+    serialize: true,
+    stateSanitizer: (state: any) => serializeState(state),
```
agurtovoy, line 138: Even with `serialize: true`, Redux Devtools doesn't fully show state slices that use `Map` or `Set` to manage the state; use `stateSanitizer` to serialize the state into a plain JSON tree.

packages/runtime/src/state/makeswift-api-client.ts
```diff
@@ -129,5 +132,20 @@ export type Dispatch = ThunkDispatch<State, unknown, Action>
 export type Store = ReduxStore<State, Action> & { dispatch: Dispatch }
 
 export function configureStore({ serializedState }: { serializedState?: SerializedState }): Store {
-  return createStore(reducer, APIResources.getInitialState(serializedState), applyMiddleware(thunk))
+  const composeEnhancers = composeWithDevToolsDevelopmentOnly({
```
agurtovoy, line 135: Redux logging is enabled in dev only.

packages/runtime/src/state/makeswift-api-client.ts
```diff
@@ -129,5 +132,20 @@ export type Dispatch = ThunkDispatch<State, unknown, Action>
...
+    name: `API client store (${new Date().toISOString()})`,
+    serialize: true,
+    stateSanitizer: (state: any) => serializeState(state),
+    actionsBlacklist: [
+      ActionTypes.BUILDER_POINTER_MOVE,
+      ActionTypes.HANDLE_POINTER_MOVE,
+      ActionTypes.ELEMENT_FROM_POINT_CHANGE,
+    ],
```
agurtovoy, line 143: Don't log actions that fire on every mouse move.

packages/runtime/src/state/react-builder-preview.ts
```diff
@@ -768,11 +770,22 @@ export function configureStore({
...
+    name: `Host store (${new Date().toISOString()})`,
+    serialize: true,
+    stateSanitizer: (state: any) => serializeState(state),
+    actionsBlacklist: [
+      ActionTypes.BUILDER_POINTER_MOVE,
+      ActionTypes.HANDLE_POINTER_MOVE,
+      ActionTypes.ELEMENT_FROM_POINT_CHANGE,
+    ],
+  })
+
```
agurtovoy, line 783: See my comments above: https://github.com/makeswift/makeswift/pull/845/files#diff-701ba0ed420276bea2f613b55a022aecc9f08aca9e519242cab81445ee070c7dR135

packages/runtime/src/utils/__tests__/__snapshots__/serializeState.test.ts.snap
```diff
@@ -0,0 +1,45 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
agurtovoy, line 1: Test for `serializeState` utility.

packages/runtime/src/utils/serializeState.ts
```diff
@@ -0,0 +1,30 @@
...
+  if (state instanceof Set) return [...state.values()].map(serializeState) as SerializedState<T>
+
+  if (state instanceof Array) return state.map(serializeState) as SerializedState<T>
+
+  if (state !== null && typeof state === 'object') {
+    return mapValues(state, serializeState) as SerializedState<T>
+  }
+
+  return state as SerializedState<T>
+}
```
agurtovoy, line 30: Recursively translate `Map` to `Record`, `Set` to `Array`.

migueloller: Should this utility be named `serializeState` even though it isn't particularly tied to our state? Seems like this is a pretty generic deep `Map` to array of entries and `Set` to array utility

packages/runtime/src/state/makeswift-api-client.ts
migueloller: If these actions become important for debugging, might it be confusing if they're missing? Is there perhaps a way to collapse these in DevTools if they occur all together?

packages/runtime/src/utils/serializeState.ts
agurtovoy: It is generic, but not generic enough to be just called `serialize`, plus the details of what it means to serialize something in general often depend on the use case. Our use case is serializing state, thus the name. We could name it something like `serializeMapSet`, but then again, if we decide to add, say `Date` serialization, we'd have to adjust the name.

> Seems like this is a pretty generic deep `Map` to array of entries and `Set` to array utility

`Map` to `Record`, `Set` to `Array`.

packages/runtime/src/state/makeswift-api-client.ts
agurtovoy: > Is there perhaps a way to collapse these in DevTools if they occur all together?

AFAIK, there isn't. 

I think it's more or less expected that "spamming" actions are by default excluded from log output, and that you need to change something in code (or the console) that get the "verbose" output.

PR #846, opened 2024-10-09
fix(runtime): don't use locale to partition non-localized resources

agurtovoy:
... plus updated tests.

`getAPIResource` and other API resource functions already enforce at compile-time whether we should pass `locale` to access an API resource of a specific type, but actions do not. However, we don't want to place the burden of conditionally passing the locale on action creators, as actions are often dispatched from a context that handles all resource types. Instead, we make it okay to pass a locale in all resource action payloads and have the reducer use the locale for scoping only if we are operating on a localized resource.

### DIFF
--- a/apps/nextjs-pages-router/next.config.mjs
+++ b/apps/nextjs-pages-router/next.config.mjs
@@ -6,8 +6,8 @@ const withMakeswift = createWithMakeswift()
 const nextConfig = {
   reactStrictMode: true,
   i18n: {
-    locales: ["en-US", "fr-FR"],
-    defaultLocale: "en-US",
+    locales: ['en-US', 'fr-FR', 'it-IT'],
+    defaultLocale: 'en-US',
   },
 }
 
--- a/packages/runtime/src/next/client.ts
+++ b/packages/runtime/src/next/client.ts
@@ -422,12 +422,11 @@ export class Makeswift {
             siteVersion,
           )
 
-          localizedResourcesMap.set(globalElementId, localizedGlobalElement?.id ?? null)
-
           if (localizedGlobalElement) {
             // Update the logic here when we can merge element trees
             elementData = localizedGlobalElement.data
 
+            localizedResourcesMap.set(globalElementId, localizedGlobalElement.id)
             localizedGlobalElements.set(localizedGlobalElement.id, localizedGlobalElement)
           }
         }
--- a/packages/runtime/src/state/modules/__tests__/api-resources.test.ts
+++ b/packages/runtime/src/state/modules/__tests__/api-resources.test.ts
@@ -163,17 +163,17 @@ describe('state / APIResources', () => {
   test('`Actions.apiResourceFulfilled` correctly populates the store', () => {
     const state = (
       [
-        [swatch, null],
+        [swatch, 'en-US'], // locale on non-localizable resources should be ignored
         [file, null],
         [typography, null],
         [pagePathnameSlice, null],
         [localizedPagePathnameSlice_fr_FR, 'fr-FR'],
-        [globalElement, null],
+        [globalElement, 'en-US'], // see above
         [localizedGlobalElement_fr_FR, 'fr-FR'],
         [localizedGlobalElement_it_IT, 'it-IT'],
         [table, null],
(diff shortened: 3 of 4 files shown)


packages/runtime/src/state/modules/api-resources.ts
```diff
@@ -1,14 +1,35 @@
...
+
+function isLocalizableAPIResourceType(resourceType: APIResourceType): boolean {
+  return resourceType in LocalizableAPIResourceType
+}
+
+function compositeId(
+  resourceId: string,
+  resourceType: APIResourceType,
+  locale?: string | null,
+): CompositeResourceId {
```
agurtovoy, line 27: Pass `resourceType` to `compositeId` so we can distinguish whether the resource needs to be locale-scoped.

packages/runtime/src/state/modules/__tests__/api-resources.test.ts
```diff
@@ -163,17 +163,17 @@ describe('state / APIResources', () => {
   test('`Actions.apiResourceFulfilled` correctly populates the store', () => {
     const state = (
       [
-        [swatch, null],
+        [swatch, 'en-US'], // locale on non-localizable resources should be ignored
```
agurtovoy, line 166: Update tests to trigger the bug we're fixing.

apps/nextjs-pages-router/next.config.mjs
```diff
@@ -6,8 +6,8 @@ const withMakeswift = createWithMakeswift()
 const nextConfig = {
   reactStrictMode: true,
   i18n: {
-    locales: ["en-US", "fr-FR"],
-    defaultLocale: "en-US",
+    locales: ['en-US', 'fr-FR', 'it-IT'],
```
agurtovoy, line 9: Allow manual testing w/ more than one non-default locale.

packages/runtime/src/next/client.ts
```diff
@@ -422,12 +422,11 @@ export class Makeswift {
             siteVersion,
           )
 
-          localizedResourcesMap.set(globalElementId, localizedGlobalElement?.id ?? null)
```
agurtovoy, line 425: Don't populate `localizedResourcesMap` w/ nulls.

packages/runtime/src/state/modules/api-resources.ts
```diff
@@ -1,14 +1,35 @@
...
 
-function compositeId(resourceId: string, locale?: string | null): CompositeResourceId {
-  return (locale != null ? `${resourceId}@${locale}` : resourceId) as CompositeResourceId
+function isValidAPIResourceType(resourceType: string): resourceType is APIResourceType {
+  return resourceType in APIResourceType
+}
+
+function isLocalizableAPIResourceType(resourceType: APIResourceType): boolean {
+  return resourceType in LocalizableAPIResourceType
+}
```
migueloller, line 21: Should we make this predicate a type guard like we did for `isValidAPIResourceType`?

packages/runtime/src/state/modules/api-resources.ts
```diff
@@ -68,6 +94,7 @@ export function getSerializedState(state: State): SerializedState {
     Array.from(resources.entries()).forEach(([compositeId, value]) => {
       if (value != null) {
         const [id, locale] = parseCompositeId(compositeId)
+        console.assert(locale == null || isLocalizableAPIResourceType(resourceType))
```
migueloller, line 97: Is this a stray console log or was the intent to keep it around?

We probably need to introduce a proper logger for our packages at some point...

migueloller (approved):
Looks good 👍🏻

packages/runtime/src/state/modules/api-resources.ts
agurtovoy: > Is this a stray console log or was the intent to keep it around?

It's intentional, it's a console **assert**, it will write an error to the console if the precondition fails: https://developer.mozilla.org/en-US/docs/Web/API/console/assert_static. I should add a message to it that allows us to identify the problematic resource, though.

> We probably need to introduce a proper logger for our packages at some point

Yes.

PR #847, opened 2024-10-14
fix(runtime): not found localized global components saved as `null`

agurtovoy:
Before:

before

After:

after

### DIFF
--- a/.changeset/fuzzy-planes-peel.md
+++ b/.changeset/fuzzy-planes-peel.md
@@ -1,0 +1,5 @@
+---
+'@makeswift/runtime': patch
+---
+
+fix: not found localized global components saved as `null` (previouly saved "not found" payload)
--- a/packages/runtime/src/next/api-handler/index.ts
+++ b/packages/runtime/src/next/api-handler/index.ts
@@ -224,15 +224,7 @@ export function MakeswiftApiHandler(
     ) {
       return client
         .getLocalizedGlobalElement(m.params.globalElementId, m.params.locale, siteVersion)
-        .then(resource => {
-          const body = resource === null ? { message: 'Not Found' } : resource
-
-          // We're not returning 404 if it's null because localized global element is nullable.
-          return match(args)
-            .with(routeHandlerPattern, () => NextResponse.json(body))
-            .with(apiRoutePattern, ([, res]) => res.json(body))
-            .exhaustive()
-        })
+        .then(handleResource)
     }
 
     if ((m = matches<{ id: string }>('/page-pathname-slices/:id'))) {


packages/runtime/src/state/makeswift-api-client.ts
```diff
@@ -119,6 +119,11 @@ async function fetchJson<T>(url: string): Promise<T | null> {
   return response.json()
 }
 
+async function fetchResource<T>(url: string): Promise<T | null> {
+  const result = await fetchJson(url)
+  return !('__typename' in result) ? null : result
+}
```
agurtovoy, line 125: Since we don't 404 on missing resources, make sure we don't return an error payload pretending to be a resource. 

Ideally should validate against a Zod schema of the resource type, but we don't have them readily available.

migueloller: Hmm, I do think we 404 on our API endpoints but don't on GraphQL endpoints. Perhaps we want to fix this at the API layer in the `/api/makeswift/` handlers?

I do agree with you that having proper parsing so we don't add garbage in the cache would also be great. I also understand why we might not want to take that on right now.

migueloller (approved):
LGTM. Left a comment about how maybe we can address this at the API layer by returning a 404 there (not api.makeswift.com, but `/api/makeswift/` handlers in the runtime) but I'm not opposed to the current approach in this PR 👍🏻

packages/runtime/src/state/makeswift-api-client.ts
agurtovoy: > Hmm, I do think we 404 on our API endpoints but don't on GraphQL endpoints. Perhaps we want to fix this at the API layer in the /api/makeswift/ handlers?

I think that might break the old versions of the runtime. Specifically for localized resources, we are relying on the fact that fetch of non-existent resource will (silently) succeed but return something w/o an ID.

agurtovoy: We can, of course, version the API layer and have the new runtime fetch from the new endpoints, but that's more than I can chew right now, haha.

agurtovoy: Never mind, I just read your PR comment, and see that you weren't suggesting to change api.makeswift.com. I don't see a `makeswift` dir in `src/api/`, though: https://github.com/makeswift/makeswift/tree/main/packages/runtime/src/api

migueloller: Well, I seem to have found the problem 🤦🏻‍♂️ 

CleanShot 2024-10-15 at 10 54 33@2x

migueloller: We need to do `res.status(404).json(body)`

migueloller: We seem to have a comment there that says this is intentional... I'm not sure why we'd do that, though.

migueloller: I wonder if we should have a single handler for global elements instead of a separate one for localized global elements 🤔

migueloller: @fikrikarim, maybe you remember why you explicitly wanted to handle this case differently?

agurtovoy: The comment (`We're not returning 404 if it's null because localized global element is nullable.`) seems to imply that returning 404 will result in a different code path for a localized resource than otherwise, but in fact in the code that makes the call to this "endpoint" (AFAIK [this is the only call site](https://github.com/makeswift/makeswift/blob/4203ec3d4d35c9229e3d81669ac1e54729b99a3a/packages/runtime/src/state/makeswift-api-client.ts#L98)), we silently translate 404 back to `null`, so I think it should be okay to not special-case here.

fetchJSON

migueloller: I agree

agurtovoy: Reworked, confirmed that this behaves as we reasoned.

PR #849, opened 2024-10-21
fix: remove id from embedded component

fikrikarim:


packages/runtime/src/next/client.ts
```diff
@@ -157,9 +155,17 @@ export type MakeswiftComponentSnapshot = {
...
 ): EmbeddedDocument {
-  const { id, data, locale, key, type, name } = componentDocument
+  const { data, locale, key, type, name } = componentDocument
+  // Because EmbeddedDocument has composite keys instead of a UUID, we need to generate a stable UUID for the key/documentKey.
+  const stableUUID = deterministicUUID({
+    type,
+    key,
+    locale,
+    dataKey: data.key,
+  })
```
fikrikarim, line 165: We need to ensure uniqueness across sites. Since we don't have `siteId` here, I'm also using the `data.key`.

One thing I don't like about this is there's an inverted effect: the `document.data.key` is affecting the `document.key`.

Any better idea?

PR #850, opened 2024-10-21
feat: add fallback prop to embedded component

arvinpoddar:
CleanShot 2024-10-22 at 14 59 10

### DIFF
--- a/packages/runtime/src/runtimes/react/components/Element.tsx
+++ b/packages/runtime/src/runtimes/react/components/Element.tsx
@@ -1,15 +1,62 @@
 'use client'
 
-import { Ref, forwardRef, memo, useCallback, useEffect, useImperativeHandle, useRef } from 'react'
-import { Element as ElementDataOrRef, isElementReference } from '../../../state/react-page'
-import { ElementImperativeHandle } from '../element-imperative-handle'
-import { useDispatch } from '../hooks/use-dispatch'
-import { useDocumentKey } from '../hooks/use-document-context'
-import { useDisableRegisterElement } from '../hooks/use-disable-register-element'
-import { mountComponentEffect, registerComponentHandleEffect } from '../../../state/actions'
+import { forwardRef, memo, ReactNode, Ref, useCallback, useImperativeHandle, useRef } from 'react'
+import { isElementReference, type Element as ElementDataOrRef } from '../../../state/react-page'
+import { ElementRegistration } from './ElementRegistration'
 import { ElementReference } from './ElementReference'
 import { ElementData } from './ElementData'
+import { ElementImperativeHandle } from '../element-imperative-handle'
 import { FindDomNode } from '../find-dom-node'
+
+type ElementWithFallbackProps = {
+  element: ElementDataOrRef
+  fallback?: ReactNode
+  isInitialData?: boolean
+}
+
+export const ElementWithFallback = memo(
+  forwardRef(function Element(
+    { element, fallback, isInitialData = false }: ElementWithFallbackProps,
+    ref: Ref<ElementImperativeHandle>,
+  ): JSX.Element | null {
+    const showFallback = fallback != null && isInitialData
+
+    const useFindDomNodeRef = useRef(true)
+    const imperativeHandleRef = useRef(new ElementImperativeHandle())
+
+    const findDomNodeCallbackRef = useCallback((current: (() => Element | Text | null) | null) => {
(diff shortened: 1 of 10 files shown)


packages/runtime/src/runtimes/react/components/Element.tsx
```diff
@@ -1,66 +1,72 @@
...
+            <ElementData key={element.key} ref={elementCallbackRef} elementData={element} />
+          )}
+        </ElementRegistration>
       </FindDomNode>
     )
   }),
 )
+
+type ElementProps = {
+  element: ReactPageElement
```
migueloller, line 62: Seems this changed from `ElementDataOrRef` to `ReactPageElement`. How come?

migueloller (approved):
This looks pretty clean!

There's some weirdness in having to drill down `fallback` and `isInitialData` through all these layers but I think this might have to do more with the original design than your particular changes. Also, having `isInitialData` as it is right now does make me wonder if instead we should have not overloaded `elementTree` or `element` to have both "fake data" and "real data" and instead we'd leave `elementTree` and `element` as the source of truth of data (would be `null` if not present) and we'd separate use an `initialElementTree` or `initialElement` prop to provide the initial value, which can then be used or not depending on whether `element` or `elementTree` is `null` or if `fallback` is present, etc. Don't know about you but that seems more like passing around the "truth" to me since we'd not be passing a computed value and overloading a property with a "fake" value sometimes (i.e., when there's initial data). Seems like things are a bit complicated as they are now.

Note, `hasFallback` being sent to the builder is a bit different of a story since that's about not being able to send the source of truth, the fallback itself, over a serialization boundary, so we're settling for `hasFallback` which is a boolean and is thus serializable. You could then argue that this isn't `metadata` but just part of the regular ol' data. So:

```
type Document = {
  elementTree: ElementTree | null
  initialElementTree: ElementTree
  fallback: boolean // I'm partial to naming this field `hasFallback`... it's probably a better name
  // ...other properties
}
```

Note how simple and clear the data structure above it.

And honestly, having `fallback` be a boolean doesn't make sense just for the builder but also because if the document ends up in Redux state, it's relevant it is serializable too. It's not like we need it from Redux state to render it, since the component that registers it has the fallback already before it passes the boolean to `useDocument`.

packages/runtime/src/runtimes/react/components/Element.tsx
arvinpoddar: This is a mistake. The underlying type is actually the same, it was just being imported under a different alias. Must have gotten switched by mistake during all the shuffling. Will address.

arvinpoddar:
> This looks pretty clean!
> 
> There's some weirdness in having to drill down `fallback` and `isInitialData` through all these layers but I think this might have to do more with the original design than your particular changes. Also, having `isInitialData` as it is right now does make me wonder if instead we should have not overloaded `elementTree` or `element` to have both "fake data" and "real data" and instead we'd leave `elementTree` and `element` as the source of truth of data (would be `null` if not present) and we'd separate use an `initialElementTree` or `initialElement` prop to provide the initial value, which can then be used or not depending on whether `element` or `elementTree` is `null` or if `fallback` is present, etc. Don't know about you but that seems more like passing around the "truth" to me since we'd not be passing a computed value and overloading a property with a "fake" value sometimes (i.e., when there's initial data). Seems like things are a bit complicated as they are now.
> 
> Note, `hasFallback` being sent to the builder is a bit different of a story since that's about not being able to send the source of truth, the fallback itself, over a serialization boundary, so we're settling for `hasFallback` which is a boolean and is thus serializable. You could then argue that this isn't `metadata` but just part of the regular ol' data. So:
> 
> ```
> type Document = {
>   elementTree: ElementTree | null
>   initialElementTree: ElementTree
>   fallback: boolean // I'm partial to naming this field `hasFallback`... it's probably a better name
>   // ...other properties
> }
> ```
> 
> Note how simple and clear the data structure above it.
> 
> And honestly, having `fallback` be a boolean doesn't make sense just for the builder but also because if the document ends up in Redux state, it's relevant it is serializable too. It's not like we need it from Redux state to render it, since the component that registers it has the fallback already before it passes the boolean to `useDocument`.

I hear you, and I like what you're suggesting. Since this PR has been around for a while, I'd like to merge this as is, but I want to sit down and have a conversation of what this should look like. Data exchanged between the builder and runtime tends to form a one-way door, so I want to make sure we get this "right" before public release.

PR #852, opened 2024-10-22
feat: use id instead of key for embedded component and update the API

fikrikarim:


packages/runtime/src/next/client.ts
```diff
@@ -140,33 +140,71 @@ export type MakeswiftPageSnapshot = {
...
 
-export type MakeswiftComponentSnapshot = {
-  document: MakeswiftComponentDocument
-  cacheData: CacheData
-}
+const makeswiftComponentDocumentFallbackSchema = z.object({
+  id: z.string(),
+  locale: z.string().nullable(),
+  data: z.null(),
+  apiKeyPrefix: z.string(),
```
migueloller, line 155: I'm a bit confused why the fallback has an API key prefix

packages/runtime/src/next/client.ts
```diff
@@ -140,33 +140,71 @@ export type MakeswiftPageSnapshot = {
...
+>
+
+export type MakeswiftComponentSnapshot =
+  | {
+      document: MakeswiftComponentDocument
+      cacheData: CacheData
+    }
+  | {
+      document: MakeswiftComponentDocumentFallback
+      cacheData: CacheData | null
```
migueloller, line 169: Instead of `cacheData | null` we should use empty cache data. Maybe we can have a factory for that, e.g., `CacheData.empty()`

migueloller (approved):
Looks good. I have some thoughts on how we model the snapshot and element tree and when we introduce the mapping to a document, etc. Happy to chat about it, but if feel like it's better to discuss another time, I'm happy to ship this as-is and discuss it later.

packages/runtime/src/next/client.ts
migueloller: Could we send the ID instead of attempting to add this to some sort of model?

fikrikarim: We need to pass the `siteId` to `deterministicUUID`. But since we only have the siteApiKey and not the `siteId` on the host (if the snapshot 404), we use the `siteApiKey` as a seed for `deterministicUUID`

migueloller: I guess what I'm trying to say is: why are we sending the inputs to the ID to the builder instead of computing it in the host and sending it the deterministic ID? Couldn't the `id` field be the deterministic ID instead of the user-provided one so just like the non-fallback document, the `id` is what the builder uses to identify it? I'm guessing there's logic in the builder right now to compute the stable ID that isn't necessary (since the host could do it)?

PR #855, opened 2024-10-24
fix: stop default rules from overriding props

arvinpoddar:
When using a Box as the root for an embedded component, the width/margin prop panels do not apply to the Box as expected, even though the data is saved correctly from the panels. The reason for this is that the CSS class being generated via the prop data is injected earlier in the stylesheet than the class for the default styles, resulting in only the latter being applied. This is not true for when the page root is present around the box, which will result in the defaults getting injected first, followed by the props class.

Since responsive styles for width/margin are always passed to the BackgroundsContainer, this fix addresses the problem by only applying the generated classname for default width/margin if a classname was not passed.

Below is an example with two pages. The one on the left has a Box as a root, with the only non-default prop being a `margin-bottom` of `100px`. The one on the right has a Root as a root, containing a Box, with the same margin settings are the first. In both cases, the Box has identical classnames. The only difference is the injection order of the CSS classes in the main style tag.

Before:

CleanShot 2024-10-24 at 18 10 25

After:

CleanShot 2024-10-24 at 18 07 00

### DIFF
--- a/packages/runtime/src/components/shared/BackgroundsContainer/index.tsx
+++ b/packages/runtime/src/components/shared/BackgroundsContainer/index.tsx
@@ -25,6 +25,11 @@ export default forwardRef<HTMLDivElement | null, Props>(function BackgroundsCont
   { backgrounds, children, className, ...restOfProps }: Props,
   ref: Ref<HTMLDivElement>,
 ) {
+  const widthAndMarginStyle = useStyle({
+    width: '100%',
+    margin: '0 auto',
+  })
+
   return (
     <div
       {...restOfProps}
@@ -32,8 +37,6 @@ export default forwardRef<HTMLDivElement | null, Props>(function BackgroundsCont
       className={cx(
         useStyle({
           position: 'relative',
-          width: '100%',
-          margin: '0 auto',
           '> *': {
             borderRadius: 'inherit',
             height: 'inherit',
@@ -42,7 +45,7 @@ export default forwardRef<HTMLDivElement | null, Props>(function BackgroundsCont
             position: 'relative',
           },
         }),
-        className,
+        className ? className : widthAndMarginStyle,
       )}
     >
       <Backgrounds backgrounds={useBackgrounds(backgrounds)} />


packages/runtime/src/components/shared/BackgroundsContainer/index.tsx
```diff
@@ -25,15 +25,18 @@ export default forwardRef<HTMLDivElement | null, Props>(function BackgroundsCont
...
+  const widthAndMarginStyle = useStyle({
+    width: '100%',
+    margin: '0 auto',
+  })
+
   return (
     <div
       {...restOfProps}
       ref={ref}
       className={cx(
```
agurtovoy, line 37: Note that Emotion's [`cx` docs](https://emotion.sh/docs/@emotion/css#cx) say that `Emotion generated styles are applied from left to right. Subsequent styles overwrite property values of previous styles.`, but this is clearly not the case here.

We could dig deeper in why not, but personally I'd argue that hard-coding defaults for CSS properties that will have their true defaults generated from prop values is redundant and makes it harder to understand how we arrive at the final component styles, and therefore needs to go either way.

agurtovoy (approved):

packages/runtime/src/components/shared/BackgroundsContainer/index.tsx
arvinpoddar: From offline: Btw, I did a lot more work that I didn't surface in my PR. I actually tested various fixes to see how the injected class order changes (I kept these test results saved, if you're interested). It turns out that while `cx` may follow the left to right rule, what's making the impact is the fact that `useStyles` has a side effect that injects classes into the cache directly, which directly determines the order they're flushed. This flush order is what we're concerned with - so if you've already generated a class and pass it to `cx`, it's too late, it's already in the cache.

migueloller (commented):
I've got _a ton_ of thoughts on how we handle CSS right now and how we need to make it better. From a perspective of reducing the amount of CSS, making it work with RSC, shipping the new Style control, and of course removing `useStyle` from components like Box so that the controls and prop controllers resolve the values outside.

This seems like a reasonable solution to address the current problem with embedded components. That being said, we might want to always apply defaults and let CSS cascade and overwrite the default values?

Right now we seem to omit the default if a class is applied, but what if it's just margin and doesn't include width?

migueloller (changes requested):
Don't forget the changeset!

agurtovoy:
> This seems like a reasonable solution to address the current problem with embedded components. That being said, we might want to always apply defaults and let CSS cascade and overwrite the default values?

@migueloller The CSS rules generated from control props will have the same specificity as the manual "defaults" and therefore it will be down to the order of the rules, which isn't correct here -- see https://github.com/makeswift/makeswift/pull/855#discussion_r1815776571

> Right now we seem to omit the default if a class is applied, but what if it's just margin and doesn't include width?

The true defaults will be generated from the prop values.

arvinpoddar:
> Don't forget the changeset!

Done

PR #851, opened 2024-10-21
fix(runtime): Pages Router links from localized to base pages don't work

agurtovoy:
## Before

Dev: 

https://github.com/user-attachments/assets/c3721867-c1b8-472a-9fc8-f39ae545e558

Prod:

https://github.com/user-attachments/assets/9b02179b-d4d7-4b66-b446-34ff1420bf99

## After

Dev:

https://github.com/user-attachments/assets/67cac23c-a773-43b7-82ca-f74f412a4562

Prod:

https://github.com/user-attachments/assets/09aa264c-13b6-4d5a-aad9-0a94e044766e

PR #857, opened 2024-10-28
feat(runtime): Next.js 15 / React 19 RC support

agurtovoy:
Next.js 15 / React 19 RC support

### Pages Router's custom `Document`

The Makeswift custom `Document` export has been moved from `@makeswift/runtime/next` to `@makeswift/runtime/next/document`. To migrate, adjust the custom `Document` import in `src/pages/_document.ts` as follows:

```diff
- export { Document as default } from '@makeswift/runtime/next'
+ export { Document as default } from '@makeswift/runtime/next/document'
```

### DIFF
--- a/apps/nextjs-app-router/.eslintrc.json
+++ b/apps/nextjs-app-router/.eslintrc.json
@@ -1,3 +1,3 @@
 {
-  "extends": ["next/babel","next/core-web-vitals"]
+  "extends": "next/core-web-vitals"
 }
--- a/apps/nextjs-app-router/package.json
+++ b/apps/nextjs-app-router/package.json
@@ -12,9 +12,9 @@
     "@formatjs/intl-localematcher": "^0.5.4",
     "@makeswift/runtime": "workspace:^",
     "negotiator": "^0.6.3",
-    "next": "14.1.0",
-    "react": "^18",
-    "react-dom": "^18"
+    "next": "15.0.2",
+    "react": "19.0.0-rc-02c0e824-20241028",
+    "react-dom": "19.0.0-rc-02c0e824-20241028"
   },
   "devDependencies": {
     "@types/negotiator": "^0.6.3",
@@ -22,8 +22,8 @@
     "@types/react": "^18",
     "@types/react-dom": "^18",
     "autoprefixer": "^10.0.1",
-    "eslint": "^8",
-    "eslint-config-next": "14.1.0",
+    "eslint": "^9.13.0",
+    "eslint-config-next": "15.0.2",
     "postcss": "^8",
     "tailwindcss": "^3.3.0",
     "typescript": "^5"
--- a/apps/nextjs-app-router/src/app/[lang]/[[...path]]/page.tsx
+++ b/apps/nextjs-app-router/src/app/[lang]/[[...path]]/page.tsx
@@ -4,10 +4,9 @@ import { getSiteVersion } from '@makeswift/runtime/next/server'
 import { notFound } from 'next/navigation'
 import { Page as MakeswiftPage } from '@makeswift/runtime/next'
 
-type ParsedUrlQuery = { lang: string; path?: string[] }
+type ParsedUrlQuery = Promise<{ lang: string; path?: string[] }>
 
 export async function generateStaticParams() {
-
   const pages = await client.getPages().toArray()
 
   return pages.flatMap((page) => [
@@ -26,7 +25,8 @@ export async function generateStaticParams() {
   ])
 }
 
-export default async function Page({ params }: { params: ParsedUrlQuery }) {
+export default async function Page(props: { params: ParsedUrlQuery }) {
+  const params = await props.params
   const path = '/' + (params?.path ?? []).join('/')
   const snapshot = await client.getPageSnapshot(path, {
     siteVersion: getSiteVersion(),
--- a/apps/nextjs-app-router/src/app/layout.tsx
+++ b/apps/nextjs-app-router/src/app/layout.tsx
@@ -1,17 +1,18 @@
 import { MakeswiftProvider } from '@/makeswift/provider'
 import { DraftModeScript } from '@makeswift/runtime/next/server'
 
-type Params = { lang: string; path?: string[] }
+type Params = Promise<{ lang: string; path?: string[] }>
 
-export default function RootLayout({
+export default async function RootLayout({
   children,
   params,
 }: Readonly<{
   children: React.ReactNode
   params: Params
 }>) {
+  const { lang } = await params
   return (
-    <html lang={params.lang}>
+    <html lang={lang}>
       <head>
         <DraftModeScript />
       </head>
--- a/apps/nextjs-app-router/tsconfig.json
+++ b/apps/nextjs-app-router/tsconfig.json
@@ -19,7 +19,8 @@
     ],
     "paths": {
       "@/*": ["./src/*"]
-    }
+    },
+    "target": "ES2017"
   },
   "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx", ".next/types/**/*.ts"],
   "exclude": ["node_modules"]
--- a/apps/nextjs-pages-router/.eslintrc.json
+++ b/apps/nextjs-pages-router/.eslintrc.json
@@ -1,3 +1,3 @@
 {
-  "extends": ["next/babel","next/core-web-vitals"]
+  "extends": "next/core-web-vitals"
 }
--- a/apps/nextjs-pages-router/package.json
+++ b/apps/nextjs-pages-router/package.json
@@ -10,7 +10,7 @@
   },
   "dependencies": {
     "@makeswift/runtime": "workspace:^",
-    "next": "14.1.0",
+    "next": "15.0.2",
     "react": "^18",
     "react-dom": "^18"
   },
--- a/apps/nextjs-pages-router/tsconfig.json
+++ b/apps/nextjs-pages-router/tsconfig.json
@@ -14,7 +14,8 @@
     "incremental": true,
     "paths": {
       "@/*": ["./src/*"]
-    }
+    },
+    "target": "ES2017"
   },
   "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx"],
   "exclude": ["node_modules"]
--- a/packages/runtime/package.json
+++ b/packages/runtime/package.json
@@ -58,6 +58,11 @@
       "import": "./dist/esm/next/index.js",
       "types": "./dist/types/next/index.d.ts"
     },
+    "./next/document": {
+      "require": "./dist/cjs/next/document.js",
+      "module": "./dist/esm/next/document.js",
+      "types": "./dist/types/next/document.d.ts"
+    },
     "./next/server": {
       "require": "./dist/cjs/next/server.js",
       "module": "./dist/esm/next/server.js",
@@ -109,9 +114,9 @@
   "peerDependencies": {
     "@types/react": "^18.0.0",
     "@types/react-dom": "^18.0.0",
-    "next": "^13.4.0 || ^14.0.0",
-    "react": "^18.0.0",
-    "react-dom": "^18.0.0"
+    "next": "^13.4.0 || ^14.0.0 || ^15.0.0",
+    "react": "^18.0.0 || ^19.0.0 || 19.0.0-rc",
+    "react-dom": "^18.0.0 || ^19.0.0 || 19.0.0-rc"
   },
   "peerDependenciesMeta": {
     "@types/react": {
@@ -187,7 +192,7 @@
     "jest-environment-jsdom": "^29.7.0",
     "madge": "^6.0.0",
     "msw": "^2.3.1",
-    "next": "14.1.0",
+    "next": "15.0.2",
     "react": "18.2.0",
     "react-dom": "18.2.0",
     "slate": "^0.91.4",
--- a/packages/runtime/src/next/api-handler/handlers/proxy-draft-mode.ts
+++ b/packages/runtime/src/next/api-handler/handlers/proxy-draft-mode.ts
@@ -2,6 +2,7 @@ import { NextApiRequest, NextApiResponse } from 'next'
 import { NextRequest, NextResponse } from 'next/server'
 import { P, match } from 'ts-pattern'
 import { cookies, draftMode } from 'next/headers'
+
 import { MAKESWIFT_DRAFT_MODE_DATA_COOKIE, MakeswiftDraftData } from '../../draft-mode'
 import { MakeswiftSiteVersion } from '../../preview-mode'
 
@@ -55,7 +56,8 @@ async function proxyDraftModeRouteHandler(
     siteVersion: MakeswiftSiteVersion.Working,
   }
 
-  draftMode().enable()
+  const draft = await draftMode()
+  draft.enable()
 
   const proxyUrl = request.nextUrl.clone()
   proxyUrl.searchParams.delete('x-makeswift-draft-mode')
@@ -63,13 +65,13 @@ async function proxyDraftModeRouteHandler(
   const proxyRequest = new NextRequest(proxyUrl, { headers: request.headers })
   proxyRequest.headers.delete('x-makeswift-draft-mode')
 
-  const draftModeCookie = cookies().get('__prerender_bypass')
+  const draftModeCookie = (await cookies()).get('__prerender_bypass')
   if (draftModeCookie) {
     proxyRequest.cookies.set(draftModeCookie)
     proxyRequest.cookies.set(MAKESWIFT_DRAFT_MODE_DATA_COOKIE, JSON.stringify(draftModeData))
   }
 
-  draftMode().disable()
+  draft.disable()
 
   const proxyResponse = await fetch(proxyRequest)
 
--- a/packages/runtime/src/next/draft-mode/draft-mode-script.tsx
+++ b/packages/runtime/src/next/draft-mode/draft-mode-script.tsx
@@ -5,8 +5,8 @@ type Props = {
   appOrigin?: string
 }
 
-export function DraftModeScript({ appOrigin = 'https://app.makeswift.com' }: Props) {
-  const { isEnabled: isDraftModeEnabled } = draftMode()
+export async function DraftModeScript({ appOrigin = 'https://app.makeswift.com' }: Props) {
+  const { isEnabled: isDraftModeEnabled } = await draftMode()
   const draftModeScript = `
 const isDraft = ${isDraftModeEnabled}
 const appOrigin = '${appOrigin.replace("'", "\\'")}'
--- a/packages/runtime/src/next/index.tsx
+++ b/packages/runtime/src/next/index.tsx
@@ -4,6 +4,5 @@ export type { MakeswiftPage, MakeswiftPageDocument, MakeswiftPageSnapshot, Sitem
 export { Makeswift } from './client'
 export type { MakeswiftPreviewData } from './preview-mode'
 export { PreviewModeScript } from './preview-mode'
-export { Document } from './document'
 export { RootStyleRegistry } from './root-style-registry'
 export { ReactRuntimeProvider } from './context/react-runtime'
--- a/packages/runtime/src/runtimes/react/lib/react-dom.ts
+++ b/packages/runtime/src/runtimes/react/lib/react-dom.ts
@@ -1,0 +1,14 @@
+import { type Component } from 'react'
+import ReactDOM from 'react-dom'
+
+// https://github.com/facebook/react/blob/main/packages/shared/ReactDOMSharedInternals.js
+const reactDOMInternals = (ReactDOM as any)
+  .__DOM_INTERNALS_DO_NOT_USE_OR_WARN_USERS_THEY_CANNOT_UPGRADE as {
+  findDOMNode: typeof ReactDOM.findDOMNode
+}
+
+export function findDOMNode(instance: Component | null | undefined): Element | Text | null {
+  return ReactDOM.findDOMNode
+    ? ReactDOM.findDOMNode(instance)
+    : reactDOMInternals.findDOMNode(instance)
+}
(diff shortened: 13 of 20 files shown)


apps/nextjs-app-router/.eslintrc.json
```diff
@@ -1,3 +1,3 @@
 {
-  "extends": ["next/babel","next/core-web-vitals"]
```
agurtovoy, line 2: Recommended ESLint config is now just `"extends": "next/core-web-vitals"`: https://nextjs.org/docs/app/building-your-application/configuring/eslint

apps/nextjs-app-router/package.json
```diff
@@ -12,9 +12,9 @@
     "@formatjs/intl-localematcher": "^0.5.4",
     "@makeswift/runtime": "workspace:^",
     "negotiator": "^0.6.3",
-    "next": "14.1.0",
-    "react": "^18",
-    "react-dom": "^18"
+    "next": "15.0.1",
+    "react": "19.0.0-rc-02c0e824-20241028",
+    "react-dom": "19.0.0-rc-02c0e824-20241028"
```
agurtovoy, line 17: Upgrade `apps/nextjs-app-router` deps to Next.js 15 and React 19's latest RC.

apps/nextjs-app-router/src/app/[lang]/[[...path]]/page.tsx
```diff
@@ -2,12 +2,11 @@ import { client } from '@/makeswift/client'
 import '@/makeswift/components'
 import { getSiteVersion } from '@makeswift/runtime/next/server'
 import { notFound } from 'next/navigation'
-import { Page as MakeswiftPage } from '@makeswift/runtime/next'
+import { Page as MakeswiftPage } from '@makeswift/runtime/next/app-router'
```
Comment (agurtovoy) [inline apps/nextjs-app-router/src/app/[lang]/[[...path]]/page.tsx:5]: Breaking `runtime` API change: had to introduce a separate, App Router-specific entry point for client-side `runtime` imports, see below.

apps/nextjs-app-router/src/app/[lang]/[[...path]]/page.tsx
```diff
@@ -26,10 +25,11 @@ export async function generateStaticParams() {
   ])
 }
 
-export default async function Page({ params }: { params: ParsedUrlQuery }) {
+export default async function Page(props: { params: ParsedUrlQuery }) {
+  const params = await props.params
```
Comment (agurtovoy) [inline apps/nextjs-app-router/src/app/[lang]/[[...path]]/page.tsx:29]: Breaking Next.js 15 change: `params` are now async: https://nextjs.org/docs/app/building-your-application/upgrading/version-15#async-request-apis-breaking-change

apps/nextjs-app-router/src/app/[lang]/[[...path]]/page.tsx
```diff
@@ -26,10 +25,11 @@ export async function generateStaticParams() {
...
   const path = '/' + (params?.path ?? []).join('/')
   const snapshot = await client.getPageSnapshot(path, {
-    siteVersion: getSiteVersion(),
+    siteVersion: await getSiteVersion(),
```
Comment (agurtovoy) [inline apps/nextjs-app-router/src/app/[lang]/[[...path]]/page.tsx:32]: Breaking `runtime` change: `getSiteVersion` is now async, as it depends on `draftMode` and `cookies`: https://github.com/makeswift/makeswift/pull/857/files#diff-a84011079bd98e3597f4073734c142606a88c2267d6478719139d3ba6fb505d0R29

apps/nextjs-app-router/src/app/layout.tsx
```diff
@@ -1,17 +1,18 @@
 import { MakeswiftProvider } from '@/makeswift/provider'
 import { DraftModeScript } from '@makeswift/runtime/next/server'
 
-type Params = { lang: string; path?: string[] }
+type Params = Promise<{ lang: string; path?: string[] }>
 
-export default function RootLayout({
+export default async function RootLayout({
```
agurtovoy, line 6: Breaking Next.js 15 change: `RootLayout` become async to access async `params`; we could keep it synchronous and wrap `params` with `React.use` (https://nextjs.org/docs/app/building-your-application/upgrading/version-15#params--searchparams), but that's also a breaking change, so I opted for async as more forward-looking / consistent with the `Page` component.

apps/nextjs-app-router/tsconfig.json
```diff
@@ -19,7 +19,8 @@
     ],
     "paths": {
       "@/*": ["./src/*"]
-    }
+    },
+    "target": "ES2017"
```
agurtovoy, line 23: Auto-added by Next.js 15

apps/nextjs-pages-router/.eslintrc.json
```diff
@@ -1,3 +1,3 @@
 {
-  "extends": ["next/babel","next/core-web-vitals"]
+  "extends": "next/core-web-vitals"
```
agurtovoy, line 2: Same as above, but for the Pages Router: https://nextjs.org/docs/pages/building-your-application/configuring/eslint

apps/nextjs-pages-router/package.json
```diff
@@ -10,7 +10,7 @@
   },
   "dependencies": {
     "@makeswift/runtime": "workspace:^",
-    "next": "14.1.0",
+    "next": "15.0.1",
```
agurtovoy, line 13: Upgrade `app/nextjs-pages-router` to Next.js 15, but keep it on React 18: https://nextjs.org/blog/next-15#pages-router-on-react-18

apps/nextjs-pages-router/tsconfig.json
```diff
@@ -14,7 +14,8 @@
     "incremental": true,
     "paths": {
       "@/*": ["./src/*"]
-    }
+    },
+    "target": "ES2017"
```
agurtovoy, line 18: Auto-added by Next.js 15

packages/runtime/package.json
```diff
@@ -153,6 +158,7 @@
     "path-to-regexp": "^6.2.1",
     "polished": "3.0.3",
     "react-player": "^1.12.0",
+    "react-reconciler": "0.31.0-rc-02c0e824-20241028",
```
agurtovoy, line 161: Use `react-reconciler` for the `findDOMNode` implementation for React 19+, see PoC here: https://github.com/makeswift/find-dom-node-spike/pull/1

apps/nextjs-app-router/src/app/[lang]/[[...path]]/page.tsx
Comment (migueloller) [reply apps/nextjs-app-router/src/app/[lang]/[[...path]]/page.tsx:5]: Instead of naming it `app-router`, what if we named the import `server` or something that's not specifically `app-router`?? This seems to be a normal distinction that's done in Next.js itself and other popular Next.js libraries.

Examples:
- https://clerk.com/docs/references/nextjs/clerk-middleware
- https://nextjs.org/docs/app/building-your-application/routing/middleware#example
- https://next-intl-docs.vercel.app/docs/getting-started/app-router/with-i18n-routing#i18n-request

apps/nextjs-app-router/src/app/[lang]/[[...path]]/page.tsx
Comment (migueloller) [reply apps/nextjs-app-router/src/app/[lang]/[[...path]]/page.tsx:32]: Is the expectation that for Next.js 14 we still expect users to `await getSiteVersion()`?

Comment (agurtovoy) [reply apps/nextjs-app-router/src/app/[lang]/[[...path]]/page.tsx:32]: For the App Router and this new `runtime` release, yes.

packages/runtime/package.json
migueloller: After what Seb mentioned, I tried looking into it and you should be able to use `ReactDOM['__DOM_INTERNALS_DO_NOT_USE_OR_WARN_USERS_THEY_CANNOT_UPGRADE'].findDOMNode`.

This quick experiment worked for me:

```js
'use client'

import React, { Component } from 'react'
import ReactDOM from 'react-dom'

export default function Page() {
  return <Div>Hi!</Div>
}

class Div extends Component {
  componentDidMount() {
    const findDOMNode =
      ReactDOM['__DOM_INTERNALS_DO_NOT_USE_OR_WARN_USERS_THEY_CANNOT_UPGRADE'].findDOMNode

    console.log(findDOMNode(this))
  }

  render() {
    return <div>{this.props.children}</div>
  }
}
```

apps/nextjs-app-router/package.json
```diff
@@ -23,7 +23,7 @@
     "@types/react-dom": "^18",
     "autoprefixer": "^10.0.1",
     "eslint": "^8",
-    "eslint-config-next": "14.1.0",
+    "eslint-config-next": "15.0.0-rc.1",
```
migueloller, line 26: Looks like the `latest` dist tag points to `15.0.1`

apps/nextjs-app-router/src/app/[lang]/[[...path]]/page.tsx
Comment (migueloller) [reply apps/nextjs-app-router/src/app/[lang]/[[...path]]/page.tsx:32]: So is this new runtime release not compatible with older versions of Next.js? Or perhaps it's compatible, but it would require users to change their code even if they're still using an older version (e.g., Next.js 14)?

apps/nextjs-app-router/src/app/[lang]/[[...path]]/page.tsx
Comment (migueloller) [reply apps/nextjs-app-router/src/app/[lang]/[[...path]]/page.tsx:5]: Also, what was the particular reason we had to introduce a separate entry point? For imports that are client-side, shouldn't we be able to have `use client` directives that avoid that problem?

apps/nextjs-pages-router/package.json
migueloller: Is there a reason not to use React 19 on this Pages Router app? Is it because it's not stable yet?

apps/nextjs-app-router/src/app/[lang]/[[...path]]/page.tsx
Comment (agurtovoy) [reply apps/nextjs-app-router/src/app/[lang]/[[...path]]/page.tsx:5]: @migueloller So the reason we have to introduce an App Router-specific endpoint is that importing and extending `Document` from `@next/document` [breaks App Router's production build](https://vercel.com/makeswift/makeswift-nextjs-app-router/AFZqEjwe3A1yZMQd1YaMuw1A9ZGG?filter=errors), and our default "client" Next.js entry point, `@makeswift/runtime/next`, has to export the (extended) `Document` for the Pages Router.

We could extend [the already existing `@makeswift/runtime/next/server` endpoint](https://github.com/makeswift/makeswift/blob/main/packages/runtime/package.json#L61) with the client-only exports we need for the App Router (`RootStyleRegistry` and `ReactRuntimeProvider`), but then it would be misnamed.

packages/runtime/package.json
```diff
@@ -109,9 +114,9 @@
   "peerDependencies": {
     "@types/react": "^18.0.0",
     "@types/react-dom": "^18.0.0",
-    "next": "^13.4.0 || ^14.0.0",
-    "react": "^18.0.0",
-    "react-dom": "^18.0.0"
+    "next": "^13.4.0 || ^14.0.0 || >=15.0.0-rc",
```
migueloller, line 117: Should we support release candidates? Also, should we use `>=` here? We wouldn't want Next.js 16 to automatically be a supported peer dependency until we explicitly support it, right?

I would say `^15.0.0` for Next.js and `^19.0.0 || 19.0.0-rc` for React

packages/runtime/src/next/api-handler/handlers/proxy-draft-mode.ts
```diff
@@ -1,7 +1,9 @@
 import { NextApiRequest, NextApiResponse } from 'next'
 import { NextRequest, NextResponse } from 'next/server'
 import { P, match } from 'ts-pattern'
-import { cookies, draftMode } from 'next/headers'
+
+import { cookies, enableDraftMode } from '../../lib/next'
```
agurtovoy, line 5: Compatibility layer for Next.js 15, see https://github.com/makeswift/makeswift/pull/857/files#diff-90b2f7383e5a1a78d0640d9e7f60d69143d095872d61c0da07309975c898d1dbR1

packages/runtime/src/next/lib/next.ts
```diff
@@ -0,0 +1,30 @@
...
+}
+
+export async function cookies() {
+  return await _cookies()
+}
+
+export async function enableDraftMode(enable: boolean = true) {
+  const dm = await draftMode()
+  return enable ? dm.enable() : dm.disable()
+}
```
migueloller, line 26: I don't understand why we'd want this utility to exist. It seems like an unnecessary layer of abstraction

packages/runtime/src/next/lib/next.ts
```diff
@@ -0,0 +1,30 @@
+// Next.js 15 compatibility layer:
+//
+// `draftMode`, `cookies` and `params` functions became async in Next.js 15
+// (https://nextjs.org/blog/next-15#async-request-apis-breaking-change),
+// but we are supporting Next.js 13, 14 and 15 in the same codebase, and are
+// compiling against Next.js 14 as a common denominator. The wrapper
+// functions below force the async use of the aforementioned primitives in
+// the rest of the codebase even though they are typed as synchronous in
+// Next.js 14. This works because awaiting a non-promise value is a no-op in
+// JavaScript.
```
migueloller, line 10: Why is this compatibility layer necessary? Can't we just do `await cookies()` or `await draftMode()` whether we're on Next.js 14 or 15?

packages/runtime/src/next/draft-mode/draft-mode-script.tsx
```diff
@@ -1,12 +1,13 @@
...
 import { ActionTypes } from '../../react'
+import { draftMode } from '../lib/next'
 
 type Props = {
   appOrigin?: string
 }
 
 export function DraftModeScript({ appOrigin = 'https://app.makeswift.com' }: Props) {
-  const { isEnabled: isDraftModeEnabled } = draftMode()
+  const { isEnabled: isDraftModeEnabled } = use(draftMode())
```
migueloller, line 10: This might be problematic if used in a root layout marked with `use client`. Not sure why anybody would do that but if somebody does, I've seen React complain about using `use` with a non-cached promise (i.e., promise created on render). It seems that we probably might want to just go ahead and make this a RSC?

packages/runtime/src/next/api-handler/handlers/proxy-draft-mode.ts
```diff
@@ -55,21 +57,21 @@ async function proxyDraftModeRouteHandler(
...
 
-  const draftModeCookie = cookies().get('__prerender_bypass')
+  const draftModeCookie = (await cookies()).get('__prerender_bypass')
   if (draftModeCookie) {
     proxyRequest.cookies.set(draftModeCookie)
     proxyRequest.cookies.set(MAKESWIFT_DRAFT_MODE_DATA_COOKIE, JSON.stringify(draftModeData))
   }
 
-  draftMode().disable()
+  await enableDraftMode(false)
```
migueloller, line 74: This is quite confusing (both _enable_ and `false` together). If passing `false` disables Draft Mode, the name of the function is completely contradictory 😅

migueloller (commented):
Ummmm, so when I was finished reviewing this PR I realized that it was in draft and you hadn't requested my review yet. I thought you had but it wasn't for this PR, it was for #851.

I already reviewed it so I'm submitting anyway, but I apologize ahead of time if my comments are irrelevant because you already changed your approach.

apps/nextjs-app-router/src/app/[lang]/[[...path]]/page.tsx
Comment (agurtovoy) [reply apps/nextjs-app-router/src/app/[lang]/[[...path]]/page.tsx:32]: The latter: it's compatible, but it'd require this change. We can allow `siteVersion` to be a `Promise` here if we want to keep the Next.js 14 upgrade path as smooth as possible.

apps/nextjs-pages-router/package.json
agurtovoy: From my reading of https://nextjs.org/blog/next-15#react-19, that seems to be a preferred community configuration. Let me know if you don't get the same read.

packages/runtime/src/next/lib/next.ts
agurtovoy: @migueloller The comment explains it:

> The wrapper functions below force the async use of the aforementioned primitives in the rest of the codebase even though they are typed as synchronous in Next.js 14.

packages/runtime/src/next/draft-mode/draft-mode-script.tsx
agurtovoy: Yep, was revising this just as you commented.

packages/runtime/src/next/api-handler/handlers/proxy-draft-mode.ts
agurtovoy: `enable(false)`, `enabled={false}`, or `enabled=false` are all [pretty](https://github.com/search?q=enabled%28false%29&type=code) [common](https://github.com/search?q=enabled%3D%7Bfalse%7D&type=code) [APIs](https://github.com/search?q=enabled%3Dfalse&type=code) out there, but happy to introduce a separate method for disabling if you find this confusing.

packages/runtime/src/next/lib/next.ts
agurtovoy: Good question, I thought `await enableDraftMode(false)` is clearer than `;(await draftMode()).enable()`, but maybe we want to get used to the semicolon here, haha

packages/runtime/src/next/api-handler/handlers/proxy-draft-mode.ts
agurtovoy: Now that I think about it more, I actually remember being appalled by `enable(false)` the first time I saw it in Windows APIs 25 years ago, hahaha.

migueloller: I'd be ok with `disableDraftMode()` and `enableDraftMode()`, hahaha

packages/runtime/src/next/lib/next.ts
migueloller: Sorry, hahaha. I read it but it didn't make sense to me so I was compelled to ask. I think I was looking for a user-facing reason, not a maintainer-facing one.

So my updated understanding here is that this is to avoid us forgetting to `await` the new APIs, not as utilities we actually export? If so, could we accomplish the same with the ESLint rule for [no floating promises](https://typescript-eslint.io/rules/no-floating-promises/)?

I feel like having the utility helps if you know the utility exists, but a maintainer could mistakenly import from `next` and that could be missed in code review.

packages/runtime/src/next/lib/next.ts
migueloller: Hahaha, I agree that would be a lot clearer. I think in this case I'd opt for what Next.js [does in their docs](https://nextjs.org/docs/app/api-reference/functions/draft-mode#enabling-draft-mode):

```ts
const draft = await draftMode()
draft.enable()
```

This is a nit and stylistic choice, though, so don't feel like this needs to be addressed to ship this thing. I just tend to avoid abstractions that just save some LOCs or characters but I can understand their appeal in certain scenarios where my personal taste wouldn't necessarily be attracted to it.

agurtovoy: For me, it's more about mental overhead of parsing and understanding "usage-land" code rather than saving LOCs/ characters. "Purity" of the API is less of a concern to me personally then readability/comprehension of the code that uses it.

apps/nextjs-app-router/package.json
```diff
@@ -12,18 +12,18 @@
     "@formatjs/intl-localematcher": "^0.5.4",
     "@makeswift/runtime": "workspace:^",
     "negotiator": "^0.6.3",
-    "next": "14.1.0",
-    "react": "^18",
-    "react-dom": "^18"
+    "next": "15.0.1",
+    "react": "19.0.0-rc-0bc30748-20241028",
+    "react-dom": "19.0.0-rc-0bc30748-20241028"
```
agurtovoy, line 17: Upgrade `apps/nextjs-app-router` deps to Next.js 15 and React 19's latest RC.

packages/runtime/src/runtimes/react/lib/react-dom.ts
```diff
@@ -0,0 +1,11 @@
...
+
+// https://github.com/facebook/react/blob/main/packages/shared/ReactDOMSharedInternals.js
+const reactDOMInternals = (ReactDOM as any)
+  .__DOM_INTERNALS_DO_NOT_USE_OR_WARN_USERS_THEY_CANNOT_UPGRADE as {
+  findDOMNode: typeof ReactDOM.findDOMNode
+}
+
+export const findDOMNode = ReactDOM.findDOMNode
+  ? ReactDOM.findDOMNode
+  : reactDOMInternals.findDOMNode
```
agurtovoy, line 11: Use publicly exposed `findDOMNode`, if any, otherwise reach into ReactDOM internals.

apps/nextjs-app-router/src/app/[lang]/[[...path]]/page.tsx
Comment (agurtovoy) [reply apps/nextjs-app-router/src/app/[lang]/[[...path]]/page.tsx]: Updated `getPageSnapshot` to allow `siteVersion` to be a `Promise` so this can literally be just a Next.js 15 compatibility release with no breaking changes.

packages/runtime/src/next/lib/next.ts
agurtovoy: Yep, I think I'm gonna go with the latest after all.

packages/runtime/src/next/index.tsx
```diff
@@ -4,6 +4,5 @@ export type { MakeswiftPage, MakeswiftPageDocument, MakeswiftPageSnapshot, Sitem
 export { Makeswift } from './client'
 export type { MakeswiftPreviewData } from './preview-mode'
 export { PreviewModeScript } from './preview-mode'
-export { Document } from './document'
```
agurtovoy, line 7: Breaking change: extracted `Document` export into its own entry point, `@makeswift/runtime/next/document`

apps/nextjs-app-router/src/app/[lang]/[[...path]]/page.tsx
Comment (agurtovoy) [reply apps/nextjs-app-router/src/app/[lang]/[[...path]]/page.tsx]: Converged on splitting out the `Document` export into its own entry point, `@makeswift/runtime/next/document`, so we can keep the `@makeswift/runtime/next` as our primary client-side import for the App Router.

PR #858, opened 2024-10-29
refactor: continuous builder document registration

arvinpoddar:
Currently, the runtime only sends builder documents during the initialization
of the builder <-> host connection. This means that any functionality that could
rely on a remount of a document after the connection has already been
established will not work, since the registration will not be communicated to
the builder.

This change addresses this by adding a document registration hook for embedded
components that sends a builder registration action on mount.

https://github.com/user-attachments/assets/d0c7a560-b3a0-4f94-92ae-f4a02fda36bc

### DIFF
--- a/packages/runtime/src/state/actions.ts
+++ b/packages/runtime/src/state/actions.ts
@@ -409,6 +409,18 @@ export function registerDocumentsEffect(
   }
 }
 
+export function registerBuilderDocumentsEffect(
+  documents: Document[],
+): ThunkAction<() => void, unknown, unknown, Action> {
+  return dispatch => {
+    documents.forEach(document => dispatch(registerBuilderDocument(document)))
+
+    return () => {
+      documents.forEach(document => dispatch(unregisterBuilderDocument(document.key)))
+    }
+  }
+}
+
 export function changeDocument(documentKey: string, operation: Operation): ChangeDocumentAction {
   return { type: ActionTypes.CHANGE_DOCUMENT, payload: { documentKey, operation } }
 }
--- a/packages/runtime/src/next/components/MakeswiftComponent.tsx
+++ b/packages/runtime/src/next/components/MakeswiftComponent.tsx
@@ -1,11 +1,10 @@
 'use client'
 
-import { memo, ReactNode, Suspense, useEffect, useMemo } from 'react'
+import { memo, ReactNode, Suspense, useMemo } from 'react'
 import { componentDocumentToRootEmbeddedDocument, MakeswiftComponentSnapshot } from '../client'
 import { DocumentRoot } from '../../runtimes/react/components/DocumentRoot'
 import { useSyncCacheData } from '../hooks/use-sync-cache-data'
-import { useDispatch } from '../../runtimes/react/hooks/use-dispatch'
-import { registerDocumentsEffect } from '../../state/actions'
+import { useRegisterDocument } from '../../runtimes/react/hooks/use-register-document'
 
 type Props = {
   snapshot: MakeswiftComponentSnapshot
@@ -15,8 +14,6 @@ type Props = {
 }
 
 export const MakeswiftComponent = memo(({ snapshot, name, type, fallback }: Props) => {
-  const dispatch = useDispatch()
-
   useSyncCacheData(snapshot.cacheData)
 
   const rootDocument = useMemo(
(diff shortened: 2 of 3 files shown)


packages/runtime/src/state/actions.ts
```diff
@@ -378,12 +378,21 @@ export function cleanUp(): CleanUpAction {
   return { type: ActionTypes.CLEAN_UP }
 }
 
-export function registerDocument(document: Document): RegisterDocumentAction {
-  return { type: ActionTypes.REGISTER_DOCUMENT, payload: { documentKey: document.key, document } }
+export function registerDocument(
+  document: Document,
+  builderInitiated = false,
+): RegisterDocumentAction {
```
arvinpoddar, line 384: With the introduction of the middleware, initiating a `registerDocument` via the builder will result in a ping-pong. As of now, this only applies to the case of focusing a global element:

1. Builder sends a `registerDocument` action via post message
2. Host receives action, registers a document. This will also kick off the middleware again, sending the document registration back to the builder.
3. When the builder receives the registration, the builder attempts to resubscribe to things since the document keys it has received have changed. This results in other actions getting missed since they are improperly buffered, preventing the global element from being interactive.

To resolve this, the builder can now call this method and pass `builderInitiated = true` to prevent the ping-ponging. Previous versions of the runtime won't have a middleware, so the ping-ponging won't happen anyways. The payload itself is backwards compatible since it's a new key.

I imagine this particular portion of the PR will be the area of most contention, so it's worth noting that I was able to get things working in the builder **without** having this carve out. However, it temporarily involves some tricks with our ShareDB reducers, and long-term (i.e., maybe this cooldown?) should use some refactoring that I've been wanting to do anyways. 

No matter what, some builder change will be required: right now, shipping the middleware alone will result in race conditions when attempting to edit global elements.

agurtovoy: I'm sure I'm missing something here, but since we are effectively translating `REGISTER_DOCUMENT` to a `registerBuilderDocument` call, can builder just dispatch `REGISTER_BUILDER_DOCUMENT` instead?

arvinpoddar: Not quite, since the builder dispatching `REGISTER_DOCUMENT` serves the purpose of populating the runtime store, which in turn changes what the host renders. `REGISTER_BUILDER_DOCUMENT` is solely for the host to communicate the mounting of a document to the builder, so dispatching that directly from the builder would serve no purpose.

fikrikarim (approved):
Looks good.

packages/runtime/src/state/actions.ts
migueloller: This is an unfortunate piece of technical debt that stems from using this action to "register" a document when a global component is focused. This is far from ideal. Instead, an action to "isolate" or "focus" a document should allow the runtime to copy the relevant document and then scope received ops to that. That might be a better solution. I'm sure there's also others that are far simpler than the status quo.

migueloller:
> Looks good. We could also use the existing `messageChannelMiddleware`, but you could argue making a new middleware is better since it's easier to understand the function of each middleware.

I think the `MessageChannel` middleware is an active part of this by making sure the dispatched actions make their way from the host to the builder and vice versa in addition to being sent to the Redux store's reducers.

packages/runtime/src/state/react-builder-preview.ts
```diff
@@ -519,6 +520,35 @@ function measureBoxModelsMiddleware(): Middleware<Dispatch, State, Dispatch> {
...
+            if (action.payload.builderInitiated) break
+            const { document } = action.payload
+            dispatch(
+              registerBuilderDocument({
+                ...document,
+                locale: document.locale ?? null,
+              }),
+            )
+            break
+          }
```
migueloller, line 538: Looking at this here seems a bit weird to me. We have a middleware that is "rewriting" actions. But if the action is being dispatched by the host, couldn't we `dispatch(registerBuilderDocument(...))` directly instead and avoid the rewrite? And if it's being dispatched from the builder, well `dispatch(registerDocument(...))` should do it. And now since the host and the builder use a different action wouldn't there be no ping-pong and also there would be no need for this "rewrite" middleware?

migueloller (approved):
This looks good but I have a small concern about whether the middleware is actually necessary or if we could simplify by dispatching the appropriate action on each side.

arvinpoddar (commented):
For now, I've settled on not using the middleware just yet, and relying on the more direct approach of dispatching the builder document registration when the component mounts. As I experiment with this, I've decided to only do this for our `MakeswiftComponent`, and not the root page. I may be revisiting this approach later.

PR #860, opened 2024-11-04
fix(runtime): `registerComponent` fails to correctly deduce resolved value type

agurtovoy:
Fixes regression introduced in 0.20.0 ("unified controls interface" release), includes a test reproducing the issue.

### DIFF
--- a/packages/runtime/src/runtimes/react/__tests__/react-runtime.test.tsx
+++ b/packages/runtime/src/runtimes/react/__tests__/react-runtime.test.tsx
@@ -1,0 +1,104 @@
+/** @jest-environment jsdom */
+import { type MouseEvent } from 'react'
+import {
+  Checkbox,
+  Combobox,
+  Color,
+  Image,
+  Link,
+  List,
+  Number,
+  Select,
+  Shape,
+  TextArea,
+  TextInput,
+} from '../../../controls'
+
+import { ReactRuntime } from '../react-runtime'
+
+type Card = {
+  imageSrc?: string
+  imageAlt: string
+  title: string
+  text: string
+  link: {
+    href: string
+    onClick: (event: MouseEvent) => void
+  }
+}
+
+type Entity = {
+  id: number
+  name: string
+}
+
+function Sandbox(props: {
+  cards: Card[]
+  entityId?: Entity['id']
+  theme: 'light' | 'dark'
+  background: string
+  boolean: boolean
+  number?: number
+}) {
+  return <div>{JSON.stringify(props)}</div>
+}
+
+describe('registerComponent', () => {
+  test("correctly deduces control definitions' resolved value types", () => {
+    ReactRuntime.registerComponent(Sandbox, {
+      type: 'sandbox',
+      label: 'Sandbox',
+      props: {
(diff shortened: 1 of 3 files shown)


packages/runtime/src/runtimes/react/__tests__/react-runtime.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+  background: string
+  boolean: boolean
+  number?: number
+}) {
+  return <div>{JSON.stringify(props)}</div>
+}
+
+describe('registerComponent', () => {
+  test("correctly deduces control definitions' resolved value types", () => {
+    ReactRuntime.registerComponent(Sandbox, {
```
agurtovoy, line 48: Reproduce for the issue, prior to the fix, this would fail with

repro

agurtovoy: I ran into this while upgrading our examples to 0.21.0. We tested 0.20.0 with both makeswift.com and bigcommerce.com implementations, but not with our own examples.

packages/runtime/src/runtimes/react/react-runtime.ts
```diff
@@ -23,7 +24,8 @@ export class ReactRuntime extends RuntimeCore {
   // We will remove them when we release a new breaking change.
   // ------------------ Deprecated API ------------------ //
   static registerComponent<
-    P extends Record<string, Descriptor>,
+    ControlDef extends UnifiedControlDefinition,
+    P extends Record<string, LegacyDescriptor | ControlDef>,
```
agurtovoy, line 28: Explicitly instruct TypeScript to deduce control definitions' types.

packages/runtime/src/runtimes/react/__tests__/react-runtime.test.tsx
arvinpoddar: Do we know what about this particular registration was resulting in the errors? I assume it was the reference to a subtype.

agurtovoy: The list of `Combobox` options was deduced as `{ id: string; value: Data; label: string }` instead of expected `{ id: string; value: number; label: string }`. If you factored out the `Combobox` prop def into a standalone variable, though, all of the types would be deduced properly:

```typescript
const entityId = Combobox({
  async getOptions() {
    return fetch(`/api/entities`)
      .then(r => r.json())
      .then((entities: Entity[]) =>
        entities.map(entity => ({
          id: entity.id.toString(),
          label: entity.name,
          value: entity.id,
        })),
      )
  },
  label: 'Entity',
})

...

ReactRuntime.registerComponent(Sandbox, {
   type: 'sandbox',
      label: 'Sandbox',
      props: {
         ...
         entityId, // this is fine
     ...
```

So the deduction was only failing in the context of `registerComponent`.

As far as what is specific about `Combobox` vs other control definitions, my guess would be is that `Combobox` is the only control which resolved value type is derived from the return type of a function.

migueloller (approved):
LGTM 👍🏻

PR #864, opened 2024-11-09
fix(controls): fix stalled calls/memory leaks on repeated function deserialization

agurtovoy:
See the added test for a reproduce:

### Before
failing test

### After
passing test

### DIFF
--- a/packages/controls/src/serialization/function-serialization.test.ts
+++ b/packages/controls/src/serialization/function-serialization.test.ts
@@ -1,0 +1,22 @@
+import {
+  deserializeFunction,
+  serializeFunction,
+} from './function-serialization'
+
+describe('function serialization', () => {
+  test('correctly deserializes serialized function', async () => {
+    const getItemLabel = (item: { label: string }) => item.label
+    const serialized = serializeFunction(getItemLabel)
+
+    const deserialized1 = deserializeFunction(serialized)
+    const deserialized2 = deserializeFunction(serialized)
+
+    expect(await deserialized1({ label: 'label 1' })).toBe('label 1')
+    expect(await deserialized1({ label: 'label 2' })).toBe('label 2')
+
+    expect(await deserialized2({ label: 'label 1' })).toBe('label 1')
+    expect(await deserialized2({ label: 'label 2' })).toBe('label 2')
+
+    serialized.close()
+  })
+})
--- a/packages/controls/src/serialization/function-serialization.ts
+++ b/packages/controls/src/serialization/function-serialization.ts
@@ -6,7 +6,16 @@ export function isFunction(value: any): value is AnyFunction {
   return typeof value === 'function'
 }
 
+type ResolveCallPromise<T extends AnyFunction> = (
+  value: SerializedFunctionReturnType<T>,
+) => void
+
+type OnMessageHandler<T extends AnyFunction> = MessagePort['onmessage'] & {
+  newCall?(resolve: ResolveCallPromise<T>): number
+}
+
 export type SerializedFunction<T extends AnyFunction> = MessagePort & {
+  onmessage: OnMessageHandler<T>
   readonly [SerializedFunctionTag]: T
 }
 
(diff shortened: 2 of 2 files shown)


packages/controls/src/serialization/function-serialization.ts
```diff
@@ -6,7 +6,16 @@ export function isFunction(value: any): value is AnyFunction {
   return typeof value === 'function'
 }
 
+type ResolveCallPromise<T extends AnyFunction> = (
+  value: SerializedFunctionReturnType<T>,
+) => void
```
agurtovoy, line 11: Simplified version of https://github.com/makeswift/makeswift/pull/864/files#diff-27828807c3388a8fc03da7ddb6fb171dc73e668a46695517bc7cd651c3cd017bL52

packages/controls/src/serialization/function-serialization.ts
```diff
@@ -44,30 +53,46 @@ export function serializeFunction<T extends AnyFunction>(
...
-
+function onmessageHandler<T extends AnyFunction>(): OnMessageHandler<T> {
+  type ResultMessageEvent = MessageEvent<
+    [CallID, SerializedFunctionReturnType<T>]
+  >
   let nextCallId = 0
-  const calls = new Map<CallID, ResolveCallPromise>()
+  const calls = new Map<CallID, ResolveCallPromise<T>>()
 
-  serializedFunction.onmessage = ({
```
agurtovoy, line 57: When `deserializeFunction` is called twice on the same `serializedFunction` instance, reassigning `serializedFunction.onmessage` here overrides/orphans any previous message handlers. That means:

- function calls to previously deserialized functions will never return, since the corresponding promises are never resolved
- these unresolved `Promises` effectively constitute a memory leak

packages/controls/src/serialization/function-serialization.ts
```diff
@@ -6,7 +6,16 @@ export function isFunction(value: any): value is AnyFunction {
   return typeof value === 'function'
 }
 
+type ResolveCallPromise<T extends AnyFunction> = (
+  value: SerializedFunctionReturnType<T>,
+) => void
+
+type OnMessageHandler<T extends AnyFunction> = MessagePort['onmessage'] & {
+  newCall?(resolve: ResolveCallPromise<T>): number
+}
```
agurtovoy, line 15: See below for why we are appending this optional method to the `onmessage` handler.

packages/controls/src/serialization/function-serialization.ts
```diff
@@ -44,30 +53,46 @@ export function serializeFunction<T extends AnyFunction>(
...
-    return new Promise((resolve) => {
-      const callId = nextCallId++
+  result.newCall = (resolve: ResolveCallPromise<T>) => {
+    const callId = nextCallId++
+    calls.set(callId, resolve)
+    return callId
+  }
 
-      calls.set(callId, resolve)
+  return result
```
agurtovoy, line 76: Use a separate closure to implicitly attach the calls map and the call counter to the `onmessage` handler. The `newCall` method allows us to append to this state from the outside.

I originally tried to attach this state directly to the `MessagePort` object, but because it's a built-in type, it cannot be extended that way.

packages/controls/src/serialization/function-serialization.test.ts
```diff
@@ -0,0 +1,22 @@
...
+  serializeFunction,
+} from './function-serialization'
+
+describe('function serialization', () => {
+  test('correctly deserializes serialized function', async () => {
+    const getItemLabel = (item: { label: string }) => item.label
+    const serialized = serializeFunction(getItemLabel)
+
+    const deserialized1 = deserializeFunction(serialized)
+    const deserialized2 = deserializeFunction(serialized)
```
agurtovoy, line 12: Deserializing the same serialized function twice used to "orphan" the previously deserialized function instances -- see below.

migueloller (approved):
Nice find. This code path was always one that I felt had a memory leak but I couldn’t quite put my finger on it. Thank you for fixing this!

PR #868, opened 2024-11-13
fix: "attempted import" errors on registering component in server-side code

agurtovoy:


packages/runtime/src/controls/rich-text-v2/rich-text-v2.ts
```diff
@@ -26,7 +26,7 @@ import { TextAlignPlugin } from '../../slate/TextAlignPlugin'
 import { BlockPlugin } from '../../slate/BlockPlugin'
 import { TypographyPlugin } from '../../slate/TypographyPlugin'
 import { InlineModePlugin } from '../../slate/InlineModePlugin'
-import { toText } from '../../slate'
+import { toText } from '../../slate/utils'
```
agurtovoy, line 29: Importing `slate/index` was the culprit here.

fikrikarim: Dang, this is such an easy mistake to make. I wonder if we can automate an eslint or another check to make sure we don't accidentally introduce something like this in the future again.

PR #869, opened 2024-11-13
feat: site theme component example

agurtovoy:


apps/nextjs-app-router/src/makeswift/client.ts
```diff
@@ -7,3 +8,12 @@ export const client = new Makeswift(MAKESWIFT_SITE_API_KEY, {
   runtime,
   apiOrigin: process.env.MAKESWIFT_API_ORIGIN,
 })
+
+export const getComponentSnapshot = async (snapshotId: string) => {
+  // const locale = await getLocale();
+
+  return await client.getComponentSnapshot(snapshotId, {
+    // locale: locale === defaultLocale ? undefined : locale,
```
joshuawootonn, line 16: Are the comments intentional?

apps/nextjs-app-router/src/app/[lang]/layout.tsx
```diff
@@ -14,17 +17,16 @@ export default async function RootLayout({
...
-          {children}
-        </MakeswiftProvider>
-      </body>
+      <MakeswiftProvider
+        previewMode={(await draftMode()).isEnabled}
+        locale={lang}
+      >
+        <head>
+          <DraftModeScript appOrigin={process.env.MAKESWIFT_APP_ORIGIN} />
+          <CssTheme colors={{}} />
```
Comment (joshuawootonn) [inline apps/nextjs-app-router/src/app/[lang]/layout.tsx:26]: thoughts on making colors optional?

apps/nextjs-app-router/src/makeswift/components/css-theme/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+import { CssTheme as CssThemeClient, PropsContextProvider } from './client'
+import { COMPONENT_TYPE } from './register'
+
+type Props = ComponentPropsWithoutRef<typeof CssThemeClient> & {
+  snapshotId?: string
+  label?: string
+}
+
+export const CssTheme = async ({
```
joshuawootonn, line 14: The naming here was hard for me to understand. Maybe this could be `HeadRootComponent` or something?

My understanding is this is the slot in our `<head/>` and we drop the other `CssComponent` into this slot and it renders the CSS.

joshuawootonn (approved):
Nice 👍🏼

apps/nextjs-app-router/src/makeswift/client.ts
agurtovoy: Yeah, it's a TODO, we need to pass a locale here for this to work w/ localization.

apps/nextjs-app-router/src/makeswift/components/css-theme/index.tsx
agurtovoy: > The naming here was hard for me to understand. Maybe this could be HeadRootComponent or something?

The pattern used here (server component wrapping a client component embedded in the snapshot) is what we settled on in Catalyst for defining all "embedded" components in general. This particular one is named for what it does (defines a CSS theme) rather where you'd use it, which I think what components should be named. Does it makes sense?

> My understanding is this is the slot in our <head/> and we drop the other CssComponent into this slot and it renders the CSS.

This a server component that is tied to a specific component snapshot, `site-theme`, and that component snapshot, once edited, will store the element tree for our client component registered under `COMPONENT_TYPE`.

agurtovoy:
Obsolete by the new iteration on the theming.

PR #872, opened 2024-11-16
fix: do not crash on incompatible prop data

agurtovoy:
https://github.com/user-attachments/assets/5e9245ed-7d0a-4e2e-833d-e7bb7591407e

### DIFF
--- a/packages/runtime/src/runtimes/react/controls/control.tsx
+++ b/packages/runtime/src/runtimes/react/controls/control.tsx
@@ -26,6 +26,8 @@ import {
   unstable_TypographyDefinition,
 } from '../../../controls'
 
+import { FallbackComponent } from '../../../components/shared/FallbackComponent'
+
 import { RenderHook } from '../components'
 import { useColorValue } from './color'
 import { useImageControlValue } from './image'
@@ -38,13 +40,26 @@ import { useSlot } from './slot'
 import { useFormattedStyle } from './style'
 import { StyleV2ControlValue } from './style-v2'
 import { useTypographyValue } from './typography'
-import { useResolvedValue } from '../hooks/use-resolved-value'
 
 type ControlValueProps = {
   definition: ControlDefinition
   data: DataType<ControlDefinition> | undefined
   children(value: ResolvedValueType<ControlDefinition>): JSX.Element
   control?: InstanceType<ControlDefinition>
+}
+
+type Result<T> = { success: true; value: T } | { success: false; error: string }
+
+export function resolveValue(
+  data: DataType<ControlDefinition> | undefined,
+  definition: ControlDefinition,
+): Result<ResolvedValueType<ControlDefinition>> {
+  try {
+    return { success: true, value: definition.resolveValue(data).readStableValue() }
+  } catch (error) {
+    console.error(`Error resolving '${definition.controlType}' prop value`, { error, data })
+    return { success: false, error: `${error}` }
+  }
 }
 
 export function ControlValue({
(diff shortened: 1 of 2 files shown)


packages/runtime/src/runtimes/react/controls/control.tsx
```diff
@@ -47,6 +48,20 @@ type ControlValueProps = {
...
 }
 
+type Result<T> = { success: true; value: T } | { success: false; error: string }
+
+export function resolveValue(
+  data: DataType<ControlDefinition> | undefined,
+  definition: ControlDefinition,
+): Result<ResolvedValueType<ControlDefinition>> {
+  try {
+    return { success: true, value: definition.resolveValue(data).readStableValue() }
```
agurtovoy, line 58: Handling errors coming from the unified controls interface only. I originally added an error boundary around `ElementData`'s `<Suspense>` boundary and let these propagate there, but in the `localhost`/dev setup this still results in React showing its (collapsed) errors banner on the page, and that didn't look right to me.

migueloller: Note, React’s error banner even happens on console.error and is controversial in the community. I wouldn’t weigh it heavily

packages/runtime/src/runtimes/react/controls/control.tsx
```diff
@@ -47,6 +48,20 @@ type ControlValueProps = {
...
+  } catch (error) {
+    console.error(`Error resolving '${definition.controlType}' prop value`, { error, data })
+    return { success: false, error: `${error}` }
```
migueloller, line 61: Will this interpolated error properly stringify?

packages/runtime/src/runtimes/react/controls/control.tsx
```diff
@@ -61,14 +76,13 @@ export function ControlValue({
     case SelectDefinition.type:
     case IconRadioGroupDefinition.type:
     case ComboboxDefinition.type:
-      return (
-        <RenderHook
-          key={definition.controlType}
-          hook={useResolvedValue}
-          parameters={[data, definition]}
-        >
+      const resolved = resolveValue(data, definition)
```
migueloller, line 79: Can we call `resolveValue` like this on render and expect proper subscriptions, etc. to happen resulting in re-renders?

migueloller (commented):
This looks good. Primary concern is calling `resolveValue` on render instead of using React hooks to “synchronize” like `useSyncExternalStore`. 

My understanding here is that if we fail to resolve the value we show a fallback. This seems reasonable. How will this impact UX? Is it possible to “fix” the data by selecting the fallback?

packages/runtime/src/runtimes/react/controls/control.tsx
agurtovoy: Great question, forgot to comment on this, this only works because so far we've merged only a "half-way" version of `resolveValue` that is effectively a straight up data transformation call. There are no subscriptions or side-effects here. The proper `useResolvedProps` hook will have to tackle this differently (or maybe fallback to the error boundary approach after all).

packages/runtime/src/runtimes/react/controls/control.tsx
agurtovoy: > React’s error banner even happens on console.error

I haven't see that.

> I wouldn’t weigh it heavily

I guess it depends on whether we consider this scenario (mismatching prop data) to be worth of polish. For all of its seeming triviality, the React banner really spoils the impression that we have everything under control.

agurtovoy:
> My understanding here is that if we fail to resolve the value we show a fallback. This seems reasonable. How will this impact UX? Is it possible to “fix” the data by selecting the fallback?

Not currently, but that's where my head went as well. Right now you have to use the builder panel to fix the offending prop(s), I have a demo of it in the video.

packages/runtime/src/runtimes/react/controls/control.tsx
agurtovoy: I think so:

error

PR #875, opened 2024-11-20
Fix tailwind configuration in `/apps`

joshuawootonn:


apps/nextjs-app-router/src/app/[lang]/[[...path]]/page.tsx
```diff
@@ -4,6 +4,8 @@ import { getSiteVersion } from '@makeswift/runtime/next/server'
 import { notFound } from 'next/navigation'
 import { Page as MakeswiftPage } from '@makeswift/runtime/next'
 
+import '@/makeswift/components'
```
Comment (agurtovoy) [inline apps/nextjs-app-router/src/app/[lang]/[[...path]]/page.tsx:7]: This one shouldn't be needed, we need this import in only two places: root layout and Makeswift provider component file.

apps/nextjs-app-router/src/app/[lang]/layout.tsx
```diff
@@ -2,6 +2,9 @@ import { MakeswiftProvider } from '@/makeswift/provider'
 import { DraftModeScript } from '@makeswift/runtime/next/server'
 import { draftMode } from 'next/headers'
 
+import '@/app/global.css'
+import '@/makeswift/components'
```
Comment (agurtovoy) [inline apps/nextjs-app-router/src/app/[lang]/layout.tsx:6]: This one is needed.

apps/nextjs-app-router/src/app/api/makeswift/[...makeswift]/route.ts
```diff
@@ -1,5 +1,6 @@
 import { MAKESWIFT_SITE_API_KEY } from '@/makeswift/env'
 import { MakeswiftApiHandler } from '@makeswift/runtime/next/server'
+import '@/makeswift/components'
```
Comment (agurtovoy) [inline apps/nextjs-app-router/src/app/api/makeswift/[...makeswift]/route.ts:3]: Not needed.

apps/nextjs-pages-router/src/pages/api/makeswift/[...makeswift].ts
```diff
@@ -2,6 +2,8 @@ import { MAKESWIFT_SITE_API_KEY } from '@/makeswift/env'
 import { runtime } from '@/makeswift/runtime'
 import { MakeswiftApiHandler } from '@makeswift/runtime/next/server'
 
+import '@/makeswift/components'
```
Comment (agurtovoy) [inline apps/nextjs-pages-router/src/pages/api/makeswift/[...makeswift].ts:5]: Not sure if this one is needed. If it is, we need to update our docs as well: https://docs.makeswift.com/developer/pages-router/installation

Comment (joshuawootonn) [reply apps/nextjs-pages-router/src/pages/api/makeswift/[...makeswift].ts:5]: It's needed if you use the Smartling integration.

Comment (joshuawootonn) [reply apps/nextjs-pages-router/src/pages/api/makeswift/[...makeswift].ts:5]: I can remove these for now since we don't use these apps for translation right now.

Comment (joshuawootonn) [reply apps/nextjs-pages-router/src/pages/api/makeswift/[...makeswift].ts:5]: Actually. I'm going to leave this for pages router. It's non trivial to figure out why translation is failing when this is not included, so I would prefer to keep it. Since app router doesn't support translations right now I'll leave it our of the `route.ts` file there.

Comment (agurtovoy) [reply apps/nextjs-pages-router/src/pages/api/makeswift/[...makeswift].ts:5]: Can we add a comment that this needed for the Smartling integration?

agurtovoy (approved):
I'd add a comment about the Smartlink issue.

PR #838, opened 2024-09-20
runtime: multi-doc support (MAIN)

agurtovoy:

PR #876, opened 2024-11-24
chore: add automation for publishing release candidates

Codeseph:
ENG-6796

For something like https://github.com/makeswift/makeswift/pull/838, we want people to be able to use @makeswift/runtime@rc instead of asking them to find a snapshot. 

What I'm trying to do here:
- create a prerelease with the distribution tag 'rc' when either (A) a PR is opened to merge a branch with '-rc-' in its name into main or (B) a commit is made to a branch with '-rc-' in the name that already has an open PR to be merged into main
- prevent the existing snapshot release process from picking up branches with '-rc-' in their name
- update the main release process (release.yml) so that it exits prerelease mode if necessary before versioning/publishing (i.e., if we merge a PR from a release candidate into main without having manually exited prerelease mode)

A few thoughts:
- I wonder if there is a branch name pattern we'd prefer over '-rc-'? We've talked about including the dist tag in the branch name (i.e., having 'release/alpha' or similar as a branch naming pattern), but right now I'm concerned about the amount of time I might spend trying to get the parsing/validation working and whether we'd get enough value out of having multiple prerelease stages for that to be worthwhile. 
- Being new to both GitHub actions and changesets, I am concerned about tinkering with the release process. I'm not sure how to test something like this and have peace of mind prior to merging. 
- I would understand if we want to handle rc publishes manually for now versus trying this out.

### DIFF
--- a/.github/workflows/rc-release.yml
+++ b/.github/workflows/rc-release.yml
@@ -1,0 +1,82 @@
+name: Release Candidate Release
+
+on:
+  pull_request:
+    # Run when PR opened and when subsequent commits are made https://github.com/orgs/community/discussions/25161 
+    # Will further restrict to only rc branches targeting the main branch below (in the 'if' under the job)
+    types: [opened, synchronize]
+
+concurrency: ${{ github.workflow}}-${{ github.ref }}
+
+jobs:
+  release_candidate_release:
+    if: contains(github.head_ref, '-rc-') && github.base_ref == 'main'
+    name: Release Candidate Release
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+    steps:
+
+      # Setup steps
+      - name: Check out git repo
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Set up pnpm 8.x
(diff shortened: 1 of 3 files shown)


migueloller (changes requested):
This does achieve what you set out to do pretty well but I've got a couple of concerns:
- It would seem that we want to opt in when we want to release a RC version as opposed to having every commit merged to a branch result in a new RC. For example, for `latest` releases, we have to decide to merge the release PR and not every commit to `main` results in a release
- When it comes to automating a release, we probably want to require either merging a PR or pushing a tag, combined with requiring PR approvals or tag protection so that a release can go through our approval workflows
- We seem to be checking if we're in prerelease mode for the release workflow, but since that workflow only runs for `main`, it seems weird that we'd be in prerelease mode in `main`. We shouldn't have to do this check since `main` should never go into prerelease mode.
- I still don't have a good grasp on what the "workflow" is that we're automating. Seems like until we have a good understanding of that, we shouldn't automate yet

Codeseph:
Will resurrect this after the current cycle. Going to publish manually for now and come back when we know exactly what we want.

PR #877, opened 2024-11-25
fix(runtime): use separate finddomnode for fallbacks

arvinpoddar:
Use a separate `FindDomNode` for rendering the fallback which won't be overruled by the element callback ref.

Before:

https://github.com/user-attachments/assets/39c61206-7868-4e1f-a519-73b7882f34cc

After:

https://github.com/user-attachments/assets/298ea89d-16bb-4da4-b9b9-9ff90e6656b7

### DIFF
--- a/packages/runtime/src/runtimes/react/components/Element.tsx
+++ b/packages/runtime/src/runtimes/react/components/Element.tsx
@@ -24,6 +24,10 @@ export const ElementWithFallback = memo(
     const useFindDomNodeRef = useRef(true)
     const imperativeHandleRef = useRef(new ElementImperativeHandle())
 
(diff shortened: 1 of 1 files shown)


migueloller (approved):
This looks good!

Something it reminds me of if a problem we have when the component's found DOM node changes. For example, imagine you had this:

```ts
function RichText({ node }) {
  switch (node.type) {
    case 'paragraph':
      return <p>Paragraph</p>

    case 'heading-one':
      return <h1>Heading 1</h1>

    default:
      return <div>Unknown rich text node</div>
  }
}
```

If you change `node.type` via a panel, then the ref is lost. I think this is because we don't properly handle this sort of change reactively in our implementation of `<FindDomNode>`.

Anyway, this seems like a separate thing from what you fix here, but it's somewhat related and I wanted to share it with you.

PR #881, opened 2024-12-04
refactor(runtime): pass initial root element directly

arvinpoddar:
Rather than passing metadata about whether the root element exists or not, we now allow the root element to be null, and also pass an initial root element field. This allows us to derive whether the root exists, as well as use this initial root to reset the data later.

### DIFF
--- a/packages/runtime/src/next/client.ts
+++ b/packages/runtime/src/next/client.ts
@@ -173,19 +173,22 @@ export function componentDocumentToRootEmbeddedDocument({
   name: string
   type: string
 }): EmbeddedDocument {
-  const { data, locale, id } = document
+  const { data: rootElement, locale, id } = document
 
-  if (data != null && data.type !== type) {
+  if (rootElement != null && rootElement.type !== type) {
     throw new Error(
-      `Type "${data.type}" does not match the expected type "${type}" from the snapshot`,
+      `Type "${rootElement.type}" does not match the expected type "${type}" from the snapshot`,
     )
   }
 
-  const rootElement = data ?? {
+  const initialRootElement = {
     // Fallback rootElement
     // Create a stable uuid so two different clients will have the same empty element data.
     // This is needed to make presence feature work for an element that is not yet created.
-    key: deterministicUUID({ id, locale, seed: document.seed }),
+    key:
+      rootElement != null
+        ? rootElement.key
+        : deterministicUUID({ id, locale, seed: document.seed }),
     type,
     props: {},
   }
@@ -193,6 +196,7 @@ export function componentDocumentToRootEmbeddedDocument({
   const rootDocument: EmbeddedDocument = {
     key: uuid(),
     rootElement,
+    initialRootElement,
     locale,
     id,
     type,
--- a/packages/runtime/src/runtimes/react/components/Document.tsx
+++ b/packages/runtime/src/runtimes/react/components/Document.tsx
@@ -3,6 +3,7 @@ import { type Document as ReactPageDocument } from '../../../state/react-page'
 import { ElementImperativeHandle } from '../element-imperative-handle'
 import { DocumentContext } from '../hooks/use-document-context'
 import { Element } from './Element'
+import { getRootElementFromDocument } from '../../../state/utils/get-root-element-from-document'
 
 type DocumentProps = {
   document: ReactPageDocument
@@ -13,9 +14,11 @@ export const Document = memo(
     { document }: DocumentProps,
     ref: Ref<ElementImperativeHandle>,
   ): JSX.Element {
+    const rootElement = getRootElementFromDocument(document)
+
     return (
       <DocumentContext.Provider value={document}>
-        <Element ref={ref} element={document.rootElement} />
+        <Element ref={ref} element={rootElement} />
       </DocumentContext.Provider>
     )
   }),
--- a/packages/runtime/src/state/actions.ts
+++ b/packages/runtime/src/state/actions.ts
@@ -2,7 +2,7 @@ import type { Operation } from 'ot-json0'
 
 import { ControlInstance } from '@makeswift/controls'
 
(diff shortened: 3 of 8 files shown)


packages/runtime/src/state/actions.ts
```diff
@@ -99,10 +99,15 @@ type DocumentPayloadBaseDocument = {
...
+  key: string
+  locale?: string | null // older versions of the runtime may not provide this field
   id: string
   type: string
   name: string
+  rootElement: Element | null
+  initialRootElement: Element
+  hasFallback?: boolean
   __type: typeof EMBEDDED_DOCUMENT_TYPE
 }
```
arvinpoddar, line 112: This is a breaking change in the data type being transacted between the host and builder. Is it permissible to make this change since embedded components are not being used?

fikrikarim: Not used by users? AFAIK, people are using it for the one-click catalyst demo.

arvinpoddar: Ah, let me rephrase: "not being used in general acceptable". Since the OCC demo users are limited, is it acceptable to tell them to upgrade their runtime instead of having to support backwards compatibility?

fikrikarim: I'm not sure about that. Let's talk about this during stand-up.

fikrikarim (dismissed):
LGTM if we're okay with the breaking change

packages/runtime/src/state/actions.ts
migueloller: What did you conclude after chatting with stakeholders @arvinpoddar?

packages/runtime/src/next/client.ts
```diff
@@ -175,35 +175,36 @@ export function componentDocumentToRootEmbeddedDocument({
...
 
-  const rootElement = data ?? {
+  const initialRootElement = {
     // Fallback rootElement
     // Create a stable uuid so two different clients will have the same empty element data.
     // This is needed to make presence feature work for an element that is not yet created.
-    key: deterministicUUID({ id, locale, seed: document.seed }),
+    key:
+      rootElement != null
+        ? rootElement?.key
```
migueloller, line 192: Optional chaining is unnecessary here since we check `rootElement` is not `null`

packages/runtime/src/runtimes/react/components/Document.tsx
```diff
@@ -15,15 +15,14 @@ export const Document = memo(
     { document, fallback }: DocumentProps,
     ref: Ref<ElementImperativeHandle>,
   ): JSX.Element {
-    const isInitialData = match(document)
-      .with({ meta: { isInitialData: true } }, () => true)
-      .otherwise(() => false)
+    const isInitialData = document.rootElement == null
+    const rootElement = getDocumentRootElement(document)
```
migueloller, line 19: Isn't `getDocumentRootElement` a selector? I don't think we should "pierce" into modules under the `state` directory as this is meant to be encapsulated by the interface the wraps the Redux store. Also, selectors should be reserved to be used with Redux state, not as general getters when we have data that matches the state. We only want to have to change selectors when the state changes, and not couple it to this part of the codebase, which isn't dealing with that state module

packages/runtime/src/state/actions.ts
arvinpoddar: @migueloller Chatted with Nathan, and the answer is **it should be possible** to force all OCC hosts to update. We'd have to do it with a script that crawls the existing projects and redeploy them. Given that there are nearly 200 OCC hosts that would be affected, I didn't feel great about this option - I would've been okay if it had been a few dozen. As much as I didn't want backwards compatibility to be a concern so early, I don't see another way.

packages/runtime/src/runtimes/react/components/Document.tsx
migueloller: I noticed later that this isn't a selector. The naming is unfortunate, though, since it follows the same naming convention there. We also seem to be using it similar to selectors elsewhere (`Documents.getRootElement`). I think we should place this in a shared place that is not the `state` directory so that we can keep things in those modules encapsulated.

migueloller (changes requested):
> This allows us to derive whether the root exists, as well as use this initial root to reset the data later.

In what cases do we need to "reset the data later"?

packages/runtime/src/runtimes/react/components/Document.tsx
arvinpoddar: Fair point, will address.

arvinpoddar:
> > This allows us to derive whether the root exists, as well as use this initial root to reset the data later.
> 
> In what cases do we need to "reset the data later"?

"Reset" refers to replacing the root element with the initial root element. This mechanic will likely be used for resetting to base locale and deletion (triggered by a message sent from the builder).

packages/runtime/src/next/client.ts
arvinpoddar: Good catch, I was trying to do a type deduction with the optional chaining at first, must've left it behind.

migueloller:
> > > This allows us to derive whether the root exists, as well as use this initial root to reset the data later.
> > 
> > 
> > In what cases do we need to "reset the data later"?
> 
> "Reset" refers to replacing the root element with the initial root element. This mechanic will likely be used for resetting to base locale and deletion (triggered by a message sent from the builder).

So in this case the initial document wouldn't be the empty element but instead the base locale's element? I thought we were going to use deletion of the element tree as reset to base so that things wouldn't stay "forked"

packages/runtime/src/runtimes/react/components/Document.tsx
arvinpoddar: Ended up making a separate directory within `state` to store this utility function. It felt weird to me to remove this from `state` when the function operates on a type that was defined in a `state` module - moreover, all usages (except for one) of this function are within `state` as well.

Codeseph (approved):

migueloller (approved):

PR #886, opened 2024-12-09
feat(runtime): utilize v2 endpoint for fetching element trees

Codeseph:
eng-7065

Updating the runtime to use the new v2 endpoint for fetching element trees that was added in eng-6954

Here we decided on making the siteVersion param mandatory: https://makeswifthq.slack.com/archives/C080BNSNEG7/p1733744344493169

### DIFF
--- a/packages/runtime/src/next/client.ts
+++ b/packages/runtime/src/next/client.ts
@@ -149,15 +149,11 @@ const makeswiftComponentDocumentSchemaV2 = z.object({
 
 export type MakeswiftComponentDocumentSchemaV2 = z.infer<typeof makeswiftComponentDocumentSchemaV2>
 
-const getMakeswiftComponentDocumentSchemaV2 = z.object({
-  elementTree: makeswiftComponentDocumentSchemaV2.nullable(),
-  parentLocaleElementTree: makeswiftComponentDocumentSchemaV2.nullable()
-})
-.and(z.union([
+const getMakeswiftComponentDocumentSchemaV2 = z.union([
   z.object({ elementTree: z.null(), parentLocaleElementTree: makeswiftComponentDocumentSchemaV2 }),
   z.object({ elementTree: makeswiftComponentDocumentSchemaV2, parentLocaleElementTree: z.null()}),
   z.object({ elementTree: makeswiftComponentDocumentSchemaV2, parentLocaleElementTree: makeswiftComponentDocumentSchemaV2 }),
-], {errorMap: (_issue, _ctx) => ({message: "One or both of 'elementTree' and 'parentLocaleElementTree' must be present."})}))
+], {errorMap: (_issue, _ctx) => ({message: "One or both of 'elementTree' and 'parentLocaleElementTree' must be present."})})
 
 const makeswiftComponentDocumentSchema = z.object({
   id: z.string(),
--- a/packages/runtime/src/next/client.ts
+++ b/packages/runtime/src/next/client.ts
@@ -683,7 +683,7 @@ export class Makeswift {
 
     // If the element tree is not found, we generate a document with null data
     if (response.status === 404) {
-      return this.getDummyDocument(id, locale)
+      return this.getSnapshotWithFallbackDocument(id, locale)
     }
 
     const json = await response.json()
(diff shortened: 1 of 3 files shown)


packages/runtime/src/next/client.ts
```diff
@@ -658,45 +674,41 @@ export class Makeswift {
...
-        document: {
-          id,
-          locale: locale ?? null,
-          data: null,
-          // We pass the seed to generate the elementKey to make sure it's unique for each site.
-          seed: apiKeyPrefix,
-        },
-        cacheData: CacheData.empty(),
-      }
+      return this.getDummyDocument(id, locale)
```
migueloller, line 692: I would like to have a more descriptive name than "dummy document"

packages/runtime/src/next/client.ts
```diff
@@ -658,45 +674,41 @@ export class Makeswift {
...
+      console.error('Failed to get component snapshot', json)
+      throw new Error(`Failed to get component snapshot with error: "${response.statusText}"`)
+    }
+
+    const elementTreeResponse = getMakeswiftComponentDocumentSchemaV2.parse(json)
+
+    // favor picking 'elementTree' if it is available.
+    const document = elementTreeResponse.elementTree ?? elementTreeResponse.parentLocaleElementTree
+
+    // This case should never be reached (response was an http 200, but neither the element tree nor the parent locale element tree were provided in the response)
```
migueloller, line 707: If this case should never be reached, we should model that in our parsing such as to have the type system enforce it. I think this is better than having a comment + fallback code

packages/runtime/src/next/client.ts
```diff
@@ -138,6 +138,27 @@ export type MakeswiftPageSnapshot = {
...
+  siteId: z.string(),
+})
+
+export type MakeswiftComponentDocumentSchemaV2 = z.infer<typeof makeswiftComponentDocumentSchemaV2>
+
+const getMakeswiftComponentDocumentSchemaV2 = z.object({
+  elementTree: makeswiftComponentDocumentSchemaV2.nullable(),
+  parentLocaleElementTree: makeswiftComponentDocumentSchemaV2.nullable()
+})
+.and(z.union([
```
Codeseph, line 156: Here I'm trying to enforce at least one of the two nullable fields being present. Basing this on a suggestion I saw here: https://github.com/colinhacks/zod/issues/61#issuecomment-1534255967

zod unions: https://zod.dev/?id=unions

packages/runtime/src/next/client.ts
Codeseph: Added to the zod type for this

packages/runtime/src/next/client.ts
```diff
@@ -652,45 +673,36 @@ export class Makeswift {
...
-      console.error('Failed to get page snapshot', json)
-      throw new Error(`Failed to get page snapshot with error: "${response.statusText}"`)
+      console.error('Failed to get component snapshot', json)
+      throw new Error(`Failed to get component snapshot with error: "${response.statusText}"`)
     }
 
-    const document = makeswiftComponentDocumentSchema.parse(json)
+    const elementTreeResponse = getMakeswiftComponentDocumentSchemaV2.parse(json)
+
+    // favor picking 'elementTree' if it is available.
```
migueloller, line 703: Super mega nit incoming...

My stance on comments is that if they describe what the code is doing, they're unnecessary. Sometimes they might even become outdated and cause confusion. I think comments should be generally reserved to explain "why". If we're having to explain "what" with comments then it's a sign we can factor the code in a better way.

packages/runtime/src/next/client.ts
migueloller: How come we're using both `.and` and `z.union`? It feels like we should use just one of those (I personally prefer `z.union`).

Also, is the very first case necessary? Seems like we're handling all valid cases within `z.union`. I see that this is how it's done in the suggestion you found. I'm wondering if it could be simplified a bit or if we have to do it that way

migueloller (approved):
LGTM! Should we add some tests?

packages/runtime/src/next/client.ts
Codeseph: definitely agree, removed this and on other

migueloller (approved):

Codeseph:
Closing this - following yesterday's discussion it sounds like we're no longer going to be utilizing the v2 endpoint, since we're not going to "merge" on reads

migueloller:
> Closing this - following yesterday's discussion it sounds like we're no longer going to be utilizing the v2 endpoint, since we're not going to "merge" on reads

🫡 ❤️

PR #888, opened 2024-12-09
refactor(runtime): remove fallback prop

arvinpoddar:
Remove support for passing a fallback to the `MakeswiftComponent` component.

### DIFF
--- a/packages/runtime/src/state/modules/read-only-documents.ts
+++ b/packages/runtime/src/state/modules/read-only-documents.ts
@@ -29,10 +29,6 @@ export type EmbeddedDocument = BaseDocument & {
   id: string
   type: string
   name: string
-  meta?: {
-    isInitialData?: boolean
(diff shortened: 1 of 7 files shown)


packages/runtime/src/state/modules/read-only-documents.ts
```diff
@@ -29,10 +29,6 @@ export type EmbeddedDocument = BaseDocument & {
   id: string
   type: string
   name: string
-  meta?: {
-    isInitialData?: boolean
-    hasFallback?: boolean
-  }
```
arvinpoddar, line 35: Not a breaking change since this is optional. Both of these fields are exclusively for fallback related builder logic.

migueloller (approved):
RIP 🪦 `fallback` 🫡

fikrikarim:
RIP 🪦 fallback 🫡

PR #889, opened 2024-12-10
feat: add site.published webhook handler for global revalidation

fikrikarim:
This adds revalidation support for <MakeswiftComponent /> on App Router

### DIFF
--- a/packages/runtime/src/next/api-handler/handlers/manifest.ts
+++ b/packages/runtime/src/next/api-handler/handlers/manifest.ts
@@ -14,6 +14,7 @@ export type Manifest = {
   siteVersions: boolean
   unstable_siteVersions: boolean
   localizedPageSSR: boolean
+  webhook: boolean
 }
 
 type ManifestError = { message: string }
@@ -67,6 +68,11 @@ export default async function handler(
     .with(apiRoutePattern, () => false)
     .exhaustive()
 
+  const supportsWebhook = match(args)
+    .with(routeHandlerPattern, () => true)
+    .with(apiRoutePattern, () => false)
+    .exhaustive()
+
   const body = {
     version: PACKAGE_VERSION,
     previewMode: supportsPreviewMode,
@@ -78,6 +84,7 @@ export default async function handler(
     siteVersions: true,
     unstable_siteVersions: true,
     localizedPageSSR: true,
+    webhook: supportsWebhook,
   }
 
   return match(args)
--- a/packages/runtime/src/next/api-handler/handlers/webhook/index.ts
+++ b/packages/runtime/src/next/api-handler/handlers/webhook/index.ts
@@ -1,0 +1,75 @@
+import { NextApiRequest, NextApiResponse } from 'next'
+import { NextRequest, NextResponse } from 'next/server'
+import { P, match } from 'ts-pattern'
+import { handleSitePublished } from './site-published'
+import {
+  sitePublishedWebhookPayloadSchema,
+  WebhookEventType,
+  WebhookPayloadSchema,
+  WebhookResponseBody,
+} from './types'
+
+type Context = { params: { [key: string]: string | string[] } }
+
+export type WebhookHandlerArgs =
+  | [request: NextRequest, context: Context, params: { apiKey: string }]
+  | [req: NextApiRequest, res: NextApiResponse<WebhookResponseBody>, params: { apiKey: string }]
+
+const routeHandlerPattern = [P.instanceOf(Request), P.any, P.any] as const
+const apiRoutePattern = [P.any, P.any, P.any] as const
+
+function getSecret(args: WebhookHandlerArgs) {
+  return match(args)
+    .with(routeHandlerPattern, ([request]) => request.nextUrl.searchParams.get('secret'))
+    .with(apiRoutePattern, ([req]) => req.query.secret)
+    .exhaustive()
+}
+
+function getRequestBody(args: WebhookHandlerArgs) {
+  return match(args)
+    .with(routeHandlerPattern, ([request]) => request.json())
+    .with(apiRoutePattern, ([req]) => req.body)
+    .exhaustive()
+}
+
+function respond(args: WebhookHandlerArgs, response: WebhookResponseBody, status?: number) {
+  return match(args)
+    .with(routeHandlerPattern, () => NextResponse.json(response, { status }))
+    .with(apiRoutePattern, ([, res]) => res.json(response))
+    .exhaustive()
+}
+
+export default async function handler(
+  request: NextRequest,
+  context: Context,
+  { apiKey }: { apiKey: string },
+): Promise<NextResponse<WebhookResponseBody>>
+export default async function handler(
+  req: NextApiRequest,
+  res: NextApiResponse<WebhookResponseBody>,
+  { apiKey }: { apiKey: string },
+): Promise<void>
+export default async function handler(
+  ...args: WebhookHandlerArgs
+): Promise<NextResponse<WebhookResponseBody> | void> {
+  const [, , { apiKey }] = args
+  const secret = getSecret(args)
+
+  if (secret !== apiKey) return respond(args, { message: 'Unauthorized' }, 401)
+
+  let payload: WebhookPayloadSchema
+
+  try {
+    const body = await getRequestBody(args)
+    payload = sitePublishedWebhookPayloadSchema.parse(body)
+  } catch (error) {
+    console.error(error)
+    return respond(args, { message: 'Invalid request body' }, 400)
+  }
+
+  const result = match(payload.type)
+    .with(WebhookEventType.SITE_PUBLISHED, () => handleSitePublished(payload))
+    .exhaustive()
+
+  return respond(args, result.body, result.status)
+}
--- a/packages/runtime/src/next/api-handler/handlers/webhook/site-published.ts
+++ b/packages/runtime/src/next/api-handler/handlers/webhook/site-published.ts
@@ -1,0 +1,10 @@
+import { revalidateTag } from 'next/cache'
+import { SitePublishedWebhookPayload, WebhookHandlerResult } from './types'
+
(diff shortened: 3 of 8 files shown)


packages/runtime/src/next/api-handler/handlers/manifest.ts
```diff
@@ -14,6 +14,7 @@ export type Manifest = {
   siteVersions: boolean
   unstable_siteVersions: boolean
   localizedPageSSR: boolean
+  webhook: boolean
```
fikrikarim, line 17: >What if the manifest provided the URL? 🤔

@migueloller We can make this dynamic and return the string value depending on where user put the api handler. 

But I think there's little value in doing that right now, given that we also hardcoded the `manifest` URL and other endpoints. I feel like it's more important to make the `manifest` URL configurable first. But then it's in the realm of the first-class host project.

packages/runtime/src/next/api-handler/index.ts
```diff
@@ -195,6 +198,13 @@ export function MakeswiftApiHandler(
         .exhaustive()
     }
 
+    if (matches('/webhook')) {
```
fikrikarim, line 201: Do we want to deprecate or remove the `/revalidate` endpoint now? 🤔

packages/runtime/src/next/api-handler/handlers/manifest.ts
alexhwoods: I second this, at least for now

packages/runtime/src/next/api-handler/handlers/webhook/site-published.ts
```diff
@@ -0,0 +1,10 @@
+import { revalidateTag } from 'next/cache'
+import { SitePublishedWebhookPayload, WebhookHandlerResult } from './types'
+
+export function handleSitePublished(_payload: SitePublishedWebhookPayload): WebhookHandlerResult {
+  revalidateTag(MAKESWIFT_CACHE_TAG)
+
+  return { body: { success: true }, status: 200 }
+}
+
+export const MAKESWIFT_CACHE_TAG = '@@makeswift'
```
alexhwoods, line 10: Should this be site specific?

For the multi-tenant case

fikrikarim: That's a good question. I didn't think about multi-tenant cases.

We could make it site-specific by using the site-api key. But I don't think it'll solve the multi-tenancy by itself. Perhaps this is also a reason why we need to expose `onPublish` option?

We also need to discuss what's the args for that `onPublish`. Pass the diff to `onPublish`?

For our default host, even right now we have a custom API handler, so I don't think we need to worry about our default host for now.

migueloller: Why do we need `onPublish` for this? We should be able to retrieve the site and use its ID as part of the tag

packages/runtime/src/next/api-handler/handlers/manifest.ts
migueloller: That sounds good to me. We have plans to rework this with first-class hosts anyway

packages/runtime/src/next/api-handler/handlers/webhook/index.ts
```diff
@@ -0,0 +1,75 @@
...
+  const [, , { apiKey }] = args
+  const secret = getSecret(args)
+
+  if (secret !== apiKey) return respond(args, { message: 'Unauthorized' }, 401)
+
+  let payload: WebhookPayloadSchema
+
+  try {
+    const body = await getRequestBody(args)
+    payload = sitePublishedWebhookPayloadSchema.parse(body)
```
migueloller, line 64: What's the status on us adding web hook signatures? If we have support for them already, we should validate the signature here

packages/runtime/src/next/api-handler/index.ts
migueloller: I think so. This is all internal, though, so it would mostly be for us to not continue to extend it

migueloller (commented):
This LGTM 👍🏻 

The revalidation is quite coarse, though. I think a good next step after this one would be to have per-site coarse revalidation and then we can move to granular, per-resource revalidation.

packages/runtime/src/next/api-handler/index.ts
fikrikarim: Sounds good. Added a new commit for this.

packages/runtime/src/next/api-handler/handlers/webhook/index.ts
fikrikarim: We don't have webhook signatures right now. We validate the request by the `siteApiKey`. Is this enough?

packages/runtime/src/next/api-handler/index.ts
fikrikarim: I changed my mind. Let's keep this endpoint for now until we conditionally disable revalidation on the backend. I don't want to remove this endpoint if the backend is still sending revalidation requests.

migueloller: Sorry, I didn't clarify that my response was "I think so" to "deprecate", not to "remove". We definitely shouldn't remove it just yet.

packages/runtime/src/next/api-handler/handlers/webhook/site-published.ts
migueloller: > We could get the siteId by fetching by siteApiKey. And then, instead of using @@makeswift as the global tag, we use siteId as the tag. When a site is published, we call revalidateTag(siteId).

Yes, that is exactly what I mean.

I'm not sure I see this as an intermediate step only. We might still want to keep support for being able to revalidate everything in an entire Makeswift site (e.g., having a button in the builder to revalidate all). Having a hardcoded string isn't quite there yet given it doesn't represent a Makeswift site, but a site ID does. Multi-tenant hosts will be able to get the site ID given the `Host` header without any issue

packages/runtime/src/next/api-handler/handlers/webhook/index.ts
migueloller: I guess we're using the site's API key as the signature. It feels weird to me that we're using the site's API key this way instead of having a web hook secret that can be used to create a signature to ensure the web hook payload is to be trusted. Sending the API key on each request (which I know we already do for revalidation) feels very insecure. Maybe there's nothing to do right now since we're using the site's API key for revalidation and we haven't implemented web hook signatures just yet but I feel dirty letting this slide...

<details>
<summary>Here's what the oracle says</summary>

When implementing webhooks, using a **webhook signature** is generally preferred over sending a **shared API key** in the URL or headers for several important security reasons:

### 1. **Reduced Risk of Exposure**
   - **API key exposure**: Sending an API key in the URL or headers makes it vulnerable to interception. URLs (including query parameters) may be logged in web server logs, browser history, or through third-party monitoring tools. This increases the risk of exposing sensitive credentials.
   - **Signature approach**: With a webhook signature, you are sending a hashed value that is generated based on the payload and a secret key. Even if the signature is intercepted, without the secret key, it cannot be verified or manipulated to produce a valid signature. This adds a layer of security by not directly exposing any sensitive data in transit.

### 2. **Integrity and Authenticity**
   - **API key validation**: API keys can be easily guessed or copied if not stored securely. An attacker who gains access to the key can impersonate the webhook provider, sending fraudulent requests.
   - **Signature validation**: Webhook signatures use cryptographic hashing (usually HMAC or similar) to ensure that the payload has not been tampered with in transit. The receiving server checks the signature against the payload and a shared secret to verify that the request genuinely comes from the expected source and that the data hasn’t been altered.

### 3. **Better Protection Against Replay Attacks**
   - **API key exposure**: If an API key is used in the query string or headers, it could potentially be reused by an attacker (replay attack), allowing them to resend requests from valid sources to the server without needing to authenticate or sign the request.
   - **Signature approach**: Signatures typically incorporate time-based or nonce-based components to prevent replay attacks. For example, a webhook signature could include a timestamp, and if a request is received with an outdated timestamp, it can be rejected as a potential replay attack.

### 4. **No Need for Storing Sensitive Data in Requests**
   - **API keys in headers/URL**: Storing sensitive keys in requests, especially in plain text or query parameters, increases the risk of unintentional exposure through logging or inspection tools.
   - **Signature approach**: The signature only reveals the result of a hashed process, not the original secret key. As long as the secret key is kept secure on the server and is not transmitted in the request, the attack surface is significantly reduced.

### 5. **Flexible Security Handling**
   - **API keys**: With API keys, any change (e.g., key rotation) would require updates to all clients, which can be cumbersome and error-prone.
   - **Signature approach**: If a secret key needs to be rotated, the update is often more straightforward since the signature is generated from a known shared secret but doesn't expose the key itself in the request. The system can easily re-sign with the new key without altering the rest of the request structure.

### 6. **Prevents Unauthorized Access**
   - **API key-based access**: Once an API key is exposed or leaked, attackers could potentially gain access to the service and impersonate the webhook provider.
   - **Signature-based security**: Since the signature is unique to the specific payload and time, unauthorized parties cannot forge the correct signature without access to the secret key. This makes it far more difficult for attackers to gain unauthorized access.

### Summary
In short, **webhook signatures** offer a more secure method for verifying the authenticity of webhook requests, ensuring the integrity of the payload, and minimizing the risk of exposure or misuse of sensitive credentials like API keys. They provide an added layer of security by relying on cryptographic signatures instead of transmitting shared secrets directly in the request.
</details>

fikrikarim: I understand. We need to do the same for all the API handlers like the manifest too. We decided not to worry about this for now.

packages/runtime/src/next/api-handler/handlers/webhook/site-published.ts
fikrikarim: This is doable, but we decided to not do this for now.

alexhwoods (approved):

PR #891, opened 2024-12-12
refactor(runtime): remove initial root element from documents

arvinpoddar:
In light of the new decisions for modeling variants, we do not have a need for communicating an initial root element to revert to on deletion.


migueloller (commented):
Doesn't the builder still need to know whether there is an initial element there to determine whether to submit a create operation? Or is that decoupled from the runtime and based on whether the ShareDB document exists? The latter seems better 🤔

arvinpoddar:
> Doesn't the builder still need to know whether there is an initial element there to determine whether to submit a create operation? Or is that decoupled from the runtime and based on whether the ShareDB document exists? The latter seems better 🤔

@migueloller It's based entirely on the latter. This field originally surfaced from the need for `fallbacks` on `MakeswiftComponent`, and then after our design meeting yesterday, there's no need for it at all since we'll be using data from the backend as the source of truth for what to replace the host element trees with.

migueloller (approved):

PR #892, opened 2024-12-13
Add slot component with fallback capability

Codeseph:
eng-7087

https://github.com/user-attachments/assets/b209e72a-df59-44cc-bec3-f49038dce6b2


packages/runtime/src/components/builtin/Slot/register.ts
```diff
@@ -0,0 +1,20 @@
...
+export function registerComponent(runtime: ReactRuntime) {
+  return runtime.registerComponent(
+    lazy(() => import('./SlotRoot')),
+      {
+        type: MakeswiftComponentType.Slot, 
+        label: 'Slot',
+        hidden: true,
+        props: {
+          children: Slot(),
+          showFallback: Checkbox({ label: "Name of checkbox test", defaultValue: true })
```
jamesqquick, line 16: I think this label needs to be updated

packages/runtime/src/components/builtin/Slot/SlotRoot.tsx
```diff
@@ -0,0 +1,21 @@
...
+  children: ReactNode
+  showFallback: boolean
+}
+
+export const SlotRoot = (props: Props) => {
+  const ctx = useContext(PropsContext)
+
+  return (
+      <>
+        { props.showFallback ? props.children : ctx.fallback}
```
jamesqquick, line 16: Out of curiosity, why does `fallback` come from a context?

packages/runtime/src/components/builtin/Slot/register.ts
Codeseph: Yikes. Thank you. Changed to 'Use Fallback' for now, but let me know if you prefer something else.

Codeseph: Yikes. Changed to 'Use Fallback' for now, but let me know if you'd prefer a different label.

packages/runtime/src/components/builtin/Slot/SlotRoot.tsx
Codeseph: The situation that prompted this is that the new slot component is used in hardcoded routes, but requires interactivity in the builder. The implication of this is that its props need to come from two different sources - some props need to come from controls (i.e. the content that is dragged into the slot), and other props need to be hardcoded (i.e. the snapshot ID). So the props that are hardcoded are being passed to SlotRoot via context, and the other props are coming from the controllers as usual.

packages/runtime/src/components/builtin/constants.ts
```diff
@@ -13,6 +13,7 @@ export const MakeswiftComponentType = {
   Image: './components/Image/index.js',
   Navigation: './components/Navigation/index.js',
   Root: './components/Root/index.js',
+  Slot: './components/Slot/index.js',
```
migueloller, line 16: The path in the types is an artifact of the original paths of the builtin components before we event had custom hosts. Let's instead use something that's not a path. How about `makeswift::components::slot::v1`?

packages/runtime/src/components/builtin/Slot/register.ts
```diff
@@ -0,0 +1,20 @@
...
+export function registerComponent(runtime: ReactRuntime) {
+  return runtime.registerComponent(
+    lazy(() => import('./SlotRoot')),
+      {
+        type: MakeswiftComponentType.Slot, 
+        label: 'Slot',
+        hidden: true,
+        props: {
+          children: Slot(),
+          showFallback: Checkbox({ label: "Use Fallback", defaultValue: true })
```
migueloller, line 16: I believe all of our panel labels are in "Sentence case" as opposed to "Title Case"

packages/runtime/src/components/builtin/Slot/SlotRoot.tsx
migueloller, line 1: We should keep the name `Slot.tsx` in the file and the component export

packages/runtime/src/components/builtin/Slot/register.ts
```diff
@@ -0,0 +1,20 @@
+import { lazy } from 'react';
+import { ReactRuntime } from '../../../react'
+import { MakeswiftComponentType } from '../constants'
+import { Checkbox } from '@makeswift/controls'
+import { Slot } from '../../../controls';
```
migueloller, line 5: If you have conflicts with this name when renaming `SlotRoot` to `Slot`, you can alias this import like so

```suggestion
import { Slot as SlotControl } from '../../../controls';
```

packages/runtime/src/next/index.tsx
```diff
@@ -1,6 +1,7 @@
 export type { PageProps } from './components/page'
 export { Page } from './components/page'
 export { MakeswiftComponent } from './components/MakeswiftComponent'
+export { MakeswiftSlot } from './components/MakeswiftSlot'
```
migueloller, line 4: I think we should keep the name as `Slot` instead of `MakeswiftSlot`. It does conflict with the control but they each have different import paths. I'm curious what you think @jamesqquick 

I want us to avoid having every single import have a Makeswift prefix in it

packages/runtime/src/next/components/MakeswiftSlot.tsx
```diff
@@ -0,0 +1,35 @@
...
+}
+
+export const PropsContext = createContext<{fallback: ReactNode}>({fallback: null});
+
+export const PropsContextProvider = ({ value, children }: { value: {fallback: ReactNode}, children: ReactNode }) => (
+  <PropsContext.Provider value={value}>{children}</PropsContext.Provider>
+);
+
+export const MakeswiftSlot = ({ label, snapshot, fallback }: Props) => {
+  const componentType = MakeswiftComponentType.Slot
```
migueloller, line 22: This intermediate variable is unnecessary

packages/runtime/src/next/components/MakeswiftSlot.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+  return (
+    <PropsContextProvider value={{fallback}}>
```
migueloller, line 25: This will create a new object on every render. We can use `useMemo` to avoid that.

packages/runtime/src/next/components/MakeswiftSlot.tsx
```diff
@@ -0,0 +1,35 @@
+'use client'
```
migueloller, line 1: We should define the context in a separate component that we can have a `<Slot>` RSC that takes the ID and calls `getComponentSnapshot` itself. The non-RSC version would be used only in Pages Router

packages/runtime/src/components/builtin/Slot/SlotRoot.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+import { ReactNode, useContext } from "react"
+import { PropsContext } from "../../../next/components/MakeswiftSlot"
+
+type Props = {
+  children: ReactNode
+  showFallback: boolean
+}
+
+export const SlotRoot = (props: Props) => {
```
migueloller, line 11: Let's use function declarations for components instead of arrow functions

packages/runtime/src/components/builtin/Slot/SlotRoot.tsx
migueloller, line 16: There's some extraneous spaces here. Do you have Prettier configured to run on save on your IDE?

migueloller, line 16: Have we verified our implementation of `findDOMNode` properly handles finding the DOM now whens switching from fallback to children?

migueloller (changes requested):
Looking good 👍🏻 

Some tests and/or a video showing the API/component in action would be much appreciated

Codeseph:
> > [eng-7087](https://linear.app/makeswift/issue/ENG-7087)
> > Screen.Recording.2024-12-16.at.4.12.55.PM.mov
> 
> Are we able to just pass `null` for the empty fallback? Or better yet, omit the fallback prop altogether?

Yeah - updated the recording, thanks

packages/runtime/src/components/builtin/Slot/SlotRoot.tsx
Codeseph: I haven't had any issues switching back and forth between them repeatedly fwiw

Codeseph: with telegraph or otherwise

Codeseph: Working for me in cosmos but not our public repo, and I don't know why. I haven't overridden any config or anything.

packages/runtime/src/next/index.tsx
jamesqquick: I do think separating it somehow from the control would be nice. I guess the difference is where it's imported from? I don't personally mind `MakeswiftSlot` I don't think. What about `SlotComponent`? Is that overly clarifying?

jamesqquick:
> Are we able to just pass `null` for the empty fallback? Or better yet, omit the fallback prop altogether?

So right now, this means the developer has to pass a fallback or nothing shows? I would personally expect to see the default box if I don't pass a fallback at all. I think that might mess with the conditional checkbox logic a bit though?

packages/runtime/src/next/index.tsx
jamesqquick: Decided on sticking with `Slot` for now.

packages/runtime/src/components/builtin/Slot/Slot.tsx
```diff
@@ -0,0 +1,21 @@
...
+  children: ReactNode
+  showFallback: boolean
+}
+
+function Slot(props: Props) {
+  const ctx = useContext(PropsContext)
+
+  return (
+      <>
+        { props.showFallback ? ctx.fallback : props.children }
```
migueloller, line 16: Seems we still have spacing inconsistencies. You can run `pnpm format` in the monorepo to format all files.

We should add this to our CI workflow 🤔

packages/runtime/src/next/components/SlotProvider.tsx
```diff
@@ -0,0 +1,17 @@
+import { createContext, ReactNode } from "react";
+
+export const PropsContext = createContext<{
+    fallback: ReactNode
+}>({
+    fallback: null
+});
+
+function SlotProvider({ value, children }: { value: {fallback: ReactNode}, children: ReactNode }) {
+    return (
```
migueloller, line 10: Looks like we've got 4 spaces for indentation here. `pnpm format` should take care of it

migueloller (dismissed):
This is looking good! All we need is to run `pnpm format`. Maybe we can slide in a CI check for it while we're at it? 😏 (half-joking)

packages/runtime/src/next/components/MakeswiftSlot.tsx
migueloller: We're not doing this right now.

packages/runtime/src/components/builtin/Slot/SlotRoot.tsx
migueloller: Are you working at the top-level of the monorepo?

migueloller: Great to hear!

Codeseph: Yeah I'm at the top level. From what I'm seeing when I run prettier using the CLI, I'm wondering if others have been having this issue too. Running prettier even just for`packages/runtime/src` gives quite a few changes
image

Odd. Anyway I'll take care of formatting the relevant files here using the CLI

Codeseph: for me, later: `pnpm prettier packages/runtime/src --write`

Codeseph:
updated the changeset message + fixed prettier formatting

migueloller (dismissed):
Looks good. Just had a minor question

packages/runtime/src/components/builtin/Slot/SlotRoot.tsx
migueloller: This is why we should enforce it in CI, hahaha

packages/runtime/src/components/builtin/index.ts
```diff
@@ -9,6 +9,7 @@ export { default as Form } from './Form'
 export { default as Image } from './Image'
 export { default as Navigation } from './Navigation'
 export { default as Root } from './Root'
+export { default as MakeswiftSlot } from './Slot'
```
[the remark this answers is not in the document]
Codeseph: Renamed to 'Slot'. Must have missed this when I was removing all the 'MakeswiftSlot' verbiage. Thanks

migueloller (approved):

PR #893, opened 2024-12-14
Require site version for fetching component snapshot

Codeseph:
For now we're looking to mirror the behavior of getPageSnapshot, see: https://makeswifthq.slack.com/archives/C080BNSNEG7/p1733744344493169


migueloller (approved):
LGTM 👍🏻

PR #895, opened 2024-12-16
Draft: Runtime changes

naqi:
Proposing some changes to allow user to force http traffic in case they are hosting inside a container which is not running nextjs in `https` mode.


naqi:
closing, merge history is suspect

PR #897, opened 2024-12-19
New shape control with label and popover panel

joshuawootonn:
Demo

https://github.com/user-attachments/assets/6c31d996-5e56-41e2-8527-09ba7456dfa1


packages/controls/src/controls/shape-v2/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './shape'
```
joshuawootonn: This commit will be "fixed down" into the commit below before I merge, just wanted to make reviewing a bit easier so you can compare the two Shape controls

packages/controls/src/controls/shape-v2/shape-v2.ts
```diff
@@ -0,0 +1,322 @@
...
+  fromData(data: DataType<C> | undefined): ValueType<C> | undefined {
+    if (data == null) return undefined
+
+    if (
+      ControlDataTypeKey in data &&
+      data[ControlDataTypeKey] === Definition.v1DataType
+    ) {
+      return mapValues(this.keyDefs, (def, key) =>
+        def.fromData((data as ShapeV2DataTypeV1<C>).value[key]),
+      )
```
agurtovoy, line 180: I _think_ we should be able to avoid casting here if we use `ts-pattern` along the lines of https://github.com/makeswift/makeswift/blob/main/packages/controls/src/controls/image/image.ts#L202-L203

agurtovoy: Let me know if you want to work through this together.

packages/controls/src/controls/shape-v2/shape-v2.ts
```diff
@@ -0,0 +1,322 @@
...
+  ): DataType<C> | undefined {
+    if (data == null) return undefined
+
+    if (
+      ControlDataTypeKey in data &&
+      data[ControlDataTypeKey] === Definition.v1DataType
+    ) {
+      return mapValues(this.keyDefs, (def, key) =>
+        def.copyData((data as ShapeV2DataTypeV1<C>).value[key], context),
+      )
```
agurtovoy, line 207: Same here: https://github.com/makeswift/makeswift/blob/466ed3b357f277fe17185822ce9320d9153c84ea/packages/controls/src/controls/image/image.ts#L265

packages/controls/src/controls/shape-v2/shape-v2.ts
```diff
@@ -0,0 +1,322 @@
...
+  label?: string
+  layout?: typeof Definition.Layout.Inline | typeof Definition.Layout.Popover
+  readonly type: Defs
+}
+
+export type ShapeV2DataTypeV0<C extends Config> = {
+  [K in keyof C['type']]?: DataType_<C['type'][K]>
+}
+
+export const ShapeV2DataTypeV1VersionId = 'shape-v2::v1' as const
```
agurtovoy, line 41: I would not pollute the global scope with this; I think we can just say `typeof Definition.v1DataType` below.

packages/controls/src/controls/shape-v2/shape-v2.ts
agurtovoy: Actually, since the only difference between v1 and v2 data is whether the value is nested inside of the `value` prop, ideally we shouldn't have two branches of the code that processes it and have something like this instead:

```typescript
const shapeData = toShapeData(data)
return mapValues(this.keyDefs, (def, key) =>
        def.fromData(data[key]),
      )
```

packages/controls/src/controls/shape-v2/shape-v2.ts
```diff
@@ -0,0 +1,322 @@
...
+      layout: L
+    }
+
+export class ShapeV2Definition<
+  C extends Config = Config,
+> extends Definition<C> {}
+
+export function ShapeV2<D extends KeyDefinitions, L extends Config['layout']>(
+  config: UserConfig<D, L>,
+): ShapeV2Definition<NormedConfig<D, L>> {
```
agurtovoy, line 314: Since our data type doesn't really depend on the `layout`, I wonder if we should bother with the complexity of trying to infer the specific type of the layout passed here.

packages/controls/src/controls/shape-v2/shape-v2.ts
```diff
@@ -0,0 +1,322 @@
...
+
+export type Config<Defs extends KeyDefinitions = KeyDefinitions> = {
+  label?: string
+  layout?: typeof Definition.Layout.Inline | typeof Definition.Layout.Popover
+  readonly type: Defs
+}
+
+export type ShapeV2DataTypeV0<C extends Config> = {
+  [K in keyof C['type']]?: DataType_<C['type'][K]>
+}
```
agurtovoy, line 39: Do we need to define/publicly expose this? I see that you are using it in the builder, but IMHO it'd be clearer/more precise to use `DataType<ShapeDefinition<C>>` for the v0 data.

packages/controls/src/controls/index.ts
```diff
@@ -16,6 +16,7 @@ export * from './text-area'
 export * from './text-input'
 export * from './font'
 export * from './typography'
+export * from './shape-v2'
```
agurtovoy, line 19: Can we nest this version under `shape/v2` the way we did it for rich text? This way our catalog of controls will stay more manageable as we add more and more versions.

packages/controls/src/controls/shape-v2/shape-v2.ts
joshuawootonn: I tried `ts-pattern` and it wasn't working I think because of v1 is a sub type of v0. 

I **love** the idea of `toShapeData` so I'll try that and also do something similar in leo

packages/controls/src/controls/shape-v2/shape-v2.ts
joshuawootonn: Ok. I need this value in cosmos, and `v1DataType` is private in the schema. Are you ok with me exposing it?

packages/controls/src/controls/shape-v2/shape-v2.ts
joshuawootonn: > IMHO it'd be clearer/more precise to use DataType<ShapeDefinition<C>> for the v0 data.
Nice. I agree.

What are your thoughts on `ShapeV2DataTypeV1`?

packages/controls/src/controls/index.ts
joshuawootonn: I can change this. 

I was kinda iffy on this decision because `RichText` has sub folders, but `StyleV2` doesn't.

packages/controls/src/controls/shape-v2/shape-v2.ts
joshuawootonn: This the error I get from doing something similar with `ts-pattern`

CleanShot 2024-12-20 at 10 19 29@2x

CleanShot 2024-12-20 at 10 19 23@2x

.changeset/hot-pumpkins-clean.md
```diff
@@ -0,0 +1,26 @@
...
+## Breaking change
+
+### Updated `Shape` control
+
+The `Shape` control has been updated with two new params `label` and `layout`.
+
+- `label?: string = "Group"`
+  - Text for the panel label in the Makeswift builder.
+  - This `label` defaults to "Group."
+- `layout?: 'Shape.Layout.Inline | Shape.Layout.Popover = Shape.Layout.Popover`
```
migueloller, line 15:
```suggestion
- `layout?: Shape.Layout.Inline | Shape.Layout.Popover = Shape.Layout.Popover`
```

.changeset/hot-pumpkins-clean.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Breaking change
+
+### Updated `Shape` control
+
+The `Shape` control has been updated with two new params `label` and `layout`.
+
+- `label?: string = "Group"`
+  - Text for the panel label in the Makeswift builder.
+  - This `label` defaults to "Group."
```
migueloller, line 14:
```suggestion
  - This `label` defaults to `"Group"`.
```

.changeset/hot-pumpkins-clean.md
```diff
@@ -0,0 +1,26 @@
...
+- `layout?: 'Shape.Layout.Inline | Shape.Layout.Popover = Shape.Layout.Popover`
+  - The configuration for how panels are rendered in the right sidebar.
```
migueloller, line 16:
```suggestion
  - The configuration for how panels are rendered in the Properties Sidebar.
```

We should use the same naming we use in our docs

.changeset/hot-pumpkins-clean.md
```diff
@@ -0,0 +1,26 @@
...
+  - The `Inline` option is the default and it is similar to the current rendering style of the `Shape` popover.
+  - The `Popover` option is a new rendering style that only shows your label with a button in the current context. When this button is clicked a popover opens to the left revealing the controls within your `Shape`.
+
+### How to migrate
+
+This change is listed as minor version because it visually changes how your panels are rendered and we no longer support the old version. Upgrading the `@makeswift/runtime` to this version will update how the right sidebar looks for a `Shape` control to include a `label` and have better visual hierarchy.
```
migueloller, line 22:
```suggestion
This change is listed as a breaking change because it visually changes how your panels are rendered and we no longer support the old version. Upgrading `@makeswift/runtime` to this version will update how the Properties Sidebar looks for a `Shape` control to include a `label` and have a better visual hierarchy.
```

.changeset/hot-pumpkins-clean.md
```diff
@@ -0,0 +1,26 @@
...
+
+This breaking change doesn't require any code changes, but we do suggest adding descriptive labels since the `Shape` control's `label` will default to "Group."
```
migueloller, line 24:
```suggestion
This breaking change doesn't require any code changes, but we do suggest adding descriptive labels since the `Shape` control's `label` will default to `"Group"`.
```

packages/controls/src/controls/shape-v2/shape-v2.ts
joshuawootonn: Ahh geez. Looks like there was an error in my implementation that could have caused this error. looking...

packages/controls/src/controls/shape-v2/shape-v2.ts
agurtovoy: We can make `v1DataType` public.  I think ideally we want to find a way to avoid checking it directly outside of the control implementation, though.

packages/controls/src/controls/shape-v2/shape-v2.ts
agurtovoy: Unfortunately, there is no way to scope types in TypeScript. We did end up exposing `RichTextDataV2` for rich text v2.

packages/controls/src/controls/index.ts
agurtovoy: Yeah, we should do this to style, too.

packages/controls/src/controls/shape-v2/shape-v2.ts
joshuawootonn: In Cosmos, we are using the `version` to differentiate data. Should I be doing something similar?

CleanShot 2024-12-20 at 13 26 28@2x

joshuawootonn: Actually that is for the version of the image control

packages/controls/src/controls/shape-v2/shape-v2.ts
joshuawootonn: Removed

packages/controls/src/controls/shape-v2/shape-v2.ts
joshuawootonn: As long as we agree on the `toShapeData` I have above then all of these such locations are now using that and don't have branching.

packages/controls/src/controls/shape-v2/shape-v2.ts
joshuawootonn: I don't think we have any references in leo to actual data versions.

packages/controls/src/controls/shape-v2/shape-v2.ts
joshuawootonn: Just kidding. I think we should update the data instead of downgrading it. Adding a. method to the definition
```

  versionData(data: DataType<C>): ShapeV2DataTypeV1<C> {
    if (data == null || !(ControlDataTypeKey in data)) {
      return {
        [ControlDataTypeKey]: Definition.v1DataType,
        value: data,
      }
    }
    return data as ShapeV2DataTypeV1<C>
  }
```

This is kinda like the `normalizeData` in the `RichTextV2` control

packages/controls/src/controls/shape-v2/shape-v2.ts
joshuawootonn: Good news. With a `versionData` static function this is no longer required.

packages/controls/src/controls/shape-v2/shape-v2.ts
joshuawootonn: With the export new static `versionData` function I no longer need this in cosmos.

.changeset/hot-pumpkins-clean.md
joshuawootonn: 💯. I agree.  I couldn't find [this page](https://docs.makeswift.com/product/builder-basics#the-basics) at the time, when I was searching there were quite a few "right sidebar" usages, but I like this change.

CleanShot 2024-12-20 at 14 45 32@2x

joshuawootonn: Ping @jamesqquick. It might be good to update these "right sidebar" instances to say Properties sidebar.

joshuawootonn:
Thanks for the feedback @migueloller / @agurtovoy 🙌🏼 . I think your comments are addressed in the commits I just pushed up.

migueloller (dismissed):
Love it! 👍🏻 

Got some minor comments, but this looks ready to me

.changeset/hot-pumpkins-clean.md
migueloller: Same with this one, if we're indicating a default value that's a string, wouldn't we want to put it as "code"? Also, I understand that the period inside quotes is grammatically the correct thing to do, but I think this would be confusing for readers and it's another reason to do '`"Group"`.'

Anyway, this is a nit, just trying to keep things clear for users reading this.

I just wasn't sure if you meant to address this given you didn't respond

packages/runtime/src/controls/index.ts
```diff
@@ -35,6 +34,9 @@ export {
   unstable_TypographyDefinition,
   Font,
   FontDefinition,
+  ShapeV2 as Shape,
+  ShapeV2Control,
+  ShapeV2Definition,
```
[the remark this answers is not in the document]
joshuawootonn: Can you explain more here? 

Are you suggesting `ShapeV2` becoming `Shape` and then the old `Shape` being called `LegacyShape` or something like that?

agurtovoy (approved):
Beautiful, ship it! :shipit:

packages/runtime/src/controls/index.ts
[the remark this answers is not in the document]
migueloller: Nah, I'm suggesting having `ShapeV1`, `ShapeV2` and then `Shape`, which uses `ShapeV2`. Once we have `ShapeV3`, then we would change `Shape` to use that

PR #900, opened 2024-12-31
runtime: getComponentSnapshot performs locale fallback

Codeseph:
eng-7269

This would make the default behavior of getComponentSnapshot be to perform locale fallback if a request for some localized element tree 404s. By "locale fallback" here we're referring to execution of a subsequent request for the same element tree ID but in the base locale.

Whether or not locale fallback is utilized is configurable via an argument 'allowLocaleFallback' to getComponentSnapshot. In the future, when getComponentSnapshot is internal to MakeswiftComponent, I'm imagining this being configurable via a prop. 

This change, along with eng-7271 (not done yet), are working towards creating the UX we want for element tree variants as described in [Localized element trees, inheritance, and fanouts](https://linear.app/makeswift/document/localized-element-trees-inheritance-and-fanouts-2e2ff9f29c60)

### DIFF
--- a/packages/runtime/src/next/client.ts
+++ b/packages/runtime/src/next/client.ts
@@ -162,6 +162,7 @@ export type MakeswiftComponentSnapshot = {
   document: MakeswiftComponentDocument | MakeswiftComponentDocumentFallback
   key: string
   cacheData: CacheData
+  localeFallbackOccurred: boolean
 }
 
 export function componentDocumentToRootEmbeddedDocument({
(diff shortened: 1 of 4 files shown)


packages/runtime/src/next/client.ts
```diff
@@ -697,6 +708,22 @@ export class Makeswift {
...
+      allowLocaleFallback = true,
+    }: { siteVersion: MakeswiftSiteVersion; locale?: string; allowLocaleFallback?: boolean },
+  ): Promise<Response> {
+    const baseLocaleWasRequested = locale == null
+    if (initialResponse.status !== 404 || !allowLocaleFallback || baseLocaleWasRequested) {
+      return initialResponse
+    }
+    return await this.fetch(`v1/element-trees/${id}`, siteVersion)
+  }
+
```
migueloller, line 726: This is some very weird level of indirection that I don't think we should have. It makes it so that the calling function doesn't know whether the fallback happened or not. This is important for us to construct the appropriate snapshot and handle what the builder should show. The snapshot should include information about the originally requested resource, and this is lost if we simply return the response from the fallback.

I would strongly recommend working on this end to end, or at least doing a spike first to understand what all the moving parts will be to avoid negative work.

PR #902, opened 2025-01-06
refactor(shape v2): introduce `itemsData`, `itemDataPath` methods

agurtovoy:
... plus related cleanup.

Prep for fixing a host of Shape v2 issues in the builder.

### DIFF
--- a/packages/controls/src/controls/shape/v2/shape-v2.ts
+++ b/packages/controls/src/controls/shape/v2/shape-v2.ts
@@ -3,10 +3,7 @@ import { z } from 'zod'
 import { mapValues } from '../../../lib/functional'
 import { safeParse, type ParseResult } from '../../../lib/zod'
 
-import {
-  ControlDataTypeKey,
-  type Data,
-} from '../../../common'
+import { ControlDataTypeKey, type Data } from '../../../common'
 import {
   type CopyContext,
   type MergeTranslatableDataContext,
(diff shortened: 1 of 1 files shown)


packages/controls/src/controls/shape/v2/shape-v2.ts
```diff
@@ -170,11 +167,9 @@ class Definition<C extends Config> extends ControlDefinition<
   fromData(data: DataType<C> | undefined): ValueType<C> | undefined {
     if (data == null) return undefined
 
-    const versionedData = Definition.versionData(data)
+    const itemsData = Definition.itemsData(data)
```
agurtovoy, line 170: Here and below, use newly introduced `itemsData` method in code that works with items data.

packages/controls/src/controls/shape/v2/shape-v2.ts
```diff
@@ -214,28 +209,50 @@ class Definition<C extends Config> extends ControlDefinition<
...
+
+    return data
+  }
+
+  static itemDataPath<C extends Config>(
+    data: DataType<C>,
+    key: string,
+  ): string[] {
+    return Definition.isShapeV2Data(data) ? ['value', key] : [key]
+  }
```
agurtovoy, line 244: Needed by the builder, implemented as a part of the `ShapeV2` interface to keep the knowledge about the inner structure of the shape data in a single place.

migueloller (approved):

PR #903, opened 2025-01-06
fix(runtime): require passing `previewMode` to the runtime provider

agurtovoy:
As long as passing `previewMode` is required for making the site editable in the builder, there is no good reason for defaulting it. See [this thread](https://makeswifthq.slack.com/archives/C057U8TLT9U/p1736182157072199?thread_ts=1735925807.739969&cid=C057U8TLT9U) for context.

### DIFF
--- a/packages/runtime/src/next/components/tests/prop-controllers/page-backgrounds-prop-controller.test.tsx
+++ b/packages/runtime/src/next/components/tests/prop-controllers/page-backgrounds-prop-controller.test.tsx
@@ -81,7 +81,7 @@ describe('Page', () => {
 
     await act(async () =>
       render(
-        <ReactRuntimeProvider runtime={runtime}>
+        <ReactRuntimeProvider runtime={runtime} previewMode={false}>
           <Page snapshot={snapshot} />
         </ReactRuntimeProvider>,
       ),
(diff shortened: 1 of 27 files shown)


packages/runtime/src/next/components/tests/prop-controllers/page-backgrounds-prop-controller.test.tsx
```diff
@@ -81,7 +81,7 @@ describe('Page', () => {
 
     await act(async () =>
       render(
-        <ReactRuntimeProvider runtime={runtime}>
+        <ReactRuntimeProvider runtime={runtime} previewMode={false}>
```
agurtovoy, line 84: A ton of identical prop controller test updates here and below, these will go away when we migrate prop controllers to the unified controls interface.

packages/runtime/src/runtimes/react/components/RuntimeProvider.tsx
```diff
@@ -13,13 +13,13 @@ const PreviewProvider = lazy(() => import('./PreviewProvider'))
...
   runtime,
-  previewMode = false,
+  previewMode,
   apiOrigin = 'https://api.makeswift.com',
   locale = undefined,
 }: {
   children: ReactNode
   runtime: ReactRuntime
-  previewMode?: boolean
+  previewMode: boolean
```
agurtovoy, line 22: The actual API change.

PR #904, opened 2025-01-07
Update document payload to support 'detach' UX for editing localized element trees that inherit from a parent locale

Codeseph:
companion PR: https://github.com/makeswift/cosmos/pull/765 

eng-7353 (part of eng-7271)

### DIFF
--- a/.changeset/young-shoes-cheer.md
+++ b/.changeset/young-shoes-cheer.md
@@ -1,0 +1,5 @@
+---
+"@makeswift/runtime": minor
+---
+
+Update `getComponentSnapshot` to perform locale fallback
--- a/packages/runtime/src/next/client.ts
+++ b/packages/runtime/src/next/client.ts
@@ -144,6 +144,7 @@ const makeswiftComponentDocumentSchema = z.object({
   locale: z.string().nullable(),
   data: Schema.element,
   siteId: z.string(),
+  inheritsFromParent: z.boolean(),
 })
 
 export type MakeswiftComponentDocument = z.infer<typeof makeswiftComponentDocumentSchema>
@@ -158,10 +159,16 @@ export type MakeswiftComponentDocumentFallback = z.infer<
   typeof makeswiftComponentDocumentFallbackSchema
 >
 
+export type MakeswiftComponentSnapshotMetadata = {
+  allowLocaleFallback: boolean
+  requestedLocale: string | null
+}
+
 export type MakeswiftComponentSnapshot = {
   document: MakeswiftComponentDocument | MakeswiftComponentDocumentFallback
   key: string
   cacheData: CacheData
+  meta: MakeswiftComponentSnapshotMetadata
 }
 
 export function componentDocumentToRootEmbeddedDocument({
@@ -169,11 +176,13 @@ export function componentDocumentToRootEmbeddedDocument({
   documentKey,
   name,
   type,
+  meta,
 }: {
   document: MakeswiftComponentDocument | MakeswiftComponentDocumentFallback
   documentKey: string
   name: string
   type: string
+  meta: MakeswiftComponentSnapshotMetadata
 }): EmbeddedDocument {
   const { data: rootElement, locale, id } = document
 
@@ -197,6 +206,7 @@ export function componentDocumentToRootEmbeddedDocument({
     id,
     type,
     name,
+    meta,
     __type: EMBEDDED_DOCUMENT_TYPE,
   }
 
@@ -658,20 +668,33 @@ export class Makeswift {
     {
       siteVersion: siteVersionPromise,
       locale,
-    }: { siteVersion: MakeswiftSiteVersion | Promise<MakeswiftSiteVersion>; locale?: string },
+      allowLocaleFallback = true,
+    }: {
+      siteVersion: MakeswiftSiteVersion | Promise<MakeswiftSiteVersion>
+      locale?: string
+      allowLocaleFallback?: boolean
+    },
   ): Promise<MakeswiftComponentSnapshot> {
     const searchParams = new URLSearchParams()
     if (locale) searchParams.set('locale', locale)
 
     const siteVersion = await siteVersionPromise
-    const response = await this.fetch(
+    const key = deterministicUUID({ id, locale, seed: this.apiKey.split('-').at(0) })
+    const baseLocaleWasRequested = locale == null
+    const canAttemptLocaleFallback = !baseLocaleWasRequested && allowLocaleFallback
+
+    let response
+    const responseForRequestedLocale = await this.fetch(
       `v1/element-trees/${id}?${searchParams.toString()}`,
       siteVersion,
     )
 
-    const key = deterministicUUID({ id, locale, seed: this.apiKey.split('-').at(0) })
+    if (responseForRequestedLocale.status === 404 && canAttemptLocaleFallback) {
+      response = await this.fetch(`v1/element-trees/${id}`, siteVersion)
+    } else {
+      response = responseForRequestedLocale
+    }
 
-    // If the element tree is not found, we generate a document with null data
     if (response.status === 404) {
       return {
         document: {
@@ -681,11 +704,14 @@ export class Makeswift {
         },
         key,
         cacheData: CacheData.empty(),
+        meta: {
+          allowLocaleFallback,
+          requestedLocale: locale ?? null,
+        },
       }
     }
 
     const json = await response.json()
-
     if (!response.ok) {
       console.error('Failed to get page snapshot', json)
       throw new Error(`Failed to get page snapshot with error: "${response.statusText}"`)
@@ -694,7 +720,15 @@ export class Makeswift {
     const document = makeswiftComponentDocumentSchema.parse(json)
     const cacheData = await this.introspect(document.data, siteVersion, locale)
 
-    return { document, cacheData, key }
+    return {
+      document,
+      cacheData,
+      key,
+      meta: {
+        allowLocaleFallback,
+        requestedLocale: locale ?? null,
+      },
+    }
   }
 
   async getSwatch(swatchId: string, siteVersion: MakeswiftSiteVersion): Promise<Swatch | null> {
--- a/packages/runtime/src/next/tests/client.get-component-snapshot.test.ts
+++ b/packages/runtime/src/next/tests/client.get-component-snapshot.test.ts
@@ -1,0 +1,206 @@
+import { Makeswift, MakeswiftComponentDocument } from '../client'
+import { http, HttpResponse, graphql } from 'msw'
+
+import { server } from '../../mocks/server'
+import { MakeswiftSiteVersion } from '../preview-mode'
+
+const TEST_API_KEY = 'myApiKey'
+const apiOrigin = 'https://api.fakeswift.com'
+const baseUrl = `${apiOrigin}/v1/element-trees`
+
+describe('getComponentSnapshot using v1 element tree endpoint', () => {
+  test('return null document data on 404', async () => {
+    // Arrange
+    const client = new Makeswift(TEST_API_KEY, { apiOrigin })
+    const treeId = 'myTree'
+    server.use(
+      http.get(`${baseUrl}/${treeId}`, () => HttpResponse.text('', { status: 404 }), {
+        once: true,
+      }),
+    )
+
+    // Act
+    const result = await client.getComponentSnapshot(treeId, {
+      siteVersion: MakeswiftSiteVersion.Working,
+    })
+
+    // Assert
+    expect(result.document).not.toBeNull()
+    expect(result.document.id).toBe(treeId)
+    expect(result.document.data).toBeNull()
+  })
+
+  test('successfully performs locale fallback by requesting base locale tree after receiving a 404 response for a locale variant tree', async () => {
+    // Arrange
+    const client = new Makeswift(TEST_API_KEY, { apiOrigin })
+    const localeToTest = 'fr-FR'
+    const baseLocale = null
+
+    // mock base locale tree document
+    const treeId = 'myTree123'
+    const elementTreeName = 'myElementTree'
+    const elementTreeKey = 'abc123'
+    const document: MakeswiftComponentDocument = {
+      id: treeId,
+      name: elementTreeName,
+      data: {
+        type: 'myType',
+        key: elementTreeKey,
+        props: {},
+      },
+      locale: baseLocale,
+      siteId: 'mySiteId',
+      inheritsFromParent: false,
+    }
+
+    /*
+        Intercept:
+        (1) initial request to v1 endpoint for the locale variant tree
+        (2) subsequent request to v1 endpoint for the base locale tree
+            - this is the step that we're really testing here - we want the logic in getComponentSnapshot to execute this subsequent request
+        (3) graphql query for introspection
+    */
+    server.use(
+      http.get(
+        `${baseUrl}/${treeId}?locale=${localeToTest}`,
+        () => HttpResponse.text('', { status: 404 }),
+        { once: true },
+      ),
+      http.get(`${baseUrl}/${treeId}`, () => HttpResponse.json(document, { status: 200 }), {
+        once: true,
+      }),
+      graphql.operation(() => {
+        return HttpResponse.json({})
+      }),
+    )
+
+    // Act
+    const result = await client.getComponentSnapshot(treeId, {
+      siteVersion: MakeswiftSiteVersion.Working,
+      locale: localeToTest,
+    })
+
+    // Assert
+    expect(result).not.toBeNull()
+    expect(result.document).not.toBeNull()
+    expect(result.document.data).not.toBeNull()
+    expect(result.document.data?.key).toBe(elementTreeKey)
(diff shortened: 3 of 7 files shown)


Codeseph:
Bit of feedback from Slack that might be worth discussing/resolving:

`In a way the unit tests are a bit too low level and maybe focus too much on implementation details instead of the result`

packages/runtime/src/next/client.ts
```diff
@@ -658,20 +663,38 @@ export class Makeswift {
...
+    const getResponse = async () => {
+      const responseForRequestedLocale = await this.fetch(
+        `v1/element-trees/${id}?${searchParams.toString()}`,
+        siteVersion,
+      )
+      if (responseForRequestedLocale.status !== 404 || !canAttemptLocaleFallback) {
+        return responseForRequestedLocale
+      }
+      const fallbackLocaleResponse = await this.fetch(`v1/element-trees/${id}`, siteVersion)
+      localeFallbackOccurred = true
```
arvinpoddar, line 691: I'd consider writing this in such a way that the result of `getResponse` returns both a response as well as a field that tells us if a fallback request was made (maybe something like `getResponse(): { response, madeFallbackRequest }`) . This way we can avoid the function having a side effect on state outside of its closure.

packages/runtime/src/next/client.ts
```diff
@@ -658,20 +663,38 @@ export class Makeswift {
...
-      `v1/element-trees/${id}?${searchParams.toString()}`,
-      siteVersion,
-    )
-
     const key = deterministicUUID({ id, locale, seed: this.apiKey.split('-').at(0) })
+    const baseLocaleWasRequested = locale == null
+    const canAttemptLocaleFallback = !baseLocaleWasRequested && allowLocaleFallback
+    let localeFallbackOccurred = false
+
+    const getResponse = async () => {
```
arvinpoddar, line 682: Maybe rename this to something like `fetchElementTree`? Also declare as a `function` rather than a `const`

packages/runtime/src/next/client.ts
```diff
@@ -658,20 +663,38 @@ export class Makeswift {
...
+      }
+      const fallbackLocaleResponse = await this.fetch(`v1/element-trees/${id}`, siteVersion)
+      localeFallbackOccurred = true
+      return fallbackLocaleResponse
+    }
 
-    // If the element tree is not found, we generate a document with null data
+    const response = await getResponse()
+
+    // If no element tree was found, we generate a document with null data
```
migueloller, line 697: I'm of the opinion that the code should be written in a way that's descriptive such that comments that explain "what" or "how", as opposed to "why", are not necessary.

packages/runtime/src/next/client.ts
```diff
@@ -694,7 +718,14 @@ export class Makeswift {
     const document = makeswiftComponentDocumentSchema.parse(json)
     const cacheData = await this.introspect(document.data, siteVersion, locale)
 
-    return { document, cacheData, key }
+    /*
+      There are two cases when 'usingInheritedData' should be true:
+      - when the requested tree did not exist for the requested locale, so we sent a subsequent request for the base locale's tree
+      - when the requested tree exists for the requested locale, and is marked in the db as inheriting
+    */
```
migueloller, line 725: See https://github.com/makeswift/makeswift/pull/904/files#r1909147338

packages/runtime/src/next/client.ts
migueloller, line 682: It feels a bit weird to me that we're abstracting the HTTP response itself instead of stepping through getting the element trees and if we don't find the first element tree, we get the other one. This function feels like it adds unnecessary complexity IMO

packages/runtime/src/next/client.ts
```diff
@@ -694,7 +718,14 @@ export class Makeswift {
...
+    const usingInheritedData = (localeFallbackOccurred || document.inheritsFromParent)
+
+    return { document: {...document, locale: locale ?? null}, cacheData, key, usingInheritedData }
```
migueloller, line 728: It feels weird that we're having to override the locale like this and modify the document. Is this necessary due to the fallback? If so, shouldn't we still be sending the base locale if we did fall back? I would imagine it's because we use the document's locale to generate an identifier for ShareDB? If so, I think we should decouple the identifier for the document vs the actual data we got and its locale.

For example, I can request a document with `(id: foo, locale: fr)` and get a 404 and thus fall back to the data for document `(id: foo, locale: en)`. The ID should continue to be `(id: foo, locale: fr)`, but the document and the data should still be `(id: foo, locale: en)`

migueloller (approved):
Changes look good but I think that this could use some work in terms of making it easier to follow for others that aren't quite as familiar

packages/runtime/src/next/tests/client.get-component-snapshot.test.ts
```diff
@@ -0,0 +1,237 @@
...
+      locale: localeToTest,
+    })
+
+    // Assert
+    expect(result).not.toBeNull()
+    expect(result.document).not.toBeNull()
+    expect(result.document.data).not.toBeNull()
+    expect(result.document.data?.key).toBe(elementTreeKey)
+    expect(result.document.locale).toBe(localeToTest)
+  }),
```
migueloller, line 122: There should be no `,` here

packages/runtime/src/next/tests/client.get-component-snapshot.test.ts
```diff
@@ -0,0 +1,237 @@
...
+    // Assert
+    expect(result.document).not.toBeNull()
+    expect(result.document.id).toBe(treeId)
+    expect(result.document.data).toBeNull()
+    expect(result.cacheData).not.toBeNull()
+    expect(result.cacheData.apiResources).toStrictEqual({})
+    expect(result.cacheData.localizedResourcesMap).toStrictEqual({})
+  })
+
+  test('throws if response is 200 but element tree result is null', async () => {
```
migueloller, line 37: This seems like a weird test case. It seems to be testing a malformed response, which I guess we do handle with our Zod schemas but generally one expects the API to follow some format. If we're testing for malformed responses, why 200 with null data and not the other infinitely many other ways the response could be malformed?

packages/runtime/src/next/tests/client.get-component-snapshot.test.ts
```diff
@@ -0,0 +1,237 @@
...
+      siteVersion: MakeswiftSiteVersion.Working,
+    })
+
+    // Assert
+    expect(result.document).not.toBeNull()
+    expect(result.document.id).toBe(treeId)
+    expect(result.document.data).toBeNull()
+    expect(result.cacheData).not.toBeNull()
+    expect(result.cacheData.apiResources).toStrictEqual({})
+    expect(result.cacheData.localizedResourcesMap).toStrictEqual({})
```
migueloller, line 34: This feels like it's testing implementation details too much. I think we should write these tests as if we were consumers of `@makeswift/runtime`, based on what we can observe.

packages/runtime/src/next/tests/client.get-component-snapshot.test.ts
```diff
@@ -0,0 +1,237 @@
...
+import { server } from '../../mocks/server'
+import { MakeswiftSiteVersion } from '../preview-mode'
+import { ZodError } from 'zod'
+
+const TEST_API_KEY = 'myApiKey'
+const apiOrigin = 'https://api.fakeswift.com'
+const baseUrl = `${apiOrigin}/v1/element-trees`
+
+describe('getComponentSnapshot using v1 element tree endpoint', () => {
+  test('return fallback document on 404', async () => {
```
migueloller, line 13: The naming of this test is a bit confusing to me as we're not testing that we fall back on 404, but that we return the "empty" 404 snapshot on 404

packages/runtime/src/next/client.ts
arvinpoddar: You're correct that the reason for this locale override is to generate a ShareDB ID. However, this ShareDB ID construction happens in the builder based on the attributes of the document. So, I advised to overwrite the locale here. To construct an ID with information different than the document's data (e.g., `fr` instead of `en`), we'd need to amend what payload gets sent to the builder.

packages/runtime/src/next/tests/client.get-component-snapshot.test.ts
```diff
@@ -0,0 +1,237 @@
...
+        allowLocaleFallback: false,
+      })
+
+      // Assert
+      expect(result).not.toBeNull()
+      expect(result.document).not.toBeNull()
+      expect(result.document.id).toBe(treeId)
+      expect(result.document.data).toBeNull()
+      expect(result.document.locale).toBe(localeToTest)
+    }),
```
migueloller, line 180: Remove `,`

migueloller (changes requested):

packages/runtime/src/next/client.ts
migueloller: I would say that the complexity here isn't worth it. In particular, I can imagine a situation where we're pulling our hair out because we're seeing attempting to debug and when we go look at the data in our API or DB it doesn't match what we're seeing in the runtime/builder. What we're doing here is a bit too "smart" for my taste and my Grug alarm is going off.

If this is a decision that was made well before this PR then perhaps it's not worth belaboring the point right now but I still think we should re-evaluate this at some point as it feels like a ticking time bomb of accidental complexity to me

packages/runtime/src/next/client.ts
Codeseph: (rewritten in such a way that getResponse no longer exists)

packages/runtime/src/next/client.ts
Codeseph: (rewritten in such a way that getResponse no longer exists)

packages/runtime/src/next/client.ts
Codeseph: removed

packages/runtime/src/next/client.ts
Codeseph: removed

packages/runtime/src/next/client.ts
Codeseph: this no longer exists in the latest rewrite

packages/runtime/src/next/client.ts
Codeseph: Updated so that this locale override no longer takes place

Arvin and I discussed quite a few scenarios yesterday while thinking through how we could get rid of this override and also keep the payload as small as possible (only including what we can't derive in the builder). We came away feeling that we could accomplish:
- detach from parent
- delete
- reset to base locale
- (in the future) edit parent 

while adding to the document payload only 2 fields:
- `allowLocaleFallback` (the original param to getComponentSnapshot, representing the user's preference for whether we show data from the inherited locale tree when no localized variant has been created)
- `requestedLocale` (needed for subscribing to the right ShareDB document)

packages/runtime/src/next/tests/client.get-component-snapshot.test.ts
Codeseph: removed, sorry

packages/runtime/src/next/tests/client.get-component-snapshot.test.ts
Codeseph: Removed. In hindsight this seems like a useless test. I have a guess for why I included it to begin with, but it doesn't help my case

packages/runtime/src/next/tests/client.get-component-snapshot.test.ts
Codeseph: Agreed - now that we're dealing with "locale fallback" this is confusing. Renamed.

packages/runtime/src/next/client.ts
```diff
@@ -694,7 +720,15 @@ export class Makeswift {
...
 
-    return { document, cacheData, key }
+    return {
+      document,
+      cacheData,
+      key,
+      meta: {
+        allowLocaleFallback,
+        requestedLocale: locale ?? null,
+      },
```
arvinpoddar, line 730: We don't necessarily have to call this field `meta` in the snapshot, since you could argue that these fields should be main citizens of the snapshot. That, or we could scope this under an object called `requestOptions`, since that's what both of these fields are.

We can leave this as `meta` in the actual document payload that gets sent over to the builder, since there, these fields are actually metadata about the resource that was delivered.

Either way, I wouldn't couple the type `MakeswiftComponentSnapshotMetadata` to what actually gets sent to the builder payload / stored in the host's redux store, since these are semantically different things. I don't think there's a need to have that type defined at all, we could just have it inlined.

Curious what you think @migueloller

packages/runtime/src/state/actions.ts
```diff
@@ -106,6 +106,7 @@ type DocumentPayloadEmbeddedDocument = {
   type: string
   name: string
   rootElement: Element
+  usingInheritedData: boolean
```
arvinpoddar, line 109: Think this needs to be updated to reflect `meta`

packages/runtime/src/state/modules/read-only-documents.ts
```diff
@@ -32,6 +33,7 @@ export type EmbeddedDocument = {
   type: string
   name: string
   rootElement: Element
+  meta: MakeswiftComponentSnapshotMetadata
```
arvinpoddar, line 36: See previous comment about coupling the snapshot type to the payload type. Assuming that holds, I think here we can have this as `meta`, and define the object type inline.

arvinpoddar, line 36: Also, curious how we feel about the name `meta`. @Codeseph and I deliberated over a few options, such as `config` (since these represent config options about the snapshot), etc. Ultimately we just decided to go with `meta` since we figured this would be something discussed here, and the fact that we already had a `meta` field before for when we supported JSX fallbacks. @migueloller any thoughts?

arvinpoddar (commented):
Awesome, thanks for tackling this

.changeset/young-shoes-cheer.md
```diff
@@ -0,0 +1,5 @@
+---
+"@makeswift/runtime": patch
+---
+
+Update getComponentSnapshot to perform locale fallback
```
migueloller, line 5:
```suggestion
Update `getComponentSnapshot` to perform locale fallback
```

.changeset/young-shoes-cheer.md
```diff
@@ -0,0 +1,5 @@
+---
+"@makeswift/runtime": patch
```
migueloller, line 2: Should this be `minor`?

packages/runtime/src/next/client.ts
migueloller: My thoughts on this is that a "snapshot" is already an opaque bag of a bunch of metadata on an element tree, so I don't think there's need for a `meta` field here since it's all basically metadata. We've got `document`, which is the element tree (?), `cacheData` which is the result of introspection and we use it to hydrate the client side cache, `key` which is the identifier of the document (?), and then the new fields.

Then again, there's also a separate way to think about this: (1) what does `<MakeswiftComponent>` need to render and (2) what does the builder need to know about what was provided to `getComponentSnapshot`. In that way, `meta` or perhaps `params` or `config` or some other name does seem reasonable. We're using the snapshot as a convenient delivery mechanism from what was passed to `getComponentSnapshot` to pass it to `<MakeswiftComponent>`.

Given that this is all part of the snapshot and we can change it more frequently than we can change the builder payload, we can move forward with `meta` and change it later if we change our minds.

packages/runtime/src/next/client.ts
```diff
@@ -658,20 +668,33 @@ export class Makeswift {
...
+      allowLocaleFallback?: boolean
+    },
   ): Promise<MakeswiftComponentSnapshot> {
     const searchParams = new URLSearchParams()
     if (locale) searchParams.set('locale', locale)
 
     const siteVersion = await siteVersionPromise
-    const response = await this.fetch(
+    const key = deterministicUUID({ id, locale, seed: this.apiKey.split('-').at(0) })
+    const baseLocaleWasRequested = locale == null
```
migueloller, line 683: Why not initialize this to `false` and then set it to `true` if the base locale is requested?

migueloller: Never mind, I didn't understand what this meant. I thought this meant that we sent a request to get the base locale as a fallback but this seems to be for whether the caller is requesting a base locale or not

packages/runtime/src/state/modules/read-only-documents.ts
migueloller: +1 from me on decoupling the two

migueloller (dismissed):

packages/runtime/src/state/modules/read-only-documents.ts
Codeseph: updated

.changeset/young-shoes-cheer.md
Codeseph: it should - updated

packages/runtime/src/state/modules/read-only-documents.ts
Codeseph: (discussed with Arvin/Miguel, went with keeping 'meta' but with an inline definition for the payload types)

migueloller (approved):

PR #905, opened 2025-01-08
fix: patched fetch in the builder to preserve existing request headers

fikrikarim:
**Steps to reproduce the issue:**
- Open a page in the builder
- Open the console on the dev tools
- Select the host console instead of `top`, 
- Send a fetch request with a custom header. 

**Before the fix, the custom header is overwritten:**
patched-fetch-headers-before

---

**After the fix, the custom header is preserved:**
patched-fetch-headers-after

### DIFF
--- a/.changeset/tall-queens-smile.md
+++ b/.changeset/tall-queens-smile.md
@@ -1,0 +1,5 @@
+---
+"@makeswift/runtime": patch
+---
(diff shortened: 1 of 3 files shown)


migueloller (approved):
LGTM 👍🏻

PR #906, opened 2025-01-09
Detach localized component UX: update runtime payload to include only fields that the builder cannot derive

Codeseph:


packages/runtime/src/next/client.ts
```diff
@@ -159,25 +159,30 @@ export type MakeswiftComponentDocumentFallback = z.infer<
   typeof makeswiftComponentDocumentFallbackSchema
 >
 
+export type MakeswiftComponentSnapshotConfig = {
```
arvinpoddar, line 162: I know I'm the one who suggested the name config, we'll probably have to think about this more. Maybe `requestMeta` or `requestConfig`, something...? Maybe `config` is already descriptive enough.

Codeseph: My preference would probably be referring to this as 'snapshot request config' in conversation, and having this type be 'MakeswiftComponentSnapshotRequestConfig'. Not a strong preference though

Codeseph: update: renamed to 'MakeswiftComponentSnapshotMetadata' for now

PR #907, opened 2025-01-10
feat(runtime): allow forced http protocol

naqi:
This is a workaround for GCP cloud run. This makes sure the proxied url has the correct protocol so our transaction is fulfilled.

### DIFF
--- a/.changeset/honest-swans-bake.md
+++ b/.changeset/honest-swans-bake.md
@@ -1,0 +1,5 @@
+---
+"@makeswift/runtime": patch
+---
+
+feat(runtime): Use headers to create proxied url
--- a/.changeset/honest-swans-bake.md
+++ b/.changeset/honest-swans-bake.md
@@ -1,5 +1,5 @@
 ---
-"@makeswift/runtime": patch
+'@makeswift/runtime': minor
 ---
 
 feat(runtime): Use headers to create proxied url
--- a/packages/runtime/src/next/api-handler/handlers/proxy-draft-mode.test.ts
+++ b/packages/runtime/src/next/api-handler/handlers/proxy-draft-mode.test.ts
@@ -2,92 +2,84 @@ import { NextRequest } from 'next/server'
 import { http, HttpResponse } from 'msw'
 import { server } from '../../../mocks/server'
 import proxyDraftMode from './proxy-draft-mode'
-import { env } from 'process';
+import { env } from 'process'
 
 jest.mock('next/headers', () => ({
   draftMode: jest.fn().mockReturnValue({ enable: jest.fn(), disable: jest.fn() }),
   cookies: jest.fn().mockReturnValue({ get: jest.fn() }),
 }))
 
-describe('ForceHTTP proxyDraftMode URL handling', () => {
-  const mockApiKey = 'test-api-key'
+const mockApiKey = 'test-api-key'
   
-  server.use(
-    http.all('*', () => {
-      return new HttpResponse()
-    })
-  )
+beforeAll(() => {
+  server.listen()
+})
 
+afterAll(() => {
+  console.log('afterAll')
+  server.close()
+})
+
+afterEach(() => {
+  console.log('afterEach')
+  server.resetHandlers()
+  jest.clearAllMocks()
+})
+
+describe('ForceHTTP proxyDraftMode URL handling', () => {
   beforeAll(() => {
-    server.listen()
+    env.FORCE_HTTP = 'true'
   })
 
   afterAll(() => {
-    server.close()
-  })
-
-  afterEach(() => {
-    server.resetHandlers()
-    jest.clearAllMocks()
+    env.FORCE_HTTP = 'false'
   })
 
   it('should force protocol to http and use host from `host` from headers instead of `x-forwarded-host`', async () => {
     const originalUrl = new URL(`https://example.com?x-makeswift-draft-mode=${mockApiKey}&keep=this`)
     
+    let capturedRequest: NextRequest | null = null
+      server.use(
+        http.all('*', ({ request }) => {
+          capturedRequest = new NextRequest(request)
+          return new HttpResponse()
+        })
+      )
+
     const request = new NextRequest(originalUrl)
     request.headers.append("X-Makeswift-Draft-Mode",mockApiKey )
     request.headers.append("x-forwarded-host", "localhost:3000")
 
     const originalProtocol = request.nextUrl.protocol
+    
     await proxyDraftMode(request, { params: {} }, { apiKey: mockApiKey })
 
 
     // Make sure original request is not changed.
     expect(request.nextUrl.protocol).toBe(originalProtocol)
+    expect(capturedRequest).not.toBeNull()
 
-    // Verify fetch was called with modified URL
-    expect(global.fetch).toHaveBeenCalled()
-    const fetchCall = (global.fetch as jest.Mock).mock.calls[0][0] as NextRequest
-    
     // Verify host is same as original host.
-    expect(fetchCall.headers.get('host')).toBe(request.headers.get('host'))
-    
+    expect(capturedRequest!.headers.get('host')).toBe(request.headers.get('host'))
+    expect(capturedRequest!.headers.has('X-Makeswift-Draft-Mode')).toBe(false)
+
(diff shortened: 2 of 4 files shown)


packages/runtime/src/next/api-handler/handlers/proxy-draft-mode.ts
```diff
@@ -59,11 +59,16 @@ async function proxyDraftModeRouteHandler(
   const draft = await draftMode()
   draft.enable()
 
-  const proxyUrl = request.nextUrl.clone()
-  proxyUrl.searchParams.delete('x-makeswift-draft-mode')
+  const forwardingHost = process.env.FORCE_HTTP ? request.headers.get('host') : request.headers.get('x-forwarded-host') ?? request.headers.get('host') 
+  const forwardingProto = process.env.FORCE_HTTP ? 'http:' : request.headers.get('x-forwarded-proto') ?? request.nextUrl.protocol
```
migueloller, line 63: It's a bit confusing to name these `forwardingHost` and `forwardingProto` but conditionally set them not to the forwarded value if `FORCE_HTTP` is set. Also, it seems `FORCE_HTTP` does more than just force the HTTP protocol; it forces proxying via localhost instead of out the internet. Maybe that doesn't require us to change the name of the env var, but I thought it would be worth pointing out.

Here's my suggestion

```ts
  const proxyUrl = request.nextUrl.clone()  
  
  if (process.env.FORCE_HTTP == null) {
    const forwardingHost =request.headers.get('x-forwarded-host') ?? request.headers.get('host') 
    const forwardingProto = request.headers.get('x-forwarded-proto') ?? request.nextUrl.protocol

    proxyUrl.protocol = forwardingProto
    proxyUrl.host = forwardingHost
  }
    
  proxyUrl.searchParams.delete('x-makeswift-draft-mode')

  const proxyHeaders = new Headers(request.headers)
  headers.delete('X-Makeswift-Draft-Mode')

  const proxyRequest = new NextRequest(proxyUrl, { headers: proxyHeaders })
```

Note, that writing it that way makes it more clear what `FORCE_HTTP` actually does: it allows usage of `x-forwarded-*` headers. And so, perhaps that's what we should name it? `USE_FORWARDED_HEADERS`? I don't love that name, but something along those lines.

Note, there seems to be a new standard header named [`Forwarded`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded) for this stuff. We might want to also consider that one in the future.

packages/runtime/src/next/api-handler/handlers/proxy-draft-mode.test.ts
```diff
@@ -0,0 +1,127 @@
...
+        // Verify fetch was called with modified URL
+        expect(global.fetch).toHaveBeenCalled()
+        const fetchCall = (global.fetch as jest.Mock).mock.calls[0][0] as NextRequest
+        expect(fetchCall.headers.has('X-Makeswift-Draft-Mode')).toBe(false)
+        expect(fetchCall.nextUrl.searchParams.has('x-makeswift-draft-mode')).toBe(false)
+        expect(fetchCall.nextUrl.searchParams.get('keep')).toBe('this')
+      })
+      
+  })
+})
```
migueloller, line 127: Missing newline at EOF

.changeset/honest-swans-bake.md
```diff
@@ -0,0 +1,5 @@
+---
+'@makeswift/runtime': patch
```
migueloller, line 2: I consider a big enough behavior change that it might warrant a `minor`

packages/runtime/src/next/api-handler/handlers/proxy-draft-mode.test.ts
```diff
@@ -0,0 +1,127 @@
...
+    const originalProtocol = request.nextUrl.protocol
+    await proxyDraftMode(request, { params: {} }, { apiKey: mockApiKey })
+
+
+    // Make sure original request is not changed.
+    expect(request.nextUrl.protocol).toBe(originalProtocol)
+
+    // Verify fetch was called with modified URL
+    expect(global.fetch).toHaveBeenCalled()
+    const fetchCall = (global.fetch as jest.Mock).mock.calls[0][0] as NextRequest
```
migueloller, line 45: I don't think we should test internal implementation details like this. Could we instead verify things are working by looking at the response? This would still force us to mock fetch such that we can check it was called with the appropriate URL and we can mock a response. We can use MSW for this (see these tests: https://github.com/makeswift/makeswift/pull/904/files#diff-c4eb5d4c00b93ba357d4e1245360a032f3f8d42018b9d05b73caa230c3604b65)

packages/runtime/src/next/api-handler/handlers/proxy-draft-mode.test.ts
```diff
@@ -0,0 +1,127 @@
+import { NextRequest } from 'next/server'
+import proxyDraftMode from './proxy-draft-mode'
+import { env } from 'process';
+
+jest.mock('next/headers', () => ({
+  draftMode: jest.fn().mockReturnValue({ enable: jest.fn(), disable: jest.fn() }),
+  cookies: jest.fn().mockReturnValue({ get: jest.fn() }),
+}))
```
migueloller, line 8: We might want to set these to useful values so that we can test that the appropriate `fetch()` request is being made

packages/runtime/src/next/api-handler/handlers/proxy-draft-mode.test.ts
```diff
@@ -0,0 +1,127 @@
...
+
+describe('ForceHTTP proxyDraftMode URL handling', () => {
+  const mockApiKey = 'test-api-key'
+  let originalFetch: typeof global.fetch
+
+  beforeAll(() => {
+    originalFetch = global.fetch
+    env.FORCE_HTTP = 'true'
+    global.fetch = jest.fn().mockResolvedValue(new Response())
```
migueloller, line 17: I think we should avoid using Jest mocks as much as possible as they cause us to test too many implementation details IMO. I think we won't have a choice mocking certain things like `draftMode()` and `cookies()`, but even those, I would prefer if there were testing utilities provided by Next.js we could use or if our tests actually ran the Next.js dev server and sent requests to it to verify the request and response payloads.

migueloller (commented):
This is looking good. I've got some suggestions about the environment variable, host we use it, and its name. I've also got some suggestions about testing and not testing too many internals.

naqi:
Force HTTP is being used here to make sure we use host from headers vs
x-forwarded-host.

The reason is because I noticed that host on cloud run is always
0.0.0.0:8080 and x-forwarded is FQDN host name.

It feels like a hack and initially I thought of using env called CLOUD_RUN
or K8N but decided to tie it to HTTP to be less specific.

packages/runtime/src/next/api-handler/handlers/proxy-draft-mode.ts
naqi: Force HTTP is being used here to make sure we use host from headers vs
x-forwarded-host.

The reason is because I noticed that host on cloud run is always
0.0.0.0:8080 and x-forwarded is FQDN host name.

It feels like a hack and initially I thought of using env called CLOUD_RUN
or K8N but decided to tie it to HTTP to be less specific.

Update: one other idea was to look for `0.0.0.0` host and force HTTP. I don't like the approach but that has the least amount of code change.

packages/runtime/src/next/api-handler/handlers/proxy-draft-mode.test.ts
naqi: Ah, I should have done MSW from start. I will update tests to use it.

naqi:
Merging this PR in to #901

packages/runtime/src/next/api-handler/handlers/proxy-draft-mode.ts
migueloller: The latter approach seems reasonable. I would like us to have something that works for most proxies, and isn't specific to Cloud Run. Using HTTP when the host is the loopback address seems reasonable but it's still possible it's running in HTTPS. So I wouldn't attempt to automatically switch to HTTP in that instance

Also, of the behavior here is about using forwarded headers or not, I think the env var name should reflect that

PR #908, opened 2025-01-10
fix(actions): Use ubuntu-22.04

naqi:
Since Jan 9th our tests have been failing. This was caused by Github Actions now pointing to Ubuntu 24 for `ubuntu-latest`. This PR switches it back to `ubuntu-22.04` since it will be supported for the next two years

### DIFF
--- a/.github/workflows/preview-deployment-playwright-tests.yml
+++ b/.github/workflows/preview-deployment-playwright-tests.yml
@@ -4,7 +4,7 @@ on:
 jobs:
   test:
     timeout-minutes: 60
-    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.changeset/three-dodos-sneeze.md
migueloller, line 1: Don't need a changeset for non-package changes

migueloller (approved):
Wow. Nice catch! We gotta start enforcing these for merging PRs

.changeset/three-dodos-sneeze.md
naqi: Got it!. I will remove it.

agurtovoy (approved):

PR #911, opened 2025-01-14
refactor: require passing `runtime` to `Makeswift` client/`MakeswiftApiHandler`

agurtovoy:
Removing a potential source of hard-to-trace issues stemming from users forgetting to pass the runtime instance to the classes in subj.


packages/runtime/src/runtimes/react/components/PreviewProvider.tsx
```diff
@@ -7,15 +7,14 @@ import * as ReactBuilderPreview from '../../../state/react-builder-preview'
...
-import { ReactRuntime } from '../react-runtime'
 
 export default function PreviewProvider({ children }: PropsWithChildren): JSX.Element {
   const runtime = useReactRuntime()
   const client = useMakeswiftHostApiClient()
   const store = useMemo(
     () =>
       ReactBuilderPreview.configureStore({
-        preloadedState: runtime ? runtime.store.getState() : ReactRuntime.store.getState(),
+        preloadedState: runtime.store.getState(),
```
agurtovoy, line 17: `runtime` is never null here

PR #915, opened 2025-01-17
fix: prevent corrupted data from stopping rendering

arvinpoddar:
Updates prop resolution to fallback to the default value for a prop (if available) on failure.

https://github.com/user-attachments/assets/279012a5-067d-4fd7-8ea6-4e950280a28e

### DIFF
--- a/packages/runtime/src/runtimes/react/hooks/use-resolved-props.ts
+++ b/packages/runtime/src/runtimes/react/hooks/use-resolved-props.ts
@@ -16,6 +16,28 @@ import { useStylesheetFactory } from './use-stylesheet-factory'
 
 import { resolvableRecord } from '../resolvable-record'
 
+function propErrorHandlingProxy(
+  resolvable: Resolvable<unknown>,
+  def: ControlDefinition,
+  propName: string,
+): Resolvable<unknown> {
+  return {
+    ...resolvable,
+    readStable: () => {
+      try {
+        return resolvable.readStable()
+      } catch (err) {
+        const defaultValue = (def.config as any)?.defaultValue
+        console.error(
+          `Error reading value for prop "${propName}", falling back to \`${defaultValue}\`.`,
+          { control: def, error: err },
+        )
+        return defaultValue
+      }
+    },
+  }
+}
+
 function useControlInstances(elementKey: string): Record<string, ControlInstance> | null {
   const documentKey = useDocumentKey()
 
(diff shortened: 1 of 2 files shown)


agurtovoy (commented):
@arvinpoddar Can we also test this on the scenario that prompted the original fix, namely having a host override the built-in button component [in an incompatible way](https://github.com/bigcommerce/catalyst/blob/integrations/makeswift/core/lib/makeswift/components/button-link/button-link.makeswift.tsx) and attempting to paste one of the templates that have a button: https://makeswifthq.slack.com/archives/C063H6R76QN/p1731963932873159?thread_ts=1731437873.649529&cid=C063H6R76QN ? 🙏

packages/runtime/src/runtimes/react/hooks/use-resolved-props.ts
```diff
@@ -16,6 +16,27 @@ import { useStylesheetFactory } from './use-stylesheet-factory'
...
+): Resolvable<unknown> {
+  return {
+    ...resolvable,
+    readStable: () => {
+      try {
+        return resolvable.readStable()
+      } catch (err) {
+        const defaultValue = (def.config as any)?.defaultValue
+        console.error(
+          `Error reading value for prop "${propName}", using \`${defaultValue}\` instead. Error: ${err}`,
```
agurtovoy, line 32: Sorry I didn't pay attention to this initially, I'd log the error as an object here and also log the control type/definition:

```typescript
console.error(
    `Error reading value for prop "${propName}", falling back to \`${defaultValue}\`.`,
    { control: def, error: err }
)
```

agurtovoy (dismissed):

agurtovoy (approved):

arvinpoddar:
> @arvinpoddar Can we also test this on the scenario that prompted the original fix, namely having a host override the built-in button component [in an incompatible way](https://github.com/bigcommerce/catalyst/blob/integrations/makeswift/core/lib/makeswift/components/button-link/button-link.makeswift.tsx) and attempting to paste one of the templates that have a button: [makeswifthq.slack.com/archives/C063H6R76QN/p1731963932873159?thread_ts=1731437873.649529&cid=C063H6R76QN](https://makeswifthq.slack.com/archives/C063H6R76QN/p1731963932873159?thread_ts=1731437873.649529&cid=C063H6R76QN) ? 🙏

We did this offline, everything worked as expected.

PR #917, opened 2025-01-18
fix: fallback to default during individual prop resolution

arvinpoddar:
Also switches to using `console.warn` for data parsing failures instead of `console.error`.

### DIFF
--- a/packages/runtime/src/runtimes/react/hooks/use-resolved-props.ts
+++ b/packages/runtime/src/runtimes/react/hooks/use-resolved-props.ts
@@ -15,28 +15,7 @@ import { useSelector } from './use-selector'
 import { useStylesheetFactory } from './use-stylesheet-factory'
 
 import { resolvableRecord } from '../resolvable-record'
-
-function propErrorHandlingProxy(
-  resolvable: Resolvable<unknown>,
-  def: ControlDefinition,
-  propName: string,
-): Resolvable<unknown> {
-  return {
-    ...resolvable,
-    readStable: () => {
-      try {
-        return resolvable.readStable()
-      } catch (err) {
-        const defaultValue = (def.config as any)?.defaultValue
-        console.error(
-          `Error reading value for prop "${propName}", falling back to \`${defaultValue}\`.`,
-          { control: def, error: err },
-        )
-        return defaultValue
-      }
-    },
-  }
-}
+import { propErrorHandlingProxy } from '../utils/prop-error-handling-proxy'
 
 function useControlInstances(elementKey: string): Record<string, ControlInstance> | null {
   const documentKey = useDocumentKey()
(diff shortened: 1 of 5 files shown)


packages/controls/src/controls/style-v2/style-v2.ts
```diff
@@ -155,10 +155,18 @@ class Definition<
     }))
 
     const getStyle = (breakpoint: Breakpoint) => {
-      const values = responsiveValues.map(({ deviceId, value }) => ({
-        deviceId,
-        value: value.readStable(),
-      }))
+      const values = responsiveValues.map(({ deviceId, value }) => {
```
agurtovoy, line 158: Since `getStyle` is called from `StableValue.read`, I wonder why the `try`/`catch` in `propErrorHandlingProxy` isn't catching this?

packages/runtime/src/runtimes/react/hooks/use-resolved-props.ts
```diff
@@ -28,7 +28,7 @@ function propErrorHandlingProxy(
         return resolvable.readStable()
       } catch (err) {
         const defaultValue = (def.config as any)?.defaultValue
-        console.error(
+        console.warn(
```
agurtovoy, line 31: Good call.

packages/prop-controllers/src/responsive-color.ts
```diff
@@ -104,7 +104,7 @@ export function getResponsiveColorDataSwatchIds(
   return data.map((override) => override.value.swatchId)
 }
 
-export function getResponsiveColorPropControllerDataSawtchIds(
+export function getResponsiveColorPropControllerDataSwatchIds(
```
agurtovoy, line 107: 🙏

packages/controls/src/controls/style-v2/style-v2.ts
arvinpoddar: Offline discussion: this is being called via `useResolvedProp`, rather than `useResolvedValue`, where the original error handling was implemented. Going to surface the error handling up there instead.

packages/runtime/src/runtimes/react/hooks/use-resolved-value.ts
```diff
@@ -2,13 +2,22 @@ import { useMemo, useEffect, useSyncExternalStore } from 'react'
...
-): T {
+  defaultValue?: T,
+): T | undefined {
   const resourceResolver = useResourceResolver()
-  const resolvable = useMemo(() => resolver(data, resourceResolver), [data, resourceResolver])
+  const resolvable = useMemo(() => {
+    return propErrorHandlingProxy(resolver(data, resourceResolver), defaultValue, error => {
+      console.warn(
+        `Error resolving data \`${JSON.stringify(data)}\`, falling back to \`${defaultValue}\`.`,
+        { error },
```
agurtovoy, line 17: Can we log `data` as an object here? E.g.

```typescript
      console.warn(
        `Error resolving data, falling back to \`${defaultValue}\`.`,
        { data, error },
      )
```

🙏

agurtovoy (approved):

packages/runtime/src/runtimes/react/hooks/use-resolved-value.ts
arvinpoddar: Good call, completely slipped my mind.

agurtovoy (approved):

PR #931, opened 2025-01-29
feat: selectively render head tags in Page

arvinpoddar:
Introduce prop to the Page component that will selectively render metadata in the page `head` with data from Makeswift, or pass booleans to blanket enable/disable this data.

### DIFF
--- a/packages/runtime/src/next/components/page.tsx
+++ b/packages/runtime/src/next/components/page.tsx
@@ -14,9 +14,11 @@ import {
 } from '../client'
 import * as ReactPage from '../../state/react-page'
 import { registerDocumentsEffect } from '../../state/actions'
+import { type PageMetadataSettings } from '../../components/page/page-seo-settings'
 
 export type PageProps = {
   snapshot: MakeswiftPageSnapshot
+  metadata?: boolean | PageMetadataSettings
 }
 
 function useRegisterPageDocument(pageDocument: MakeswiftPageDocument): ReactPage.Document {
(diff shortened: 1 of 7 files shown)


packages/runtime/src/next/components/page.tsx
```diff
@@ -31,7 +33,7 @@ function useRegisterPageDocument(pageDocument: MakeswiftPageDocument): ReactPage
   return rootDocuments[0]
 }
 
-export const Page = memo(({ snapshot, ...props }: PageProps) => {
+export const Page = memo(({ snapshot, makeswiftMetadata = true, ...props }: PageProps) => {
```
agurtovoy, line 36: @arvinpoddar Can we add a test for this? 🙏

arvinpoddar: Just added some

migueloller, line 36: Given this component is for the Makeswift SDK, I feel like the `makeswift` prefix here is redundant. And I know, we do it with `MakeswiftComponent` but that's a top-level import and it wold conflict with React's `Component` if we named it that. This is a prop on a Makeswift component so I think we could do without the prefix and name it `meta`

migueloller (approved):
Looks great!

packages/runtime/src/next/components/page.tsx
agurtovoy: I agree with @migueloller

agurtovoy: Beautiful

arvinpoddar: Makes sense to me, renamed to `metadata`

PR #936, opened 2025-01-30
fix: encode component ID when fetching its element tree

agurtovoy:
Fixes "lost" edits to slots/embedded components with slashes in their IDs (a likely scenario for dynamic pages).

### Before

https://github.com/user-attachments/assets/34247fc1-c36a-42ce-95e4-7bef4dbf42a0

### After

https://github.com/user-attachments/assets/494f9a3c-e252-49cb-8de6-af3c93b8baa3

### DIFF
--- a/packages/runtime/src/next/client.ts
+++ b/packages/runtime/src/next/client.ts
@@ -685,12 +685,12 @@ export class Makeswift {
 
     let response
     const responseForRequestedLocale = await this.fetch(
-      `v1/element-trees/${id}?${searchParams.toString()}`,
+      `v1/element-trees/${encodeURIComponent(id)}?${searchParams.toString()}`,
       siteVersion,
     )
 
     if (responseForRequestedLocale.status === 404 && canAttemptLocaleFallback) {
-      response = await this.fetch(`v1/element-trees/${id}`, siteVersion)
+      response = await this.fetch(`v1/element-trees/${encodeURIComponent(id)}`, siteVersion)
     } else {
       response = responseForRequestedLocale
     }
--- a/packages/runtime/src/next/tests/client.get-component-snapshot.test.ts
+++ b/packages/runtime/src/next/tests/client.get-component-snapshot.test.ts
@@ -37,63 +37,73 @@ describe('getComponentSnapshot using v1 element tree endpoint', () => {
     expect(result.document.data).toBeNull()
   })
 
-  test('successfully performs locale fallback by requesting base locale tree after receiving a 404 response for a locale variant tree', async () => {
-    // Arrange
-    const client = createTestClient()
-    const localeToTest = 'fr-FR'
-    const baseLocale = null
+  test.each([
+    { treeId: 'myTree123', locale: 'fr-FR' },
+    { treeId: 'unsafe:url;chars=@&?❔🤷', locale: 'fr-FR' },
+    { treeId: '/blog/slug', locale: 'en-US' },
+  ])(
+    "successfully performs locale fallback by requesting base locale tree after receiving a 404 response for the element tree '$treeId' with the '$locale' locale",
+    async ({ treeId, locale }) => {
+      // Arrange
+      const client = createTestClient()
+      const localeToTest = locale
+      const baseLocale = null
 
-    // mock base locale tree document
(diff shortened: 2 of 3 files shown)


packages/runtime/src/next/client.ts
```diff
@@ -685,12 +685,12 @@ export class Makeswift {
 
     let response
     const responseForRequestedLocale = await this.fetch(
-      `v1/element-trees/${id}?${searchParams.toString()}`,
+      `v1/element-trees/${encodeURIComponent(id)}?${searchParams.toString()}`,
```
agurtovoy, line 688: If the component ID contained slashes, this `fetch` would 404 and we'd always return an empty snapshot.

As an aside, I think we can benefit from using a type-safe library for constructing URL literals, but I couldn't find one with an ergonomic API that was also well-maintained. Out of the two candidates that I found, [`urlcat`](https://github.com/balazsbotond/urlcat) has a bunch of stale deps and hasn't been updated in 2 years, and [`mouri`](https://github.com/Brecert/mouri) is virtually unknown and hasn't been touched for 3 years.

packages/runtime/src/next/client.ts
```diff
@@ -685,12 +685,12 @@ export class Makeswift {
...
       siteVersion,
     )
 
     if (responseForRequestedLocale.status === 404 && canAttemptLocaleFallback) {
-      response = await this.fetch(`v1/element-trees/${id}`, siteVersion)
+      response = await this.fetch(`v1/element-trees/${encodeURIComponent(id)}`, siteVersion)
```
agurtovoy, line 693: Same for the locale fallback `fetch`.

packages/runtime/src/next/tests/client.get-component-snapshot.test.ts
```diff
@@ -37,63 +37,73 @@ describe('getComponentSnapshot using v1 element tree endpoint', () => {
...
-      siteId: 'mySiteId',
-      inheritsFromParent: false,
-    }
-
-    /*
+  test.each([
+    { treeId: 'myTree123', locale: 'fr-FR' },
+    { treeId: 'unsafe:url;chars=@&?❔🤷', locale: 'fr-FR' },
+    { treeId: '/blog/slug', locale: 'en-US' },
+  ])(
```
agurtovoy, line 44: Expanded tests to probe the API with "problematic" IDs (review with the "Hide whitespace" toggle on to see the actual diffs).

packages/runtime/src/next/tests/client.get-component-snapshot.test.ts
```diff
@@ -37,63 +37,73 @@ describe('getComponentSnapshot using v1 element tree endpoint', () => {
...
-    // Assert
-    expect(result).not.toBeNull()
-    expect(result.document).not.toBeNull()
-    expect(result.document.data).not.toBeNull()
-    expect(result.document.data?.key).toBe(elementTreeKey)
-    expect(result.document.locale).toBeNull()
-  })
+      server.use(
+        http.get(
+          `${baseUrl}/${encodeURIComponent(treeId)}?locale=${localeToTest}`,
```
agurtovoy, line 77: Funny enough, we have to apply the same fix to the test setup.

packages/runtime/src/next/tests/client.get-component-snapshot.test.ts
agurtovoy: passing tests

Codeseph:
Thanks Aleksey - I'm sorry we missed this

arvinpoddar (commented):
Looking good, great catch

packages/runtime/src/next/client.ts
migueloller: Oof. Nice find. I remember being a bit concerned about constructing URLs like this and the possibility of forgetting to `encodeURIComponent`.

I think that is a great idea. An API similar to the motivating example of [Type-Level TypeScript](https://type-level-typescript.com/#chapters) would be great.

```typescript
url('/v1/element-trees/:id', { params: { id }, searchParams })

// or

url('/v1/element-trees/:id?locale=:locale', { id, locale })

// or variants thereof
```

PR #938, opened 2025-02-03
fix: don't pass `locale` prop to `NextLink` when running under App Router

agurtovoy:
Passing `locale={false}` to `NextLink` in the App Router context is silently ignored in Next 13/14, but [generates a console error in Next 15](https://github.com/vercel/next.js/discussions/74703).

### DIFF
--- a/packages/runtime/src/components/shared/Link/index.tsx
+++ b/packages/runtime/src/components/shared/Link/index.tsx
@@ -6,6 +6,8 @@ import NextLink from 'next/link'
 import { LinkData } from '@makeswift/prop-controllers'
 
 import { Link as LinkDef } from '../../../controls/link'
+
+import { useIsPagesRouter } from '../../../next/hooks/use-is-pages-router'
 import { useResolvedValue } from '../../../runtimes/react/hooks/use-resolved-value'
 
 type BaseProps = {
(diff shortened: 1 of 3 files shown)


packages/runtime/src/components/shared/Link/index.tsx
```diff
@@ -70,7 +74,7 @@ export const Link = forwardRef<HTMLAnchorElement, Props>(function Link(
         target={target}
         onClick={handleClick}
         href={href}
-        locale={false}
+        {...(isPagesRouter ? { locale: false } : {})}
```
agurtovoy, line 77: Pass `locale` only when rendering under the Pages Router.

packages/runtime/src/next/hooks/use-is-pages-router.ts
```diff
@@ -1,10 +1,15 @@
...
 export function useIsPagesRouter() {
-  try {
-    useRouter()
-    return true
-  } catch (error) {
-    return false
-  }
+  // switch to `next/compat/router` once we drop support for Next.js 14:
+  // https://nextjs.org/docs/pages/api-reference/functions/use-router#the-nextcompatrouter-export
+  return useMemo(() => {
```
agurtovoy, line 7: Memoize the result so that we are not throwing exceptions on every rerender.

fikrikarim (approved):
Thank you 🙏

arvinpoddar (approved):

packages/runtime/src/next/hooks/use-is-pages-router.ts
agurtovoy: This effectively calls the `useRouter` hook conditionally, which is a no-no.

PR #941, opened 2025-02-04
fix: missing `Carousel` styles

agurtovoy:
Fix missing padding, arrow hover styles on `Carousel` when running under the App Router. Note that this has been broken since we introduced support for the App Router. See inline comments below for details.

## Before

before

before hover

## After

after

after hover

### DIFF
--- a/packages/runtime/src/components/builtin/Carousel/Carousel.tsx
+++ b/packages/runtime/src/components/builtin/Carousel/Carousel.tsx
@@ -78,6 +78,10 @@ type Props = {
 const SWIPE_THRESHOLD = 20
 const swipePower = (dx: number, velocity: number) => dx * (1 + velocity)
 
+// constructs a CSS [class selector](https://developer.mozilla.org/en-US/docs/Web/CSS/Class_selectors),
+// returns a compound class selector if the `classname` string includes multiple class names
+const classSelector = (classname: string) => `.${classname.split(' ').join('.')}`
+
 const Carousel = forwardRef(function Carousel(
   {
     images = [],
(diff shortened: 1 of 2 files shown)


packages/runtime/src/components/builtin/Carousel/Carousel.tsx
```diff
@@ -174,17 +178,19 @@ const Carousel = forwardRef(function Carousel(
       useResponsiveStyle([slideAlignment] as const, ([alignItems = 'center']) => ({ alignItems })),
     ),
   )
+
   const reelClassName = cx(
     useStyle({ display: 'flex', position: 'relative', flexWrap: 'nowrap' }),
     useStyle(
       useResponsiveStyle([gap] as const, ([gap = { value: 0, unit: 'px' }]) => ({
         margin: `0 ${`${-gap.value / 2}${gap.unit}`}`,
-        [`& > .${slideClassName}`]: {
```
agurtovoy, line 182: The old code assumed that the `slideClassName` string, which is defined using Emotion's `cx` utility [in line 164 above](https://github.com/makeswift/makeswift/pull/941/files#diff-c12ce42256d2481cc5fb39511d1b141a02982f1eec7e67618114c25d474f6299L164), will contain a single class name, which is [not guaranteed](https://emotion.sh/docs/@emotion/css#cx). In the "worst case" scenario, `cx` will simply concatenate the classes passed to it as arguments, and the `& > .${slideClassName}` expression will result in an ill-formed selector that will be ignored by Emotion and never inserted into the document.

However, in this specific case, `cx` does happen to return a single classname combining the styles of all the passed classes, **but only when running under the Pages Router** (the way Emotion styles are registered and inserted into the document differs between the router setups).

It returns multiple classnames when running under the App Router, which is the root case of the bug we're fixing here.

packages/runtime/src/components/builtin/Carousel/Carousel.tsx
```diff
@@ -174,17 +178,19 @@ const Carousel = forwardRef(function Carousel(
...
+        [`& > ${classSelector(slideClassName)}`]: {
```
agurtovoy, line 187: Use the [newly minted `classSelector` utility](https://github.com/makeswift/makeswift/pull/941/files#diff-c12ce42256d2481cc5fb39511d1b141a02982f1eec7e67618114c25d474f6299R83) to construct a valid slide selector.

packages/runtime/src/components/builtin/Carousel/Carousel.tsx
```diff
@@ -252,13 +260,14 @@ const Carousel = forwardRef(function Carousel(
     useStyle({
       left: 0,
 
-      [`&:hover > .${arrowClassName}`]: {
+      [`&:hover > ${classSelector(arrowClassName)}`]: {
```
agurtovoy, line 263: Same for the arrows.

arvinpoddar (approved):
Is this our only builtin that attempts to construct a selector from `cx` output?

agurtovoy:
> Is this our only builtin that attempts to construct a selector from cx output?

@arvinpoddar Great question, forgot to comment on it! I did a regex search through the codebase and this is the only builtin that uses this pattern.

PR #950, opened 2025-02-14
fix: cache key mismatch when default locale is passed to getPageSnapshot

fikrikarim:
When we read from the API resource reducer, we're using the locale from `useDocumentLocale`, which is coming from `snapshot.document.locale`. We have a legacy behavior in our API server where if the `locale === defaultLocale`, the server will return the locale as `null`.

When we write to the API resource reducer, we're using the `cacheData` from the introspect function, which gets the locale from `getPageSnapshot` argument directly, without checking if it's the defaultLocale or not.

When we pass the default locale to `getPageSnapshot`:
```
const snapshot = getPageSnapshot({ locale: 'en-US' })
```
The snapshot:
```
{ document: { locale: null } }
```
The cacheData:
```
{ "id": "...", "value": { ... }, "locale": "en-US" }
```

This commit fixes the issue by using the locale returned from the snapshot, instead of from the `getPageSnapshot` argument, to make sure there is no locale mismatch on the document locale and the `cacheData`.

Before:

https://github.com/user-attachments/assets/651034bb-8d67-4d77-b2fa-a192351297cd

After:

https://github.com/user-attachments/assets/1db74c71-7390-4524-9174-a3de656bb62f

### DIFF
--- a/packages/runtime/src/next/client.ts
+++ b/packages/runtime/src/next/client.ts
@@ -474,7 +474,7 @@ export class Makeswift {
   private async introspect(
     element: Element,
     siteVersion: MakeswiftSiteVersion,
-    locale?: string,
+    locale: string | null,
   ): Promise<CacheData> {
     const runtime = this.runtime
     const descriptors = getPropControllerDescriptors(runtime.store.getState())
@@ -654,7 +654,10 @@ export class Makeswift {
     const cacheData = await this.introspect(
       baseLocalizedPage?.data ?? document.data,
       siteVersion,
-      localeInput,
+      // The /v3/pages endpoint returns null for document.locale when the requested locale is the default.
+      // This legacy behavior is set to change with the upcoming /v4/pages endpoint.
+      // We rely on document.locale when reading from the API cache, so ensure the cache is built during introspection using the same value.
+      document.locale,
     )
 
     return {
@@ -718,7 +721,7 @@ export class Makeswift {
     }
 
     const document = makeswiftComponentDocumentSchema.parse(json)
-    const cacheData = await this.introspect(document.data, siteVersion, locale)
+    const cacheData = await this.introspect(document.data, siteVersion, locale ?? null)
 
     return {
       document,
@@ -814,7 +817,7 @@ export class Makeswift {
   async getPagePathnameSlices(
     pageIds: string[],
     siteVersion: MakeswiftSiteVersion,
-    { locale }: { locale?: string },
+    { locale }: { locale?: string | null },
   ): Promise<(PagePathnameSlice | null)[]> {
     if (pageIds.length === 0) return []
 
--- a/.changeset/new-sheep-matter.md
+++ b/.changeset/new-sheep-matter.md
@@ -1,0 +1,5 @@
+---
+'@makeswift/runtime': patch
+---
+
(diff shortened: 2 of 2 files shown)


packages/runtime/src/next/client.ts
```diff
@@ -814,7 +814,7 @@ export class Makeswift {
   async getPagePathnameSlices(
     pageIds: string[],
     siteVersion: MakeswiftSiteVersion,
-    { locale }: { locale?: string },
+    { locale }: { locale?: string | null },
```
arvinpoddar, line 817: Does `locale` still need to be optional here now that you've explicitly said that it's a `string | null`?

fikrikarim: Unfortunately, this is a "public" API since it's available on the Makeswift class. For internal methods, I'd prefer to have locale required to avoid mistakes, while for external methods, I think it's better to be optional so users don't need to pass locale if they don't need it.

migueloller: Hmm, it makes me wonder if we should come up with a plan to reduce the surface area here. I don't think this API is used by anybody. Something for us to consider whenever we decide to work on refactoring our API client

migueloller (commented):
Changes look good to me. The only thing that comes to mind is whether having `null` as a cache key for locale would result in problems? Do we need the document response to include the default locale in the response instead of `null`?

packages/runtime/src/next/client.ts
fikrikarim: Agree 💯

fikrikarim:
> Changes look good to me. The only thing that comes to mind is whether having null as a cache key for locale would result in problems? Do we need the document response to include the default locale in the response instead of null?

I'd love to have that. But I think it's better to do it from the API, and if we want to do that, we need to make a new endpoint.

Thoughts on the PR @agurtovoy?

packages/runtime/src/next/client.ts
```diff
@@ -654,7 +654,7 @@ export class Makeswift {
     const cacheData = await this.introspect(
       baseLocalizedPage?.data ?? document.data,
       siteVersion,
-      localeInput,
+      document.locale,
```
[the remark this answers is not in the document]
agurtovoy: Can we add a comment here along the lines of

 We use `document.locale` when reading from the API cache, make sure we build the cache using the exact same value (as opposite to `localeInput`, which might not be equal to the `document.locale` as the server will return `null` for `document.locale` if the requested locale === the default locale).

?

agurtovoy (dismissed):
Thank you for correcting my blunder here 🙏🙏

agurtovoy:
@fikrikarim If you have an appetite for a new endpoint that doesn't switch out the locale from under us, I'd also prefer that.

packages/runtime/src/next/client.ts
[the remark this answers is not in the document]
fikrikarim: Sounds good. Added the comment 👍

fikrikarim:
> @fikrikarim If you have an appetite for a new endpoint that doesn't switch out the locale from under us, I'd also prefer that.

I'd love to, but I don't have an appetite for that right now. This issue is blocking the BC web team, so I want to unblock them first.

PR #952, opened 2025-02-17
feat(runtime): makeswift draft middleware

arvinpoddar:
Introduces a Makeswift middleware and middleware decorator, which serve as an replacement to the draft/preview mode proxy.


packages/runtime/src/next/api-handler/handlers/bypass-token.ts
```diff
@@ -0,0 +1,106 @@
...
+export type BypassTokenResponse = BypassTokenError | Response
+
+type BypassTokenHandlerArgs =
+  | [request: NextRequest, context: Context, params: { apiKey: string }]
+  | [req: NextApiRequest, res: NextApiResponse<BypassTokenResponse>, params: { apiKey: string }]
+
+const routeHandlerPattern = [P.instanceOf(Request), P.any, P.any] as const
+const apiRoutePattern = [P.any, P.any, P.any] as const
+
+export default async function bypassTokenHandler(
```
arvinpoddar, line 27: This is a single API endpoint that returns the necessary cookies to bypass draft/preview mode and view working site data. The draft middleware consumes this endpoint to patch the request with the necessary cookies.

packages/runtime/src/next/middleware/index.ts
```diff
@@ -0,0 +1,15 @@
...
+export function withMakeswiftMiddleware(
+  config: { apiKey: string },
+  middleware: NextMiddleware,
+): NextMiddleware {
+  return withDraftMiddleware(config, middleware)
+}
+
+export function makeswiftMiddleware(config: { apiKey: string }): NextMiddleware {
+  return withMakeswiftMiddleware(config, req => {
+    return NextResponse.next({ headers: req.headers })
```
arvinpoddar, line 13: Any subsequent middlewares need to pass on the request headers, otherwise draft data is not shown.

apps/nextjs-pages-router/src/middleware.ts
```diff
@@ -0,0 +1,18 @@
+import { makeswiftMiddleware } from '@makeswift/runtime/next/middleware'
+
+export default makeswiftMiddleware({
+  apiKey: process.env.MAKESWIFT_SITE_API_KEY!,
+})
```
arvinpoddar, line 5: For a vanilla host, this is all they would need to add for their `middleware.ts`

.changeset/kind-dryers-beam.md
migueloller, line 1: This change will require documentation updates and we probably want a pretty strong upgrade guide and maybe even a blog post.

cc: @jamesqquick

apps/nextjs-app-router/src/middleware.ts
```diff
@@ -1,23 +1,36 @@
...
-  return NextResponse.redirect(request.nextUrl)
+  return NextResponse.redirect(request.nextUrl, { headers: request.headers })
 }
 
+export const middleware = withMakeswiftMiddleware(
+  { apiKey: process.env.MAKESWIFT_SITE_API_KEY! },
+  // @ts-ignore Type mismatch on NextRequest due to conflicting React versions
+  // in this app vs. runtime
+  localizationMiddleware,
+)
```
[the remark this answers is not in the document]
migueloller: I don't love that for this composition to work we had to make sure we forwarded `headers` when calling `NextResponse.next` or `NextResponse.redirect`. This feels like something our users would probably not know to do

apps/nextjs-pages-router/src/middleware.ts
```diff
@@ -0,0 +1,18 @@
+import { makeswiftMiddleware } from '@makeswift/runtime/next/middleware'
+
+export default makeswiftMiddleware({
+  apiKey: process.env.MAKESWIFT_SITE_API_KEY!,
```
migueloller, line 4: Probably unrelated to this PR, but I don't love that we're using a non-null assertion for environment variables

packages/makeswift-next-plugin/index.js
```diff
@@ -48,58 +44,9 @@ module.exports =
       },
       async rewrites() {
```
migueloller, line 45: We should be able to remove all of `rewrites` here, no?

packages/runtime/package.json
```diff
@@ -58,6 +58,11 @@
       "import": "./dist/esm/next/index.js",
       "types": "./dist/types/next/index.d.ts"
     },
+    "./next/middleware": {
+      "require": "./dist/cjs/next/middleware/index.js",
```
migueloller, line 62: I wonder if we should drop support for CJS and only support ESM for middleware 🤔

packages/runtime/src/next/api-handler/handlers/bypass-token.ts
```diff
@@ -0,0 +1,106 @@
...
+  _context: Context,
+  { apiKey }: { apiKey: string },
+): Promise<NextResponse<BypassTokenResponse>> {
+  const secret =
+    request.nextUrl.searchParams.get(SECRET_HEADER) ?? request.headers.get(SECRET_HEADER)
+
+  if (secret !== apiKey) return new NextResponse('Unauthorized', { status: 401 })
+
+  const draft = await draftMode()
+  const cookie = await cookies()
```
migueloller, line 57: nit: rename to `cookieStore`

packages/runtime/src/next/api-handler/handlers/bypass-token.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  draft.enable()
+  const prerenderBypassToken = cookie.get(PRERENDER_BYPASS_COOKIE)
+
+  if (prerenderBypassToken?.value == null)
+    return new NextResponse('Failed to retrieve token', { status: 500 })
```
migueloller, line 63: nit: add block to multiline `if`

packages/runtime/src/next/api-handler/handlers/bypass-token.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  return NextResponse.json({
+    cookies: [
+      { name: PRERENDER_BYPASS_COOKIE, value: prerenderBypassToken.value },
+      {
+        name: DRAFT_DATA_COOKIE,
+        value: JSON.stringify({ makeswift: true, siteVersion: 'Working' }),
+      },
```
migueloller, line 71: It seems weird to couple Makeswift data with the endpoint to get the bypass token

migueloller: I'm guessing this is because we're using Preview Mode for Pages Router and are trying to have the same API in both but perhaps we want to have separate routes for this stuff

packages/runtime/src/next/middleware/draft-middleware.ts
```diff
@@ -0,0 +1,89 @@
...
+
+const HeaderNames = {
+  DraftMode: 'X-Makeswift-Draft-Mode',
+  PreviewMode: 'X-Makeswift-Preview-Mode',
+}
+
+const SearchParams = {
+  DraftMode: 'x-makeswift-draft-mode',
+  PreviewMode: 'x-makeswift-preview-mode',
+}
```
migueloller, line 19: Should we use `as const` here since these are enum-like?

packages/runtime/src/next/middleware/draft-middleware.ts
```diff
@@ -0,0 +1,89 @@
...
+  { apiKey }: { apiKey: string },
+  middleware: NextMiddleware,
+): NextMiddleware {
+  return async (request, event) => {
+    const secret = getSecretFromRequest(request)
+
+    if (secret == null) {
+      return await middleware(request, event)
+    } else if (secret != null && secret !== apiKey) {
+      console.error('Incorrect Makeswift middleware secret, ignoring')
```
migueloller, line 71: Do we want to ignore in this case as opposed to failing with a 4XX error?

Generally, I think making errors be handle-able or not swallowing them makes it easier to avoid confusing from end-users

packages/runtime/src/next/middleware/draft-middleware.ts
```diff
@@ -0,0 +1,89 @@
...
+      return await middleware(request, event)
+    }
+
+    const tokenUrl = new URL('/api/makeswift/bypass-token', request.nextUrl.origin)
+    tokenUrl.searchParams.set('x-makeswift-api-key', encodeURIComponent(secret))
+    const response = await fetch(tokenUrl, { headers: { 'x-makeswift-api-key': secret } })
+    const parsedResponse = bypassTokenResponseSchema.safeParse(await response.json())
+
+    if (!parsedResponse.success) {
+      console.error('Could not parse bypass token response')
```
migueloller, line 81: Same here, wouldn't we want to 500?

packages/runtime/src/next/middleware/draft-middleware.ts
```diff
@@ -0,0 +1,89 @@
...
+  proxiedRequest.headers.delete(HeaderNames.PreviewMode)
+
+  bypassResponse.cookies.forEach(({ name, value }) => {
+    proxiedRequest.cookies.set(name, value)
+  })
+
+  return proxiedRequest
+}
+
+export function withDraftMiddleware(
```
migueloller, line 61: A lot of the logic of this middleware can be reduced to:
- Create a proxied response
- If we have a proxied response, terminate the middleware and use that
- Otherwise, continue

I wonder if in addition to the middleware we can provide the utility to create the proxied response and if that's there then a user can return it early. This seems less difficult to explain to users than a middleware decorator that requires forwarding headers. What do you think?

packages/runtime/src/next/middleware/index.ts
migueloller: I don't love that requirement 😞

packages/runtime/src/next/middleware/draft-middleware.ts
arvinpoddar: For this and the error below, I'm fine with throwing errors. We've been burned by error throwing in the past, but those incidences are mostly related to rendering, not something as high level as a missing/incorrect secret.

packages/runtime/src/next/middleware/draft-middleware.ts
arvinpoddar: Resolving this to continue discussion in other error thread.

packages/runtime/src/next/api-handler/handlers/bypass-token.ts
arvinpoddar: Yes, it does ultimately boil down to preview mode specifically. `previewData` is set via a JWT, so we have to rely on the `setPreviewData` API rather than attempting to tack on the cookie ourselves. 

Here's the original path I went down, which I ended up refining into what you see now because I thought it was too complex:

1. Have two separate routes for `/draft-mode` and `/preview-mode`. Since we must use the response `setPreviewData` method, this `/preview-mode` route must set `previewData` and the prerender bypass token, so our Makeswift data will be coupled. For consistency's sake then, we might as well set the custom Makeswift data on a cookie in the `/draft-mode` route as well.
2. In our middleware, conditionally call one of these routes. Based on the route we call, parse the respective cookies that we're looking for. Personally felt like this was too much peeking into the internals of the route.

I felt that having the singular route fixed a lot of this complexity and duplicate packing/unpacking logic. Rather, we have one route as a source of truth that will give us everything we need to enter draft mode and view working data.

packages/runtime/src/next/middleware/draft-middleware.ts
arvinpoddar: Wouldn't terminating with the proxied respones not compose well? That's why I hand off the proxy response to down-the-line middlewares instead. Otherwise, I could have just returned a response there directly.

Also, I personally don't think the middleware decorator itself is that hard to explain - in most cases, you're just writing your own middleware within the decorator, and you don't need to worry about what the decorator is doing. It's also not an uncommon pattern,(ex: that's what NextAuth offers). With the right docs and examples, it shouldn't be too confusing. In comparison, offering primitives to construct proxy responses to me feels like it's exposing too much of our internals, and things will start to get hairier with our future plans to attach cookie warnings, etc.

Is the source of the qualm here with the fact that other middlewares have to forward headers? If we could do this without forwarding headers, would you be less opposed to the decorator?

packages/runtime/src/next/api-handler/handlers/bypass-token.ts
migueloller: I don't love that we're coupling the behavior of the route even more into Next.js due to Pages Router. I feel like we should provide the necessary primitives to be able to bypass the static cache and also provide the relevant Makeswift headers. For Pages Router, it happens that the Makeswift headers has to be done with a special cookie, but for App Router that's not necessary and we can use a standard Makeswift header that is independent from the static bypass.

Perhaps something along these lines:

In App Router (the request data can be read in the component):
```ts
export default function middleware(request) {
  const draftModeHeaders = await getDraftModeHeaders(request)

  if (draftModeHeaders == null) return NextResponse.next()

  return NextResponse.next({ headers: draftModeHeaders })
}
```

In Pages Router (the request data has to be encoded in the :
```ts
  const previewModeHeaders = await getPreviewModeHeaders(request)

  if (previewModeHeaders == null) return NextResponse.next()

  return NextResponse.next({ headers: previewModeHeaders })
```

Then again, it begs the question: why have different utilities if the way they are called is the same? Perhaps the explicitness is nice. It makes me wonder if the issue here is how we're naming the endpoint (`/bypass-token`)?

packages/runtime/src/next/api-handler/handlers/bypass-token.ts
```diff
@@ -0,0 +1,107 @@
...
+    .exhaustive()
+}
+
+async function bypassTokenRouteHandler(
+  request: NextRequest,
+  _context: Context,
+  { apiKey }: { apiKey: string },
+): Promise<NextResponse<BypassTokenResponse>> {
+  const secret =
+    request.nextUrl.searchParams.get(SECRET_HEADER) ?? request.headers.get(SECRET_HEADER)
```
migueloller, line 52: I think that we should have a separate constant for the search param and the header. It's a bit weird we're using the `SECRET_HEADER` constant for reading a URL search param

packages/runtime/src/next/api-handler/handlers/bypass-token.ts
```diff
@@ -0,0 +1,107 @@
...
+  if (secret !== apiKey) return new NextResponse('Unauthorized', { status: 401 })
+
+  const draft = await draftMode()
+  const cookieStore = await cookies()
+
+  draft.enable()
+  const prerenderBypassToken = cookieStore.get(PRERENDER_BYPASS_COOKIE)
+
+  if (prerenderBypassToken?.value == null) {
+    return new NextResponse('Failed to retrieve token', { status: 500 })
```
migueloller, line 63: Can we make this error message a bit more information-dense. It would be great if it was obvious where this was coming from when this was logged

migueloller: Perhaps we even want to have error codes so that when our middleware calls this endpoint, it can handle the errors and show detailed logs as to what went wrong

packages/runtime/src/next/api-handler/handlers/bypass-token.ts
migueloller: I think the name is throwing me off here since it does more than return the bypass token

packages/runtime/src/next/api-handler/handlers/bypass-token.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  if (secret !== apiKey) return res.status(401).send('Unauthorized')
+
+  const setCookie = res
+    .setPreviewData({ makeswift: true, siteVersion: 'Working' })
+    .getHeader('set-cookie')
+
+  res.removeHeader('set-cookie')
+
+  const parsedCookies = parseSetCookie(Array.isArray(setCookie) ? setCookie : '')
```
migueloller, line 92: Do we need to handle parsing errors here?

packages/runtime/src/next/api-handler/handlers/bypass-token.ts
```diff
@@ -0,0 +1,107 @@
...
+  draft.enable()
+  const prerenderBypassToken = cookieStore.get(PRERENDER_BYPASS_COOKIE)
+
+  if (prerenderBypassToken?.value == null) {
+    return new NextResponse('Failed to retrieve token', { status: 500 })
+  }
+
+  return NextResponse.json({
+    cookies: [
+      { name: PRERENDER_BYPASS_COOKIE, value: prerenderBypassToken.value },
```
migueloller, line 68: Should we use `prerenderBypassToken.name` here?

packages/runtime/src/next/api-handler/handlers/bypass-token.ts
```diff
@@ -0,0 +1,107 @@
...
+  const previewDataCookie = parsedCookies.find(c => c.name === PREVIEW_DATA_COOKIE)
+
+  if (prerenderBypassCookie?.value == null || previewDataCookie?.value == null) {
+    return res.status(500).send('Failed to retrieve token')
+  }
+
+  return res.json({
+    cookies: [
+      { name: PRERENDER_BYPASS_COOKIE, value: prerenderBypassCookie.value },
+      { name: PREVIEW_DATA_COOKIE, value: previewDataCookie.value },
```
migueloller, line 104: Should we use the cookie `.name` fields here? Seems more semantically appropriate

packages/runtime/src/next/middleware/draft-middleware.ts
migueloller: Sorry, I mistakenly thought we could terminate because I was still thinking about proxying the request through a proxy endpoint, which would be terminating.

> Is the source of the qualm here with the fact that other middlewares have to forward headers? If we could do this without forwarding headers, would you be less opposed to the decorator?

Yes. If we at least provided a helper to get the request, then it is very clear that unless you use the request you're not getting any new functionality. Something like this:

```ts
export default function middleware(request) {
  const makeswiftMiddlewareRequest = getMakeswiftMiddlewareRequest(request)

  return NextResponse.next({ headers: makeswiftMiddlewareRequest.headers })
}
```

You could argue, though, that this prevents us from making our middleware terminating in the future if we needed to. If there's other Next.js libraries out there that follow the decorator pattern and work around the pitfall of having to forward headers with documentation, then I might be convinced otherwise.

.changeset/kind-dryers-beam.md
```diff
@@ -0,0 +1,65 @@
+---
+'@makeswift/next-plugin': minor
+'@makeswift/runtime': minor
+---
+
+refactor: replaces the runtime draft/preview mode proxy endpoint with
```
migueloller, line 6: I wouldn't consider this a "refactor" since we didn't just move internals around without impacting public APIs—we're changing behavior by moving from rewrites to middleware

.changeset/kind-dryers-beam.md
```diff
@@ -0,0 +1,65 @@
...
+
+To upgrade, you'll need to add the new Makeswift middleware to your project. If
+you don't already have middleware for your site, create a `middleware.ts` file
+at the same level as where your `app/` or `pages/` directory is located.
+
+Then, use the new `withMakeswiftMiddleware` function to create a middleware and
+export the result. You will need to pass your API key:
+
+```ts
+import { MAKESWIFT_SITE_API_KEY } from '@/makeswift/env'
```
migueloller, line 20: Could we point to a place in the docs that shows where this value comes from? Otherwise readers might think this is some sort of special way to get the Makeswift API key

packages/runtime/src/next/middleware/branding-middleware.ts
migueloller, line 1: "branding middleware" is a confusing name. I originally thought this was about the branded badge we automatically add for free users on the default renderer

packages/runtime/src/next/middleware/index.ts
```diff
@@ -0,0 +1,36 @@
...
+  const middlewares = otherMiddlewareWrappers.reduce(
+    (accumulatedMiddlewares, nextMiddleware) => middleware =>
+      accumulatedMiddlewares(nextMiddleware(middleware)),
+    firstMiddlewareDecorator,
+  )
+
+  return middlewares(baseMiddleware)
+}
+
+const passThroughMiddleware: NextMiddleware = (_req, _event) => NextResponse.next()
```
migueloller, line 21: Doesn't this have to pass `request` to `NextResponse.next` for it to be a real passthrough?

migueloller (commented):
Looking great! The one thing I feel like this is missing is tests for the new API surface area

arvinpoddar:
Closing this PR after learning that the approach of patching request cookies in middleware to include the prerender bypass token does not work on statically generated pages when deployed on Vercel. We were able to get around this problem by converting our middleware to a terminating middleware that constructs a proxy request, fetches the request, then returns back the response - effectively doing what our existing proxy endpoint does. Although equivalent in function, this change would now be adding middleware to our integration surface area, adding yet another integration point. Since we don't want to rely on proxying for draft mode long term (and have plans for decoupling site version retrieval vs. cache bypassing), for now the plan is to just ship some unstable proxying util functions to fix hosts relying on request patching middleware.

PR #953, opened 2025-02-19
fix(runtime): correctly handle reset element ops

arvinpoddar:
Do not show an error message for reset ops with an empty path. Also correctly handle element deletion on resets.

### DIFF
--- a/packages/runtime/src/state/modules/element-trees.ts
+++ b/packages/runtime/src/state/modules/element-trees.ts
@@ -31,8 +31,8 @@ export function getInitialState(
   return state
 }
 
-function getElementTree(state: State, id: string): ElementTree | null {
-  return state.get(id) ?? null
+function getElementTree(state: State, documentKey: string): ElementTree | null {
+  return state.get(documentKey) ?? null
 }
 
 export function getElements(state: State, documentKey: string): Map<string, Element> {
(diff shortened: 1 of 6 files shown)


packages/runtime/src/state/modules/element-trees.ts
```diff
@@ -243,10 +247,10 @@ function hasChildren(
 function deleteElement(
   { elements, elementIds }: ElementTree,
   deletedElement: Element,
-  propName: string,
+  propName: string | null,
   descriptors: DescriptorsByComponentType,
 ) {
-  if (hasChildren(deletedElement, propName, descriptors)) {
+  if (propName == null || hasChildren(deletedElement, propName, descriptors)) {
     for (const element of traverseElementTree(deletedElement, descriptors)) {
```
agurtovoy, line 254: This change made me realize that the existing code is suboptimal in that it always reindexes all element children rather than just those under the deleted/inserted prop, but we don't have to address this in this PR.

agurtovoy (approved):
Beautiful 🙏

packages/runtime/src/state/modules/element-trees.ts
arvinpoddar: To be clear, you're referring to the traversal step (which isn't scoped by prop), right?

agurtovoy: Correct.

PR #955, opened 2025-02-20
fix: trigger resolution on dirty props

arvinpoddar:
Trigger resolution whenever a prop is dirty. For resource-related props, like Colors and Files, this will result in attempting to fetch the resource whenever the prop value changes (from the cache or network).

https://github.com/user-attachments/assets/54fbbb1d-9688-4bdf-a322-cb339a68a5b9


packages/runtime/src/runtimes/react/hooks/use-resolvable-record.tsx
```diff
@@ -0,0 +1,63 @@
...
+          dirtyProps: Record<string, boolean>
+          resolved: Record<string, unknown>
+        }>(
+          ({ dirtyProps, resolved }, [propName, subscription]) => {
+            const lastPropValue = lastResolvedValues.current[propName]
+            const propValue = subscription.readStable()
+            const propIsDirty = propValue !== lastPropValue
+
+            return {
+              dirtyProps: { ...dirtyProps, [propName]: propIsDirty },
```
arvinpoddar, line 35: The difference of this hook vs `resolvable-record`: 
- Local state is held in `refs` instead of the function closure. Because the function was reinvoked whenever the new resolvables were passed, the closure state was lost.
- Tracks whether each prop is dirty, compared to a single boolean flag. We use this information to trigger resolution for dirty props.

packages/runtime/src/runtimes/react/hooks/use-resolvable-record.tsx
```diff
@@ -0,0 +1,63 @@
+import { type Resolvable } from '@makeswift/controls'
+import { useMemo, useRef } from 'react'
+
+export function useResolvableRecord(
+  resolvables: Record<string, Resolvable<unknown>>,
+): Resolvable<Record<string, unknown>> {
+  const lastResolvedValues = useRef<Record<string, unknown>>({})
+  const propIsDirty = useRef<Record<string, boolean>>({})
```
agurtovoy, line 8: I think we really need just a list of "drity" props, which would also simplify some of the code below. Also, since we are calculating/updating these two props together, I think we can make the code more structured by doing something like

```typescript
  type State = {
    resolvedValues: Record<string, unknown>
    dirtyProps: string[]
  }

  const state = useRef<State>({ resolvedValues: {}, dirtyProps: [] })
```

Then we can also use the `State` type in the `readStable`'s reduce to calculate the new state.

packages/runtime/src/runtimes/react/hooks/use-resolvable-record.tsx
```diff
@@ -0,0 +1,63 @@
...
+      name: 'resolvable-record',
+      subscribe: (onUpdate: () => void): (() => void) => {
+        const unsubscribes = Object.values(resolvables).map(s => s.subscribe(onUpdate))
+        return () => {
+          unsubscribes.forEach(u => u())
+        }
+      },
+
+      readStable: () => {
+        const { dirtyProps, resolved } = Object.entries(resolvables).reduce<{
```
agurtovoy, line 25: ... something like

```typescript
        const newState = Object.entries(resolvables).reduce<State>(
       ...
        )

        if (newState.dirtyProps.length > 0) {
          state.current = newState
        }

        return state.current
```

packages/runtime/src/runtimes/react/hooks/use-resolvable-record.tsx
```diff
@@ -0,0 +1,63 @@
...
+        if (hasDirtyProp) {
+          lastResolvedValues.current = resolved
+          propIsDirty.current = dirtyProps
+        }
+
+        return lastResolvedValues.current
+      },
+
+      triggerResolve: async () => {
+        return await Promise.all(
```
agurtovoy, line 53: ... and here:

```typescript
const { resolvedValues, dirtyProps } = state.current
return await Promise.all(
  dirtyProps.map(propName =>
    resolvables[propName].triggerResolve(resolvedValues[propName]),
  ),
)
```

packages/runtime/src/runtimes/react/hooks/use-resolvable-record.tsx
```diff
@@ -0,0 +1,63 @@
...
+  resolvables: Record<string, Resolvable<unknown>>,
+): Resolvable<Record<string, unknown>> {
+  const lastResolvedValues = useRef<Record<string, unknown>>({})
+  const propIsDirty = useRef<Record<string, boolean>>({})
+
+  function shouldResolveProp(propName: string): boolean {
+    return propIsDirty.current[propName] || lastResolvedValues.current[propName] == null
+  }
+
+  const mergedResolvables = useMemo<Resolvable<Record<string, unknown>>>(() => {
```
agurtovoy, line 14: Since we immediately return `mergedResolvables` w/o using it elsewhere, I'd just return `useMemo` here directly.

packages/runtime/src/runtimes/react/hooks/use-resolvable-record.tsx
agurtovoy: On a more high-level note, I wonder if by calling `triggerResolve` on "dirty" props only, we are introducing a subtle bug for (potential) resolvables that return `undefined` from `readStable` when their dependencies haven't been fetched.

packages/runtime/src/runtimes/react/hooks/use-resolvable-record.tsx
arvinpoddar: I intentionally separated the direct return because it was helpful for logging the output of the memo. Unless you're strongly against it, I'd like to keep it.

packages/runtime/src/runtimes/react/hooks/use-resolvable-record.tsx
arvinpoddar: I actually encountered this, so I had a carve out in the logic for triggering whenever the `readStable()` is `nullish`. Since our other course of action was triggering prop resolution on all props, I didn't view this as a serious cost:

```
function shouldResolveProp(propName: string): boolean {
    return propIsDirty.current[propName] || lastResolvedValues.current[propName] == null
  }
```

packages/runtime/src/runtimes/react/hooks/use-resolvable-record.tsx
agurtovoy: I just don't find `const name = ...; return name;` to be a helpful pattern in terms of aiding understanding, so I prefer to eliminate these unless there is a reason not to. Logging is a valid reason, but in this case the return value itself is just an object with functions, there is no state on the memo itself; were you logging something else?

packages/runtime/src/runtimes/react/hooks/use-resolvable-record.tsx
arvinpoddar: See [my other comment](https://github.com/makeswift/makeswift/pull/955#discussion_r1966597585) on why I stored more than just dirty prop names.

agurtovoy: Got it. I think a `Set` would be a bit clearer than `Record<string, boolean>` here.

packages/runtime/src/runtimes/react/hooks/use-resolvable-record.tsx
agurtovoy: Ah, interesting, I missed that bit.

packages/runtime/src/runtimes/react/hooks/use-resolvable-record.tsx
arvinpoddar: I like that, addressed.

agurtovoy (approved):
🙏

PR #957, opened 2025-02-27
refactor(runtime): decouple api handler endpoints from Next.js draft mode

arvinpoddar:
Refactor the API Handler to no longer rely on API requests being proxied. Instead, the host client is configured to retrieve certain versions of resources based on an explicit version passed to it.

### DIFF
--- a/packages/runtime/src/api/react.ts
+++ b/packages/runtime/src/api/react.ts
@@ -19,6 +19,7 @@ import {
   CreateTableRecordMutationResult,
   CreateTableRecordMutationVariables,
 } from './graphql/generated/types'
+import { MakeswiftSiteVersion } from './site-version'
 
 export type CacheData = MakeswiftApiClient.SerializedState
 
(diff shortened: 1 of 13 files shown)


packages/runtime/src/api/react.ts
```diff
@@ -51,17 +52,28 @@ export const CacheData = {
...
   graphqlClient: GraphQLClient
   makeswiftApiClient: MakeswiftApiClient.Store
   subscribe: MakeswiftApiClient.Store['subscribe']
 
-  constructor({ uri, cacheData, locale }: { uri: string; cacheData?: CacheData; locale?: string }) {
+  constructor({
+    uri,
+    cacheData,
+    locale,
+    siteVersion = MakeswiftSiteVersion.Live,
```
agurtovoy, line 64: I would make the site version required and not default it here. It's a breaking change either way; if you don't provide the version, the client behavior will be different from what it currently is.

agurtovoy (dismissed):
This is great 🙏

packages/runtime/src/api/react.ts
arvinpoddar: Addressed

agurtovoy (approved):

packages/runtime/src/state/makeswift-api-client.ts
```diff
@@ -124,6 +128,7 @@ async function fetchJson<T>(url: string): Promise<T | null> {
 export function fetchAPIResource<T extends APIResourceType>(
   resourceType: T,
   resourceId: string,
+  siteVersion: MakeswiftSiteVersion,
   locale?: APIResourceLocale<T>,
 ): Thunk<Promise<Extract<APIResource, { __typename: T }> | null>> {
```
migueloller, line 133: At this point I would start using named parameters for this method

migueloller (commented):
Looks great 👍🏻 

Just have one minor comment that isn't urgent

PR #959, opened 2025-03-03
feat: introduce draft proxy utility functions

arvinpoddar:
Adds utility functions for getting draft/preview data and bypass tokens and constructing a proxy response. Will be used for bringing proxying capabilities to hosts that are currently patching requests with a bypass token in middleware, which does not work for SSGd pages.

### DIFF
--- a/packages/runtime/package.json
+++ b/packages/runtime/package.json
@@ -58,6 +58,11 @@
       "import": "./dist/esm/next/index.js",
       "types": "./dist/types/next/index.d.ts"
     },
+    "./next/middleware": {
+      "require": "./dist/cjs/next/middleware/index.js",
+      "import": "./dist/esm/next/middleware/index.js",
+      "types": "./dist/types/next/middleware/index.d.ts"
+    },
     "./next/document": {
       "require": "./dist/cjs/next/document.js",
       "module": "./dist/esm/next/document.js",
@@ -143,6 +148,7 @@
     "@types/uuid": "^9.0.1",
     "@use-gesture/react": "^10.2.24",
     "color": "^3.1.3",
+    "cookie": "^1.0.2",
     "corporate-ipsum": "^1.0.1",
     "cors": "^2.8.5",
     "css-box-model": "^1.2.1",
@@ -163,6 +169,7 @@
     "redux-thunk": "^2.3.0",
     "reselect": "^5.1.1",
     "scroll-into-view-if-needed": "^2.2.20",
+    "set-cookie-parser": "^2.7.1",
     "slate": "^0.91.4",
     "slate-hyperscript": "^0.77.0",
     "slate-react": "^0.91.7",
@@ -189,6 +196,8 @@
     "@types/node": "^17.0.21",
     "@types/react": "^18.2.14",
     "@types/react-dom": "^18.2.6",
+    "@types/set-cookie-parser": "^2.4.10",
+    "@types/web": "^0.0.208",
     "concurrently": "^5.3.0",
     "expect-type": "^0.19.0",
     "jest": "^29.7.0",
--- a/packages/runtime/src/next/api-handler/handlers/draft-mode.ts
+++ b/packages/runtime/src/next/api-handler/handlers/draft-mode.ts
@@ -1,0 +1,106 @@
+import { NextApiRequest, NextApiResponse } from 'next'
+import { NextRequest, NextResponse } from 'next/server'
+import { P, match } from 'ts-pattern'
+import { cookies, draftMode } from 'next/headers'
+import { MakeswiftSiteVersion } from '../../../api/site-version'
+
+const PRERENDER_BYPASS_COOKIE = '__prerender_bypass'
+const DRAFT_DATA_COOKIE = 'x-makeswift-draft-data'
+
+type Context = { params: { [key: string]: string | string[] } }
+
+type DraftModeError = string
+
+type Response = { __brand: 'DraftModeResponse' }
+
+export type DraftModeResponse = DraftModeError | Response
+
+type DraftModeHandlerArgs =
+  | [request: NextRequest, context: Context, params: { apiKey: string }]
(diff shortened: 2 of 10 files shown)


packages/runtime/package.json
```diff
@@ -189,6 +196,8 @@
     "@types/node": "^17.0.21",
     "@types/react": "^18.2.14",
     "@types/react-dom": "^18.2.6",
+    "@types/set-cookie-parser": "^2.4.10",
+    "@types/web": "^0.0.208",
```
arvinpoddar, line 200: This was necessary to prevent a type error on `getSetCookie()` not being available on `Headers`.

packages/runtime/src/next/api-handler/handlers/draft-mode.ts
```diff
@@ -0,0 +1,101 @@
...
+export type DraftModeResponse = DraftModeError | Response
+
+type DraftModeHandlerArgs =
+  | [request: NextRequest, context: Context, params: { apiKey: string }]
+  | [req: NextApiRequest, res: NextApiResponse<DraftModeResponse>, params: { apiKey: string }]
+
+const routeHandlerPattern = [P.instanceOf(Request), P.any, P.any] as const
+const apiRoutePattern = [P.any, P.any, P.any] as const
+
+export default async function draftModeHandler(
```
arvinpoddar, line 25: The draft-mode handler (and preview-mode) handler use the built-in Next APIs to enable draft/preview mode, and patch the cookies to be iframe friendly. We'll be able to reuse these endpoints for our fully decoupled toolbar solution, but we'll still need to rely on these Next APIs given that preview mode only permits passing data through `previewData`.

In the future, these endpoints will be modified to accept a requested "site version" that will be set via a cookie / preview data respectively.

packages/runtime/src/next/middleware/request-utils.ts
```diff
@@ -0,0 +1,193 @@
...
+  } else if (isPreviewModeRequest(request)) {
+    return await fetchPreviewModeCookies(request.nextUrl.origin, draftSecret)
+  }
+
+  return []
+}
+
+class MakeswiftDraftRequest extends NextRequest {
+  isMakeswiftDraftRequest = true as const
+}
```
arvinpoddar, line 134: Custom class to prevent arbitrary NextRequests from being passed to fetchDraftProxyResponse. This class is only returned by createDraftRequest.

packages/runtime/package.json
migueloller: That's a weird requirement. I wonder if we can tune our tsconfig to address this 🤔

packages/runtime/src/next/middleware/request-utils.ts
```diff
@@ -0,0 +1,193 @@
...
+    throw new InvariantDraftRequestError(
+      "Request can't include draft mode and preview mode secrets",
+    )
+  }
+  return draftSecret ?? previewSecret
+}
+
+async function fetchCookies(url: URL, draftSecret: string): Promise<Cookie[]> {
+  const requestUrl = new URL(url)
+  requestUrl.searchParams.set('secret', encodeURIComponent(draftSecret))
```
migueloller, line 69: The `URLSearchParams` API should automatically encode the arguments of `set` so `encodeURIComponent` shouldn't be necessary here

packages/runtime/src/next/middleware/request-utils.ts
```diff
@@ -0,0 +1,193 @@
...
+      throw new UnknownDraftFetchRequestError(
+        'Encountered unknown error while fetching draft cookies',
+      )
+    })
+    .catch(err => {
+      if (err instanceof MiddlewareError) throw err
+      const exception = new UnknownDraftFetchRequestError(
+        'Encountered unknown error while fetching draft cookies',
+      )
+      exception.cause = err
```
migueloller, line 91: Why mutate the exception instead of passing the cause as a constructor parameter?

packages/runtime/src/next/middleware/request-utils.ts
```diff
@@ -0,0 +1,193 @@
...
+      throw exception
+    })
```
migueloller, line 93: I will admit that now looking at it, I'm not sure the `.then`/`.catch` looks any better than `try`/`catch` with `async`/`await`

packages/runtime/src/next/middleware/request-utils.ts
```diff
@@ -0,0 +1,193 @@
...
+  draftRequest.nextUrl.searchParams.delete(SearchParams.DraftMode)
+  draftRequest.nextUrl.searchParams.delete(SearchParams.PreviewMode)
+  draftRequest.headers.delete(HeaderNames.DraftMode)
+  draftRequest.headers.delete(HeaderNames.PreviewMode)
+
+  return draftRequest
+}
+
+export async function fetchDraftProxyResponse(
+  draftRequest: MakeswiftDraftRequest,
```
migueloller, line 168: If we're requiring a certain instance we should validate it at runtime as well. Remember not all consumers of our SDK use TypeScript

packages/runtime/src/next/middleware/request-utils.ts
migueloller: How come we're adding the extra field here? Couldn't we check at runtime by calling `request instanceof MakeswiftDraftRequest`?

migueloller (approved):
Looks good 👍🏻 

Two things I would like to see:
- Tests for the new functionality
- More descriptive errors for the new Draft Mode and Preview Mode handlers

packages/runtime/package.json
arvinpoddar: From my tinkering with this, the only solutions I could find were upgrading the TS version or adding this library. I tried various tsconfig adjustments, but that did not yield anything.

Since TS itself comes with `@types/web`, chances are that the current version we're on is simply missing the updated definitions. To avoid any potential downstream affects of upgrading TS, I'll stick with installing `@types/web` for now.

PR #960, opened 2025-03-04
draft: test middleware util on deployments

arvinpoddar:


packages/runtime/src/next/middleware/request-utils.ts
```diff
@@ -0,0 +1,193 @@
...
+  if (isDraftModeRequest(request)) {
+    return await fetchDraftModeCookies(request.nextUrl.origin, draftSecret)
+  } else if (isPreviewModeRequest(request)) {
+    return await fetchPreviewModeCookies(request.nextUrl.origin, draftSecret)
+  }
+
+  return []
+}
+
+class MakeswiftDraftRequest extends NextRequest {
```
arvinpoddar, line 132: Custom class to prevent arbitrary `NextRequest`s from being passed to `fetchDraftProxyResponse`. This class is only returned by `createDraftRequest`

PR #961, opened 2025-03-07
fix: kitchen sink app router 404 page

arvinpoddar:
Introduces a `not-found` so that the 404 page is properly rendered.

Before:

CleanShot 2025-03-07 at 09 44 17

After: 
CleanShot 2025-03-07 at 09 51 18

### DIFF
--- a/apps/nextjs-app-router/src/app/[lang]/not-found.tsx
+++ b/apps/nextjs-app-router/src/app/[lang]/not-found.tsx
@@ -1,0 +1,11 @@
+export default function NotFound() {
+  return (
+    <div className="bg-white dark:bg-black h-screen w-screen flex justify-center items-center">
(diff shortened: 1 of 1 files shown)


fikrikarim (dismissed):

migueloller:
> What's weird about this is that generally on localized pages the root layout is _under_ `[lang]` so that you can do `<html lang={lang}>`. Are we not doing this in our root layout?

Perhaps we address this by moving `not-found.tsx` under `[lang]`?

arvinpoddar:
> What's weird about this is that generally on localized pages the root layout is _under_ `[lang]` so that you can do `<html lang={lang}>`. Are we not doing this in our root layout?

Ah, good callout. We can't do this in the root layout since it won't have access to the `lang` param. I may just have to introduce a `not-found.tsx` somewhere so that the app doesn't blow up.

PR #963, opened 2025-03-10
feat: introduce native draft mode

arvinpoddar:
Remove proxying functionality from the runtime in favor of setting draft mode and version cookies on the client. Draft mode is now set via redirect. In the case a user is able to set draft cookies on the live site, this feature introduces a toolbar UI that allows the user to exit the preview.

https://github.com/user-attachments/assets/872fcd33-0d14-4dba-b90e-0caa8a553d26


packages/runtime/src/next/api-handler/handlers/redirect-draft.ts
```diff
@@ -0,0 +1,144 @@
...
+  { apiKey }: { apiKey: string },
+): Promise<NextResponse<RedirectDraftResponse>>
+export default async function redirectDraftHandler(
+  req: NextApiRequest,
+  res: NextApiResponse<RedirectDraftResponse>,
+  { apiKey }: { apiKey: string },
+): Promise<void>
+export default async function redirectDraftHandler(
+  ...args: RedirectDraftHandlerArgs
+): Promise<NextResponse<RedirectDraftResponse> | void> {
```
arvinpoddar, line 46: These handlers are essentially identical to the former `/draft-mode` and `/preview-mode`. The only difference is they return a redirect.

packages/runtime/src/next/components/tests/controls/style-control/__snapshots__/server.test.tsx.snap
```diff
@@ -6,4 +6,4 @@ exports[`renders a style: component styles 1`] = `
 ]
 `;
 
-exports[`renders a style: resolvedValue 1`] = `"css-1cny9vj u-R0-propKey"`;
+exports[`renders a style: resolvedValue 1`] = `"css-1cny9vj u-R1-propKey"`;
```
arvinpoddar, line 9: The snapshot changed because react's component IDs change based on the component rendering structure - we added a toolbar to the provider, hence new IDs.

packages/makeswift-next-plugin/index.js
```diff
@@ -27,6 +27,8 @@ const NEXT_IMAGE_REVIEW_APP_REMOTE_PATTERNS = [
   },
 ]
 
+const DraftRedirectPath = '/api/makeswift/redirect-draft'
```
migueloller, line 30: Thoughts on having a separate handler for Preview Mode?

Also, minor nit, but should we include the word "redirect" in the URL? I think it's standard to do a redirect after enabling Draft Mode or Preview Mode (It's what I remember from Next.js docs). So perhaps `/api/makeswift/draft` and `/api/makeswift/preview`?

Reference:
- https://nextjs.org/docs/pages/building-your-application/configuring/preview-mode
- https://nextjs.org/docs/app/building-your-application/configuring/draft-mode

packages/runtime/src/next/api-handler/handlers/clear-draft.ts
```diff
@@ -0,0 +1,94 @@
...
+  {}: { apiKey: string },
+): Promise<NextResponse<ClearDraftResponse>> {
+  const headers = new Headers()
+
+  headers.append(
+    SET_COOKIE_HEADER,
+    clearCookiesHeader([PRERENDER_BYPASS_COOKIE, MAKESWIFT_DRAFT_DATA_COOKIE]),
+  )
+
+  return NextResponse.json({ __brand: 'ClearDraftResponse' }, { headers })
```
migueloller, line 80: Why not use the Next.js APIs to clear Draft Mode? I feel like we should avoid any references to the "internal" `__prerender_bypass` cookie

packages/runtime/src/next/api-handler/handlers/redirect-draft.ts
```diff
@@ -0,0 +1,144 @@
...
+    },
+  ]
+
+  const destinationUrl = new URL(request.nextUrl)
+  destinationUrl.searchParams.delete(SearchParams.DraftMode)
+
+  const headers = new Headers()
+  draftCookies.forEach(({ name, value }) => {
+    headers.append(SET_COOKIE_HEADER, serializeCookie(name, value, { ...cookieSettingOptions }))
+  })
```
migueloller, line 95: We seem to be doing a heck of a lot here for a "vanilla" Draft Mode implementation

migueloller: Ah, I'm guessing it's to partition the cookie?

packages/runtime/src/next/api-handler/index.ts
```diff
@@ -174,31 +170,17 @@ export function MakeswiftApiHandler(
...
-        .with(routeHandlerPattern, args => draftMode(...args, { apiKey }))
-        .with(apiRoutePattern, args => draftMode(...args, { apiKey }))
-        .exhaustive()
-    }
-
-    if (matches('/preview-mode')) {
-      return match(args)
-        .with(routeHandlerPattern, args => previewMode(...args, { apiKey }))
-        .with(apiRoutePattern, args => previewMode(...args, { apiKey }))
+        .with(routeHandlerPattern, args => clearDraft(...args, { apiKey }))
```
migueloller, line 182: It's a bit confusing we have `clearDraft` when it's for Preview Mode

migueloller (commented):
Looks great! Left some comments. We should talk to @andrewreifman to get some design feedback

packages/runtime/src/next/api-handler/handlers/redirect-draft.ts
arvinpoddar: Bingo. `draftMode.enable` does not accept params to configure cookie settings, so we have to manually patch up the cookies ourselves.

packages/runtime/src/next/api-handler/handlers/clear-draft.ts
migueloller: I guess we can't avoid the references given we're partitioning the cookies. I guess we could use `draftMode().disable()` here but we'll still reference `__prerender_bypass` when setting it

arvinpoddar: Good question. Recall from https://github.com/makeswift/makeswift/pull/963#discussion_r1989322431 that we have to manually patch the cookies so that they're set as partitioned (plus other properties). When clearing cookies, you _have to_ set the cookies with the identical attributes it was originally set with and a `Date(0)` expiration in order for the cookie to be cleared. 

Since we modified how `draftMode.enable()` sets cookies to be partitioned, we can't rely on `draftMode.disable()`.

packages/runtime/src/next/api-handler/index.ts
arvinpoddar: I use the word `draft` to refer to the idea of "bypassing the cache" (this is also how Next.js refers to this concept in their docs). `DraftMode` refers specifically to the app router mechanism, and `PreviewMode` refers specifically to the preview mode mechanism.

packages/makeswift-next-plugin/index.js
arvinpoddar: Sure, not opposed to renaming the handler. However, what do we gain from having a separate Preview Mode handler?

arvinpoddar:
@mig

> Looks great! Left some comments. We should talk to @andrewreifman to get some design feedback

Had a call with @andrewreifman, here's the button after some refinement:

https://github.com/user-attachments/assets/7b814521-97c8-4f96-9305-241b2c2952ab

packages/makeswift-next-plugin/index.js
arvinpoddar: Partially addressed by renaming the handler to `/draft`

migueloller: I like the separation of concerns and explicitness. Having a single handler means that we're having to do detection of whether we're in App Router or Pages Router to determine whether we use one or the other, etc. and it feels a bit weird to me rather than being explicit about `/api/makeswift/draft` enables Draft Mode and redirects you and `/api/makeswift/preview` enables Preview Mode and redirects you. And both also store Makeswift draft data but one stores it in preview data and the other in a cookie (?)

packages/runtime/src/next/api-handler/handlers/clear-draft.ts
migueloller: > When clearing cookies, you have to set the cookies with the identical attributes it was originally set with and a Date(0) expiration in order for the cookie to be cleared.

Ah, I was not aware of this requirement. That's surprising to me. Thanks for clarifying 👍🏻

packages/runtime/src/next/api-handler/index.ts
migueloller: From that perspective, what's weird is that we're not clearing the draft itself (i.e., the data), but the _mode_. It was a bit confusing for me as I was reading it

migueloller (approved):

.changeset/tasty-ads-clean.md
```diff
@@ -0,0 +1,5 @@
+---
+"@makeswift/runtime": minor
+---
+
+Removes the DraftModeScript and PreviewModeScript from the runtime. These components are no longer needed for integrating a site with Makeswift, and can be safely removed from any existing code.
```
migueloller, line 5:
```suggestion
Removes the `DraftModeScript` and `PreviewModeScript` from the runtime. These components are no longer needed for integrating a site with Makeswift, and can be safely removed from any existing code.
```

apps/nextjs-app-router/src/app/[lang]/layout.tsx
```diff
@@ -33,9 +32,7 @@ export default async function RootLayout({
       lang={lang}
       className={`${GrenzeGotischFont.variable} ${GrenzeFont.variable}`}
     >
-      <head>
-        <DraftModeScript appOrigin={process.env.MAKESWIFT_APP_ORIGIN} />
-      </head>
+      <head></head>
```
Comment (migueloller) [inline apps/nextjs-app-router/src/app/[lang]/layout.tsx:35]: This line can be removed

packages/runtime/src/next/document.tsx
```diff
@@ -33,13 +30,9 @@ export class Document extends NextDocument<DocumentProps> {
...
 
   render() {
-    const { isPreview } = this.props.__NEXT_DATA__
-
     return (
       <Html>
-        <Head>
-          <PreviewModeScript isPreview={isPreview} appOrigin={this.props.appOrigin} />
-        </Head>
+        <Head></Head>
```
migueloller, line 35: Can we remove this line? If not, let's change to a self-closing tag. i.e., `<Head />`

packages/runtime/src/runtimes/react/components/hooks/use-builder-handshake.ts
```diff
@@ -0,0 +1,20 @@
...
+  const isPagesRouter = useIsPagesRouter()
+
+  const handshakeInitiationMessage = useMemo(() => {
+    if (isPagesRouter) return PAGES_ROUTER_HANDSHAKE_MESSAGE
+    return APP_ROUTER_HANDSHAKE_MESSAGE
+  }, [isPagesRouter])
+
+  useEffect(() => {
+    if (window.parent !== window) {
+      window.parent.postMessage({ type: handshakeInitiationMessage }, appOrigin)
```
migueloller, line 17: nit: if we use the object format for the second parameter, it might help future readers as to what the parameter is. i.e.:

```ts
postMessage({ type }, { targetOrigin: appOrigin })
```

packages/runtime/src/runtimes/react/components/hooks/use-builder-connection-ping.ts
```diff
@@ -0,0 +1,45 @@
+import { useEffect, useRef } from 'react'
+import { ActionTypes } from '../../../../react'
+
+const CONNECTION_PING_INTERVAL_MS = 20
+
+export function useBuilderConnectionPing(appOrigin: string) {
+  const sanitizedAppOrigin = appOrigin.replace("'", "\\'")
```
migueloller, line 7: I believe this is unnecessary and is an artifact of `appOrigin` previously being interpolated in a string

packages/runtime/src/runtimes/react/components/hooks/use-builder-connection-ping.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  const connectionInterval = useRef<number | null>(null)
```
migueloller, line 9: Couldn't we keep the interval ID in a closure instead of using a ref?

packages/runtime/src/runtimes/react/components/hooks/use-builder-connection-ping.ts
```diff
@@ -0,0 +1,45 @@
...
+      }
+    }
+
+    function messageHandler(event: MessageEvent): void {
+      if (
+        event.origin === sanitizedAppOrigin &&
+        event.data.type === ActionTypes.MAKESWIFT_CONNECTION_INIT
+      ) {
+        if (connectionInterval.current != null) {
+          window.clearInterval(connectionInterval.current)
```
migueloller, line 30: Should we be setting `connectionInterval.current = null` here?

packages/runtime/src/runtimes/react/components/hooks/use-builder-connection-ping.ts
```diff
@@ -0,0 +1,45 @@
...
+  useEffect(() => {
+    if (window.parent !== window) {
+      window.addEventListener('message', messageHandler)
+      window.parent.postMessage({ type: ActionTypes.MAKESWIFT_CONNECTION_INIT }, appOrigin)
+    }
+
+    return () => {
+      window.removeEventListener('message', messageHandler)
+      if (connectionInterval.current != null) {
+        window.clearInterval(connectionInterval.current)
```
migueloller, line 20: Should we be setting `connectionInterval.current = null` here?

migueloller (approved):
Looks great!

PR #966, opened 2025-03-12
Update Basic TypeScript Example

jamesqquick:
- update all packages, namely Next.js, React, and Makeswift runtime
- update built-in component to use `Group` control instead of deprecated `Shape` control
- update code setup to use `src` directory and match the setup in the manual Installation docs
- ignore `.env` file and rename to `.env.example`


examples/basic-typescript/src/app/[[...path]]/page.tsx
```diff
@@ -3,21 +3,20 @@ import { notFound } from 'next/navigation'
 import { Page as MakeswiftPage } from '@makeswift/runtime/next'
 import { getSiteVersion } from '@makeswift/runtime/next/server'
 
-import { client } from '@/lib/makeswift/client'
+import { client } from '@/makeswift/client'
```
Comment (agurtovoy) [inline examples/basic-typescript/src/app/[[...path]]/page.tsx:6]: @jamesqquick Can we keep the directory structure of this example and update the docs instead? I agree 100% that these two should be synchronized. `@/lib/makeswift` is not only our standard convention in the rest of the examples, though (all of which can be used to bootstrap your project through the CLI), but also our convention in Catalyst. It also makes it easier to distinguish between a global (`@makeswift/runtime/next`) and a local import (`@/lib/makeswift/client`).

Comment (jamesqquick) [reply examples/basic-typescript/src/app/[[...path]]/page.tsx]: Alright, updated to go back to the original folder structure. We'll worry about updating docs later on

agurtovoy (approved):
🙏

PR #971, opened 2025-03-17
feat: Dato example integration

drewpledger:
Here is my implementation of Dato support for Blogs. There are 2 main components to demonstrate blogging using Dato: `BlogRichText` and `BlogFeed.tsx`. The README should have the instructions required for anyone to start using this example.


examples/dato/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+This guide assumes that you're familiar with Makeswift and setting up a custom host. Read here to learn more: https://docs.makeswift.com/developer/app-router/installation
+
+## Using this example
+
+To quickly try this example either [deploy to Vercel](#deploy-this-example-to-vercel) or [use our CLI](#use-this-example-locally-with-the-makeswift-cli).
+
+In addition to your Makeswift API key, add your Dato API key in either `.env.local` or your environment variables in vercel.
+
+`NEXT_PUBLIC_DATO_CMS_API_TOKEN=your_token`
```
jamesqquick, line 14: Can we clarify where to find this token inside of Dato? Is there any required step to generate the token or does it come by default?

Can we also move this to be listed as a requirement before we mention the two ways to deploy?

Lastly, in this section you say add to `.env.local` but we show two different ways to try the example: Vercel dpeloy button and running locally. Adding to `.env.local` is only relevant if running locally. Maybe we start by saying "You need to get an environment variable" in this section, then leave the "where do you put the environment variable" context to the two specific sections below?

jamesqquick: Realized you actually mentioned both. I do think it makes more sense to include the details of where to put the envirnoment variable in the appropriate section

examples/dato/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+---
+
+### Deploy this example to Vercel
```
jamesqquick, line 18:
```suggestion
### Deploy to Vercel
```

examples/dato/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+The deploy link below includes integrations with Dato and Makeswift.
+
+[![Deploy with Vercel](https://vercel.com/button)](https://vercel.com/new/clone?repository-url=https%3A%2F%2Fgithub.com%2Fmakeswift%2Fmakeswift%2Ftree%2Fmain%2Fexamples%2Fdato&project-name=dato-makeswift-example&repository-name=dato-makeswift-example&redirect-url=https%3A%2F%2Fapp.makeswift.com&integration-ids=oac_51ryd7Pob5ZsyTFzNzVvpsGq,oac_nsrwzogJLEFglVwt2060kB0y&external-id=dato-makeswift)
+
+### Use this example locally with the Makeswift CLI
```
jamesqquick, line 24:
```suggestion
### Running locally with the Makeswift CLI
```

examples/dato/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+```bash
+npx makeswift@latest init --example=dato
+```
+
+## Local Development
+
+Run `pnpm codegen-ts` after copying the following Dato schema
+
+![alt text](https://github.com/user-attachments/assets/8f235b4b-ed0f-4b99-a018-15ceaa56f1ac)
```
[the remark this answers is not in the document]
jamesqquick: This step needs to happen before running locally or Vercel right? I think this, along with finding the API key should be the first steps that get listed here before running locally or Vercel. That way, we know they've completed the necessary steps.

examples/dato/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+We recommend following the same schema because the `.graphql` files provided assume the same schema.
+
+In order to start visually editing these components in your dynamic page, use the builder's URL bar to navigate directly to the route (i.e. `/blog/my-blog`). This page renders an embedded MakeswiftComponent that represents the Blog Page. You will then be able to use the following component that dynamically renders off the slug:
+
+- `Blog/BlogRichText` which has controls to determine which RichText field to render. Should you extend the Blog schema to include more RichText fields, you will see them inside this dropdown. See `BlogRichText.makeswift.ts` for more information.
+
+  ![alt text](https://github.com/user-attachments/assets/278fd44c-b4c0-4f29-a666-911798c4f5a1)
+
+In addition to this we included `Blog/Feed` which queries all the blogs in Dato, renders them on the page, and paginates.
```
jamesqquick, line 48: I think this deserves a screenshot to explicitly show. what we mean by a feed.

examples/dato/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+Run `pnpm codegen-ts` after copying the following Dato schema
+
+![alt text](https://github.com/user-attachments/assets/8f235b4b-ed0f-4b99-a018-15ceaa56f1ac)
+
+Note that the Structured Text type ("body") has a validation field under "Specify the blocks allowed" and "Image" in the combobox.
+
+We recommend following the same schema because the `.graphql` files provided assume the same schema.
+
+In order to start visually editing these components in your dynamic page, use the builder's URL bar to navigate directly to the route (i.e. `/blog/my-blog`). This page renders an embedded MakeswiftComponent that represents the Blog Page. You will then be able to use the following component that dynamically renders off the slug:
```
jamesqquick, line 42: This feels like it assumes some Makeswift knowledge specifically of `MakeswiftComponent`. I think we should show an abbreviated code snippet and explain a bit what we've actually done there.

The idea of "dynamically renders off the slug" is pretty high level. I think this is goig to be a new concept for people that view this and we should hand hold them a bit in explaining with this is. Maybe could say, "If you have three blog posts with titles x, y, and z, this will automatically the three associated pages with the following routes".

examples/dato/README.md
```diff
@@ -0,0 +1,59 @@
...
+npx makeswift@latest init --example=dato
+```
+
+## Local Development
+
+Run `pnpm codegen-ts` after copying the following Dato schema
+
+![alt text](https://github.com/user-attachments/assets/8f235b4b-ed0f-4b99-a018-15ceaa56f1ac)
+
+Note that the Structured Text type ("body") has a validation field under "Specify the blocks allowed" and "Image" in the combobox.
```
jamesqquick, line 38: What's the impact of that specific validation field?

examples/dato/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+We recommend following the same schema because the `.graphql` files provided assume the same schema.
```
jamesqquick, line 40: We could probably make this a bit stronger of a recommendation. Instead of "we recommend", we could be clear that our graphql queries are dependent on the same schema. Otherwise, it will break.

examples/dato/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+- `Blog/BlogRichText` which has controls to determine which RichText field to render. Should you extend the Blog schema to include more RichText fields, you will see them inside this dropdown. See `BlogRichText.makeswift.ts` for more information.
+
+  ![alt text](https://github.com/user-attachments/assets/278fd44c-b4c0-4f29-a666-911798c4f5a1)
+
+In addition to this we included `Blog/Feed` which queries all the blogs in Dato, renders them on the page, and paginates.
+
+## Next steps
+
+With Makeswift, you can give your marketing team hand-crafted, CMS building blocks to create a custom website.
```
jamesqquick, line 52: A good next CTA might be our guide on creating custom elements. That way, they can learn a little bit more about how to write code themselves to customize things.

jamesqquick (changes requested):

Issue #964, opened 2025-03-11
Emotion key


paulmabe:
### Describe the bug

Multiple instances of emotion used by different packages and scripts with the same default cache key seems to cause errors when attempting to insert stylesheets.

In general, when using the makeswift runtime, which uses a cache key of 'css', with any other service or script that creates an emotion cache with the same key, 'css', may overwrite stylesheets or produce a NotFoundError error due to mismatched insertBefore container and reference object parent elements.

In my case, the other instance of emotion is from the BigCommerce headless b2b buyer portal. The emotion cache key that the buyer portal uses is also 'css'.
```
https://cdn.bundleb2b.net/b2b/production/storefront/headless.js
```

Briefly describing my stack in context of this issue, we are utilizing Catalyst with makeswift and the B2B Buyer portal.

locally, I edited the makeswift runtime module code usage of the emotion createCache with a unique key in this file and the issue seems to have been resolved.

https://github.com/makeswift/makeswift/blob/71daac3a55e464292745ad11fe4dd2494b90757e/packages/runtime/src/next/root-style-registry.tsx#L11

### Steps to reproduce

**Steps:**
1. Implement Makeswift runtime, render a snapshot.
2. Implement some third party that uses emotion with the same key, 'css', as Makeswift runtime.
3. Trigger a stylesheet insert, click around the page, use buttons.

**Expected Behavior:**
Stylesheets are inserted properly and as expected in the head of the DOM, or container if emotion createCache container is utilized.

**Actual Behavior:**
Stylesheets are overwritten or a NotFoundError error is produced due to mismatched insertBefore container and reference object parent elements.

### Environment

- **Binaries**
  - Node.js version: `20.17.0`
  - Package Manager version: `pnpm@9.15.0`
  - Makeswift Runtime version: `0.23.4`
- **Operating system**
  - Name: `macOS (Intel chip)`
  - Version: `Sequoia 15.3.2`
- **Browser**
  - Name: `chrome` 
  - Version: `133.0.6943.142`

agurtovoy:
Good catch @paulmabe, thanks for the detailed report!

paulmabe:
@agurtovoy No problem! Curious what your thoughts are on a potential fix? I'd settle for a simple makeswift branded unique string to replace 'css' or a more meaningful change that allows for us to set the key with a ReactRuntime constructor option or client constructor argument.

agurtovoy:
[...]

PR #973, opened 2025-03-20
fix: default makeswift emotion cache key and RootStyleRegistry cacheK…

paulmabe:
Following a suggested update from @agurtovoy, [here](https://github.com/makeswift/makeswift/issues/964#issuecomment-2730518088), I updated the emotion cache key to reduce the chance of conflict with other emotion instances and added an optional prop to [`<RootStyleRegistry />`](https://github.com/makeswift/makeswift/blob/main/packages/runtime/src/next/root-style-registry.tsx#L34) so the cache key prefix can be set as needed.

### DIFF
--- a/.changeset/khaki-bags-lead.md
+++ b/.changeset/khaki-bags-lead.md
@@ -1,0 +1,5 @@
+---
+'@makeswift/runtime': minor
+---
+
+Default emotion cache key update and RootStyleRegistry optional prop to set emotion cache key
--- a/.changeset/khaki-bags-lead.md
+++ b/.changeset/khaki-bags-lead.md
@@ -1,5 +1,5 @@
 ---
-'@makeswift/runtime': minor
+'@makeswift/runtime': patch
 ---
 
 Default emotion cache key update and RootStyleRegistry optional prop to set emotion cache key
--- a/packages/runtime/src/next/root-style-registry.tsx
+++ b/packages/runtime/src/next/root-style-registry.tsx
@@ -7,10 +7,10 @@ import { ReactNode, createContext, useContext, useState } from 'react'
 
 const CacheContext = createContext(cache)
 
-const createRootStyleCache = (cacheKeyPrefix?: string) => {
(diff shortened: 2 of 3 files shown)


packages/runtime/src/next/root-style-registry.tsx
```diff
@@ -7,8 +7,12 @@ import { ReactNode, createContext, useContext, useState } from 'react'
 
 const CacheContext = createContext(cache)
 
-const createRootStyleCache = () => {
-  const cache = createCache({ key: 'css' })
+const createRootStyleCache = (cacheKeyPrefix?: string) => {
```
migueloller, line 10: What do you think if we made this a named parameter and also had it define the entire cache key instead of just the prefix?

migueloller (commented):
@paulmabe, thank you so much for the contribution! We'll need a changeset and possibly tests before we ship this. For the changeset, it should be easy to figure out how to generate one by following the [changesets](https://github.com/changesets/changesets) docs. For the tests, I don't think we have a test suite for this particular behavior right now so you might need some support from @agurtovoy on that

.changeset/khaki-bags-lead.md
```diff
@@ -0,0 +1,5 @@
+---
+'@makeswift/runtime': minor
```
migueloller, line 2: Given we're pre-1.0 we would consider this a non-breaking change and this a `patch`

migueloller (commented):
Thanks for addressing the comments!

.changeset/khaki-bags-lead.md
paulmabe: Understood, updated to `patch`

agurtovoy (approved):
Thank you for putting this together, @paulmabe!

I temporarily set the default cache prefix back to `css` to avoid breaking our existing tests; will follow up with a separate PR that updates the prefix and the test snapshots.

Really appreciate your contribution ❤️

PR #983, opened 2025-03-26
fix: avoid document context rerenders

arvinpoddar:
As part of the document context, we were passing the document root element. This means that every edit would result in the context value changing, resulting in all children rerendering. For larger element trees, this means that all props in the document would be re-resolved for any prop change. This fix only passes the necessary fields for the document (key and locale).

## Demos

Before:

https://github.com/user-attachments/assets/8504cd9e-d4be-489f-98b8-6fdad4e9a450

After:

https://github.com/user-attachments/assets/aeba54cb-2725-4975-9e49-0fdca7db4847

## Measured Evidence (Rendering Profiling):

All measurements were taken on the same element tree by modifying the same prop.

Before regression, runtime `v0.22.3`  (7.2ms per render):

pre-regression-profile-data

Regression (71.8ms per render, order of magnitude higher than before):

regression-profile-data

With Fix (5.2ms per render)

post-fix-profile-data

### DIFF
--- a/.changeset/calm-apples-smell.md
+++ b/.changeset/calm-apples-smell.md
@@ -1,0 +1,5 @@
+---
+"@makeswift/runtime": patch
+---
+
+Fixes a prop editing performance regression introduced in `v0.23.0`.
--- a/packages/runtime/src/runtimes/react/components/Document.tsx
+++ b/packages/runtime/src/runtimes/react/components/Document.tsx
@@ -1,7 +1,7 @@
 import { Ref, forwardRef, memo } from 'react'
 import { type Document as ReactPageDocument, getRootElement } from '../../../state/react-page'
 import { ElementImperativeHandle } from '../element-imperative-handle'
-import { DocumentContext } from '../hooks/use-document-context'
+import { DocumentKeyContext, DocumentLocaleContext } from '../hooks/use-document-context'
 import { Element } from './Element'
 
 type DocumentProps = {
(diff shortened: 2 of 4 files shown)


packages/runtime/src/runtimes/react/components/Document.tsx
```diff
@@ -14,9 +14,11 @@ export const Document = memo(
     ref: Ref<ElementImperativeHandle>,
   ): JSX.Element {
     return (
-      <DocumentContext.Provider value={document}>
-        <Element ref={ref} element={getRootElement(document)} />
-      </DocumentContext.Provider>
+      <DocumentKeyContext.Provider value={document.key}>
```
agurtovoy, line 17: This is fixing a regression introduced [here](https://github.com/makeswift/makeswift/commit/12b0123e59d403cd36c5b3afd11e262fea9906ca#diff-be45a66a0ddef1e58d09bcd2ab646b3c5fbac719f24d97252d37f3e836a241dfL17).

.changeset/calm-apples-smell.md
```diff
@@ -0,0 +1,5 @@
+---
+"@makeswift/runtime": patch
+---
+
+fix: prop editing performance improvements via correcting context value
```
agurtovoy, line 5: I'd add that this fixes a performance regression introduced in 0.23.

agurtovoy (approved):
Thank you for tracking down and fixing my blunder!

.changeset/calm-apples-smell.md
arvinpoddar: Done

PR #985, opened 2025-03-27
Create/register prop controllers for documents added after (host <--> builder) connection established

Codeseph:
After:
https://github.com/user-attachments/assets/10d6797e-b224-457c-8d2b-69c861db7beb

### DIFF
--- a/packages/runtime/src/runtimes/react/hooks/use-register-document.ts
+++ b/packages/runtime/src/runtimes/react/hooks/use-register-document.ts
@@ -1,8 +1,9 @@
-import { useEffect } from 'react'
+import { useEffect, useLayoutEffect, useMemo } from 'react'
 import { type Document } from '../../../state/react-page'
 import { useDispatch } from './use-dispatch'
 import { useIsInBuilder } from './use-is-in-builder'
 import { registerBuilderDocumentsEffect, registerDocumentsEffect } from '../../../state/actions'
+import { isServer } from '../../../utils/is-server'
 
 /**
  * @param document Document to register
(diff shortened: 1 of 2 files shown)


arvinpoddar:
Excellent digging. However, I think there's something more sneaky going wrong here. If the prop controllers are failing to be created because the document hasn't been registered in the runtime store yet, that's a bigger problem and I feel that relying on effects may be a bandaid solution rather than a cure.

packages/runtime/src/runtimes/react/hooks/use-register-document.ts
```diff
@@ -11,7 +12,17 @@ export function useRegisterDocument(document: Document): void {
...
+  if (isServer()) {
+    dispatch(registerDocument(document))
+  }
+
+  /*
+    Layout effect is to ensure that the document registration happens prior to the
+    attempted creation/registration of prop controllers in child components.
+  */
+  useLayoutEffect(() => {
+    dispatch(registerDocument(document))
```
arvinpoddar, line 24: We still need to rely on returning `registerDocumentsEffect` here since we need the returned cleanup function that unregisters the document whenever the root component unmounts. `registerDocument` doesn't give us a cleanup function.

packages/runtime/src/runtimes/react/hooks/use-register-document.ts
```diff
@@ -11,7 +12,17 @@ export function useRegisterDocument(document: Document): void {
   const isInBuilder = useIsInBuilder()
   const dispatch = useDispatch()
 
-  useEffect(() => dispatch(registerDocumentsEffect([document])), [document])
+  if (isServer()) {
+    dispatch(registerDocument(document))
```
arvinpoddar, line 16: See comment on `registerDocumentsEffect`. I think you'll need to introduce the memoized documents array, and then pass that to both this case and the layout effect.

arvinpoddar (approved):

packages/runtime/src/runtimes/react/hooks/use-register-document.ts
```diff
@@ -11,12 +12,24 @@ export function useRegisterDocument(document: Document): void {
...
+
+  if (isServer()) {
+    dispatch(registerDocumentsEffect(documentArray))
+  }
+
+  /*
+    Layout effect is to ensure that the document registration happens prior to the
+    attempted creation/registration of prop controllers in child components.
+  */
+  useLayoutEffect(() => {
```
agurtovoy, line 25: Sorry I haven't reviewed this earlier, but I *think* using `useLayoutEffect` here unconditionally will result in warnings during SSR—see, e.g. this commit: https://github.com/makeswift/makeswift/commit/1683722. @migueloller is that correct?

We should probably use our `useIsomorphicLayoutEffect` wrapper instead.

migueloller: Yes, that's correct. I don't particularly love that we're using a layout effect to ensure document registrations happens before creation/registration of prop controllers in child components because it makes an assumption about behavior that is not local. Particularly, it makes an assumption that we're using `useEffect` there.

Removing the requirement for order would be more robust. Would that be a practical thing to pursue?

packages/runtime/src/runtimes/react/hooks/use-register-document.ts
```diff
@@ -11,12 +12,24 @@ export function useRegisterDocument(document: Document): void {
   const isInBuilder = useIsInBuilder()
   const dispatch = useDispatch()
 
-  useEffect(() => dispatch(registerDocumentsEffect([document])), [document])
+  const documentArray = useMemo(() => [document], [document])
+
+  if (isServer()) {
+    dispatch(registerDocumentsEffect(documentArray))
```
migueloller, line 18: Dispatching an effect on the server feels inappropriate to me given that we ignore the returned cleanup function

PR #989, opened 2025-04-03
Change server dispatch for document registration

Codeseph:
(just showing that hot reload dropzones are still working)
https://github.com/user-attachments/assets/3e198f5e-7d39-4ad1-b357-e6334163089b

### DIFF
--- a/.changeset/flat-nails-mate.md
+++ b/.changeset/flat-nails-mate.md
@@ -1,0 +1,5 @@
+---
+"@makeswift/runtime": patch
+---
+
+chore: refactor use-register-document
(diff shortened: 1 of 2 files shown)


arvinpoddar (commented):
Btw, I think we can do away with the memoization of the document now (getting rid of `documentArray` in favor of defining the document array inline in the effects). Makes the code easier to read IMO:

```
useIsomorphicLayoutEffect(() => {
    return dispatch(registerDocumentsEffect([document]))
  }, [dispatch, document])

  useEffect(() => {
    if (!isInBuilder) return
    return dispatch(registerBuilderDocumentsEffect([document]))
  }, [isInBuilder, document])
```

arvinpoddar (approved):
Looking good! Don't forget to add a changeset.

Codeseph:
whoops added the changeset

arvinpoddar (approved):

PR #990, opened 2025-04-04
feat: add onPublish callback to MakeswiftApiHandler

ferrata:
## Summary

This PR adds `onPublish` callback to the `MakeswiftApiHandler` function. It will allow users to implement a custom logic when site content is getting published.

### `onPublish` callback

```ts
onPublish: () => void | Promise<void>;
```

The callback will be called once per a site publish event. The callback can be synchronous or asynchronous.

> [!IMPORTANT]
> The publish event only works for a deployed site. E.g. it won't works for a local development environment.

> [!NOTE]
> Any errors thrown in the callback will be logged and ignored.

### Example

To use the `onPublish` callback, implement it in the `MakeswiftApiHandler` call:

```diff
 const handler = MakeswiftApiHandler(process.env.MAKESWIFT_SITE_API_KEY, {
   runtime,
+  events: {
+    onPublish() {
+      console.log(`Published content`);
+    },
+  }
 });
 ```

### CODE REVIEW

**Review by migueloller — COMMENTED** [review]
@ferrata, the structure here looks great! As for the source of the event that calls the callback, I would prefer that we use a "publish event" instead of "hooking into" the revalidate call. 

IIRC, @fikrikarim set up the web hook endpoint for tag-based revalidation in App Router. Perhaps we can use the mechanism instead?

I am reviewing this pull request from a plane so I can't easily check the source without losing progress on this comment. So I will comment back with more details once I find them.

**Comment by migueloller** [comment]
@ferrata, I have found the relevant handler here: packages/runtime/src/next/api-handler/handlers/webhook/site-published.ts

**Comment by migueloller** [comment]
One more note. It's ok if the callback has no arguments for now. It can simply be used as a signal for when a publish has happened.

What's problematic about revalidation, by the way, is its coupling to Makeswift-defined pages, which does not cover builtin elements.

File: packages/runtime/src/next/api-handler/handlers/webhook/site-published.ts [hunk 1]
```diff
@@ -1,9 +1,23 @@
...
+  _payload: SitePublishedWebhookPayload,
+  { onPublish }: SitePublishedParams
+): Promise<WebhookHandlerResult> {
   revalidateTag(MAKESWIFT_CACHE_TAG)
 
+  try {
+    await onPublish?.()
+  } catch (error) {
+    // log and ignore any error in user-provided onPublish
+    console.error(error)
```
agurtovoy, line 18: I'd add context to the error log here, e.g. `console.error("Unhandled exception in the 'onPublish' callback:", error)`

packages/runtime/src/next/api-handler/index.ts
```diff
@@ -33,6 +33,7 @@ type MakeswiftApiHandlerConfig = {
   appOrigin?: string
   apiOrigin?: string
   getFonts?: GetFonts
+  onPublish?: OnPublish
```
migueloller, line 36: I wonder if we should put this under a `webhook` field so that we have future webhook callbacks grouped in there 🤔

packages/runtime/src/next/tests/server.makeswift-api-handler.test.ts
```diff
@@ -0,0 +1,207 @@
...
+    consoleErrorSpy = jest.spyOn(console, 'error').mockImplementation(() => { });
+  });
+
+  describe('/api/makeswift/revalidate', () => {
+    function revalidateRequest(
+      query: Record<string, string | string[]> = {},
+    ) {
+      return createRequest<NextApiRequest>({
+        method: 'GET',
+        query: { makeswift: ['revalidate'], ...query },
```
migueloller, line 44: These should probably be set in the URL not directly in the query? Unless I'm missing how this utility function is intended to work?

What I mean is, the URL of the request should be `/api/makeswift/revalidate`. Passing it in the query makes some assumptions about Next.js routing internals (and honestly probably shouldn't work)

migueloller (approved):

packages/runtime/src/next/api-handler/index.ts
ferrata: That's a good idea!

fikrikarim (dismissed):
This is great, @ferrata!

Something that came to mind a few months ago about this feature is whether the `onPublish` hook should override or extend the built-in handler.

I lean towards extending it, since I can’t think of a reason why a user would want to disable the default built-in handler.

packages/runtime/src/next/tests/server.makeswift-api-handler.test.ts
ferrata: I could not find a tool that can be used to simulate how Next.js does this thing. I will make a little helper to make it use URL instead of a structure.

migueloller:
> The publish event only works for a deployed site. E.g. it won't works for a local development environment.

In the future, I would love to have support for this locally via the Makeswift CLI using a method similar to [Stripe's](https://docs.stripe.com/webhooks#local-listener)

migueloller (approved):
Looking good 👍🏻 

Nice work on your first contribution to the runtime!

agurtovoy (approved):
🙏

migueloller (approved):

PR #993, opened 2025-04-10
wip: explicitly provide registered components

migueloller:
This PR attempts to show how we can make the dependency of Makeswift components more explicit by adding a `components` option to `ReactRuntime` and thus removing the need for side effects at the module level with `regsiterComponent`.

> [!CAUTION]
> This has not been tested thoroughly!


arvinpoddar (approved):
Very straightforward

PR #994, opened 2025-04-11
fix: replace react lazy with next/dynamic

fikrikarim:


fikrikarim:
The snapshot release action is failing. Here's the current snapshot version:
```
0.0.0-snapshot-20250411073228
```

PR #1000, opened 2025-04-18
chore: pass app origin as named param in connection hooks

arvinpoddar:
Pass the app origin as the `targetOrigin` for all `postMessage` calls for explicitness.

### DIFF
--- a/packages/runtime/src/runtimes/react/components/hooks/use-builder-connection-ping.ts
+++ b/packages/runtime/src/runtimes/react/components/hooks/use-builder-connection-ping.ts
@@ -9,7 +9,10 @@ export function useBuilderConnectionPing({ appOrigin }: { appOrigin: string }) {
 
     if (window.parent !== window) {
       window.addEventListener('message', messageHandler)
-      window.parent.postMessage({ type: ActionTypes.MAKESWIFT_CONNECTION_INIT }, appOrigin)
+      window.parent.postMessage(
+        { type: ActionTypes.MAKESWIFT_CONNECTION_INIT },
+        { targetOrigin: appOrigin },
+      )
     }
 
     return () => {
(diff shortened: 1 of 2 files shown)


packages/runtime/src/runtimes/react/components/hooks/use-builder-connection-ping.ts
```diff
@@ -9,7 +9,10 @@ export function useBuilderConnectionPing({ appOrigin }: { appOrigin: string }) {
 
     if (window.parent !== window) {
       window.addEventListener('message', messageHandler)
-      window.parent.postMessage({ type: ActionTypes.MAKESWIFT_CONNECTION_INIT }, appOrigin)
+      window.parent.postMessage(
+        { type: ActionTypes.MAKESWIFT_CONNECTION_INIT },
+        { targetOrigin: appOrigin },
```
agurtovoy, line 14: @arvinpoddar Can you elaborate on what this affects/why we are changing it? 🙏

agurtovoy (approved):

packages/runtime/src/runtimes/react/components/hooks/use-builder-connection-ping.ts
arvinpoddar: Practically, it does not change anything. The second argument to `postMessage` can either be a string or an options object. If it's a string, it's assumed to be the intended origin of the message, or target origin. If it's an options object, one of the options is `targetOrigin`, which is the intended origin of the message. All this change does is use the named option to be more explicit about what's happening here.

agurtovoy: Got it, thanks for clarifying!

PR #1005, opened 2025-04-20
feat: localized pages are online by default, fallback to the base locale

agurtovoy:
Last piece of [ENG-7803](https://linear.app/makeswift/issue/ENG-7803/make-localized-pages-online-by-default).

### DIFF
--- a/.changeset/lemon-ravens-divide.md
+++ b/.changeset/lemon-ravens-divide.md
@@ -1,0 +1,19 @@
+---
+'@makeswift/runtime': minor
+---
+
+BREAKING: Upgrading to this runtime version will opt your site into the new localized pages behavior, both in the builder and on the live site.
+
+Localized pages are no longer silently created when navigated to in the builder. Instead, they now automatically fall back to the base locale by default. To create a localized page, users must take explicit action in the builder.
+
+Localized pages that are explicitly marked as Offline will remain offline.
+
+You can disable fallback behavior on a per-page basis by passing `allowLocaleFallback: false` to the `client.getPageSnapshot` call:
+
+```typescript
+const snapshot = await client.getPageSnapshot(path, {
+  siteVersion: await getSiteVersion(),
+  locale,
+  allowLocaleFallback: false,
+})
+```
--- a/packages/runtime/src/next/client.ts
+++ b/packages/runtime/src/next/client.ts
@@ -662,16 +662,23 @@ export class Makeswift {
     {
       siteVersion: siteVersionPromise,
       locale,
-    }: { siteVersion: MakeswiftSiteVersion | Promise<MakeswiftSiteVersion>; locale?: string },
+      allowLocaleFallback = true,
+    }: {
+      siteVersion: MakeswiftSiteVersion | Promise<MakeswiftSiteVersion>
+      locale?: string
+      allowLocaleFallback?: boolean
+    },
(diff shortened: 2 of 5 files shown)


packages/runtime/src/next/client.ts
```diff
@@ -662,16 +662,23 @@ export class Makeswift {
...
   ): Promise<MakeswiftPageSnapshot | null> {
-    const searchParams = new URLSearchParams()
-    if (locale) {
-      searchParams.set('locale', locale)
+    const queryParams = (): string => {
+      const params = new URLSearchParams()
+      if (locale) params.set('locale', locale)
+      if (allowLocaleFallback != null) params.set('allowLocaleFallback', `${allowLocaleFallback}`)
+      return params.toString()
     }
```
arvinpoddar, line 677: Curious, why'd we put this in a function instead of having it inline?

arvinpoddar (approved):

arvinpoddar:
🙏

arvinpoddar:
Also, curious how we arrived at the product decision to make this behavior unconditional (unless opt-out). I'm guessing it's some combination of it being a better experience than how we handle localization for pages now, and wanting UX consistency with builtin elements?

.changeset/lemon-ravens-divide.md
```diff
@@ -0,0 +1,19 @@
+---
+'@makeswift/runtime': minor
+---
+
+BREAKING: Upgrading to this runtime version will opt your site into the new localized pages behavior, both in the builder and on the live site.
+
+Localized pages are no longer silently created when navigated to in the builder. Instead, they now automatically fall back to the base locale by default. To create a localized page, users must take explicit action in the builder.
+
+Localized pages that are explicitly marked as Offline will remain offline.
```
arvinpoddar, line 9: Should we specify that this is "marked as Offline **in the builder**"?

packages/runtime/src/next/client.ts
agurtovoy: It's a technique for limiting the scope / lifting the level of abstraction at this level. A function basically gives you a named entity you can skip over/collapse until/unless you are ready to examine what exactly hides under that name.

.changeset/lemon-ravens-divide.md
agurtovoy: We can, although IMHO this is understood and also the previous para ends with "in the builder."

packages/runtime/src/next/client.ts
arvinpoddar: Let me elaborate, what I meant was _why_ we decided to abstract this. I don't think it's a bad choice, I've just been in similar situations where I decided the implementation was light enough not to. Trying to develop a sense for when something should be encapsulated like this.

agurtovoy: My general guideline for writing function-level code is that all statements within a function should operate at the same level of abstraction. I was adding an extra if statement to this logic, and what previously felt like an okay-ish "inlined abstraction" started to feel too in-the-weeds to keep inlined.

agurtovoy:
> I'm guessing it's some combination of it being a better experience than how we handle localization for pages now, and wanting UX consistency with builtin elements?

@arvinpoddar yes, we want to gradually phase out the old localized pages experience in the builder.

PR #1007, opened 2025-04-22
Add helper for introspecting without controls

Codeseph:
Companion PR: https://github.com/makeswift/cosmos/pull/1266

### DIFF
--- a/packages/runtime/src/builder/unstructured-introspection/unstructured-introspection.ts
+++ b/packages/runtime/src/builder/unstructured-introspection/unstructured-introspection.ts
@@ -1,0 +1,69 @@
+import { Base64 } from 'js-base64'
+import { Element } from '@makeswift/controls'
+
+import { z } from 'zod'
+import { APIResourceType } from '../../api'
+
+function isAPIResourceType(val: string): val is APIResourceType {
+  return APIResourceType[val as keyof typeof APIResourceType] != null
+}
+
+const parseResourceIdSchema = z.string().transform((val, ctx) => {
+  try {
+    const match = Base64.decode(val).match(/^([^:]+):([^:]+)$/)
+    if (!match) throw new TypeError(`NodeID cannot represent value: ${String(val)}`)
+    const [, typeName, key] = match
+    if (isAPIResourceType(typeName)) {
+      return { key, typeName }
+    }
+    throw new TypeError(`decoded type '${typeName}' is not a valid APIResourceType`)
+  } catch (error) {
+    ctx.addIssue({
+      code: z.ZodIssueCode.custom,
+      message: `${String(val)} is not a valid node resource ID`,
+    })
+    return z.NEVER
+  }
+})
+
+export function unstructuredIntrospection(element: Element): {
+  swatchIds: Set<string>
+  typographyIds: Set<string>
+  fileIds: Set<string>
+  tableIds: Set<string>
+  pageIds: Set<string>
+  globalElementIds: Set<string>
+} {
+  const typeIdsToFound: Map<APIResourceType, Set<string>> = new Map()
+  Object.values(APIResourceType).forEach(resourceType => {
+    typeIdsToFound.set(resourceType, new Set())
+  })
+  function findResourceIds(curr: unknown): void {
+    if (typeof curr === 'string') {
+      const parseResourceResult = parseResourceIdSchema.safeParse(curr)
+      if (parseResourceResult.success) {
+        typeIdsToFound.get(parseResourceResult.data.typeName)?.add(curr)
+      }
+    } else if (Array.isArray(curr)) {
+      for (const item of curr) {
+        findResourceIds(item)
+      }
+    } else if (curr != null && typeof curr === 'object') {
+      for (const key in curr) {
+        if (Object.prototype.hasOwnProperty.call(curr, key)) {
+          findResourceIds(curr[key as keyof typeof curr])
+        }
+      }
+    }
+  }
+
+  findResourceIds(element)
+  return {
+    swatchIds: typeIdsToFound.get(APIResourceType.Swatch) ?? new Set(),
+    typographyIds: typeIdsToFound.get(APIResourceType.Typography) ?? new Set(),
+    fileIds: typeIdsToFound.get(APIResourceType.File) ?? new Set(),
+    tableIds: typeIdsToFound.get(APIResourceType.Table) ?? new Set(),
+    pageIds: typeIdsToFound.get(APIResourceType.Page) ?? new Set(),
+    globalElementIds: typeIdsToFound.get(APIResourceType.GlobalElement) ?? new Set(),
+  }
+}
--- a/.changeset/heavy-rice-complain.md
+++ b/.changeset/heavy-rice-complain.md
@@ -1,0 +1,5 @@
+---
+"@makeswift/runtime": patch
+---
+
+Add helper function for introspecting without utilizing controls
--- a/packages/runtime/package.json
+++ b/packages/runtime/package.json
@@ -160,6 +160,7 @@
     "html-react-parser": "^5.0.10",
     "immutable": "^4.0.0-rc.12",
     "is-hotkey": "^0.1.4",
+    "js-base64": "^3.7.7",
     "ot-json0": "^1.1.0",
     "parse5": "^7.1.2",
     "path-to-regexp": "^6.2.1",
--- a/packages/runtime/src/builder/index.ts
+++ b/packages/runtime/src/builder/index.ts
@@ -1,1 +1,2 @@
 export * from './serialization'
+export * from './unstructured-introspection'
--- a/packages/runtime/src/builder/unstructured-introspection/__fixtures__/element-containing-resources.ts
+++ b/packages/runtime/src/builder/unstructured-introspection/__fixtures__/element-containing-resources.ts
@@ -1,0 +1,884 @@
+import { Element } from '@makeswift/controls'
+
+/*
+  Below is a fairly minimal example of an Element containing one of each:
+    - Swatch
+    - File
+    - Typography
+    - Global Element
+    - Page Link (called 'Page' in APIResourceType)
+    - Table
+*/
+export const elementContainingResources: Element = {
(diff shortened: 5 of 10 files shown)


packages/runtime/src/builder/raw-json-introspection/raw-json-introspection.ts
```diff
@@ -0,0 +1,37 @@
...
+ * Introspects a raw JSON string without utilizing controls.
+ *
+ * Iterates over members of the JSON string and attempts to Base64 decode
+ * each value. If successful, and the decoded result corresponds to a known APIResourceType,
+ * then the decoded resource type/ID are added to the result map.
+ *
+ * @param json - A raw JSON string
+ * @returns A map from APIResourceType(s) found in the JSON string to the Set of IDs found for each.
+ */
+export function introspectRawJson(json: string): Map<APIResourceType, Set<string>> {
```
arvinpoddar, line 21: Two things:
- I think we should call this function `unstructuredIntrospection` (with appropriate file renames).
- Second, I think the input to this function should be restricted to `ElementData`, rather than a JSON string. We have some existing schemas/types that you can leverage here, see `packages/controls/src/common/schema.ts` and `packages/controls/src/common/types.ts`
- Can we make the returned `Map` always have every resource type? If a particular resource type isn't found, it's simply an empty `Set`.

packages/runtime/src/index.ts
```diff
@@ -17,6 +17,8 @@ export {
   setLocale,
   setLocalizedResourceId,
 } from './state/actions'
+export { APIResourceType } from './api'
+export { introspectRawJson } from './builder/raw-json-introspection/raw-json-introspection'
```
arvinpoddar, line 21: I thought about this a bit more. I think the "right" place to expose this is the `runtime/builder` entry point, since for now these are intended to be consumed by the builder only.

Codeseph: updated

packages/runtime/src/builder/raw-json-introspection/__fixtures__/raw-json-containing-resources.ts
```diff
@@ -0,0 +1,152 @@
+export const fileIds = ['RmlsZTpjNzc2MjI3ZC1iNmZjLTQ3MGYtYTY1OC00NWQ4NzA2NDVkMDM=']
+export const globalElementIds = [
+  'R2xvYmFsRWxlbWVudDoxMjZiNjgwYy0yZDNmLTRjYWMtOTQ3NS01ZmJkNGJlZTkyY2Y=',
+]
+export const typographyIds = ['VHlwb2dyYXBoeTo1ODVkMGE1ZS0zYmUyLTQyNzEtOTczYS1mNWNjN2UzMzk4ODg=']
+
+export const rawJsonContainingResources = JSON.stringify({
```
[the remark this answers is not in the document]
Codeseph: Yeah in hindsight this wasn't a particularly effective test case. I put together a fairly minimal page containing the 6 resources we care about for our dialog:
- Swatch
- File
- Typography
- GlobalElement
- Table
- Page (called 'Page' but is really a pathname)

and copied the JSON into a new fixture

packages/runtime/src/builder/raw-json-introspection/raw-json-introspection.ts
Codeseph: (third bullet here) - the result Mapping will now always contain values of our `APIResourceType` as keys, even if there are no occurrences of the corresponding resource in the introspected tree 

image

Codeseph: How strongly do you feel about the first two bullet points? I wonder if there's a benefit to narrowing the input type?

Codeseph: I can't think of a situation in which we'd be introspecting JSON that isn't from an element tree, but still

packages/runtime/src/builder/raw-json-introspection/__fixtures__/raw-json-containing-resources.ts
```diff
@@ -0,0 +1,152 @@
+export const fileIds = ['RmlsZTpjNzc2MjI3ZC1iNmZjLTQ3MGYtYTY1OC00NWQ4NzA2NDVkMDM=']
+export const globalElementIds = [
+  'R2xvYmFsRWxlbWVudDoxMjZiNjgwYy0yZDNmLTRjYWMtOTQ3NS01ZmJkNGJlZTkyY2Y=',
+]
+export const typographyIds = ['VHlwb2dyYXBoeTo1ODVkMGE1ZS0zYmUyLTQyNzEtOTczYS1mNWNjN2UzMzk4ODg=']
```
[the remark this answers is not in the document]
Codeseph: Looking into this, might have some questions soon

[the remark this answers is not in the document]
Codeseph: Done. That looks much better, thanks for the suggestion

packages/runtime/src/builder/raw-json-introspection/raw-json-introspection.ts
arvinpoddar: Fairly strong? Introspection is an element data only concept, so it doesn't really make sense to have an introspection function apply to anything else. Do you have any counterpoints?

Codeseph: I'm thinking of the element data in terms of 2 regions:

Region (A) - outside of a 'props' value, where we know the structure (keys are 'type', 'key', 'props', or 'value')
Region (B) - inside of 'props', where we know nothing about the structure without having control definitions to tell us the expected keys

[Client introspection](https://github.com/makeswift/makeswift/blob/4181243788c38e6fbd25e3def00caaf01a1fa09d/packages/runtime/src/next/client.ts#L508) makes sense to me as a version of introspection that knows the full tree structure (knows the keys of both regions above). It doesn't strike me as inappropriate to have a version of introspection that knows neither/can be used to find resource IDs in any bit of JSON. To me calling a version of introspection "unstructured" would almost imply that. But maybe that's just semantics/I should only be applying "structured" or "unstructured" in reference to 'props'... Anyway I think a good argument _against_ accepting a string for this function could be avoiding JSON.stringify(). Although there is somewhat of a price to be paid in that we won't be able to rely on JSON.parse() to crawl through 'props'. Bummer. I can rework this to accept only ElementData (or whatever the runtime type is) tomorrow, I just hope it's the right move

arvinpoddar: That is a good point, my counterargument for limiting the input is mostly an issue of establishing an interface that introspection is a "element data" concept - the underlying implementation will still have to treat it like an arbitrary JSON object. It doesn't make sense to pass an arbitrary JSON string other than being able to use JSON.parse for the recursion: for example, `'null'` is a JSON string, why should we introspect that? 

To simplify implementation, I wouldn't begin to rely on the JSON methods, and rather enforce that our function takes in something that has the shape of an object, and then deeply recurse over the values. We have things that can be referenced for this like `deepEquals`.

packages/runtime/src/builder/unstructured-introspection/unstructured-introspection.ts
```diff
@@ -0,0 +1,42 @@
...
+  return { key, typeName }
+}
+
+export function unstructuredIntrospection(element: Element) {
+  const typeToIdsFound: Map<APIResourceType, Set<string>> = new Map()
+  Object.values(APIResourceType).forEach(resourceType => {
+    typeToIdsFound.set(resourceType, new Set())
+  })
+
+  function findResourceIds(curr: any) {
```
arvinpoddar, line 18: Can this be `unknown` instead of `any`?

packages/runtime/src/builder/unstructured-introspection/unstructured-introspection.ts
```diff
@@ -0,0 +1,42 @@
+import { Base64 } from 'js-base64'
+import { APIResourceType } from '../../api'
+import { Element } from '@makeswift/controls'
+
+function parseNodeId(nodeId: string) {
```
arvinpoddar, line 5: Can we add explicit return types to this and the `unstructuredIntrospection` function? Also, see if we can convert this to a zod schema for parsing, in which case this function won't exist anymore.

arvinpoddar (commented):
Looking great!

packages/runtime/src/builder/unstructured-introspection/unstructured-introspection.ts
```diff
@@ -0,0 +1,42 @@
...
+  const typeToIdsFound: Map<APIResourceType, Set<string>> = new Map()
+  Object.values(APIResourceType).forEach(resourceType => {
+    typeToIdsFound.set(resourceType, new Set())
+  })
+
+  function findResourceIds(curr: any) {
+    if (typeof curr === 'string') {
+      try {
+        const parsedNodeId = parseNodeId(curr)
+        const typeName = parsedNodeId.typeName as APIResourceType
```
arvinpoddar, line 22: We should be able to avoid this cast if we use a type guard function somewhere

packages/runtime/src/index.ts
```diff
@@ -17,9 +17,11 @@ export {
   setLocale,
   setLocalizedResourceId,
 } from './state/actions'
+export { APIResourceType } from './api'
```
arvinpoddar, line 20: Can we move this so that it's only exposed via `runtime/builder`?

packages/runtime/src/index.ts
```diff
@@ -17,9 +17,11 @@ export {
...
 export type { Operation } from './state/modules/read-write-documents'
 export type { ComponentMeta } from './state/modules/components-meta'
 export { ComponentIcon } from './state/modules/components-meta'
+export { unstructuredIntrospection } from './builder'
```
arvinpoddar, line 24: This should be removed

arvinpoddar (commented):
Also, make sure the commit follows a conventional commit format (feat: ...)

Codeseph:
Thanks for all the tips here @arvinpoddar

PR #1008, opened 2025-04-24
feat: add `enableCssReset` option to `RootStyleRegistry` to support Tailwind v4 and other `@layer`-based CSS framework

fikrikarim:
## What happened?

[Tailwind v4 use @layer](https://tailwindcss.com/blog/tailwindcss-v4#designed-for-the-modern-web). This caused the CSS reset to override the Tailwind styles.
Screenshot 2025-04-24 at 19 40 09
[source](https://developer.mozilla.org/en-US/docs/Web/CSS/@layer)

## Fix

Add `enableCssReset` option to `RootStyleRegistry` to disable the CSS reset.

## Proof

https://github.com/user-attachments/assets/73bb766e-600f-4c14-b124-76fefd686081

### DIFF
--- a/.changeset/shy-actors-grow.md
+++ b/.changeset/shy-actors-grow.md
@@ -1,0 +1,5 @@
+---
+'@makeswift/runtime': patch
+---
+
+Add `enableCssReset` prop to `RootStyleRegistry` (default `true`). Set it to `false` when using a `@layer`-based CSS framework such as Tailwind v4.
--- a/packages/runtime/src/next/root-style-registry.tsx
+++ b/packages/runtime/src/next/root-style-registry.tsx
@@ -7,6 +7,9 @@ import { ReactNode, createContext, useContext, useState } from 'react'
 
 const CacheContext = createContext(cache)
 
+const DEFAULT_CSS_RESET_ENABLED = true
+const CSSResetEnabledContext = createContext(DEFAULT_CSS_RESET_ENABLED)
+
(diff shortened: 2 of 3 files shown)


.changeset/shy-actors-grow.md
```diff
@@ -0,0 +1,5 @@
+---
+'@makeswift/runtime': patch
```
fikrikarim, line 2: Patch or minor?

fikrikarim: This is technically a breaking change. But AFAIK this only affects people that already use `@layer` and they rely on the reset CSS to override something.

packages/runtime/src/next/root-style-registry.tsx
```diff
@@ -7,6 +7,10 @@ import { ReactNode, createContext, useContext, useState } from 'react'
 
 const CacheContext = createContext(cache)
 
+// Aligned with Tailwind's lowest priority layer.
+const DEFAULT_RESET_LAYER_NAME = 'theme'
```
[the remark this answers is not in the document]
fikrikarim: Yeah, that's what I'm still exploring right now. We can only have the lower priority layer if:

1. User define the `@layer makeswift` on their `global.css` before Tailwind import.
2. We find a way to inject `@layer makeswift` before Tailwind style tag, which is on `global.css`.

chanceaclark (approved):

arvinpoddar (approved):

PR #1012, opened 2025-04-30
fix(pages router): localized pages redirect to the base locale in the builder

agurtovoy:
Fixes a regression introduced in 0.24.0 when we removed [this code](https://github.com/makeswift/makeswift/blob/a645d1e83bc0fb4422ff14387738b3aa93c91c60/packages/runtime/src/next/api-handler/handlers/proxy-preview-mode.ts#L92).

https://github.com/user-attachments/assets/da1dfb57-8bd6-4648-b3fb-144e771eeca8

### DIFF
--- a/packages/makeswift-next-plugin/index.js
+++ b/packages/makeswift-next-plugin/index.js
@@ -70,6 +70,7 @@ module.exports =
                 },
               ],
               destination: '/api/makeswift/preview',
+              locale: false,
             },
           ]
           return {
--- a/packages/runtime/src/next/api-handler/handlers/redirect-preview.ts
+++ b/packages/runtime/src/next/api-handler/handlers/redirect-preview.ts
@@ -51,7 +51,7 @@ export default async function redirectPreviewHandler(
 async function redirectPreviewRouteHandler(
   _request: NextRequest,
   _context: Context,
-  {}: { apiKey: string },
+  { }: { apiKey: string },
 ): Promise<NextResponse<RedirectPreviewResponse>> {
   const message =
     'Cannot request preview endpoint from an API handler registered in `app`. Move your Makeswift API handler to the `pages/api` directory'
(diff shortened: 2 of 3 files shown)


packages/makeswift-next-plugin/index.js
```diff
@@ -70,6 +70,7 @@ module.exports =
                 },
               ],
               destination: '/api/makeswift/preview',
+              locale: false,
```
agurtovoy, line 73: `locale: false` disables "transparent" locale handling and instructs Next to include the locale prefix in the path: https://nextjs.org/docs/pages/api-reference/config/next-config-js/rewrites#rewrites-with-i18n-support

packages/runtime/src/next/api-handler/handlers/redirect-preview.ts
```diff
@@ -51,30 +51,34 @@ export default async function redirectPreviewHandler(
...
 
+
 async function redirectPreviewApiRouteHandler(
   req: NextApiRequest,
   res: NextApiResponse<RedirectPreviewResponse>,
   { apiKey }: { apiKey: string },
 ): Promise<void> {
   const secret = req.query[SearchParams.PreviewMode]
+  // `query.path` is filled in by the preview mode rewrite here: https://github.com/makeswift/makeswift/blob/main/packages/makeswift-next-plugin/index.js#L64
+  const pathname = req.query.path as string | undefined
```
agurtovoy, line 70: Next translates rewrite URL params (e.g. `:path`) into corresponding query params.

packages/runtime/src/next/api-handler/handlers/redirect-preview.ts
```diff
@@ -51,30 +51,37 @@ export default async function redirectPreviewHandler(
 async function redirectPreviewRouteHandler(
   _request: NextRequest,
   _context: Context,
-  {}: { apiKey: string },
+  { }: { apiKey: string },
```
migueloller, line 54: This looks like maybe something Prettier should've caught?

migueloller (approved):

packages/runtime/src/next/api-handler/handlers/redirect-preview.ts
agurtovoy: It *is* a Prettier reformat.

packages/runtime/src/next/api-handler/handlers/redirect-preview.ts
```diff
@@ -99,7 +106,7 @@ async function redirectPreviewApiRouteHandler(
 
   res.setHeader(SET_COOKIE_HEADER, patchedCookies)
 
-  const destinationUrl = new URL(req.url, 'http://test.com')
+  const destinationUrl = new URL(pathname, 'http://test.com')
```
jamesqquick, line 109: Don't have context for this, but should `http://test.com` be referenced here?

arvinpoddar: The `http://test.com` is only there because an origin is required to construct the URL. In the following lines, we pull out the pathname from this URL, which is what we actually need for the redirect.

packages/runtime/src/next/api-handler/handlers/redirect-preview.ts
arvinpoddar: Good find! Much better than that hack

packages/runtime/src/next/api-handler/handlers/redirect-preview.ts
agurtovoy: We don't run Prettier checks on CI.

migueloller: We probably should 😄

PR #1016, opened 2025-05-02
Explicitly provide registered components

fikrikarim:
Based on Miguel's work on PR https://github.com/makeswift/makeswift/pull/993


packages/runtime/src/runtimes/react/react-runtime.ts
```diff
@@ -19,7 +20,40 @@ function validateComponentType(type: string, component?: ComponentType): void {
...
+}
+
+type ComponentRegistration<P extends Record<string, LegacyDescriptor | UnifiedControlDefinition>> =
+  {
+    component: ComponentType<{ [K in keyof P]: DescriptorValueType<P[K]> }>
+    meta: ComponentRegistrationMeta<P>
+  }
+
 export class ReactRuntime extends RuntimeCore {
+  static connect<
```
fikrikarim, line 40: Any better name for this? Some alternatives:
- `createComponent`
- `defineComponent`
- `component`

One thing I'm not sure about connect is that it can be interpreted as a verb that has side effect of connecting the component to the runtime, even though in this case it's connecting the component to its prop.

apps/nextjs-pages-router/src/pages/api/makeswift/[...makeswift].ts
```diff
@@ -2,12 +2,6 @@ import { MAKESWIFT_SITE_API_KEY } from '@/makeswift/env'
 import { runtime } from '@/makeswift/runtime'
 import { MakeswiftApiHandler } from '@makeswift/runtime/next/server'
 
-// This import is required for the Smartling integration.
-// To translate a page, we need to iterate the component data finding all translatable strings.
-// Without control definitions, we can't traverse the component data.
-// Importing components here allows us to get translatable strings and merge them back once they are translated.
-import '@/makeswift/components'
```
Comment (migueloller) [inline apps/nextjs-pages-router/src/pages/api/makeswift/[...makeswift].ts:9]: I would love to see some tests that validate this still works when using the new API

packages/runtime/src/runtimes/react/__tests__/react-runtime.test.tsx
```diff
@@ -43,10 +44,45 @@ function Sandbox(props: {
...
+
+    expect(reg).toEqual({
+      component: Component,
+      meta: expect.objectContaining(meta),
+    })
+  })
+})
+
+describe('ReactRuntime()', () => {
+  it('accepts components registration', () => {
```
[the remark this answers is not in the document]
migueloller: Maybe you can simulate it by passing a client reference instead of trying to use `use client`?

packages/runtime/src/runtimes/react/react-runtime.ts
migueloller: In the future, I would like the API to return a component that can be rendered directly and internally uses `getComponentSnapshot` and `MakeswiftComponent`.

I took the `connect` name from Figma's [Code Connect](https://www.figma.com/code-connect-docs/) and that one is indeed a side effect.

I don't mind `createComponentRegistration` and later on we can introduce `createMakeswiftComponent` once the thing that's returned is renderable.

`createComponent` feels misleading since we're not really creating a component. Unless we want to go ahead and return a `MakeswiftComponent` component that has `type` bound to the `meta.type`? That shouldn't be a heavy lift. We could also have `MakeswiftComponent.meta` have the appropriate values.

We've then be able to use it in builtin elements like so:

```ts
const MakeswiftNav = ReactRuntime.createComponent(Nav, {
  type: 'nav',
  label: 'Nav',
  props: {},
})

const snapshot = await getComponentSnapshot('nav')

<MakeswiftNav snapshot={snapshot} label="Nav" /> // No `type`!
```

packages/runtime/src/runtimes/react/react-runtime.ts
```diff
@@ -50,9 +84,30 @@ export class ReactRuntime extends RuntimeCore {
     }
   }
 
-  constructor({ breakpoints }: { breakpoints?: BreakpointsInput } = {}) {
+  constructor({
+    breakpoints,
+    components,
+  }: {
+    breakpoints?: BreakpointsInput
+    components?: Record<string, ComponentRegistration<any>>
```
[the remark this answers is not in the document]
migueloller: I think that we should accept any iterable. I didn't do that in my PR because it was a very quick draft. I spent like 10 minutes on it.

packages/runtime/src/runtimes/react/react-runtime.ts
```diff
@@ -50,9 +84,30 @@ export class ReactRuntime extends RuntimeCore {
...
+  } = {}) {
```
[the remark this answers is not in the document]
migueloller: I see no problems with deprecating it and releasing the new API and making the breaking change later. If we decide to deprecate it, we _should_ mark `registerComponent` as `@deprecated` using JSDoc and in there document the alternative.

packages/runtime/src/runtimes/react/react-runtime.ts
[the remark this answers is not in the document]
fikrikarim: Why do we need to accept any iterable? Isn't it simpler if we only accept either an Array or a Record?

packages/runtime/src/runtimes/react/react-runtime.ts
```diff
@@ -50,9 +85,30 @@ export class ReactRuntime extends RuntimeCore {
...
     super({ breakpoints })
 
     registerBuiltinComponents(this)
```
fikrikarim, line 97: What if, when you pass `components`, we don't register the built-in components by default, and you have to include the default components manually?

We need a way to do this for the [Remove built-in components from runtime](https://linear.app/makeswift/project/remove-built-in-components-from-runtime-fe5104cd48a3) project

packages/runtime/src/runtimes/react/react-runtime.ts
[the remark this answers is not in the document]
migueloller: Why don't we prepare those things before we make this release so that we can release this change together with those updates and deprecate the API?

It's ok releasing it without deprecating `registerComponent`, but I feel like once we put the API out there it's easy to get lazy and then leave the two APIs there for a long time since we can just use the new one and ignore the old one. This happened with `ReactRuntime` APIs before where we had a "new, better" version of using it but it took us forever to deprecate.

packages/runtime/src/runtimes/react/react-runtime.ts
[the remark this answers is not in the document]
migueloller: That's what I mean… Also, I'm not sure it's "simpler" if we specify and handle objects and records explicitly, but I digress.

packages/runtime/src/runtimes/react/__tests__/react-runtime.test.tsx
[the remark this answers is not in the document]
migueloller: Perfect is the enemy of good 🙂

packages/runtime/src/runtimes/react/react-runtime.ts
migueloller: Would we export the builtin components so you can do something like `[…myComponents, …builtinComponents]`?

packages/runtime/src/runtimes/react/react-runtime.ts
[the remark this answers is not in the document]
fikrikarim: Sure let's deprecate `registerComponent` in this PR. 

I'll create the appropriate PRs once we finalize the public API.

packages/runtime/src/runtimes/react/react-runtime.ts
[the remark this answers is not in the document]
fikrikarim: Wait, I don't understand. Do we want to accept an array
```
ReactRuntime({
  components: [Accordion, Slider]
})
```

an object
```
ReactRuntime({
  components: { Accordion, Slider }
})
```

or **any iterable** including `Map` and `Set`?

packages/runtime/src/runtimes/react/react-runtime.ts
fikrikarim: Yes, we need to export the builtin components, or at least the registration.

We might want to make it explicit though, so users can easily remove components that they don't need, like how we want to exclude Form on Catalyst.

packages/runtime/src/runtimes/react/react-runtime.ts
fikrikarim: > What do you mean by "revamp" here?

I thought we're also changing the `MakeswiftComponent`.

We still need to pass the return value to `ReactRuntime`, regardless whether they're a component or not.

Returning a component might introduce an error if they try to use the component directly, but they forget to pass the component to `ReactRuntime`.

packages/runtime/src/runtimes/react/react-runtime.ts
migueloller: Yeah, perhaps we can have an option, `disableAutomaticBuiltinsRegistration` on the runtime constructor

packages/runtime/src/runtimes/react/react-runtime.ts
```diff
@@ -19,7 +20,41 @@ function validateComponentType(type: string, component?: ComponentType): void {
...
+
+export type ComponentRegistration<
+  P extends Record<string, LegacyDescriptor | UnifiedControlDefinition>,
+> = {
+  component: ComponentType<{ [K in keyof P]: DescriptorValueType<P[K]> }>
+  meta: ComponentRegistrationMeta<P>
+}
+
 export class ReactRuntime extends RuntimeCore {
+  static connect<
```
agurtovoy, line 41: The more I think about this change, the more I question whether it makes sense to ship it incrementally, or whether it would be better folded into our broader effort to remake the runtime and decouple it from Next.js.

While this doesn’t significantly *increase* the complexity of the component registration model, it also doesn’t reduce the number of concepts and entities users need to learn or define to get started with Makeswift. At the same time, it introduces another round of API changes for existing users, without offering a clear, immediate benefit.

If we weren’t already planning a larger runtime overhaul in the near future, I’d feel more comfortable polishing and shipping this as a standalone improvement. But given that we are, I think it’s worth raising the question.

migueloller: I'm a month late responding here, so I apologize...

I'm not opposed to incorporating something like this later but I disagree that it doesn't offer a clear, immediate benefit to existing and new users. The current API makes it very easy for new component registrations to be missed. Yes, once you have things set up properly and there's a single file full of imports it reduces the number of places you need to care about. But you still have to remember to do that and this change removes that. For new users, the benefit is even clearer.

Given that we currently haven't prioritized when we'll do the runtime overhaul, I feel like we're holding back on something that we know is clearly better based on something that we're not sure when we'll do.

agurtovoy: > I feel like we're holding back on something that we know is clearly better

Yeah, I don't feel that way, which is why I wrote what I wrote. I tried to polish this into a shippable state and imagine trying to justify this change to the users, and I couldn't sell it to myself. It might be just me, but that's where my original comment was coming from.

migueloller: That's alright. I'm ok with us closing this and bringing it back later when it makes more sense, then. I'll delegate to you guys on this.

PR #1019, opened 2025-05-07
feat: support clearing context in data copying

arvinpoddar:
Updates the `ReplacementContext` to support removing resource references while copying data.


packages/controls/src/context.ts
```diff
@@ -60,3 +72,71 @@ export type CopyContext = {
...
+    case ContextResource.Typography:
+      return hasRemoveTag(id, ctx.replacementContext.typographyIds)
+    case ContextResource.Table:
+      return hasRemoveTag(id, ctx.replacementContext.tableIds)
+    case ContextResource.TableColumn:
+      return hasRemoveTag(id, ctx.replacementContext.tableColumnIds)
+    case ContextResource.Page:
+      return hasRemoveTag(id, ctx.replacementContext.pageIds)
+    case ContextResource.GlobalElement:
+      return hasRemoveTag(id, ctx.replacementContext.globalElementIds)
```
agurtovoy, line 112: I'd go even further and do this: https://github.com/makeswift/makeswift/pull/1021

packages/controls/src/controls/slot/slot.ts
```diff
@@ -99,7 +101,19 @@ abstract class Definition<RuntimeNode> extends ControlDefinition<
 
     return {
       ...data,
-      elements: data.elements.map((element) => context.copyElement(element)),
+      elements: data.elements.flatMap((element) => {
```
agurtovoy, line 104: I think using a `filter` would be more idiomatic here:

```typescript
data.elements
    .filter((element) => !isElementReference(element) || !shouldRemoveResource(ContextResource.GlobalElement, element.value, context))
    .map((element) => context.copyElement(element))
```

packages/controls/src/controls/typography/typography.ts
```diff
@@ -125,12 +154,9 @@ class Definition extends ControlDefinition<
                 color: {
                   ...override.value.color,
                   alpha: override.value.color.alpha ?? null,
-                  swatchId:
-                    context.replacementContext.swatchIds.get(
-                      override.value.color.swatchId ?? '',
-                    ) ??
-                    override.value.color.swatchId ??
-                    null,
+                  swatchId: replaceOverrideSwatchId(
```
agurtovoy, line 157: I wonder if the logic here should be more along the lines of

```typescript
...(override.value.color == null || shouldRemoveResource(ContextResource.Swatch, override.value.color.swatchId, context)
? {}
: ...
```

since that's what we do for the standalone color.

agurtovoy: ... and in `copyColorData` [below](https://github.com/makeswift/makeswift/pull/1019/commits/8647b90bf38197f24896bdbe5ef71d2dcd8ab135#diff-c55d05c33333d8d9ae3308170932088a5859445f0c96631ef5e067c4ffe59973R23).

packages/prop-controllers/src/grid/grid.ts
```diff
@@ -144,7 +149,19 @@ export function getGridPropControllerElementChildren(
 function copyGridData(data: GridData, context: CopyContext): GridData {
   return {
     ...data,
-    elements: data.elements.map((element) => context.copyElement(element)),
+    elements: data.elements.flatMap((element) => {
```
agurtovoy, line 152: Here, similarly to https://github.com/makeswift/makeswift/pull/1019/commits/8647b90bf38197f24896bdbe5ef71d2dcd8ab135#r2078437155, I think using `filter` would make the intent clearer.

packages/prop-controllers/src/table-form-fields/table-form-fields.ts
```diff
@@ -136,12 +141,27 @@ function copyTableFormFieldsData(
 
   return {
     ...data,
-    fields: data.fields.map((field) => ({
-      ...field,
-      tableColumnId:
-        context.replacementContext.tableColumnIds.get(field.tableColumnId) ??
-        field.tableColumnId,
-    })),
+    fields: data.fields.flatMap((field) => {
```
agurtovoy, line 144: Maybe `filter` here as well?

packages/controls/src/controls/image/image.ts
```diff
@@ -253,20 +258,34 @@ class Definition<C extends Config = DefaultConfig> extends ControlDefinition<
...
-    const replaceFileId = (fileId: string) =>
-      replacementContext.fileIds.get(fileId) ?? fileId
+    const fileResourceId = match(data satisfies z.infer<typeof this.dataSchema>)
+      .with(P.string, (id) => id)
+      .with(Definition.dataSignature.makeswiftFile, ({ id }) => id)
+      .otherwise(() => undefined)
+
+    if (
+      fileResourceId != null &&
+      shouldRemoveResource(ContextResource.File, fileResourceId, ctx)
```
agurtovoy, line 272: We have enough cases where we have to check whether the resource ID is null before passing it to `shouldRemoveResource` that I think we should just enable the ID argument to be null.

packages/controls/src/controls/typography/typography.ts
arvinpoddar: The general rule of thumb I've been following is "remove the resource reference at the finest granularity permitted by the data." In the case of `ColorData`, there cannot be a case where `swatchId` is nullish, so I have to ignore the entire data and return `undefined`. In the case of typography, however, the data does permit for a nullish swatch ID, so I keep the other information (alpha) and only clear the swatch ID.

arvinpoddar: Done

packages/controls/src/controls/image/image.ts
arvinpoddar: It would feel weird for me to allow this since the method name starts to make less sense. shouldRemoveResource, but I can pass an argument that has nothing to do with a resource.

packages/controls/src/controls/color/color.ts
```diff
@@ -167,22 +172,28 @@ class Definition<C extends Config> extends ControlDefinition<
...
+      return undefined
+    }
 
     const inputSchema = this.dataSchema.optional()
     return match(data satisfies z.infer<typeof inputSchema>)
       .with(Definition.dataSignature.v1, (val) => ({
         ...val,
-        swatchId: replaceSwatchId(val.swatchId),
+        swatchId:
+          getReplacementResourceId(ContextResource.Swatch, val.swatchId, ctx) ??
```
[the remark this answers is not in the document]
arvinpoddar: I went for the first level of simplification. While I like the idea of making `CopyContext` a class, I don't know if that's something I'd want to stuff into one PR.

packages/controls/src/controls/image/image.ts
agurtovoy: I guess it depends on how you look at it. To me, when a resource is nullable, null is a resource ID.

arvinpoddar: Hm, fair point.

agurtovoy (approved):
🙏

PR #1024, opened 2025-05-12
Decoupling Next.js Sandbox

fikrikarim:


packages/runtime/src/remix/api-handler.ts
```diff
@@ -0,0 +1,52 @@
...
+      const apiHandler = createMakeswiftApiHandler(apiKey, {
+        ...userConfig,
+        manifest: {
+          previewMode: true,
+        },
+        draftCookieNames: [previewModeCookie.name],
+        revalidationHandler,
+        runtime,
+        client,
+      })
```
migueloller, line 47: 🍹 Could this be moved outside of `loader`?

PR #1025, opened 2025-05-13
fix: `<script>` tag error in the console when a site is loaded in the builder

agurtovoy:
Fix `Cannot render a sync or defer <script> outside of the main document` console error in the App Router sites.

Verified for both App and Pages Router.

### DIFF
--- a/packages/runtime/src/next/components/tests/__snapshots__/makeswift-page-metadata-rendering.test.tsx.snap
+++ b/packages/runtime/src/next/components/tests/__snapshots__/makeswift-page-metadata-rendering.test.tsx.snap
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`MakeswiftPage metadata and seo tag rendering does NOT render any metadata when passing all options as false 1`] = `
+exports[`MakeswiftPage metadata and seo tag rendering does NOT render any page metadata when passing all options as false 1`] = `
 <head>
   <style
     href="makeswift-base-styles"
@@ -15,10 +15,14 @@ exports[`MakeswiftPage metadata and seo tag rendering does NOT render any metada
         }
         
   </style>
+  <meta
+    content="{"draft":false,"inBuilder":false}"
+    name="makeswift-draft-info"
+  />
 </head>
 `;
 
-exports[`MakeswiftPage metadata and seo tag rendering does NOT render any metadata when passing empty options 1`] = `
+exports[`MakeswiftPage metadata and seo tag rendering does NOT render any page metadata when passing empty options 1`] = `
 <head>
   <style
     href="makeswift-base-styles"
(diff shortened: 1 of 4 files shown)


packages/runtime/src/runtimes/react/components/draft-switcher/draft-switcher.tsx
```diff
@@ -46,12 +48,15 @@ export function DraftSwitcher({ isDraft }: { isDraft: boolean }) {
...
-      <script
+      {/* Insert draft mode information into the DOM to make it easier to debug draft mode-related 
+          issues on production sites */}
+      <Script
         type="application/json"
         id="makeswift-draft-info"
         dangerouslySetInnerHTML={{
           __html: JSON.stringify({ draft: isDraft, inBuilder: isInBuilder }),
         }}
+        strategy="afterInteractive"
```
agurtovoy, line 59: Note that one of the suggested solutions in the error message is to make the `<script>` tag async, which was my first preference since it'd avoid introducing yet another dependency on Next, but that did not resolve the issue.

arvinpoddar (approved):

arvinpoddar:
Thank you 🙏

packages/runtime/src/runtimes/react/components/draft-switcher/draft-switcher.tsx
```diff
@@ -35,8 +35,6 @@ export function DraftSwitcher({ isDraft }: { isDraft: boolean }) {
     }
   }, [showToolbar, shadowRoot, setShadowRoot])
 
-  if (!showToolbar) return null
-
```
agurtovoy, line 39: Always render the debug `<meta>` tag; toolbar visibility is already conditioned on `showToolbar` (below).

packages/runtime/src/runtimes/react/components/draft-switcher/draft-switcher.tsx
```diff
@@ -46,13 +45,11 @@ export function DraftSwitcher({ isDraft }: { isDraft: boolean }) {
...
-        id="makeswift-draft-info"
-        dangerouslySetInnerHTML={{
-          __html: JSON.stringify({ draft: isDraft, inBuilder: isInBuilder }),
-        }}
-      />
+      {/* Insert draft mode information into the DOM to make it easier to debug draft mode-related 
+          issues on production sites */}
+      <PageMeta
+        name="makeswift-draft-info"
+        content={JSON.stringify({ draft: isDraft, inBuilder: isInBuilder })} />
```
agurtovoy, line 52: Replaced `<script>` with `<meta>`. In addition to triggering the original error, the `<script>` element also contained incorrect draft mode values due to the fact React only injects `<script>` content on the first render.

Note that, on the other hand, this introduces a new dependency on Next so that the `<meta>` tag gets properly hoisted to the head in React 18. It's worth considering whether we want to require React 19 for the decoupled version of the runtime.

cc: @fikrikarim

packages/runtime/src/next/components/tests/__snapshots__/makeswift-page-metadata-rendering.test.tsx.snap
```diff
@@ -15,10 +15,14 @@ exports[`MakeswiftPage metadata and seo tag rendering does NOT render any metada
         }
         
   </style>
+  <meta
+    content="{"draft":false,"inBuilder":false}"
+    name="makeswift-draft-info"
+  />
```
agurtovoy, line 21: Update page metadata test snapshots, see below.

arvinpoddar (approved):

packages/runtime/src/runtimes/react/components/draft-switcher/draft-switcher.tsx
fikrikarim: Thank you!

PR #1027, opened 2025-05-19
fix: surround elements with error boundaries

arvinpoddar:
Prevent rendering errors from breaking entire pages by surrounding each element component with an error boundary. This allows the user to delete any erroring elements.

Before:

https://github.com/user-attachments/assets/1401174b-436e-4001-a401-034ac87d345c

After:

https://github.com/user-attachments/assets/2ce0af18-a508-47c8-8ccd-4e6a689a397c

### DIFF
--- a/.changeset/fair-crabs-deny.md
+++ b/.changeset/fair-crabs-deny.md
@@ -1,0 +1,5 @@
+---
+"@makeswift/runtime": patch
+---
+
+Surround all elements with an error boundary to prevent rendering errors from causing page-wide failures.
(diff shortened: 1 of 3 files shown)


joshuawootonn (approved):
LGTM, but probably wait for another review, as I don't have context on the potential perf problems you mentioned.

agurtovoy (approved):
🙏

migueloller (approved):
This is LGTM 👍🏻 

I did want to point out, though, that in the future I would like to investigate how we can avoid having `<Suspense>` or error boundaries wrapping our elements. Ideally we would allow users to be able to leverage both of these features on their components while composing with Slot. As it stands right now, though, our `<Suspense<` and error boundaries will "catch" first.

For example, imagine the user wanted to have various components that suspended and then a component that used a Slot that had a visually editable pending state. Then, they'd wrap their suspending components with the fallback component. Same goes for the error case.

All that said, I don't think we're in a position to do that right now given how we rely on `<Suspense>` for performance and the importance of handling errors for users _right now_. Also, what I'm describing is a pretty niche use case...

PR #1030, opened 2025-05-30
feat: framework-independent API handler implementation

agurtovoy:
Incorporates [comprehensive tests](https://github.com/makeswift/makeswift/pull/1062) for all `api/makeswift/` routes. Includes a fix for the hanging CORS preflight request promise in the Pages Router.


packages/runtime/src/api-handler/request-response.ts
migueloller, line 1: What do you think if we modeled our standard request and response interface after the web `Request` and `Response` interfaces?

agurtovoy: I gave it a try, see the current version below.

packages/runtime/package.json
```diff
@@ -204,7 +204,7 @@
     "@types/react": "^18.2.14",
     "@types/react-dom": "^18.2.6",
     "@types/set-cookie-parser": "^2.4.10",
-    "@types/web": "^0.0.208",
+    "@typescript/lib-dom": "npm:@types/web@^0.0.243",
```
agurtovoy, line 209: Update TypeScript's DOM lib [type definitions](https://www.npmjs.com/package/@types/web) to include `Request.json` ([requires Node 18+)](https://developer.mozilla.org/en-US/docs/Web/API/Response/json_static#browser_compatibility).

packages/runtime/src/api-handler/index.ts
```diff
@@ -0,0 +1,177 @@
...
+      'The Makeswift API handler must be passed a valid Makeswift site API key. ' +
+        `Received "${apiKey}" instead.`,
+    )
+  }
+
+  return async function (req: ApiRequest, route: string): Promise<ResponseType> {
+    const res =
+      req.method.toUpperCase() !== 'OPTIONS'
+        ? await apiRouteHandler(req, route)
+        : new Response(null, { status: 204, headers: [['Content-Length', '0']] })
```
agurtovoy, line 90: Handle CORS preflight requests: https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request

packages/runtime/src/next/tests/server.api-handler-cors.test.ts
```diff
@@ -1,60 +1,56 @@
...
-  // FIXME: testing App Router only, the Pages Router currently hangs a promise on preflight requests
-  describe.each([{ fixture: appRouterApiRequestFixture, router: 'app' }])(
-    '[$router router] CORS requests',
-    ({ fixture }) => {
-      test.each(['/api/makeswift/manifest', '/api/makeswift/fonts'])(
-        'responds with 204 and appropriate CORS headers to a preflight request (%s)',
-        async path => {
-          // Arrange
-          const { testApiRequest } = fixture()
+  describe.each(apiRequestFixtures)('[$router router] CORS requests', ({ fixture }) => {
```
agurtovoy, line 8: Best reviewed with "Hide whitespace" toggle on.

packages/runtime/src/next/tests/server.api-handler-cors.test.ts
```diff
@@ -1,60 +1,56 @@
...
-          expect(headers.get('access-control-allow-headers')).toBe('Content-Type, Authorization')
-          expect(await textBody).toBe('')
-        },
-      )
+        // Assert
+        expect(statusCode).toBe(204)
+        expect(headers.get('content-length')).toBe('0')
+        expect(headers.get('access-control-allow-origin')).toBe('https://app.fakeswift.com')
+        expect(headers.get('access-control-allow-methods')).toBe(
+          'GET, HEAD, PUT, POST, DELETE, PATCH, OPTIONS',
```
agurtovoy, line 26: New implementation returns a more comprehensive list of allowed methods.

packages/runtime/src/next/tests/server.api-handler-manifest.test.ts
```diff
@@ -51,7 +51,7 @@ describe('MakeswiftApiHandler', () => {
         siteVersions: true,
         unstable_siteVersions: true,
         localizedPageSSR: true,
-        webhook: !pagesRouter,
+        webhook: true,
```
agurtovoy, line 54: `onPublish` event should be called in both Pages and App Router, see [this thread](https://makeswifthq.slack.com/archives/C014SCS4G75/p1753224257561999?thread_ts=1750462168.060599&cid=C014SCS4G75) for more details.

packages/runtime/src/next/tests/server.api-handler-webhook.test.ts
```diff
@@ -109,7 +109,7 @@ describe('MakeswiftApiHandler', () => {
 
       // Assert
       expect(statusCode).toBe(200)
-      if (router === 'app' || true) {
+      if (router === 'app') {
```
agurtovoy, line 112: `revalidateTag` [should not be called](https://makeswifthq.slack.com/archives/C014SCS4G75/p1750678367865609?thread_ts=1750462168.060599&cid=C014SCS4G75) when running in the Pages Router.

packages/runtime/src/next/tests/server.api-handler-clear-draft.test.ts
```diff
@@ -36,9 +36,13 @@ describe('MakeswiftApiHandler', () => {
...
         router == 'pages'
-          ? '__prerender_bypass=; Path=/; Expires=Thu, 01 Jan 1970 00:00:00 GMT; HttpOnly; Secure; Partitioned; SameSite=None, __next_preview_data=; Path=/; Expires=Thu, 01 Jan 1970 00:00:00 GMT; HttpOnly; Secure; Partitioned; SameSite=None'
+          ? [
+              '__prerender_bypass=; Path=/; Expires=Thu, 01 Jan 1970 00:00:00 GMT; HttpOnly; Secure; Partitioned; SameSite=None',
+              '__next_preview_data=; Path=/; Expires=Thu, 01 Jan 1970 00:00:00 GMT; HttpOnly; Secure; Partitioned; SameSite=None',
+            ]
           : [
-              '__prerender_bypass=; Path=/; Expires=Thu, 01 Jan 1970 00:00:00 GMT; HttpOnly; Secure; Partitioned; SameSite=None, x-makeswift-draft-data=; Path=/; Expires=Thu, 01 Jan 1970 00:00:00 GMT; HttpOnly; Secure; Partitioned; SameSite=None',
+              '__prerender_bypass=; Path=/; Expires=Thu, 01 Jan 1970 00:00:00 GMT; HttpOnly; Secure; Partitioned; SameSite=None',
+              'x-makeswift-draft-data=; Path=/; Expires=Thu, 01 Jan 1970 00:00:00 GMT; HttpOnly; Secure; Partitioned; SameSite=None',
```
agurtovoy, line 45: `headers.getSetCookie()` in the new implementation always returns the `Set-Cookie` headers as an array; otherwise, the actual headers' content is the same.

packages/runtime/src/next/api-handler/index.ts
```diff
@@ -1,308 +1,225 @@
...
+          existingValue.push(value)
+        } else {
+          acc[key] = [existingValue, value]
         }
+      } else {
+        acc[key] = value
+      }
+      return acc
+    },
+    {} as Record<string, string | string[]>,
```
arvinpoddar: We can avoid the casting here by specifying a type to the reduce: `.reduce<Record<string, string | string[]>`

packages/runtime/src/next/api-handler/index.ts
```diff
@@ -1,308 +1,225 @@
...
-): (...args: MakeswiftApiHandlerArgs) => Promise<NextResponse<MakeswiftApiHandlerResponse> | void> {
-  const cors = Cors({ origin: appOrigin })
-
-  if (typeof apiKey !== 'string') {
-    throw new Error(
-      'The Makeswift Next.js API handler must be passed a valid Makeswift site API key: ' +
-        "`MakeswiftApiHandler('<makeswift_site_api_key>')`\n" +
-        `Received "${apiKey}" instead.`,
-    )
+function requestHeaders(headers: NextApiRequest['headers']): Headers {
```
arvinpoddar: Should we move this and the other utility functions into a separate file so it's easier to reason about what this file is actually for?

arvinpoddar (dismissed):
This may be my favorite runtime PR ever.

packages/runtime/src/next/api-handler/index.ts
agurtovoy: Good point.

packages/runtime/src/next/api-handler/index.ts
agurtovoy: These have no utility outside the Pages Router handler configuration in this file. That said, this is a relatively large file that mixes App and Pages Router configurations, and your question made me wonder if it might actually make sense to factor these two out into separate files. I'm going to try it out.

agurtovoy: I think this is an improvement:

spit-up-pages-and-app-router-handlers

packages/runtime/src/api-handler/index.ts
migueloller: Should we perhaps factor the code in such a way that it's more obvious to the reader that we're handling CORS preflight requests? Perhaps a comment would be enough if we don't want to extract this into a function?

agurtovoy: Good point, will do.

PR #1031, opened 2025-06-03
fix: only clear typography id when marked for removal

arvinpoddar:
The current behavior for clearing typography data from the copy context is to not return any data altogether. However, since the typography data permits for an undefined ID, we should follow the rule of smallest granularity to only clear the typography ID, rather than the entire data.

https://github.com/user-attachments/assets/c6e5abc6-c371-458f-999e-47dec52508c1

### DIFF
--- a/.changeset/few-hornets-wink.md
+++ b/.changeset/few-hornets-wink.md
@@ -1,0 +1,5 @@
+---
+"@makeswift/controls": patch
+---
+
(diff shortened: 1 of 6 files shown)


agurtovoy (approved):
@arvinpoddar could you clarify what the user-visible effects of this change are, if any?

Codeseph:
After these changes is it still possible to fully discard a typography that's been copied?

PR #1032, opened 2025-06-03
spike: added support for optional descriptions

tymakeswift:


agurtovoy:
@tymakeswift this can be closed, right?

PR #1034, opened 2025-06-04
chore: remove unused useBuilderHandshake

fikrikarim:
The `useBuilderHandshake` was used to trigger the `DraftModeScript`:
1. Runtime would send `makeswift_draft_mode` message to the builder.
2. After receiving the message, the builder then send `makeswift_draft_mode` message back to the host with a secret.
3. The [DraftModeScript](https://github.com/makeswift/makeswift/blob/%40makeswift/runtime%400.23.0/packages/runtime/src/next/draft-mode/draft-mode-script.tsx#L20) then will patch fetch.

Since we don't have `DraftModeScript` anymore, then we don't need the `useBuilderHandshake`.


arvinpoddar (dismissed):
As long as we've thoroughly tested this, this LGTM. Excited to see a free decoupling opportunity. 

Maybe let's keep this change in our pocket for a bit until we starting digging a bit more into our previewing layers of the runtime, just so that we can be 100% confident?

agurtovoy (dismissed):
Even though this doesn't affect any user-facing APIs, I'd add a changeset so it's easier to find in the release notes if we ever need to.

agurtovoy:
@fikrikarim Do you want to add a changeset and merge it? 🙏

PR #1038, opened 2025-06-09
example: Makeswift + Prismic with Blogs & Authors

drewpledger:
This PR implements a blog and author integration with Prismic and Makeswift

Key Features:

Dynamic blog routing using Next.js App Router (/blog/[slug])
Integration with Prismic for content management
Custom blog feed component for displaying posts
Utilizes MakeswiftComponent for dynamic, visual page building
Type-safe content fetching with generated Prismic types through GraphQL codegen


examples/prismic/app/blog/[slug]/page.tsx
```diff
@@ -0,0 +1,47 @@
...
+export default async function Page({ params }: { params: { slug: string } }) {
+  const { slug } = params
+
+  if (!slug) {
+    return notFound()
+  }
+
+  const componentSnapshot = await MakeswiftClient.getComponentSnapshot('blog-post-component', {
+    siteVersion: await getSiteVersion(),
+  })
```
Comment (arvinpoddar) [inline examples/prismic/app/blog/[slug]/page.tsx]: Why is the snapshot id the same for all pages? Is it because the blog data itself changes per slug? Doesn't that mean a change to one blog visually will apply to all of them?

examples/prismic/app/blog/[slug]/page.tsx
```diff
@@ -0,0 +1,47 @@
...
+}
+
+export default async function Page({ params }: { params: { slug: string } }) {
+  const { slug } = params
+
+  if (!slug) {
+    return notFound()
+  }
+
+  const componentSnapshot = await MakeswiftClient.getComponentSnapshot('blog-post-component', {
```
Comment (arvinpoddar) [inline examples/prismic/app/blog/[slug]/page.tsx]: Can we move the `blog-post-component` to a different `const` so we avoid magic strings here? I know you used this pattern in the contentful example.

examples/prismic/components/Prismic/common/PrismicText/PrismicText.makeswift.ts
```diff
@@ -0,0 +1,36 @@
+import { Color, Number, Select, Style } from '@makeswift/runtime/controls'
+
+export const props = {
+	className: Style({
```
arvinpoddar: Formatting: using tabs instead of spaces.

examples/prismic/app/blog/[slug]/page.tsx
Comment (drewpledger) [reply examples/prismic/app/blog/[slug]/page.tsx]: That's the idea!

examples/prismic/.env
```diff
@@ -0,0 +1,3 @@
+MAKESWIFT_SITE_API_KEY=
+NEXT_PUBLIC_PRISMIC_API_TOKEN=
```
agurtovoy, line 2: @drewpledger Is there a need for this to be a client-side accessible token?

examples/prismic/README.md
```diff
@@ -0,0 +1,171 @@
...
+- **Body | Rich text** - The main content as rich text
+- **Hero | Media type** - An optional hero image for the post
+
+Be sure to add a few blog posts for testing purposes, and connect them to an Author
+
+4. Generate Prismic types:
+
+   ```bash
+   pnpm codegen-ts
+   ```
```
agurtovoy, line 81: I followed the instructions above, and I'm getting these errors on the codegen step:

codegen errors

migueloller:
I'm a bit confused here @drewpledger.

> [example: Makeswift + Prismic with Blogs & Authors](https://github.com/makeswift/makeswift/pull/1038#top)

The PR title says Prismic

> This PR implements a blog and author integration with Contentful and Makeswift
>
> Key Features:
>
> Dynamic blog routing using Next.js App Router (/blog/[slug])
> Integration with Dato for content management
> Custom blog feed component for displaying posts
> Utilizes MakeswiftComponent for dynamic, visual page building
> Type-safe content fetching with generated Dato types through GraphQL codegen

The PR description mentions Contentful and Dato.

Which is it? 😕

examples/prismic/.env
```diff
@@ -0,0 +1,3 @@
...
+PRISMIC_REPOSITORY_NAME=
```
migueloller, line 3: Missing a newline at the end of the file

examples/prismic/app/[[...path]]/page.tsx
```diff
@@ -0,0 +1,28 @@
+import { notFound } from 'next/navigation'
+
+import { Page as MakeswiftPage } from '@makeswift/runtime/next'
+import { getSiteVersion } from '@makeswift/runtime/next/server'
+
+import { client } from '@/lib/makeswift/client'
+
+type ParsedUrlQuery = { path?: string[] }
```
Comment (migueloller) [inline examples/prismic/app/[[...path]]/page.tsx:8]: 🍹 I generally try to follow Next.js docs—specially when crafting examples—and in their latest docs they inline the prop types for page components. This avoids having to deal with naming, which feels a bit off to me here since this isn't exactly a "parsed URL query" and instead are matched dynamic route params

examples/prismic/app/blog/[slug]/page.tsx
```diff
@@ -0,0 +1,50 @@
+import { notFound } from 'next/navigation'
+
+import { MakeswiftComponent } from '@makeswift/runtime/next'
+import { getSiteVersion } from '@makeswift/runtime/next/server'
+
+import { BLOG_POST_EMBEDDED_COMPONENT_ID } from '@/components/Builtins/BlogPost/BlogPost.makeswift'
```
Comment (migueloller) [inline examples/prismic/app/blog/[slug]/page.tsx:6]: 🍹 What if instead of exporting the component ID, we exported a server component from `BlogPost.makeswift.ts` that rendered `<MakeswiftComponent>` with the appropriate type applied and fetched its own `snapshot`? We'd only have to pass in the ID

examples/prismic/app/blog/[slug]/page.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { GetBlogDocument } from '@/generated/prismic'
+import { client as MakeswiftClient } from '@/lib/makeswift/client'
+import { client } from '@/lib/prismic/client'
+import { getAllBlogs } from '@/lib/prismic/fetchers'
+import { PrismicProvider } from '@/lib/prismic/provider'
+
+export async function generateStaticParams() {
+  const blogs = await getAllBlogs()
+
+  return blogs.map(blog => ({ slug: blog?._meta?.uid }))
```
Comment (migueloller) [inline examples/prismic/app/blog/[slug]/page.tsx:16]: Shouldn't we be filtering out entries here where `blog?._meta?.uid` is `undefined`? This will otherwise result in a build or runtime error

examples/prismic/app/blog/[slug]/page.tsx
```diff
@@ -0,0 +1,50 @@
...
+}
+
+export default async function Page({ params }: { params: { slug: string } }) {
```
Comment (migueloller) [inline examples/prismic/app/blog/[slug]/page.tsx:19]: Related to an earlier comment, we seem to inline types here and that's yet another reason to inline the types in the top-level catch-all page: consistency

examples/prismic/app/blog/[slug]/page.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const { slug } = params
+
+  if (!slug) {
+    return notFound()
+  }
```
Comment (migueloller) [inline examples/prismic/app/blog/[slug]/page.tsx:24]: This check is unnecessary—`slug` is always defined

examples/prismic/app/blog/[slug]/page.tsx
```diff
@@ -0,0 +1,50 @@
...
+  })
+
+  if (!blogData.blogpost) return notFound()
+
+  return (
+    <PrismicProvider value={blogData.blogpost}>
+      <MakeswiftComponent
+        snapshot={componentSnapshot}
+        label="Blog Post"
+        type={BLOG_POST_EMBEDDED_COMPONENT_ID}
```
Comment (migueloller) [inline examples/prismic/app/blog/[slug]/page.tsx:46]: I think it's confusing that we're using a single value for the component's type and ID. These two things aren't the same and using the same value for it feels off.

examples/prismic/app/blog/[slug]/page.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+export default async function Page({ params }: { params: { slug: string } }) {
+  const { slug } = params
+
+  if (!slug) {
+    return notFound()
+  }
+
+  const componentSnapshot = await MakeswiftClient.getComponentSnapshot(
+    BLOG_POST_EMBEDDED_COMPONENT_ID,
```
Comment (migueloller) [inline examples/prismic/app/blog/[slug]/page.tsx:27]: Shouldn't we be using the slug here so that the builtin element can be different per blog post? Or is the idea that there's a single template for all blog posts?

examples/prismic/app/blog/page.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+import { BLOG_EMBEDDED_COMPONENT_ID } from '@/components/Builtins/Blog/Blog.makeswift'
+import { client as MakeswiftClient } from '@/lib/makeswift/client'
+import { getAllBlogs } from '@/lib/prismic/fetchers'
+import { PrismicProvider } from '@/lib/prismic/provider'
+
+export default async function Page() {
+  const blogs = await getAllBlogs()
+
+  const componentSnapshot = await MakeswiftClient.getComponentSnapshot('blog-component', {
```
migueloller, line 14: 🍹 The ID `blog-component` is a bit confusing as this is an _element_, not a component. Maybe we use `blog-index`?

examples/prismic/app/blog/[slug]/page.tsx
Comment (migueloller) [reply examples/prismic/app/blog/[slug]/page.tsx:46]: We also seem to be inconsistent here as in the blog index we separate the type from the ID but don't do that here. We should do what we do there, here. For example, we can pass in `blog-template` as the key and there's no need for a shared constant.

examples/prismic/app/layout.tsx
```diff
@@ -0,0 +1,31 @@
...
+import '@/lib/makeswift/components'
+import { MakeswiftProvider } from '@/lib/makeswift/provider'
+
+import './globals.css'
+
+const inter = Inter({ subsets: ['latin'] })
+
+export const metadata: Metadata = {
+  title: 'Create Next App',
+  description: 'Generated by create next app',
```
migueloller, line 14: Should we update these to reflect the example?

examples/prismic/app/layout.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+export default async function RootLayout({
+  children,
+}: Readonly<{
+  children: React.ReactNode
+}>) {
+  return (
+    <html lang="en">
+      <body className={inter.className}>
+        <MakeswiftProvider previewMode={(await draftMode()).isEnabled}>
```
migueloller, line 25: @arvinpoddar, please correct me if I'm wrong here, but am I misremembering that we don't need this anymore after we introduced "native" Draft Mode?

examples/prismic/components/Builtins/Blog/Blog.makeswift.ts
```diff
@@ -0,0 +1,16 @@
+import { Slot, Style } from '@makeswift/runtime/controls'
+
+import { runtime } from '@/lib/makeswift/runtime'
+
+import Blog from './Blog'
+
+export const BLOG_EMBEDDED_COMPONENT_ID = 'blog-component' //unique id for the registered component
```
migueloller, line 7: See previous comment about not using the ID for both the type and the snapshot ID. Also, I think that we can keep this type encapsulated here by exporting an RSC that fetches the snapshot:

```tsx
export async function MakeswiftBlog({ id, label = "Blog Post" }: { id: string; label?: string }) {
  const snapshot = await client.getComponentSnapshot(id)

  return <MakeswiftComponent type={TYPE} label={label} snapshot={snapshot} />
}
```

examples/prismic/components/Builtins/Blog/Blog.tsx
```diff
@@ -0,0 +1,9 @@
+'use client'
+
+type Props = {
+  content?: React.ReactNode
+}
+
+export default function Blog({ content }: Props) {
+  return content
```
migueloller, line 8: 🍹 Do we need to wrap this in a React fragment? What if `React.ReactNode` is `undefined`?

examples/prismic/utils/isDevelopment.ts
migueloller, line 1: This file is using tabs instead of spaces

examples/prismic/tailwind.config.js
migueloller, line 1: 🍹 Should we be using Tailwind 4 for these new examples?

examples/prismic/pnpm-lock.yaml
migueloller, line 1: This lockfile seems extraneous since the top-level lockfile was udpated so I'm assuming we're using pnpm workspaces?

examples/prismic/graphql.config.ts
```diff
@@ -0,0 +1,83 @@
...
+  // if there is a leading underscore but it's not in the converted string, add it
+  if (str.indexOf('_') === 0 && result.substring(0, 1) !== '_') {
+    return `_${result}`
+  }
+  return result
+}
+
+async function getConfig() {
+  assert(process.env.PRISMIC_REPOSITORY_NAME, 'PRISMIC_REPOSITORY_NAME is required')
+  assert(process.env.NEXT_PUBLIC_PRISMIC_API_TOKEN, 'NEXT_PUBLIC_PRISMIC_API_TOKEN is required')
```
migueloller, line 25: Why are we doing assertions here when we have the `prismic/env.ts` file?

PR #1043, opened 2025-06-17
add runtime support for component, control, and built-in element descriptions

tymakeswift:


packages/controls/src/controls/checkbox/checkbox.test.types.ts
```diff
@@ -44,6 +45,7 @@ describe('Checkbox Types', () => {
     type ConfigType = typeof def.config
     expectTypeOf<ConfigType>().toEqualTypeOf<{
       label?: string | undefined
+      description?: string | undefined
```
arvinpoddar: You should not need to specify `| undefined` since the property is already optional.

packages/controls/src/controls/color/color.test.types.ts
```diff
@@ -19,6 +19,7 @@ describe('Color Types', () => {
       expectTypeOf<ConfigType>().toEqualTypeOf<{
         label?: string | undefined
         defaultValue?: string | undefined
+        description?: string | undefined
```
arvinpoddar: You should not need to specify `| undefined` since the property is already optional.

packages/controls/src/controls/color/color.test.types.ts
```diff
@@ -43,6 +44,7 @@ describe('Color Types', () => {
       type ConfigType = typeof def.config
       expectTypeOf<ConfigType>().toEqualTypeOf<{
         defaultValue: string
+        description?: string | undefined
```
arvinpoddar: You should not need to specify `| undefined` since the property is already optional.

packages/controls/src/controls/group/group.test.types.ts
```diff
@@ -59,13 +59,15 @@ describe('Group Types', () => {
       type ConfigWithoutType = Omit<Config, 'props'>
       expectTypeOf<ConfigWithoutType>().toEqualTypeOf<{
         readonly label?: string | undefined
+        readonly description?: string | undefined
```
arvinpoddar: You should not need to specify `| undefined` since the property is already optional.

packages/controls/src/controls/group/group.test.types.ts
```diff
@@ -59,13 +59,15 @@ describe('Group Types', () => {
...
         readonly preferredLayout?:
           | typeof Group.Layout.Popover
           | typeof Group.Layout.Inline
       }>()
 
       type ConfigWithoutLabel = Omit<Config, 'label'>
       expectTypeOf<ConfigWithoutLabel>().toEqualTypeOf<{
+        readonly description?: string | undefined
```
arvinpoddar: You should not need to specify `| undefined` since the property is already optional.

packages/controls/src/controls/image/image.ts
```diff
@@ -397,7 +398,7 @@ type UserConfig<F extends Config['format']> = Config & {
 
 type NormedConfig<F extends Config['format']> = undefined extends F
   ? Config
-  : { label?: string | undefined; format: F }
+  : { label?: string | undefined; description?: string | undefined; format: F }
```
arvinpoddar: You should not need to specify `| undefined` since the property is already optional.

packages/controls/src/controls/style/v2/style-v2.test.types.ts
```diff
@@ -22,6 +22,7 @@ describe('StyleV2 Types', () => {
     expectTypeOf<Config>().toEqualTypeOf<{
       type: CheckboxDefinition<{
         defaultValue: boolean
+        description?: string | undefined
```
arvinpoddar: You should not need to specify `| undefined` since the property is already optional.

packages/controls/src/controls/text-input/text-input.test.types.ts
```diff
@@ -25,6 +25,7 @@ describe('TextInput Types', () => {
       type Config = typeof def.config
       expectTypeOf<Config>().toEqualTypeOf<{
         label?: string
+        description?: string | undefined
```
arvinpoddar: You should not need to specify `| undefined` since the property is already optional.

packages/controls/src/controls/text-input/text-input.test.types.ts
```diff
@@ -47,6 +48,7 @@ describe('TextInput Types', () => {
...
```
arvinpoddar: You should not need to specify `| undefined` since the property is already optional.

packages/runtime/src/next/client.ts
```diff
@@ -177,11 +177,13 @@ export function componentDocumentToRootEmbeddedDocument({
   documentKey,
   name,
   type,
+  description,
   meta,
 }: {
   document: MakeswiftComponentDocument | MakeswiftComponentDocumentFallback
   documentKey: string
   name: string
+  description: string | undefined
```
arvinpoddar: We can mark this as optional, rather than explicitly specifying `| undefined`. Try `description?: string`

packages/runtime/src/state/modules/read-only-documents.ts
```diff
@@ -32,6 +32,7 @@ export type EmbeddedDocument = {
   id: string
   type: string
   name: string
+  description?: string
```
arvinpoddar: Most of the root fields here are for identifying the document itself. Let's move this `description` inside of the `meta` objection field (still as an optional string)?

PR #1044, opened 2025-06-18
feat: framework-independent `MakeswiftClient` implementation

agurtovoy:
Extract a framework-independent `MakeswiftClient` implementation and keep the Next.js-specific implementation backward-compatible.


packages/runtime/src/client/index.ts
```diff
@@ -0,0 +1,990 @@
+import { z } from 'zod'
```
agurtovoy, line 1: Mostly moved from `next/client.ts`, leaving out Next-specific parts. See https://github.com/makeswift/makeswift/pull/1044/files#diff-47636c8874de1aaa13ff42ed6ee56694d80e56961280ccd545dbc5cee42ff203

arvinpoddar: As we start pulling out base implementations, what are your thoughts on renaming something like this to `base-client.ts`? One of my personal frustrations working in the runtime is that many files have identical names, so it's difficult to navigate via search. Having a specific name would alleviate this. We could also go the other way and rename the `next/client.ts` to `next-client.ts` (while not changing how it's exported).

Curious to hear what you think.

.changeset/stupid-ducks-appear.md
arvinpoddar, line 1: Nice file name

packages/runtime/package.json
```diff
@@ -38,6 +39,11 @@
       "import": "./dist/esm/box-model.js",
       "types": "./dist/types/box-model.d.ts"
     },
+    "./client": {
+      "require": "./dist/cjs/client/index.cjs",
+      "import": "./dist/esm/client/index.mjs",
+      "types": "./dist/types/client/index.d.ts"
+    },
```
arvinpoddar, line 46: Why do we need a separate entry point here?

packages/runtime/src/client/index.ts
fikrikarim: [Conversation](https://makeswifthq.slack.com/archives/C07DAL1D8JK/p1735861427975999) that might be related.

packages/runtime/package.json
fikrikarim: Same question, do we plan to expose the base `MakeswiftClient` class on `@makeswift/runtime/client`, and other framework will use that?

Also, once we finish refactoring , do we want Next.js users to install both the `@makeswift/runtime` and `@makeswift/next`, or do they only need to install `@makeswift/next`?

If we require them to install both `@makeswift/runtime` and `@makeswift/next`, they might accidentally import the wrong `MakeswiftClient` from the `@makeswift/runtime/client`. This is not an issue if Next.js users only need to install one package `@makeswift/next`.

agurtovoy: > Why do we need a separate entry point here?

We need to expose `MakeswiftClient` from _some_ entry point to be able to build PoCs for other frameworks. I didn’t want to expose it from `@makeswift/runtime` to avoid accidental misuse, and none of the existing entry points seemed appropriate.

Overall, I’m not concerned about adding a new entry point, since we’ll be reshuffling them once we're ready to separate out the Next.js-specific layer.

agurtovoy: > do we plan to expose the base MakeswiftClient class on @makeswift/runtime/client, and other framework will use that?

Yes.

> Also, once we finish refactoring , do we want Next.js users to install both the @makeswift/runtime and @makeswift/next, or do they only need to install @makeswift/next?

They would only need to install `@makeswift/next`, just as they currently only need to install `@makeswift/runtime`.

packages/runtime/src/client/index.ts
agurtovoy: > [Conversation](https://makeswifthq.slack.com/archives/C07DAL1D8JK/p1735861427975999) that might be related.

Yes, thanks for linking that @fikrikarim. Also, https://github.com/bigcommerce/catalyst/pull/2418.

@arvinpoddar I personally prefer the naming conventions used in Catalyst, where we use simple, consistent file names that make sense within the immediate context of the component or directory, without worrying about keeping them unique across the entire repo. It does take time to get used to prefixing the file with the directory name when navigating in the IDE, and there is a cost even when you master that, but I find the toll of coming up with and maintaining unique, ad-hoc, or overly verbose names to be worse.

PR #1049, opened 2025-07-10
fix: support preflight requests in app router

arvinpoddar:
Update the API handler to support OPTIONS requests


packages/runtime/src/next/api-handler/index.ts
```diff
@@ -92,6 +92,18 @@ export function MakeswiftApiHandler(
   ): Promise<NextResponse<MakeswiftApiHandlerResponse> | void> {
     return match(args)
       .with(routeHandlerPattern, async args => {
+        const [req] = args
+        if (req.method === 'OPTIONS') {
+          return new NextResponse(null, {
+            status: 200,
```
agurtovoy: Funny enough, I questioned the lack of OPTIONS request handling here in my API handlers re-write, and ended up [adding it for both App and Pages Router](https://github.com/makeswift/makeswift/pull/1030/files#r2198518869), but it didn't occur to me that this might be the root cause when we were looking into this yesterday :(

agurtovoy: I think 204 (No Content) is what's technically expected here/is the de-facto standard in CORS middleware:

- https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request
- https://github.com/expressjs/cors/blob/2c44de3296d8aaf605315d48f972e73f467b138d/lib/index.js#L12
- https://github.com/tinyhttp/cors/blob/76b709e0c7350fa126c1c7a5fb4c431bb6a5254a/src/index.ts#L88

... together with `'Content-Length: 0` workaround for Safari: https://github.com/expressjs/cors/blob/2c44de3296d8aaf605315d48f972e73f467b138d/lib/index.js#L179

agurtovoy (approved):
🙏

PR #1050, opened 2025-07-10
Preview URL support

arvinpoddar:
Adds support for preview URLs to the runtime


packages/runtime/src/client/index.ts
```diff
@@ -365,19 +381,19 @@ export class MakeswiftClient {
...
+    siteVersion = null,
     ...params
   }: {
-    siteVersion?: MakeswiftSiteVersion
+    siteVersion?: SiteVersion | null
   } & GetPagesParams = {}): Promise<MakeswiftGetPagesResult> => {
     const queryParams = getPagesQueryParams(params)
 
-    const response = await this.fetch(`v4/pages?${queryParams.toString()}`, siteVersion)
+    const response = await this.fetch(`v5_unstable/pages?${queryParams.toString()}`, siteVersion)
```
agurtovoy: Are we still replacing `vX_unstable` with `vX` before merging?

packages/runtime/src/next/api-handler/config/app-router.ts
```diff
@@ -18,19 +19,19 @@ export const argsPattern = [P.instanceOf(Request), P.any] as const
 export async function config({
   req,
   context,
-  apiKey,
+  apiKey: _apiKey,
```
agurtovoy: I think we can remove the `apiKey` from the internal config params since it's been replaced by the `client`.

packages/runtime/src/next/api-handler/draft.ts
agurtovoy, line 1: Since we're converging on "preview" as the unifying terminology, I'd rename this file to `preview.ts` or `preview-mode.ts`.

packages/runtime/src/next/draft-mode/index.tsx
agurtovoy, line 1: I wonder if something like `app-router-site-version.ts` would be a more appropriate file name here.

packages/runtime/src/next/api-handler/handlers/redirect-draft.test.ts
agurtovoy, line 1: I'd rename this test to match the file it's testing: `app-router-redirect-preview.test.ts`.

packages/runtime/src/next/tests/server.api-handler-clear-draft.test.ts
agurtovoy, line 1: Should we rename the `clear-draft` endpoint `exit-preview`?

packages/runtime/src/runtimes/react/components/draft-switcher/draft-toolbar.tsx
agurtovoy, line 1: And rename this one `preview-toolbar.tsx`?

agurtovoy (approved):
Love the further convergence between the app and pages routers and overall code simplification here. See inline comments for some naming suggestions.

packages/runtime/src/client/index.ts
arvinpoddar: Yes, this will be the final thing changed in this PR prior to merging.

.changeset/ninety-lamps-act.md
```diff
@@ -0,0 +1,96 @@
...
+
+The `<ReactRuntimeProvider>`'s `previewMode` prop has been replaced with the
+required `siteVersion` prop. The value passed for this prop should be set
+by the same `getSiteVersion` function.
+
+#### App Router
+
+If you're using App Router, use the `getSiteVersion` function to get the value
+to be passed to the `siteVersion` prop of `<ReactRuntimeProvider>`. If you
+followed our installation guide and created a `<MakeswiftProvider>` component,
```
fikrikarim: Do we also want to add the diff for `<MakeswiftProvider>` here @arvinpoddar? I was confused that we show the diff for the root layout but not the `<MakeswiftProvider>` itself.

image

arvinpoddar: Added

packages/runtime/src/client/index.ts
arvinpoddar: Done

PR #1054, opened 2025-07-13
test: fix MSW warnings about redundant usage of query parameters

agurtovoy:


packages/runtime/src/client/tests/client.get-component-snapshot.test.ts
```diff
@@ -93,19 +93,18 @@ describe('getComponentSnapshot using v1 element tree endpoint', () => {
             - this is the step that we're really testing here - we want the logic in getComponentSnapshot to execute this subsequent request
         (3) graphql query for introspection
     */
+      const httpHandler = jest.fn(({ request }) => {
```
agurtovoy, line 96: Here and below, follow MSW recommendations to use explicit request URL parameters parsing: https://mswjs.io/docs/http/intercepting-requests/query-parameters.

Wrapping the handler in `jest.fn` so we can assert on the number of calls and/or parameters.

packages/runtime/src/client/tests/client.get-component-snapshot.test.ts
```diff
@@ -156,15 +158,18 @@ describe('getComponentSnapshot using v1 element tree endpoint', () => {
...
     server.use(
-      http.get(
-        `${baseUrl}/${treeId}?locale=${localeToTest}`,
-        () => HttpResponse.text('', { status: 404 }),
-        { once: true },
-      ),
-      http.get(`${baseUrl}/${treeId}`, () => HttpResponse.json(document, { status: 200 }), {
-        once: true,
-      }),
+      http.get(`${baseUrl}/${encodeURIComponent(treeId)}`, httpHandler),
```
agurtovoy, line 172: Added `encodeURIComponent` here and below.

PR #1055, opened 2025-07-13
fix: add `OPTIONS` request hanlding to all of the App Router examples

agurtovoy:
Following up https://github.com/makeswift/makeswift/pull/1049


arvinpoddar:
🙏

arvinpoddar:
Wait a sec, don't we have to update the runtime version in each of these examples?

PR #1048, opened 2025-07-06
example: Contentful without granular GraphQL control on Entry Type components.

drewpledger:
Following discussions with James and Lindsay, we've redesigned our Makeswift + Contentful integration to prioritize ease of adoption over flexibility. This new example trades the granular GraphQL control of our previous implementation for a broad, approachable solution that empowers developers to quickly build production-ready blog websites.

🎯 What This Solves
Before: Complex integration requiring deep GraphQL knowledge, intimidating for newcomers
After: Intuitive, comprehensive example that developers can understand and extend within minutes

✨ Key Features

🏗️ Complete Blog System
Optionally dynamic blog post generation with unique Makeswift layouts per post
Automated content fetching from Contentful with type-safe GraphQL
SEO-optimized static generation with generateStaticParams

🎨 Visual Page Building
Unique Component Snapshots: Each blog post gets its own editable layout (blog-content-with-slot-${slug})
Drag-and-drop content sections below blog posts
Live preview in Makeswift builder with real Contentful data

🚀 Production-Ready Architecture
Modern Next.js App Router with Server Components
Automatic TypeScript generation from Contentful schema
Environment variable validation with @t3-oss/env-nextjs
Pre-built UI components from VIBES for consistent design

📦 Plug-and-Play Setup
Single command installation: `npx makeswift@latest init --example=contentful-simple`
Clear content model definitions (Author + BlogPost)

🎭 Developer Experience
Zero Configuration: Works out of the box with minimal setup
Clear Documentation: Step-by-step guides with code examples
Modern Stack: Latest Next.js, TypeScript, and VIBES

🎯 Target Audience
This example is designed for developers who want to:
Build content-driven websites quickly
Combine visual editing with headless CMS benefits
Learn Makeswift + Contentful integration patterns

PR #1059, opened 2025-07-21
example: Strapi without granular GraphQL control on Entry Components

drewpledger:
This PR reflects the patterns established in this PR (https://github.com/makeswift/makeswift/pull/1048)

### DIFF
--- a/examples/strapi-simple/README.md
+++ b/examples/strapi-simple/README.md
@@ -32,21 +32,15 @@ strapi-simple/
 
 ## Quick Start
 
-### 1. Clone the repository
+### 1. Set up Strapi locally
 
-```bash
-   npx makeswift@latest init --example=strapi-simple
-```
-
-### 2. Set up Strapi locally
-
-Before jumping into Strapi with Makeswift, you'll need to create a new Strapi project. This will serve as your content management system, where you'll create and manage blog posts and authors.
+Before integrating with Makeswift, you'll need to create a new Strapi project. This will serve as your content management system, where you'll create and manage blog posts and authors.
 
 #### Install Strapi Server
 
 Before you begin, make sure you have Node.js version 20.19 or higher installed.
 
-For this guide, we’ll use the Strapi server as a standalone setup.
+For this guide, we'll use the Strapi server as a standalone setup.
 
 In the desired parent directory, run the following command to create your Strapi project:
 
@@ -56,7 +50,18 @@ npx create-strapi-app@latest my-strapi-project
 
 You can read the full guide on Strapi [here](https://docs.strapi.io/cms/quick-start)
 
-Navigate to the Strapi project directory and start the Strapi development server:
+#### Install the GraphQL plugin
+
+Strapi's GraphQL plugin is required for this integration. Before starting the Strapi server, install the plugin:
+
+```bash
+cd my-strapi-project
+npm install @strapi/plugin-graphql
+```
+
+#### Start the Strapi development server
+
+Now start the Strapi development server:
 
 ```bash
 npm run develop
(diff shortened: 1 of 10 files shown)


examples/strapi-simple/README.md
```diff
@@ -0,0 +1,206 @@
...
+
+```bash
+npm run dev
+```
+
+Your site will be available at `http://localhost:3000`. Ensure the strapi server continues running.
+
+## Content Model
+
+The Strapi project comes pre-configured with all the necessary content types for a blog:
```
Codeseph: Hey Drew - I've either missed something or the guide might be missing something. But I've reached this point (having followed the [Strapi Quick Start Guide](https://docs.strapi.io/cms/quick-start)) and do not have these blog types. The guide led me to having the following types

image

examples/strapi-simple/README.md
```diff
@@ -0,0 +1,206 @@
...
+STRAPI_ACCESS_TOKEN=your_strapi_api_token_from_step_3
+NEXT_PUBLIC_STRAPI_SERVER_URL=http://localhost:1337
+```
+
+### 5. Generate TypeScript types
+
+Install dependencies and generate TypeScript types from your Strapi GraphQL schema:
+
+```bash
+npm install
```
Codeseph: Are you using npm or pnpm? 

If I try to actually use npm, I get an error from the blocks-react-renderer package trying to run a postinstall script. Using pnpm works, apparently because it is more reluctant to run postinstall scripts. I don't know much about this. Anyway here's a relevant issue from their GitHub: https://github.com/strapi/blocks-react-renderer/issues/30

examples/strapi-simple/README.md
```diff
@@ -0,0 +1,206 @@
...
+
+### Content Formatting
+
+Blog posts are transformed from Strapi's structure to match the API expected by the VIBES components using utility functions in [`lib/strapi/format.ts`](lib/strapi/format.ts).
+
+## Troubleshooting
+
+Common issues and solutions:
+
+- **Type generation fails**
```
Codeseph, line 238: Maybe people coming here from a Strapi background would know better, but for me the flow of trying to follow this guide was such that I didn't learn I needed the GraphQL plugin for Strapi until failing to follow the early sections of the guide (due to codegen errors), being confused for a little while, then happening upon this section here while scrolling

examples/strapi-simple/README.md
```diff
@@ -0,0 +1,206 @@
...
+  - Check that your Strapi access token is correct in `.env.local`.
+  - Verify that the GraphQL plugin is enabled in Strapi.
+  - Ensure that the relevant content is published in Strapi.
+
+- **Blog posts not showing**
+  - Confirm that the slug format matches your entries in Strapi.
+  - Make sure the posts are published.
+  - Ensure required fields—**slug**, **cover**, and **blocks**—are populated.
+
+- **Makeswift builder issues**
```
Codeseph, line 252: Is it normal for us to recommend clearing the browser cache? This seems extreme to me for what I imagine most people would categorize as "builder issues". Off the top of my head I actually can't think of a builder-related issue that I've encountered where the solution was to clear my cache. Just curious, but I feel like clearing the cache is more of an "if all else fails" kind of thing, since most issues that people hit are probably going to be related to their host configuration

examples/strapi-simple/README.md
drewpledger: I just experimented with saying Y/N on the prompt for pre-populated data, here are the results:

CleanShot 2025-10-10 at 15 10 59
CleanShot 2025-10-10 at 15 19 12

Interesting that you got Restaurant.. did you remember what you said to that prompt?

drewpledger: Just gonna rework the readme as if we're building from scratch 👍

Codeseph: ah interesting. As for how I got Restaurant - it gets created if you follow the quick start guide through section B, which I did

image

examples/strapi-simple/README.md
```diff
@@ -0,0 +1,275 @@
...
+This will:
+
+- Start Strapi on `http://localhost:1337`
+- Open the admin panel in your browser
+
+#### Install the GraphQL plugin
+
+Strapi's GraphQL plugin is required for this integration. To enable it:
+
+1. Stop the Strapi development server if it's running (Ctrl+C)
```
Codeseph: Potential flow issue - it seems we finish the 'Install Strapi Server' section above by asking them to run the local Strapi server, then immediately ask them to stop the Strapi server to install the GraphQL plugin

examples/strapi-simple/README.md
```diff
@@ -0,0 +1,275 @@
...
+
+During installation, you'll be prompted to provide the following environment variables. The installer will automatically create a `.env.local` file with your values.
+
+#### Where to find these values:
+
+2. **STRAPI_ACCESS_TOKEN**: The API token you created in Step 1
+
+   - Use the token you just generated and copied
+   - If you need to retrieve it again, go to **Settings** → **API Tokens** in your Strapi admin panel
+   - Note: For security reasons, the token is only shown once when created
```
Codeseph: While this is commonly done, it doesn't seem to be the case here - I seem to be able to revisit a token that was previously created and click 'View token' whenever I want

I'm guessing you added line 107 and meant to delete 108

examples/strapi-simple/README.md
```diff
@@ -0,0 +1,275 @@
...
+
+```bash
+npx makeswift@latest init --example=strapi-simple
+```
+
+During installation, you'll be prompted to provide the following environment variables. The installer will automatically create a `.env.local` file with your values.
+
+#### Where to find these values:
+
+2. **STRAPI_ACCESS_TOKEN**: The API token you created in Step 1
```
Codeseph: This numbered list starts at 2. Presumably there was an entry in the list for the `MAKESWIFT_SITE_API_KEY`. I imagine we want to add that entry back (even if that value is already set in .env.local when generated), probably with some mention of where it can be found in the builder

examples/strapi-simple/README.md
```diff
@@ -0,0 +1,275 @@
...
+3. **NEXT_PUBLIC_STRAPI_SERVER_URL**: The URL where your Strapi server is running
+
+   - Use: `http://localhost:1337` for local development
+   - For production, use your deployed Strapi URL
+
+4. **NEXT_PUBLIC_SITE_URL**: The URL where your Next.js site will be running
+   - Use: `http://localhost:3000` for local development
+   - For production, use your deployed Next.js URL
+
+Your site will be available at `http://localhost:3000`. Ensure the Strapi server continues running.
```
Codeseph: I wonder if it's worth explicitly mentioning the run command for the host, since this is something we do for the Strapi server under the 'Install Strapi Server' section

examples/strapi-simple/README.md
```diff
@@ -0,0 +1,275 @@
...
+
+Your content is now accessible via the GraphQL API!
+
+## Building Blog Pages in Makeswift
+
+This integration includes a dynamic `/blog/[slug]` route that automatically retrieves and renders individual blog posts from Strapi based on their slug. You don’t need to manually add or publish each blog post in Makeswift—once a post is published in Strapi, it will automatically appear on your site.
+
+By default, the blog post layout includes the essential elements you'd expect: title, date, author, hero image, and body content. However, the integration also allows you to visually add custom sections in Makeswift below the blog content and apply them to every blog page:
+
+1. In the [Makeswift builder](https://docs.makeswift.com/product/builder-basics), navigate to your any of your blog post pages by entering the post's URL (e.g., `/blog/welcome`) in the builder's URL bar.
```
Codeseph: accidental 'navigate to your any of your' instead of 'navigate to any of your'

What we're doing here is editing the path in the builder (you cannot edit the full URL in the builder, since the scheme/hostname are locked to the custom host value given in your site settings) to contain the **slug** of a blog post that we want to visit. I think we should be more explicit about this, versus saying "entering the post's URL"

examples/strapi-simple/README.md
```diff
@@ -0,0 +1,275 @@
...
+
+This project uses pre-built components from [VIBES](https://vibes.site/) for consistent, modern UI. See the [`vibes/README.md`](vibes/README.md) for more details.
+
+- **BlogPostList** – Displays a grid of blog post cards.
+- **BlogPostContent** – Renders individual blog post content with breadcrumbs.
+- **SectionLayout** – Provides consistent spacing and layout structure.
+
+### Content Formatting
+
+Blog posts are transformed from Strapi's structure to match the API expected by the VIBES components using utility functions in [`lib/strapi/format.ts`](lib/strapi/format.ts).
```
Codeseph, line 232: I see in `format.ts` that there are certain fields which are required (slug and banner URL) for the VIBES component. But we didn't mention during that the 'Setting Up Content Types' section that these fields should be created as required fields in the Strapi Content-Type Builder.

This seems especially problematic for the blog feed component, since as things stand a single blog post which is missing one of these required fields would result in the entire feed blowing up

PR #1061, opened 2025-07-22
Necessary dependencies for Vercel builds added to examples + ContentfulRichText fix

drewpledger:
During my deployment for these examples using Vercel, I encountered build errors involving these dependencies. I also discovered a bug with the ContentfulRichText component, which is addressed.


pnpm-lock.yaml
```diff
@@ -119,6 +119,378 @@ importers:
         specifier: ^5
         version: 5.1.6
 
+  examples/contentful-simple:
```
agurtovoy: I think we should develop our examples using `npm` so that (a) our development environment more closely matches the isolated environment users will experience when trying them out, and (b) we avoid polluting the repo’s `pnpm-lock`.

As far as this PR is concerned, it means dropping the changes to `pnpm-lock.yaml`.

agurtovoy (approved):
Looks good, but see my comment on `pnpm-lock.yaml`.

PR #1062, opened 2025-07-25
tests: API handlers

agurtovoy:
Tests for all `api/makeswift/` routes. No semantics changes. Prep for https://github.com/makeswift/makeswift/pull/1030.


packages/runtime/src/client/tests/client.get-component-snapshot.test.ts
```diff
@@ -15,7 +15,15 @@ function createTestClient() {
...
 }
 
+afterEach(() => {
+  jest.resetAllMocks()
+})
+
 describe('getComponentSnapshot using v1 element tree endpoint', () => {
+  beforeEach(() => {
+    jest.spyOn(console, 'error').mockImplementation(() => {})
+  })
```
agurtovoy, line 25: I should have put this in a separate PR, this suppresses a `console.error` message in the `getComponentSnapshot` test output.

packages/runtime/src/jest-setup.ts
```diff
@@ -42,3 +42,5 @@ if (typeof window !== 'undefined') {
     })),
   })
 }
+
+;(global as any).PACKAGE_VERSION = require('../package.json').version
```
agurtovoy, line 46: `PACKAGE_VERSION` is never defined, but is substituted at build time by `tsup`: https://github.com/makeswift/makeswift/blob/2135f656acd1b7c19a7c195f270785bb6d41bfc6/packages/runtime/tsup.config.ts#L18

Define it as a global constant for tests.

PR #1064, opened 2025-07-25
tests: uid mocks return a distinct id for every call within a test

agurtovoy:


packages/runtime/src/jest-setup.ts
```diff
@@ -12,9 +12,10 @@ if (typeof window !== 'undefined') {
   expect.addSnapshotSerializer(createSerializer())
 }
 
+let uidSuffix = 100000000000
 jest.mock('uuid', () => ({
-  v4: jest.fn(() => 'xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx'),
-  v5: jest.fn(() => 'xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx'),
+  v4: jest.fn(() => `xxxxxxxx-xxxx-xxxx-xxxx-${uidSuffix++}`),
+  v5: jest.fn(() => `xxxxxxxx-xxxx-xxxx-xxxx-${uidSuffix++}`),
```
agurtovoy, line 18: This also brings these mocks in sync with uid mocks in the `controls` package: https://github.com/makeswift/makeswift/blob/main/packages/controls/src/jest-setup.ts

arvinpoddar (approved):

PR #1069, opened 2025-07-30
support preview token URLs

arvinpoddar:
Updates the runtime to support displaying new site versions via preview tokens.


.changeset/ninety-lamps-act.md
arvinpoddar, line 1: Very much a WIP. Will be corrected once the details of this PR have been ironed out.

packages/runtime/src/client/index.ts
```diff
@@ -943,4 +965,42 @@ export class MakeswiftClient {
   mergeTranslatedData(elementTree: ElementData, translatedData: Record<string, Data>): Element {
     return this.runtime.mergeTranslatedData(elementTree, translatedData)
   }
+
+  async readPreviewToken(token: string): Promise<PreviewTokenPayload | null> {
```
arvinpoddar, line 969: New client method that verifies a given preview token

packages/runtime/src/client/index.ts
```diff
@@ -361,19 +377,19 @@ export class MakeswiftClient {
...
+    siteVersion = null,
     ...params
   }: {
-    siteVersion?: MakeswiftSiteVersion
+    siteVersion?: MakeswiftVersionData | null
   } & GetPagesParams = {}): Promise<MakeswiftGetPagesResult> => {
     const queryParams = getPagesQueryParams(params)
 
-    const response = await this.fetch(`v4/pages?${queryParams.toString()}`, siteVersion)
+    const response = await this.fetch(`v5_unstable/pages?${queryParams.toString()}`, siteVersion)
```
arvinpoddar, line 392: Here and below, many of our resource retrieval methods are now using a `vX_unstable` version of the endpoint. These will be adjusted prior to merging to `main` to not have the `unstable` prefix.

apps/nextjs-pages-router/src/pages/[[...path]].tsx
```diff
@@ -43,17 +44,15 @@ export async function getStaticProps({
   GetStaticPropsResult<PageProps>
 > {
   const path = '/' + (params?.path ?? []).join('/')
-  const snapshot = await client.getPageSnapshot(path, {
-    siteVersion: Makeswift.getSiteVersion(previewData),
-    locale,
-  })
+  const siteVersion = Makeswift.getSiteVersion(previewData)
```
Comment (arvinpoddar) [inline apps/nextjs-pages-router/src/pages/[[...path]].tsx:47]: `getSiteVersion` now returns an object containing a `token` and a `version`, rather than just `Live` or `Working`.

packages/makeswift-next-plugin/index.js
```diff
@@ -54,23 +54,11 @@ module.exports =
               has: [
                 {
                   type: 'query',
-                  key: 'x-makeswift-draft-mode',
-                  value: '(?<secret>.+)',
+                  key: 'x-makeswift-preview-token',
```
arvinpoddar, line 57: Only one query param. No distinction between URLs used by app router and pages router.

packages/runtime/src/client/index.ts
```diff
@@ -344,14 +352,22 @@ export class MakeswiftClient {
...
+    const requestUrl = new URL(path, this.apiOrigin)
+
+    const requestHeaders = new Headers({
+      'X-API-Key': this.apiKey,
+      'Makeswift-Site-API-Key': this.apiKey,
+    })
+
+    if (siteVersion?.token) {
+      requestUrl.searchParams.set('version', siteVersion.version)
+      requestHeaders.set('makeswift-preview-token', siteVersion.token)
```
arvinpoddar, line 365: The Host API expects the `version` via a query param. The token is added as a separate header. If the token does not authorize access to the version in the query param, then the endpoint will return a 401.

.changeset/ninety-lamps-act.md
```diff
@@ -0,0 +1,42 @@
+---
+'@makeswift/next-plugin': minor
+'@makeswift/runtime': minor
+---
+
+Adds support for viewing various site versions with preview tokens.
+
+## Breaking Changes
+
+We've fully deprecated `MakeswiftSiteVersion`. All APIs that accept a
```
agurtovoy: I think "deprecated" means that API is still available but is discouraged. If we are removing/replacing it, I think we have to say that explicitly.

apps/nextjs-app-router/src/makeswift/provider.tsx
```diff
@@ -6,22 +6,23 @@ import { runtime } from '@/makeswift/runtime'
 import {
   ReactRuntimeProvider,
   RootStyleRegistry,
+  type MakeswiftVersionData,
```
agurtovoy: Right now, we have what’s arguably an inconsistent naming convention: some of our public symbols are prefixed with `Makeswift`, while others are not, without a discernible rationale.

We're not unique in this; for example, [Supabase JS SDK](https://github.com/supabase/supabase-js/blob/master/src/index.ts) has the same, although less pronounced, problem. 

That said, many high-quality SDKs out there *are* consistent, do not prefix their symbols with the namespace, and let users handle it in on import as they prefer ([Sanity JS SDK](https://github.com/sanity-io/client/blob/main/src/index.ts), [Notion JS SDK](https://github.com/makenotion/notion-sdk-js/blob/main/src/index.ts), [Slack Web API JS SDK](https://github.com/slackapi/node-slack-sdk/blob/main/packages/web-api/src/index.ts)). I'd love for us to adopt/move towards that standard. 

With that in mind, I suggest we rename `MakeswiftVersionData` to just `SiteVersion`.

packages/runtime/src/api-handler/index.ts
```diff
@@ -98,7 +98,14 @@ export function createApiHandler(
   }
 
   async function apiRouteHandler(req: ApiRequest, route: string): Promise<ResponseType> {
-    const siteVersion = parseSiteVersion(req.headers.get(API_HANDLER_SITE_VERSION_HEADER))
+    const versionHeader = req.headers.get(ApiHandlerHeaders.SiteVersion)
+    const tokenHeader = req.headers.get(ApiHandlerHeaders.PreviewToken)
+
+    const siteVersion: MakeswiftVersionData | null =
+      versionHeader != null && tokenHeader != null
+        ? { version: versionHeader, token: tokenHeader }
```
agurtovoy: Do we want to parse/validate these headers?

packages/runtime/src/next/api-handler/config/app-router.ts
```diff
@@ -38,8 +41,8 @@ export async function config({
       }
     },
     customRoutes: async (route: string) => {
-      if (route === '/draft') {
-        return { res: await redirectDraftHandler(req, context, { apiKey }) }
+      if (route === '/redirect-draft') {
```
agurtovoy: Are intentionally changing the route here?

agurtovoy: Ah, never mind, this is the downstream of this change: https://github.com/makeswift/makeswift/pull/1069/files#diff-19f9b2611ef669911a4219dfef480818b939bbfdb19116162d7fdad0f6aaee79R61

packages/runtime/src/next/middleware/request-utils.ts
```diff
@@ -1,30 +1,28 @@
...
+}
 
-function getDraftModeSecret(request: NextRequest): string | null {
+export function hasLegacyDraftModeCookies(request: NextRequest): boolean {
   return (
-    request.nextUrl.searchParams.get(SearchParams.DraftMode) ??
-    request.headers.get(HeaderNames.DraftMode) ??
-    null
+    request.cookies.has(PRERENDER_BYPASS_COOKIE) &&
+    request.cookies.has(MAKESWIFT_VERSION_DATA_COOKIE)
```
agurtovoy: Is this supposed to check for `MAKESWIFT_DRAFT_DATA_COOKIE`? If not, why is called "legacy"?

packages/runtime/src/next/tests/server.api-handler-redirect-draft.test.ts
```diff
@@ -42,43 +52,46 @@ describe('MakeswiftApiHandler', () => {
...
           get: jest.fn(() => null),
         } as any)
 
+        server.use(
+          http.post(
+            `${TestOrigins.apiOrigin}/${READ_PREVIEW_TOKEN_PATH}`,
+            () => HttpResponse.json(SUCCESS_TOKEN_VERIFICATION_RESPONSE, { status: 200 }),
+            { once: true },
+          ),
+        )
```
agurtovoy: Should we move this to a `beforeEach` block?

agurtovoy (approved):

packages/runtime/src/next/tests/server.api-handler-redirect-draft.test.ts
arvinpoddar: I decided not to because it forces the test writer to be explicit in each test about whether the call succeeds or fails. Do you have a strong feeling about sharing the behavior across tests?

apps/nextjs-app-router/src/makeswift/provider.tsx
arvinpoddar: I'm fine with that! I was not a fan of `MakeswiftVersionData`, so thanks for the alternative.

packages/runtime/src/api-handler/index.ts
arvinpoddar: Can you elaborate on what you mean? Are we parsing/validating that they're strings? Or do you mean parsing/validating that they're valid version refs/tokens?

packages/runtime/src/next/middleware/request-utils.ts
arvinpoddar: This is a mistake on my end, I did not mean to rename this method. Good catch.

.changeset/ninety-lamps-act.md
arvinpoddar: Thank you, yes you're correct.

packages/runtime/src/api-handler/index.ts
agurtovoy: The latter. They will always be strings, but I was just pondering whether it's desirable/feasible to validate them beyond that, since the current code does validate the version string. And if we don't validate them here, at what point do they get validated/how do we handle incorrect values?

Currently, an incorrect version is silently treated as Live.

packages/runtime/src/next/tests/server.api-handler-redirect-draft.test.ts
agurtovoy: No, I was just wondering if this was boilerplate that was the same across all tests. I see now that it's not 🙏.

packages/runtime/src/api-handler/index.ts
arvinpoddar: Sadly, the only true way to validate them is via network request to our token verification endpoint, which I don't consider an option here. The other options are bring in schemas for our refs/JWTs to the runtime, which I also would like to avoid since I want these values to be largely opaque to the runtime.

Perhaps what we do is the same technique we use for serializing/deserializing the data from the draft mode cookie, and stick everything into one header. How does that sound?

arvinpoddar: Discussed offline: since this is an internal interface, we're fine moving forward with the approach of putting everything into one header. To further emphasize that this internal (and to more philosophically align with HTTP headers only containing one string, rather than smashing multiple strings together in a serialized object), we're also Base64 encoding the serialized `SiteVersion`.

PR #1073, opened 2025-08-06
fix: "Do not add <script> tags using next/head" warning in Pages Router

agurtovoy:


packages/runtime/src/next/components/framework-provider/app-router/HeadSnippet.tsx
```diff
@@ -5,19 +5,19 @@ import { type Snippet } from '../../../../client'
 
 import {
   BaseHeadSnippet,
-  snippetToElement,
+  snippetToElements,
```
agurtovoy, line 8: Here and below, renamed `snippetToElement` to `snippetToElements` to reflect its actual return type (`(string | JSX.Element)[]`)

packages/runtime/src/next/components/framework-provider/pages-router/HeadSnippet.tsx
```diff
@@ -19,7 +24,15 @@ export function HeadSnippet({ snippet }: { snippet: Snippet }) {
     }
   }, [snippet])
 
-  const headSnippetElement = snippetToElement(snippet)
+  const headSnippetElements = snippetToElements(snippet)
+  const [scripts, snippets] = partition(headSnippetElements, isScriptElement)
```
agurtovoy: Separate `<script>` tags from everything else.

packages/runtime/src/next/components/framework-provider/pages-router/HeadSnippet.tsx
```diff
@@ -19,7 +24,15 @@ export function HeadSnippet({ snippet }: { snippet: Snippet }) {
...
 
-  return <Head>{headSnippetElement}</Head>
+  return (
+    <>
+      <Head>{snippets}</Head>
+      {scripts.map(({ key, props }) => (
+        <Script key={key} {...props} strategy="beforeInteractive" />
+      ))}
+    </>
```
agurtovoy, line 37: I also considered and tried a more complex implementation below, thinking that it would preserve the relative order of elements within the snippet, but in practice it turned out to produce the exact same HTML, so I went with a simpler/cleaner implementation.

```tsx
  return (
    <>
      {headSnippetElements.map((element, i) =>
        isScriptElement(element) ? (
          <Script key={element.key} {...element.props} strategy="beforeInteractive" />
        ) : (
          <Head key={isElement(element) ? element.key : i}>{element}</Head>
        ),
      )}
    </>
  )
```

packages/runtime/src/next/components/framework-provider/pages-router/HeadSnippet.tsx
```diff
@@ -1,15 +1,20 @@
...
 
 import {
-  snippetToElement,
+  snippetToElements,
   getSnippetElementsFromDOM,
   cleanUpSnippet,
 } from '../../../../runtimes/react/components/page/HeadSnippet'
 
+const isScriptElement = (element: string | JSX.Element): element is ReactElement<any, 'script'> =>
+  element instanceof Object && element.type === 'script'
```
migueloller: Should we use [`isValidElement`](https://react.dev/reference/react/isValidElement) here?

migueloller (dismissed):
Looks good 👍🏻 

Do we want to invest a bit on tests here?

packages/runtime/src/next/components/framework-provider/pages-router/HeadSnippet.tsx
agurtovoy: Good point!

arvinpoddar (dismissed):

packages/runtime/src/next/components/framework-provider/pages-router/HeadSnippet.tsx
agurtovoy: Went back to the more complex but conceptually correct implementation for the sake of tests.

agurtovoy:
> Do we want to invest a bit on tests here?

We don't have the infrastructure for proper App/Pages Router end-to-end tests in the runtime, but partial tests are better than none, so I added some. Thanks for the prompt!

migueloller:
I appreciate that! 🙏 

Seems like something worth investing into either during cooldown or as a project we bet on…

PR #1075, opened 2025-08-07
chore: decouple `Slot`/`MakeswiftComponent` from Next.js

agurtovoy:
They weren't really coupled to Next.js in the first place, so mostly just moving them under `runtimes/react`. No user-visible changes.

### DIFF
--- a/packages/runtime/src/runtimes/react/components/Slot.tsx
+++ b/packages/runtime/src/runtimes/react/components/Slot.tsx
@@ -1,10 +1,11 @@
 'use client'
 
 import { ReactNode, useMemo } from 'react'
-import { MakeswiftComponentSnapshot } from '../../client'
-import { MakeswiftComponentType } from '../../components'
+import { MakeswiftComponentSnapshot } from '../../../client'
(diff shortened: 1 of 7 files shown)


packages/runtime/src/runtimes/react/components/SlotProvider.tsx
```diff
@@ -6,7 +6,7 @@ export const PropsContext = createContext<{
   fallback: null,
 })
 
-function SlotProvider({
+export function SlotProvider({
```
agurtovoy, line 9: Replaced default export with named export.

packages/runtime/src/runtimes/react/components/Slot.tsx
```diff
@@ -21,5 +22,3 @@ export const Slot = ({ label, snapshot, fallback }: Props) => {
     </SlotProvider>
   )
 }
-
-export default Slot
```
agurtovoy, line 25: Removed redundant default export.

arvinpoddar (approved):

PR #1077, opened 2025-08-07
tests: built-in `Image` component rendering tests

agurtovoy:
Basic rendering test for the built-in `Image` component. No semantic changes. Prep for `next/image` decoupling.

See also https://github.com/makeswift/makeswift/pull/1078.

### DIFF
--- a/packages/runtime/src/next/components/tests/__fixtures__/resources/files.ts
+++ b/packages/runtime/src/next/components/tests/__fixtures__/resources/files.ts
@@ -1,0 +1,14 @@
+import { type File } from '../../../../../api'
+
+export const file1: File = {
+  __typename: 'File',
+  id: 'RmlsZTplMGJlNzVmMC01MjhmLTRiM2EtYTY4YS1kNjhlMTJjODMwZDI=',
+  name: 'Hill Chart',
+  publicUrl:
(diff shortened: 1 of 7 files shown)


packages/runtime/src/next/components/tests/__fixtures__/resources/files.ts
```diff
@@ -0,0 +1,14 @@
+import { type File } from '../../../../../api'
+
+export const file1: File = {
+  __typename: 'File',
+  id: 'RmlsZTplMGJlNzVmMC01MjhmLTRiM2EtYTY4YS1kNjhlMTJjODMwZDI=',
+  name: 'Hill Chart',
+  publicUrl:
+    'https://storage.googleapis.com/s.mkswft.com/RmlsZTplMGJlNzVmMC01MjhmLTRiM2EtYTY4YS1kNjhlMTJjODMwZDI=/Hill%20Chart.png',
```
arvinpoddar, line 8: Are there any risks to using a real image that points to our bucket?

agurtovoy: I don't think so, that's what you'd see in the HTML output for any Makeswift site.

PR #1079, opened 2025-08-08
example: Algolia

drewpledger:
This project exemplifies the usage of Algolia in Makeswift. It provides a registered component that communicates with the Algolia client configured by the environment variables. This example includes a built-in Navigation component that has the Algolia search component hard-coded. Instructions in the readme show how to use the Algolia crawler with Makeswift pages.

https://github.com/user-attachments/assets/69c62dcb-a93d-40d8-a7bf-c91fa70a1f94

To clarify something unspecified in the video, you're replacing the "placeholder" with the site verification key.

CleanShot 2025-12-30 at 12 29 32

This is the finished crawler that I "demonstrated"

EDIT: Since recording, I've specified to include the domain as an environment variable for the sitemap.

### DIFF
--- a/examples/algolia/README.md
+++ b/examples/algolia/README.md
@@ -1,0 +1,275 @@
+# Makeswift + Algolia Integration
+
+This integration demonstrates how to combine the power of Algolia's search capabilities with Makeswift's visual page building platform. Build visually editable websites with fast, intelligent search powered by Algolia's crawler that automatically indexes your Makeswift pages.
+
+## Prerequisites
+
+- [Node.js](https://nodejs.org/) 20.x or later
+- [Makeswift](https://www.makeswift.com/) account (for visual page building) [Sign up for free](https://app.makeswift.com/signup)
+- [Algolia](https://www.algolia.com/) account (for search indexing) [Sign up for free](https://www.algolia.com/users/sign_up)
+- Basic familiarity with [Next.js App Router](https://nextjs.org/docs/app)
+
+This guide assumes that you're familiar with Makeswift and setting up a custom host. Read here to learn more: https://docs.makeswift.com/developer/app-router/installation
+
+## Project Structure
+
+```
+algolia-demo/
+├── app/                          # Next.js App Router pages
+│   ├── [[...path]]/              # Dynamic Makeswift routes
+│   ├── api/makeswift/            # Makeswift API routes
+│   ├── layout.tsx                # Root layout with Makeswift provider
+│   ├── globals.css               # Global styles
+│   └── sitemap.ts                # Auto-generated sitemap for SEO
+├── components/                   # Makeswift components
+│   ├── AlgoliaSearch/            # Search component (used by Navigation)
+│   │   └── client.tsx            # Main search UI component
+│   └── Navigation/               # Navigation component with search
+├── lib/                          # Utility functions and configs
+│   ├── algolia/                  # Algolia client configuration
+│   ├── makeswift/                # Makeswift runtime & provider setup
+│   └── utils.ts                  # Shared utilities
+├── vibes/soul/                   # Pre-built UI components
+└── env.ts                        # Environment variable validation
+```
+
+## Quick Start
+
+### 1. Set up Algolia
+
+Before cloning the repository, you'll need to set up your Algolia account and get your API credentials:
+
+#### Create an Algolia Account
+
+1. [Sign up for a free Algolia account](https://www.algolia.com/users/sign_up) if you don't have one
+2. If you need to create a new application, go to **Settings** → **Applications** and click **New Application**
+   - You can use an existing application if you already have one
+
+#### Get Your API Credentials
+
+1. In your Algolia dashboard, navigate to **Settings** → **API Keys**
+2. Note down the following credentials (you'll need them when initializing the project):
+   - **Application ID** (`NEXT_PUBLIC_ALGOLIA_APP_ID`) - Found at the top of the API Keys page
+   - **Search API Key** (`NEXT_PUBLIC_ALGOLIA_SEARCH_API_KEY`) - Found in the "Your API Keys" section
+
+> **Important**: Use the **Search API Key**, not the Admin API Key. The search key is safe to use in your frontend code.
+
+> **Note**: You don't need to manually create an index yet. When you create your Algolia Crawler in step 8, it will automatically create the necessary indices to store your crawled content.
+
+### 2. Clone the example with the Makeswift CLI
+
+Initialize the project with the Makeswift CLI:
+
+```bash
+npx makeswift@latest init --example=algolia
+```
+
+### 3. Configure environment variables (initial setup)
+
+Here is how your `.env.local` should look once setup is finished:
+
+```
+MAKESWIFT_SITE_API_KEY=your_makeswift_api_key
+NEXT_PUBLIC_ALGOLIA_APP_ID=your_algolia_app_id
+NEXT_PUBLIC_ALGOLIA_SEARCH_API_KEY=your_algolia_search_api_key
+ALGOLIA_SITE_VERIFICATION=placeholder
+NEXT_PUBLIC_SITE_URL=https://example.com
+```
+
+**MAKESWIFT_SITE_API_KEY**: Automatically applied, found in your Makeswift site settings
+
+You will be prompted for:
+
+- **NEXT_PUBLIC_ALGOLIA_APP_ID**: Your Algolia Application ID from step 1
+- **NEXT_PUBLIC_ALGOLIA_SEARCH_API_KEY**: Your Algolia Search API Key from step 1
+- **ALGOLIA_SITE_VERIFICATION**: Use a placeholder like `placeholder` for now. You'll get the real value in step 5 after deploying
+- **NEXT_PUBLIC_SITE_URL**: Use a placeholder like `https://example.com` for now. Update this with your deployed domain after step 4. This is used for generating the sitemap
+
+The CLI should start up your development environment automatically, but if you need to run the server manually use:
+
+```bash
+npm run dev
+```
+
+### 4. Deploy your site and set up Makeswift host
+
+> **Note**: You need to deploy first to get a domain URL. The Algolia domain verification will be configured after deployment.
+
+1. Deploy your site to Vercel
+   - It's okay that `ALGOLIA_SITE_VERIFICATION` is using a placeholder for now
+   - Copy the generated Vercel URL (e.g., `https://your-site.vercel.app`)
+2. Configure your custom host in Makeswift:
+   - Open the [Makeswift builder](https://app.makeswift.com)
+   - Navigate to **Settings** → **Host** (accessible from the left sidebar)
+   - Set the **Custom host URL** field to your deployed domain (e.g., `https://your-site.vercel.app`)
+   - Save your changes
+3. Create some pages in Makeswift
+   - **Important**: Make sure your pages have a title and description set for them to be properly crawled and indexed by Algolia
+   - **Important**: Add links to your other pages. The Algolia crawler discovers pages by following links from your start URL, so any page you want indexed must be reachable through navigation links
+4. Add links to those new pages within the Navigation component -- it should look something like this:
+   <img width="1710" height="743" alt="Image" src="https://github.com/user-attachments/assets/d3f005de-f3e8-447f-bb63-25b64105371d" />
+
+### 5. Get Algolia domain verification token
+
+Now that you have a deployed domain, you can get the verification token:
+
+1. Navigate to the [Algolia Crawler Getting Started guide](https://www.algolia.com/doc/tools/crawler/getting-started/create-crawler/)
+2. Follow the "Add domains" section to add your deployed domain (e.g., `https://your-site.vercel.app`)
+3. With the **Meta tag** tab selected, copy the token from the meta tag (the value on content=`verification-code`)
+
+### 6. Update environment variables and redeploy
+
+1. Update your **local** `.env.local` with the verification token and your deployed site URL:
+
+   ```bash
+   ALGOLIA_SITE_VERIFICATION=your_actual_verification_token
+   NEXT_PUBLIC_SITE_URL=https://your-actual-site.vercel.app
+   ```
+
+2. Update the `ALGOLIA_SITE_VERIFICATION` and `NEXT_PUBLIC_SITE_URL` environment variables in your Vercel project settings
+
+The verification meta tag is included in `app/layout.tsx`:
+
+```tsx
+export const metadata: Metadata = {
+  other: {
+    'algolia-site-verification': env.ALGOLIA_SITE_VERIFICATION,
+  },
+}
+```
+
+3. **Redeploy your site** so the verification meta tag appears on your live site and the sitemap uses the correct domain
+
+### 7. Verify domain ownership in Algolia
+
+1. Once your site is redeployed with the verification meta tag, return to the Algolia Crawler dashboard under **Data sources**
+2. Click **Verify now** next to your domain
+3. Algolia will confirm ownership by detecting the meta tag on your live site
+
+### 8. Create Algolia Crawler
+
+After your domain is verified, you can create the crawler. **Important**: Creating a crawler automatically creates the associated index. The index name is automatically generated from the crawler name with the following transformations:
+
+- Spaces are converted to underscores
+- `_pages` is appended to the end
+- Example: Crawler name "My Blog" → Index name `my_blog_pages`
+
+1. In your Algolia dashboard, navigate to **Data sources** → **Crawlers** (at the bottom of the left sidebar)
+2. Click **New Crawler**
+3. Enter your crawler configuration:
+   - **Crawler name**: Choose a descriptive name (e.g., "Makeswift Site", "My Blog", or "Production Site")
+   - **Start URL**: Your deployed domain's home page (e.g., `https://your-site.vercel.app`)
+   - **What types of content do you want to crawl?**: Select the content types that match your site:
+     - Check **General web pages** for standard Makeswift pages
+     - Check **Articles** if you have blog posts or article pages
+     - Leave **Products** and **Technical documentation** unchecked unless applicable
+   - **Index prefix**: Leave this field **empty** unless you want an additional prefix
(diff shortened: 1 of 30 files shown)


examples/algolia/README.md
```diff
@@ -0,0 +1,234 @@
...
+export const metadata: Metadata = {
+  other: {
+    'algolia-site-verification': env.ALGOLIA_SITE_VERIFICATION,
+  },
+}
+```
+
+### 4. Deploy your site and set up Makeswift host
+
+1. Deploy your site to your hosting platform (Vercel, Netlify, etc.)
```
Codeseph: I'm running into a build issue that appears to be related to an incompatibility between the NextJS version and the Makeswift runtime version used by this example:

> 2025-10-13T21:02:40.803Z  Type error: Type 'typeof import("/vercel/path0/app/api/makeswift/[...makeswift]/route")' does not satisfy the constraint 'RouteHandlerConfig<"/api/makeswift/[...makeswift]">'.
2025-10-13T21:02:40.804Z    Types of property 'GET' are incompatible.
2025-10-13T21:02:40.805Z      Type '(...args: MakeswiftApiHandlerArgs) => Promise<void | NextResponse<MakeswiftApiHandlerResponse>>' is not assignable to type '(request: NextRequest, context: { params: Promise<{ makeswift: string[]; }>; }) => void | Response | Promise<void | Response>'.
2025-10-13T21:02:40.805Z        Types of parameters 'args' and 'request' are incompatible.
2025-10-13T21:02:40.805Z          Type '[request: NextRequest, context: { params: Promise<{ makeswift: string[]; }>; }]' is not assignable to type 'MakeswiftApiHandlerArgs'.
2025-10-13T21:02:40.805Z            Type '[request: NextRequest, context: { params: Promise<{ makeswift: string[]; }>; }]' is not assignable to type '[NextApiRequest, NextApiResponse<MakeswiftApiHandlerResponse>]'.
2025-10-13T21:02:40.805Z              Type at position 0 in source is not compatible with type at position 0 in target.
2025-10-13T21:02:40.805Z                Type 'NextRequest' is missing the following properties from type 'NextApiRequest': query, env, aborted, httpVersion, and 67 more.
2025-10-13T21:02:40.805Z
2025-10-13T21:02:40.840Z  Next.js build worker exited with code: 1 and signal: null
2025-10-13T21:02:40.906Z   ELIFECYCLE  Command failed with exit code 1.
2025-10-13T21:02:40.924Z  Error: Command "pnpm run build" exited with 1

Arvin explains this here: https://makeswifthq.slack.com/archives/C014SCS4G75/p1755828772730989

It sounds like we either need to bump the runtime to 0.25 (preferred) or use a fixed version of Next (15.4.7)

Codeseph (changes requested):

examples/algolia/README.md
Codeseph: And just to be clear, this wouldn't have been an issue back when you wrote the example. The error above wouldn't have appeared until later in August when Next 15.5 was released. I get that this is probably frustrating if you've been waiting for one of us to review, so I'm sorry

examples/algolia/README.md
```diff
@@ -0,0 +1,234 @@
...
+
+## Quick Start
+
+### 1. Clone the repository
+
+```bash
+npx makeswift@latest init --example=algolia-demo
+```
+
+### 2. Configure environment variables
```
Codeseph: I noticed the `.env.example` only has the Makeswift site API key. Even if we have the CLI generating the `.env.local `(after your upcoming changes), I imagine we want the example file to have the other required environment variables

examples/algolia/README.md
```diff
@@ -0,0 +1,234 @@
...
+NEXT_PUBLIC_ALGOLIA_SEARCH_API_KEY=your_algolia_search_api_key
+NEXT_PUBLIC_ALGOLIA_INDEX_NAME=your_algolia_index_name
+ALGOLIA_SITE_VERIFICATION=your_algolia_site_verification
+```
+
+1. **MAKESWIFT_SITE_API_KEY**: Found in your Makeswift site settings.
+2. **NEXT_PUBLIC_ALGOLIA_APP_ID**: Your Algolia Application ID, found in your Algolia dashboard settings under API Keys.
+3. **NEXT_PUBLIC_ALGOLIA_SEARCH_API_KEY**: Also found in the Algolia dashboard settings under API Keys.
+4. **NEXT_PUBLIC_ALGOLIA_INDEX_NAME**: The name of the Algolia index you want to use for search (create or select one in your Algolia dashboard).
+5. **ALGOLIA_SITE_VERIFICATION**: The verification code from the Algolia Crawler meta tag (see next step for instructions).
```
Codeseph: Up to you of course but I wonder if we should reorder this so that we wait to create the `.env.local` until after we've followed more of the Algolia "Getting Started" guide.

Right now for the `ALGOLIA_SITE_VERIFICATION` environment variable we say "see next step for instructions", but it feels like this is applicable to the other environment variables - in my case I had never worked with Algolia before, so I didn't have an Algolia app ID or search API key until following the steps below. And the index name that I ultimately ended up using was not created until following the steps for creating a crawler.

examples/algolia/README.md
```diff
@@ -0,0 +1,234 @@
...
+    'algolia-site-verification': env.ALGOLIA_SITE_VERIFICATION,
+  },
+}
+```
+
+### 4. Deploy your site and set up Makeswift host
+
+1. Deploy your site to your hosting platform (Vercel, Netlify, etc.)
+2. In your Makeswift settings, set your deployed domain as the custom host
+3. Create and publish some pages in Makeswift to have content for indexing
```
Codeseph: The troubleshooting section says that pages need a title and description to be crawled. If that's correct (I don't know) then I would include that information here

examples/algolia/README.md
```diff
@@ -0,0 +1,234 @@
...
+  other: {
+    'algolia-site-verification': env.ALGOLIA_SITE_VERIFICATION,
+  },
+}
+```
+
+### 4. Deploy your site and set up Makeswift host
+
+1. Deploy your site to your hosting platform (Vercel, Netlify, etc.)
+2. In your Makeswift settings, set your deployed domain as the custom host
```
Codeseph: I think we should be explicit about what 'Makeswift settings' refers to (host settings that are accessed via the builder), and what we're actually setting (the custom host URL)

examples/algolia/README.md
```diff
@@ -0,0 +1,234 @@
...
+### 6. Create Algolia Crawler
+
+After your domain is verified, you can create the crawler:
+
+1. In the [Algolia Crawler](https://crawler.algolia.com/) dashboard, click "New Crawler"
+2. Enter your crawler configuration:
+   - **Crawler name**: A descriptive name
+   - **App ID**: Your Algolia application ID
+   - **Start URL**: Your domain's home page
+   - **Crawler template**: Select the default template
```
Codeseph: Here are the options I had:

image

examples/algolia/README.md
```diff
@@ -0,0 +1,234 @@
...
+
+1. Once your site is deployed with the verification meta tag, return to the Algolia Crawler dashboard
+2. Click **Verify now** next to your domain
+3. Algolia will confirm ownership by detecting the meta tag on your live site
+
+### 6. Create Algolia Crawler
+
+After your domain is verified, you can create the crawler:
+
+1. In the [Algolia Crawler](https://crawler.algolia.com/) dashboard, click "New Crawler"
```
Codeseph: The crawler dashboard (left) suggests going to the Algolia dashboard (right) to do this:

image

examples/algolia/README.md
```diff
@@ -0,0 +1,234 @@
...
+2. Enter your crawler configuration:
+   - **Crawler name**: A descriptive name
+   - **App ID**: Your Algolia application ID
+   - **Start URL**: Your domain's home page
+   - **Crawler template**: Select the default template
+3. Click **Create** to run the initial test crawl
```
Codeseph: Doesn't this create a new Algolia search index, which means we now need to update the corresponding environment variable `NEXT_PUBLIC_ALGOLIA_INDEX_NAME` and redeploy?

examples/algolia/README.md
```diff
@@ -0,0 +1,234 @@
...
+- URLs
+
+### 7. Install dependencies and run development server
+
+```bash
+pnpm install
+pnpm dev
+```
+
+Your site will be available at `http://localhost:3000`.
```
Codeseph: Why are we suggesting that they run a local dev host here?

examples/algolia/README.md
```diff
@@ -0,0 +1,234 @@
...
+
+## Building Search Pages in Makeswift
+
+This integration includes a powerful search component that you can add to any Makeswift page:
+
+### Using the Search Component
+
+You can find the search component being used in the navigation bar, but you can also find it available in the component tray.
```
Codeseph: It feels weird to me to have the custom Algolia search component be at the top of every page (via layout.tsx) while also allowing it to be dragged/dropped from the component toolbar.

It would feel less weird to me if we registered the component such that we could specify which search index to use (via controls), but this isn't the case right now

examples/algolia/README.md
```diff
@@ -0,0 +1,234 @@
...
+
+1. In the [Makeswift builder](https://docs.makeswift.com/product/builder-basics), open any page or create a new one
```
Codeseph: It feels weird to just now be linking to the docs for the Makeswift builder, since by the time someone reaches this part of the guide we've already asked them to do several things (setting the custom host URL, publishing content, etc.) there

examples/algolia/README.md
```diff
@@ -0,0 +1,234 @@
...
+2. In the builder's left sidebar, find the **Algolia Search** component and drag it onto your page
```
Codeseph: I think we should use the term '[component toolbar](https://docs.makeswift.com/product/builder-basics#component-toolbar)' instead of 'left sidebar'

examples/algolia/README.md
```diff
@@ -0,0 +1,234 @@
...
+3. Configure the component properties:
+   - **Placeholder text**: Customize the search input placeholder (default: "Search...")
+   - **Initial results to show**: Number of results displayed initially (default: 8)
+   - **Results per pagination**: Number of additional results to load when users click "Show more" (default: 8)
+
+### Search Component Features
+
+The Algolia Search component includes:
+
+- **Real-time search**: Results update as users type (debounced for performance)
```
Codeseph, line 208: I think we should take another look at the debouncing implementation:

https://github.com/user-attachments/assets/d18d370c-9a14-4455-b8c3-17b1c44d301e

Codeseph: Here you can see how jarring this is when there are actual results - the video below is using a sample search index that Algolia allows you to add during setup

I haven't dug into the component implementation, but the flashing here is jarring. It feels like there's more rerendering going on than necessary, even discounting the debounce issue. 

https://github.com/user-attachments/assets/8ec361ed-017e-49be-93e0-e9e49797408e

examples/algolia/README.md
```diff
@@ -0,0 +1,234 @@
...
+- **Pagination**: "Show more" button to load additional results
+- **Typo tolerance**: Algolia's built-in typo tolerance for better search experience
+- **Responsive design**: Works on all device sizes
+
+### Automatic Page Discovery
+
+The integration automatically helps Algolia discover all your Makeswift pages through:
+
+1. **Sitemap Generation**: The `app/sitemap.ts` file automatically generates a sitemap containing all published Makeswift pages
+2. **Crawler Configuration**: Point your Algolia crawler to your sitemap URL for automatic page discovery
```
Codeseph: I think it's worth noting that there's a specific Algolia config for sitemap URLs

image


I'm not sure how significant this is versus just adding the sitemap URL as a starting URL. It looks like that's [all that's happening behind the scenes](https://www.algolia.com/doc/tools/crawler/apis/configuration/sitemaps):
image

examples/algolia/README.md
```diff
@@ -0,0 +1,234 @@
...
+3. **Continuous Updates**: When you publish new pages in Makeswift, they're automatically added to the sitemap
+
+### Index Population Process
+
+1. **Page Publishing**: When you publish a page in Makeswift, it becomes available at its URL
+2. **Sitemap Update**: The sitemap automatically includes the new page
+3. **Crawler Detection**: The Algolia crawler detects the sitemap changes (if configured)
+4. **Content Extraction**: The crawler visits the new page and extracts searchable content
+5. **Index Update**: The extracted content is added to your Algolia search index
```
Codeseph, line 229: Maybe worth noting that whether extracted content is searchable is dependent on your index settings:

image

I ended up investigating this because I published a page that had a bunch of text content, saw that the text content had been crawled, but didn't see that content reflected in search results

image


Maybe this is outside of the scope of what you want to handle for this sample integration. But I would worry that most people who are testing this would do so by trying to search for text content in a page. And it doesn't seem that this text content is searchable by default

examples/algolia/README.md
```diff
@@ -0,0 +1,234 @@
...
+The integration automatically helps Algolia discover all your Makeswift pages through:
+
+1. **Sitemap Generation**: The `app/sitemap.ts` file automatically generates a sitemap containing all published Makeswift pages
+2. **Crawler Configuration**: Point your Algolia crawler to your sitemap URL for automatic page discovery
+3. **Continuous Updates**: When you publish new pages in Makeswift, they're automatically added to the sitemap
+
+### Index Population Process
+
+1. **Page Publishing**: When you publish a page in Makeswift, it becomes available at its URL
+2. **Sitemap Update**: The sitemap automatically includes the new page
```
Codeseph: I have two published pages: `/page-a` and `/page-b`. `/page-b` is not reachable from `/page-a`, but it is included in the sitemap. Despite this, it doesn't seem that `/page-b` is being crawled. I'm not sure if I've missed something in the setup, but I would have expected `/page-b` to be crawled. 

image

Note that `/page-a` is used explicitly as a startUrl for the crawler. So it isn't clear to me that the sitemap URLs are actually being crawled. It's also entirely possible that I've misconfigured something

Codeseph (changes requested):

examples/algolia/README.md
Codeseph: (I bumped the runtime to 0.25 locally to work around this. Maybe we want to cherrypick my commits?)

examples/algolia/README.md
drewpledger: Naw, an index is the database, the crawler is something entirely different

examples/algolia/README.md
drewpledger: Image

Image

Interesting, I was able to get all the links from the sitemap crawled

examples/algolia/README.md
drewpledger: The debounce is set to 50ms, quite fast

examples/algolia/README.md
Codeseph: @drewpledger when I create a new crawler, a new search index gets created for the corresponding crawled content. And if I want the search results in my custom search component to reflect the content which was discovered by the new crawler that I just created, don't I need to update which index I'm pointing to in the environment? 

image

drewpledger: You are so right.. I even deployed an example of this

drewpledger: This is interesting. It's a bit of an odd workflow for this acute example that we'd need to deploy all the pages and verify the domain before creating the crawler and index. For example, when someone is creating this example, they'd need the index name for the CLI prompt, just to deploy it, and then create the actual index. I'll need to think about this more

examples/algolia/lib/makeswift/component.tsx
```diff
@@ -0,0 +1,20 @@
...
+  type: string
+  label: string
+  snapshotId: string
+}) => {
+  const snapshot = await client.getComponentSnapshot(snapshotId, {
+    siteVersion: await getSiteVersion(),
+  })
+
+  return <MakeswiftComponent label={label} snapshot={snapshot} {...props} />
+}
```
arvinpoddar: Missing newline

examples/algolia/lib/makeswift/provider.tsx
```diff
@@ -0,0 +1,20 @@
...
+}: {
+  children: React.ReactNode
+  siteVersion: SiteVersion | null
+}) {
+  return (
+    <ReactRuntimeProvider siteVersion={siteVersion} runtime={runtime}>
+      <RootStyleRegistry enableCssReset={false}>{children}</RootStyleRegistry>
+    </ReactRuntimeProvider>
+  )
+}
```
arvinpoddar: Missing newline

examples/algolia/components/Navigation/client.tsx
```diff
@@ -0,0 +1,206 @@
...
+                >
+                  {item.label}
+                </Link>
+              </li>
+            ))}
+          </ul>
+
+          {/* Mobile Menu Button - Only show on mobile when there are links */}
+          {links.length > 0 && (
+            <button
```
arvinpoddar, line 82: Could we replace this with a [button from vibes](https://vibes.site/docs/soul/button)? Or the [icon](https://vibes.site/docs/soul/icon)?

examples/algolia/components/Navigation/client.tsx
```diff
@@ -0,0 +1,206 @@
...
+              viewBox="0 0 24 24"
+            >
+              <path
+                strokeLinecap="round"
+                strokeLinejoin="round"
+                strokeWidth={2}
+                d="M21 21l-6-6m2-5a7 7 0 11-14 0 7 7 0 0114 0z"
+              />
+            </svg>
+          </button>
```
arvinpoddar, line 103: Same as https://github.com/makeswift/makeswift/pull/1079/changes#r2654156313

examples/algolia/README.md
```diff
@@ -0,0 +1,273 @@
...
+
+The verification meta tag is automatically included in `app/layout.tsx`:
+
+```tsx
+export const metadata: Metadata = {
+  other: {
+    'algolia-site-verification': env.ALGOLIA_SITE_VERIFICATION,
+  },
+}
+```
```
arvinpoddar, line 139: I think this example should be moved within the subsequent step ("Update environment variables and redeploy"), since that's where we set the environment variable. Maybe after Step 2?

PR #1066, opened 2025-07-26
fix: `backgrounds` control replaces/introspects video background swatch

agurtovoy:
See inline comments.

PR #1086, opened 2025-08-12
fix: null swatch ID for backgrounds prop controller

arvinpoddar:
Updates the backgrounds prop controller data schema to support null swatch IDs. 

Following up https://github.com/makeswift/makeswift/pull/1066.

### DIFF
--- a/packages/prop-controllers/src/backgrounds/backgrounds.ts
+++ b/packages/prop-controllers/src/backgrounds/backgrounds.ts
@@ -13,21 +13,28 @@ import {
   Schema,
 } from '../prop-controllers'
 import { P, match } from 'ts-pattern'
-import {
-  ColorData,
-  colorDataSchema,
-  imageDataV0Schema,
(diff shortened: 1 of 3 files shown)


packages/prop-controllers/src/backgrounds/backgrounds.ts
```diff
@@ -13,21 +13,28 @@ import {
...
-function copyColorData(
-  data: ColorData | null,
+import { imageDataV0Schema, imageDataV1Schema } from '../data'
+
+const backgroundsColorDataSchema = z.object({
+  swatchId: z.string().nullable(),
+  alpha: z.number(),
+})
+
+type BackgroundsColorData = z.infer<typeof backgroundsColorDataSchema>
```
arvinpoddar, line 23: A local schema for the backgrounds prop controller that supports null swatch IDs.

agurtovoy (approved):
🙏

PR #1090, opened 2025-08-14
chore: upgrade from `redux`/`redux-thunk` to `@reduxjs/toolkit`

agurtovoy:
ESM/Remix compatibility.


packages/runtime/package.json
```diff
@@ -177,8 +177,6 @@
     "path-to-regexp": "^6.2.1",
     "polished": "3.0.3",
     "react-player": "^2.16.0",
-    "redux": "^4.0.5",
-    "redux-thunk": "^2.3.0",
```
agurtovoy, line 181: `redux` and `redux-thunk` come bundled with `@reduxjs/toolkit`, so we don't need to explicitly import them.

packages/runtime/src/state/__tests__/react-builder-preview.test.ts
```diff
@@ -23,10 +24,15 @@ describe('propControllerHandlesMiddleware', () => {
     // Arrange
     const documentKey = 'documentKey'
     const element: ReactPage.Element = { key: 'elementKey', type: 'type', props: {} }
-    const store = createStore(
+    const store = configureReduxStore({
       reducer,
-      applyMiddleware(thunk, ReactPage.elementTreeMiddleware(), propControllerHandlesMiddleware()),
-    )
+      middleware: getDefaultMiddleware =>
```
agurtovoy, line 29: Here and below, new pattern for configuring the store's middleware: https://redux-toolkit.js.org/usage/migrating-rtk-2#configurestoremiddleware-must-be-a-callback

packages/runtime/src/state/actions.ts
```diff
@@ -803,3 +803,10 @@ export function setLocalizedResourceId({
 export function updateAPIClientCache(payload: APIClientCache): UpdateAPIClientCache {
   return { type: ActionTypes.UPDATE_API_CLIENT_CACHE, payload }
 }
+
+export function isKnownAction(action: unknown): action is Action {
+  return (
+    isAction(action) &&
+    Object.hasOwn(ActionTypes, action.type) != null
+  )
+}
```
agurtovoy, line 811: The two most significant (internal) breaking changes in this upgrade are:
- [Middleware `action` and `next` are typed as `unknown`](https://redux-toolkit.js.org/usage/migrating-rtk-2#middleware-type-changed---middleware-action-and-next-are-typed-as-unknown)
- [Reducers are expected to take `UnknownAction`](https://redux-toolkit.js.org/usage/migrating-rtk-2#anyaction-deprecated-in-favour-of-unknownaction)

`isKnownAction` allows us to limit the impact of these, see its usage in the reducers and in `toolkit.ts` below: https://github.com/makeswift/makeswift/pull/1090/files#diff-2d819ffea223200758ded5b0f0c1e50bb87cb0a227fb7428c93e211d72092ca0

packages/runtime/src/state/makeswift-api-client.ts
```diff
@@ -30,12 +31,13 @@ import {
 import { MAKESWIFT_CACHE_TAG } from '../next/cache'
 import { API_HANDLER_SITE_VERSION_HEADER, MakeswiftSiteVersion } from '../api/site-version'
 
-const reducer = combineReducers({
+const reducer = combineActionReducers({
```
agurtovoy: `combineActionReducers` is our wrapper around `combineReducers`: https://github.com/makeswift/makeswift/pull/1090/files#diff-2d819ffea223200758ded5b0f0c1e50bb87cb0a227fb7428c93e211d72092ca0R40

packages/runtime/src/state/makeswift-api-client.ts
```diff
@@ -211,17 +213,13 @@ export function fetchAPIResource<T extends APIResourceType>(
   }
 }
 
-export type Dispatch = ThunkDispatch<State, unknown, Action>
```
agurtovoy, line 214: Moved to the top of the file.

packages/runtime/src/state/toolkit.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    return state
+  }
+
+export function combineActionReducers<RMap extends Record<string, Reducer<any, Action>>>(
+  reducers: RMap,
+) {
+  const wrapped = {} as {
+    [K in keyof RMap]: Reducer<SliceStateMap<RMap>[K], Action | UnknownAction>
+  }
```
arvinpoddar: Not a big deal for something like this, but could we avoid the cast here just by typing the const?

packages/runtime/src/state/makeswift-api-client.ts
```diff
@@ -211,17 +213,13 @@ export function fetchAPIResource<T extends APIResourceType>(
...
-
 // FIXME: this middleware can be removed once we've upgraded the builder
 // to always provide the locale when dispatching resource actions
 function defaultLocaleMiddleware(
   defaultLocale: string | undefined,
-): Middleware<Dispatch, State, Dispatch> {
-  return () => (next: ReduxDispatch<Action>) => {
-    return (action: Action): Action => {
+): ThunkMiddleware<State, UnknownAction> {
+  return actionMiddleware(() => next => {
```
agurtovoy, line 217: Adjust types, wrap with [`actionMiddleware`](https://github.com/makeswift/makeswift/pull/1090/files#diff-2d819ffea223200758ded5b0f0c1e50bb87cb0a227fb7428c93e211d72092ca0R15) to avoid spreading `isKnownAction` checks throughout the codebase.

packages/runtime/src/state/makeswift-api-client.ts
```diff
@@ -245,26 +243,28 @@ export function configureStore({
 }: {
   defaultLocale: string | undefined
   serializedState?: SerializedState
-}): Store {
-  const composeEnhancers = composeWithDevToolsDevelopmentOnly({
```
agurtovoy, line 249: Replaced with [`devToolsConfig`](https://github.com/makeswift/makeswift/pull/1090/files#diff-2d819ffea223200758ded5b0f0c1e50bb87cb0a227fb7428c93e211d72092ca0R63) call below.

packages/runtime/src/state/react-builder-preview.ts
```diff
@@ -490,99 +492,95 @@ export function initialize(
...
-                registerMeasurable(
-                  action.payload.documentKey,
-                  action.payload.elementKey,
-                  action.payload.componentHandle,
-                ),
-              )
-            }
-
-            break
+  return actionMiddleware(({ dispatch }) => next => {
```
agurtovoy, line 492: Wrap with `actionMiddleware`.

packages/runtime/src/state/react-builder-preview.ts
```diff
@@ -490,99 +492,95 @@ export function initialize(
...
-          case ActionTypes.REGISTER_BUILDER_COMPONENT: {
-            const { transferables, ...forwardedAction } = action
-            channel.postMessage(forwardedAction, transferables)
-            break
-          }
-
-          case ActionTypes.UNREGISTER_BUILDER_COMPONENT:
-            channel.postMessage(action)
-            break
+  return actionMiddleware(({ dispatch }) => next => {
```
agurtovoy, line 522: Wrap with `actionMiddleware`.

packages/runtime/src/state/react-builder-preview.ts
```diff
@@ -490,99 +492,95 @@ export function initialize(
...
+    if (typeof window === 'undefined') return (action: Action) => next(action)
```
agurtovoy, line 523: This is the only semantic change: skip only this middleware, don't terminate the whole middleware chain when not running in the browser.

packages/runtime/src/state/react-builder-preview.ts
agurtovoy, line 1: This file is best reviewed with "Hide whitespace" toggle on.

packages/runtime/src/state/react-builder-preview.ts
```diff
@@ -730,53 +727,54 @@ function setupMessageChannel(channel: MessageChannel): ThunkAction<void, State,
...
-  const composeEnhancers = composeWithDevToolsDevelopmentOnly({
-    name: `Host store (${new Date().toISOString()})`,
-    serialize: true,
-    stateSanitizer: (state: any) => serializeState(state),
-    actionsDenylist: [
-      ActionTypes.BUILDER_POINTER_MOVE,
-      ActionTypes.HANDLE_POINTER_MOVE,
-      ActionTypes.ELEMENT_FROM_POINT_CHANGE,
-    ],
-  })
```
agurtovoy, line 756: Replaced with [`devToolsConfig`](https://github.com/makeswift/makeswift/pull/1090/files#diff-2d819ffea223200758ded5b0f0c1e50bb87cb0a227fb7428c93e211d72092ca0R63) call below.

packages/runtime/src/state/react-builder-preview.ts
```diff
@@ -730,53 +727,54 @@ function setupMessageChannel(channel: MessageChannel): ThunkAction<void, State,
...
+    preloadedState: initialState,
+
+    middleware: getDefaultMiddleware =>
+      getDefaultMiddleware(middlewareOptions).concat(
         ReactPage.elementTreeMiddleware(),
         measureBoxModelsMiddleware(),
         messageChannelMiddleware(channel),
         propControllerHandlesMiddleware(),
         makeswiftApiClientSyncMiddleware(client),
       ),
```
agurtovoy, line 750: Same list of middleware as before except for `thunk`, which is included by default.

packages/runtime/src/state/makeswift-api-client.ts
```diff
@@ -245,26 +243,28 @@ export function configureStore({
...
-    actionsDenylist: [
-      ActionTypes.BUILDER_POINTER_MOVE,
-      ActionTypes.HANDLE_POINTER_MOVE,
-      ActionTypes.ELEMENT_FROM_POINT_CHANGE,
-    ],
-  })
-
-  return createStore(
+}) {
+  return configureReduxStore({
```
agurtovoy, line 243: From https://redux.js.org/usage/migrating-to-modern-redux#modernizing-redux-logic-with-redux-toolkit:

> You should always start by replacing the legacy createStore call with configureStore. This is a one-time step, and all of the existing reducers and middleware will continue to work as-is.

packages/runtime/src/state/toolkit.ts
```diff
@@ -0,0 +1,71 @@
...
+
+export function actionMiddleware<S, A extends UnknownAction = UnknownAction>(
+  middleware: (
+    api: MiddlewareAPI<ThunkDispatch<S, unknown, A>, S>,
+  ) => (next: (action: unknown) => unknown) => (action: Action) => unknown,
+): Middleware<unknown, S, ThunkDispatch<S, unknown, A>> {
+  return api => next => {
+    const m = middleware(api)(next)
+    return (action: unknown) => {
+      return isKnownAction(action) ? m(action) : next(action)
```
agurtovoy, line 21: Filter out unknown actions in middleware.

packages/runtime/src/state/toolkit.ts
```diff
@@ -0,0 +1,71 @@
...
+
+export function combineActionReducers<RMap extends Record<string, Reducer<any, Action>>>(
+  reducers: RMap,
+) {
+  const wrapped = {} as {
+    [K in keyof RMap]: Reducer<SliceStateMap<RMap>[K], Action | UnknownAction>
+  }
+
+  for (const key in reducers) {
+    wrapped[key] = actionReducer(reducers[key])
```
agurtovoy: Filter out unknown actions in reducers.

packages/runtime/src/state/react-builder-preview.ts
arvinpoddar: What else is in this chain that was originally being terminated? Looking at the diff, I see:

```
ReactPage.elementTreeMiddleware(),
measureBoxModelsMiddleware(),
messageChannelMiddleware(channel),
propControllerHandlesMiddleware()
```

All of these are for builder interaction — can you elaborate why we wouldn't terminate these on the server?

packages/runtime/src/state/react-builder-preview.ts
```diff
@@ -730,53 +727,54 @@ function setupMessageChannel(channel: MessageChannel): ThunkAction<void, State,
...
-export type Store = ReduxStore<State, Action> & { dispatch: Dispatch } & SetupTeardownMixin
-
 export function configureStore({
   preloadedState,
   client,
 }: {
   preloadedState: Partial<State>
   client: MakeswiftHostApiClient
-}): Store {
+}) {
```
arvinpoddar, line 732: Did we intend to get rid of the explicit typing here?

packages/runtime/src/state/toolkit.ts
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  return combineReducers(wrapped)
+}
+
+export const middlewareOptions = {
+  serializableCheck: false,
+  immutableCheck: false,
+} as const
```
arvinpoddar, line 29: Just confirming that these match the default behaviors from before

packages/runtime/src/state/toolkit.ts
```diff
@@ -0,0 +1,71 @@
...
+
+type DevToolsOptions = Exclude<ConfigureStoreOptions['devTools'], boolean>
+
+const isDevToolsEnabled = () => process.env.NODE_ENV === 'development'
+
+export function devToolsConfig(options: DevToolsOptions): boolean | DevToolsOptions {
```
arvinpoddar, line 35: Kind of a weird return signature, but I'm not going to lose sleep over it.

packages/runtime/src/state/toolkit.ts
agurtovoy: Yep, there weren't any built-in checks in plain Redux. Also, these are just debugging checks, they don't affect actual store behavior.

We have to disable `serializableCheck` because we routinely put `Map` and `Set` in our state, neither of which are serializable. I disabled `immutableCheck` because it's performance-intensive.

packages/runtime/src/state/toolkit.ts
agurtovoy: It's coming from Redux Toolkit:

Screenshot 2025-08-14 at 5 03 42 PM

packages/runtime/src/state/toolkit.ts
agurtovoy: Unfortunately, no, because that type is a lie, haha. It's the type of the resulting object with all the keys, not the initial empty literal. If we make it a partial, it'll mess up the type of the combined reducer.

We have a [`mapValues`](https://github.com/makeswift/makeswift/blob/main/packages/controls/src/lib/functional.ts#L21) utility that would be perfect for this:

```ts
export function combineActionReducers<RMap extends Record<string, Reducer<any, Action>>>(
  reducers: RMap,
) {
  return combineReducers(mapValues(reducers, reducer => actionReducer(reducer)))
}
```

... but unfortunately it also fails to properly deduce the return type.

packages/runtime/src/state/react-builder-preview.ts
agurtovoy: Yes, that type is unwieldy, and the Redux Toolkit docs recommend deducing most of these types: https://redux-toolkit.js.org/usage/usage-with-typescript

packages/runtime/src/state/react-builder-preview.ts
agurtovoy: It's not so much that we want to run these on the server, but rather the arbitrary nature of this particular middleware silently interrupting the rest of the middleware chain out of nowhere. 

If we want to explicitly disable all of these on the server, we should do so, but that's not what this was doing previously: `messageChannelMiddleware` is not at the beginning or the end of the list here, so it'd literally let half of these execute and then ghost the rest.

packages/runtime/src/state/actions.ts
```diff
@@ -803,3 +803,10 @@ export function setLocalizedResourceId({
 export function updateAPIClientCache(payload: APIClientCache): UpdateAPIClientCache {
   return { type: ActionTypes.UPDATE_API_CLIENT_CACHE, payload }
 }
+
+export function isKnownAction(action: unknown): action is Action {
+  return (
+    isAction(action) &&
+    Object.hasOwn(ActionTypes, action.type) != null
+  )
```
fikrikarim: [Object.hasOwn()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn) returns a boolean?
```ts
return isAction(action) && Object.hasOwn(ActionTypes, action.type)
```

packages/runtime/src/state/toolkit.ts
fikrikarim: Would this interfere with [Redux internal actions](https://github.com/reduxjs/redux/blob/master/src/utils/actionTypes.ts)?

packages/runtime/src/state/actions.ts
agurtovoy: Good catch, thank you!

packages/runtime/src/state/toolkit.ts
agurtovoy: No, this simply skips the middleware that is being wrapped. It's no different than writing `if (!isKnownAction(action)) return next(action)` inside of the middleware, which is what we would have to do w/o this wrapper.

packages/runtime/src/state/toolkit.ts
arvinpoddar: Haha wow

packages/runtime/src/state/toolkit.ts
agurtovoy: So this is not a problem for middleware, but filtering out Redux internal actions *is* a problem for reducers, since Redux relies on the reducer returning the initial state slice for populating the state.

packages/runtime/src/state/actions.ts
agurtovoy: Fixed, wrote a test: https://github.com/makeswift/makeswift/pull/1090/files#diff-49a9a28d82272bc23e5889f34026530a515e479896fc4c2cd9f464f3af397c58

agurtovoy: Funny enough, I thought I fumbled it when I copied it from the builder, but we have the same bug there.

packages/runtime/src/state/toolkit.ts
agurtovoy: This doesn't work, see https://github.com/makeswift/makeswift/pull/1090#discussion_r2279536276

packages/runtime/src/state/modules/api-resources.ts
```diff
@@ -134,7 +134,9 @@ export function getAPIResource<T extends APIResourceType>(
     : null
 }
 
-export function reducer(state: State = getInitialState(), action: Action): State {
+export function reducer(state: State = getInitialState(), action: Action | UnknownAction): State {
+  if (!isKnownAction(action)) return state
+
```
agurtovoy, line 139: Here and below, adjust reducer's type to acknowledge that it might receive an unknown action, pre-filter actions using the `isKnownAction` predicate to keep the `switch`/`case` below properly typed.

In contrast to the middleware, the `isKnownAction` check has to happen *inside* of the reducer to allow for the state slice initialization (the `state: State = getInitialState()` default in the reducer definition).

packages/runtime/src/state/modules/components-meta.ts
```diff
@@ -27,7 +27,12 @@ export const ComponentIcon = {
 
 export type ComponentIcon = (typeof ComponentIcon)[keyof typeof ComponentIcon]
 
-export type ComponentMeta = { label: string; icon: ComponentIcon; hidden: boolean, description?: string }
+export type ComponentMeta = {
+  label: string
+  icon: ComponentIcon
+  hidden: boolean
+  description?: string
+}
```
agurtovoy, line 35: Prettier reformat.

packages/runtime/src/state/toolkit.ts
agurtovoy: https://github.com/makeswift/makeswift/pull/1090/files#r2279692559

PR #1091, opened 2025-08-14
chore: upgrade to the latest TypeScript (5.1.6 -> 5.9.2)

agurtovoy:
Upgrading to get the improved type inlining / [declaration tracking for dependencies](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-5.html). No usage of any newer TS features is being introduced, so this should be a no-op from the user’s standpoint.

Prep for https://github.com/makeswift/makeswift/pull/1090.

### DIFF
--- a/packages/runtime/package.json
+++ b/packages/runtime/package.json
@@ -225,7 +225,7 @@
     "slate-hyperscript": "^0.77.0",
     "slate-react": "^0.91.7",
     "tsup": "^8.0.1",
-    "typescript": "^5.1.6",
+    "typescript": "^5.9.2",
     "undici": "^6.19.2"
   }
 }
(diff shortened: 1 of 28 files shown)


packages/runtime/src/slate/BlockPlugin/__tests__/clearBlockKeyForDevice.test.tsx
```diff
@@ -2,7 +2,6 @@
 /** @jsx jsx */
 
 import { BlockActions } from '../'
-// @ts-expect-error: 'jsx' is declared but its value is never read.
```
agurtovoy, line 5: Here and below, remove now-unnecessary error suppression.

packages/runtime/package.json
```diff
@@ -225,7 +225,7 @@
     "slate-hyperscript": "^0.77.0",
     "slate-react": "^0.91.7",
     "tsup": "^8.0.1",
-    "typescript": "^5.1.6",
```
agurtovoy, line 228: Upgrading `runtime` only for now.

arvinpoddar (approved):

PR #1093, opened 2025-08-15
Add exit preview functionality via query param

arvinpoddar:
Introduces a query param that will redirect to the same path, but without any preview cookies set.


packages/runtime/src/next/api-handler/config/app-router.ts
```diff
@@ -41,6 +40,8 @@ export async function config({
     customRoutes: async (route: string) => {
       if (route === '/redirect-preview') {
         return { res: await appRouterRedirectPreviewHandler(req, context, client) }
+      } else if (route === '/redirect-live') {
+        return { res: await appRouterRedirectLiveHandler(req, context) }
```
arvinpoddar, line 44: In a separate commit, I'd like to update the config to require handlers (similar to revalidation) for redirecting preview tokens and redirecting to live. For now, I'm just adding them as custom routes.

packages/runtime/src/next/api-handler/handlers/get-request-url.ts
arvinpoddar, line 1: Utils for deriving the original request URL. These are used in both the handlers for preview tokens and for redirecting to live.

PR #1094, opened 2025-08-16
Add query param to clear preview cookies

arvinpoddar:


packages/runtime/src/api-handler/request-response.ts
```diff
@@ -19,6 +19,20 @@ export const ApiResponse = {
...
   },
+
+  redirect(location: string | URL, init?: ResponseInit): ApiResponse<null> {
+    const headers = new Headers(init?.headers)
+    headers.set('Location', location.toString())
+
+    // You cannot pass custom headers to a Redirect. We have to manually craft a
+    // Response with a redirect status code and Location header to achieve the
+    // redirect behavior on the client.
+    return new Response(null, {
```
arvinpoddar, line 31: See: https://community.cloudflare.com/t/make-redirect-with-additional-headers/249365. This also leads to some limitations as far as having the response be a perfect analog to a redirect, since the `readonly redirected` attribute cannot be set.

This does work perfectly on the client, and also allows for having redirects to relative paths (which `Response.redirect` also does not allow)

agurtovoy: I don't think `Response.redirected` matters on the server.

packages/runtime/src/api-handler/request-response.ts
```diff
@@ -19,6 +19,20 @@ export const ApiResponse = {
       headers,
     }) as ApiResponse<T>
   },
+
+  redirect(location: string | URL, init?: ResponseInit): ApiResponse<null> {
+    const headers = new Headers(init?.headers)
+    headers.set('Location', location.toString())
+
+    // You cannot pass custom headers to a Redirect. We have to manually craft a
```
agurtovoy: Maybe something like:

```ts
// Manually constructing a redirect response. We can’t use the `Response.redirect` shorthand here, 
// since it doesn’t allow passing custom headers; see https://community.cloudflare.com/t/make-redirect-with-additional-headers/249365
```

agurtovoy (approved):
Nice work

PR #1097, opened 2025-08-17
chore: don't use `use-sync-external-store/shim`

agurtovoy:
See inline comments.

### DIFF
--- a/packages/runtime/src/components/hooks/useMediaQuery.ts
+++ b/packages/runtime/src/components/hooks/useMediaQuery.ts
@@ -1,6 +1,4 @@
-import { useSyncExternalStore } from 'use-sync-external-store/shim'
-import { useCallback } from 'react'
-
+import { useCallback, useSyncExternalStore } from 'react'
 import {
   type Breakpoints,
   type DeviceOverride,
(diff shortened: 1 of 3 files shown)


packages/runtime/src/components/hooks/useMediaQuery.ts
```diff
@@ -1,6 +1,4 @@
-import { useSyncExternalStore } from 'use-sync-external-store/shim'
```
agurtovoy, line 1: Here and below, migrate stale imports of `useSyncExternalStore` from `use-sync-external-store/shim` to [React 18's native implementation](https://github.com/reactwg/react-18/discussions/86#discussioncomment-1567159). 

(We already do that in most of the new code).

packages/runtime/src/runtimes/react/hooks/use-selector.ts
```diff
@@ -1,4 +1,5 @@
-import { useSyncExternalStoreWithSelector } from 'use-sync-external-store/shim/with-selector'
+import { useSyncExternalStoreWithSelector } from 'use-sync-external-store/with-selector'
```
agurtovoy, line 1: React doesn't export `useSyncExternalStoreWithSelector`, so we have to keep this as a direct import from `use-sync-external-store` (but not the shim).

arvinpoddar (dismissed):

PR #1098, opened 2025-08-17
refactor: make `flush` a method of the style cache

agurtovoy:
See inline comments.


packages/runtime/src/next/root-style-registry.tsx
```diff
@@ -21,24 +21,19 @@ type Props = {
 }
 
 export function NextRootStyleRegistry({ children, cacheKey, enableCssReset }: Props) {
-  const [{ cache, flush }] = useState(() => createRootStyleCache({ key: cacheKey }))
+  const [cache] = useState(() => createRootStyleCache({ key: cacheKey }))
```
agurtovoy, line 24: The new API simplifies usage at the call site.

packages/runtime/src/runtimes/react/root-style-registry.tsx
```diff
@@ -9,13 +9,23 @@ const CacheContext = createContext(cache)
...
 const CSSResetEnabledContext = createContext(DEFAULT_CSS_RESET_ENABLED)
 
-export const createRootStyleCache = ({ key }: { key?: string } = {}) => {
+export type StyleCache = EmotionCache & {
+  /**
+   * Flush the inserted styles.
+   * @returns A list of class names and the corresponding CSS rules that were flushed.
+   */
+  flush: () => { classNames: string[]; css: string }
+}
```
agurtovoy, line 18: `flush` was already a de facto method of root style cache, since it modified its internal state; making it official so it can be accessed through the cache context.

packages/runtime/src/runtimes/react/root-style-registry.tsx
```diff
@@ -9,33 +9,51 @@ const CacheContext = createContext(cache)
...
+
+export const createRootStyleCache = ({ key }: { key?: string } = {}): StyleCache => {
   const cache = createCache({ key: key ?? 'mswft' })
   cache.compat = true
 
-  const prevInsert = cache.insert
+  // additional state to track inserted style names
```
agurtovoy, line 24: Comment up/reorg for easier understanding.

packages/runtime/src/runtimes/react/root-style-registry.tsx
```diff
@@ -9,33 +9,51 @@ const CacheContext = createContext(cache)
...
+      // reset our own state, leave `cache.inserted` intact
+      inserted = []
+
+      return {
+        classNames,
+        css: classNames.reduce((css, name) => {
+          return css + cache.inserted[name]
+        }, ''),
+      }
+    },
```
agurtovoy, line 50: Make `flush` a method, return structured, ready-to-use results.

packages/runtime/src/next/root-style-registry.tsx
```diff
@@ -21,24 +21,19 @@ type Props = {
...
-    if (names.length === 0) return null
-    let styles = ''
-    for (const name of names) {
-      styles += cache.inserted[name]
-    }
-
-    return (
+    const { classNames, css } = cache.flush()
+
+    return classNames.length > 0 ? (
```
fikrikarim, line 29: nit: should we use an early return here instead?
```
if (classNames.length === 0) return null
```

packages/runtime/src/next/root-style-registry.tsx
fikrikarim: nice

packages/runtime/src/next/root-style-registry.tsx
agurtovoy: It wouldn't be an "early return" since the following statement would also be a `return` :). In cases like this, I generally find the ternary operator cleaner and more expressive.

PR #1103, opened 2025-08-19
chore: `react-player` Vite SSR/ESM compatibility

agurtovoy:
Prep for the Remix implementation. Resolves `TypeError: __vite_ssr_import_3__.ReactPlayer.canPlay is not a function at BackgroundVideo` etc.

### DIFF
--- a/packages/runtime/package.json
+++ b/packages/runtime/package.json
@@ -176,7 +176,7 @@
     "parse5": "^7.1.2",
     "path-to-regexp": "^6.2.1",
     "polished": "3.0.3",
-    "react-player": "^2.16.0",
+    "react-player": "^2.16.1",
     "reselect": "^5.1.1",
     "scroll-into-view-if-needed": "^2.2.20",
     "set-cookie-parser": "^2.7.1",
--- a/packages/runtime/src/components/builtin/Video/Video.tsx
+++ b/packages/runtime/src/components/builtin/Video/Video.tsx
@@ -2,8 +2,8 @@
 
 import { cx } from '@emotion/css'
 import { forwardRef, Ref, useEffect, useState } from 'react'
-import ReactPlayer from 'react-player'
 
+import { ReactPlayer } from '../../shared/react-player'
 import { useStyle } from '../../../runtimes/react/use-style'
 import { placeholders } from '../../utils/placeholders'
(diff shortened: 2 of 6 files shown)


packages/runtime/package.json
```diff
@@ -176,7 +176,7 @@
     "parse5": "^7.1.2",
     "path-to-regexp": "^6.2.1",
     "polished": "3.0.3",
-    "react-player": "^2.16.0",
+    "react-player": "^2.16.1",
```
agurtovoy, line 179: Upgrade to the [latest non-ESM version](https://github.com/cookpete/react-player/releases/tag/v2.16.1).

packages/runtime/src/components/builtin/Video/Video.tsx
```diff
@@ -2,8 +2,8 @@
 
 import { cx } from '@emotion/css'
 import { forwardRef, Ref, useEffect, useState } from 'react'
-import ReactPlayer from 'react-player'
 
+import { ReactPlayer } from '../../shared/react-player'
```
agurtovoy, line 6: Here and below, point imports to an intermediate file that provides the workaround.

packages/runtime/src/components/shared/react-player.ts
```diff
@@ -0,0 +1,7 @@
+import Player from 'react-player'
+
+import { esmCompat } from '../../utils/esm-compat'
+
+// Vite SSR/ESM compatibility; upgrading to the latest ESM-only version of
+// `react-player` removes the need for this workaround but breaks Jest
+export const ReactPlayer = esmCompat(Player)
```
agurtovoy, line 7: We really need to switch from Jest to vitest, but it's more than I can chew at the moment.

packages/runtime/src/utils/esm-compat.ts
```diff
@@ -0,0 +1,2 @@
+// see https://www.npmjs.com/package/vite-plugin-cjs-interop
+export const esmCompat = <T>(x: T): T => (x as any).default ?? x
```
agurtovoy, line 2: Factored this out in a separate file for clarity/easier search of all places where we apply it.

arvinpoddar (approved):

PR #1106, opened 2025-08-21
chore: export preview-related symbols for internal consumption

agurtovoy:
See inline comments.

### DIFF
--- a/packages/runtime/src/api-handler/preview.ts
+++ b/packages/runtime/src/api-handler/preview.ts
@@ -1,0 +1,5 @@
+export const MAKESWIFT_SITE_VERSION_COOKIE = 'makeswift-site-version'
+
+export const SearchParams = {
+  PreviewToken: 'makeswift-preview-token',
+} as const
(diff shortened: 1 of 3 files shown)


packages/runtime/src/api-handler/preview.ts
```diff
@@ -0,0 +1,5 @@
+export const MAKESWIFT_SITE_VERSION_COOKIE = 'makeswift-site-version'
+
+export const SearchParams = {
+  PreviewToken: 'makeswift-preview-token',
+} as const
```
agurtovoy, line 5: Moved here from the [Next-specific implementation](https://github.com/makeswift/makeswift/pull/1106/files#diff-f0d260034da049c2c8d11d3185d6a2d6918543739374220708d3a1498a0cbf7dL4); other framework implementations should default to the same search param/cookie names.

arvinpoddar (approved):
Feel free to merge, but note I'll be renaming this symbol in a future PR to omit the `x-` prefix

PR #1108, opened 2025-08-21
chore: make API handlers' manifest override optional

agurtovoy:
Our manifest defaults don't require any overrides in most cases.


agurtovoy:
@arvinpoddar I don't think so, this is purely internal at this point.

PR #1110, opened 2025-08-21
refactor: encapsulate Emotion style tag creation

agurtovoy:
Code outside of the `runtimes/react` dir should not have knowledge / depend on the fact that we're using Emotion.

### DIFF
--- a/packages/runtime/src/runtimes/react/root-style-registry.tsx
+++ b/packages/runtime/src/runtimes/react/root-style-registry.tsx
@@ -51,6 +51,24 @@ export const createRootStyleCache = ({ key }: { key?: string } = {}): StyleCache
   }
 }
 
+type StyleTagProps = {
+  key: string
+  classNames: string[]
+  css: string
+}
+
+export const StyleTagSSR = ({ key, classNames, css }: StyleTagProps) => (
+  <style
+    data-emotion={`${key} ${classNames.join(' ')}`}
+    dangerouslySetInnerHTML={{
+      __html: css,
+    }}
+  />
+)
+
+export const styleTagHtml = ({ key, classNames, css }: StyleTagProps): string =>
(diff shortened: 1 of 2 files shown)


packages/runtime/src/runtimes/react/root-style-registry.tsx
```diff
@@ -51,6 +51,24 @@ export const createRootStyleCache = ({ key }: { key?: string } = {}): StyleCache
...
+}
+
+export const StyleTagSSR = ({ key, classNames, css }: StyleTagProps) => (
+  <style
+    data-emotion={`${key} ${classNames.join(' ')}`}
+    dangerouslySetInnerHTML={{
+      __html: css,
+    }}
+  />
+)
```
agurtovoy, line 67: Rendering through JSX.

packages/runtime/src/runtimes/react/root-style-registry.tsx
```diff
@@ -51,6 +51,24 @@ export const createRootStyleCache = ({ key }: { key?: string } = {}): StyleCache
...
+
+export const styleTagHtml = ({ key, classNames, css }: StyleTagProps): string =>
+  `<style data-emotion="${key} ${classNames.join(' ')}">${css}</style>`
```
agurtovoy, line 70: Rendering to a stream.

arvinpoddar: Where is this being used?

arvinpoddar (approved):

packages/runtime/src/runtimes/react/root-style-registry.tsx
agurtovoy: Currently used in the Remix and Express proxy implementation; likely to be used for any other "DYI" server-side rendering frameworks, e.g. Parcel.

PR #1113, opened 2025-08-22
fix: api route handler type compatibility

arvinpoddar:


packages/runtime/src/next/api-handler/config/app-router.ts
```diff
@@ -11,17 +11,21 @@ import { MAKESWIFT_SITE_VERSION_COOKIE, PRERENDER_BYPASS_COOKIE } from '../previ
...
+type Params = { [key: string]: string | string[] }
+type Context = { params: Promise<Params> }
 
-export type ApiHandlerArgs = [NextRequest, Context]
-export const argsPattern = [P.instanceOf(Request), P.any] as const
+export type ApiHandlerArgs = [NextRequest | Request, Context]
+export const argsPattern = [
+  P.union(P.instanceOf(Request), P.instanceOf(NextRequest)),
+  P.any,
+] as const
```
arvinpoddar, line 21: We are now required to explicitly handle `NextRequest` or `Request`. Updating the args type to support the discriminated union of these two request types. Also updating the ts-pattern matcher, which already matched against `Request`, to support `NextRequest` as well.

packages/runtime/src/next/api-handler/config/app-router.ts
```diff
@@ -39,7 +43,12 @@ export async function config({
     },
     customRoutes: async (route: string) => {
       if (route === '/redirect-preview') {
-        return { res: await appRouterRedirectPreviewHandler(req, context, client) }
+        // Convert any Request to NextRequest, if necessary. Due to an issue in
+        // how NextRequests are instantiated, we have to pass the original
+        // request to the constructor twice. See
+        // https://github.com/vercel/next.js/issues/52967
+        const request = req instanceof NextRequest ? req : new NextRequest(req, req)
```
arvinpoddar, line 50: Since we're explicitly handling `Request` now, this converts any inbound requests to `NextRequest`. Since it doesn't modify the request if it was already a `NextRequest`, this should not result in any new behavior.

packages/runtime/src/next/api-handler/config/app-router.ts
```diff
@@ -11,17 +11,21 @@ import { MAKESWIFT_SITE_VERSION_COOKIE, PRERENDER_BYPASS_COOKIE } from '../previ
 import { validateApiRoute, type ApiHandlerConfig } from './base'
 import { MakeswiftClient } from '../../../client'
 
-type Context = { params: { [key: string]: string | string[] } }
+type Params = { [key: string]: string | string[] }
+type Context = { params: Promise<Params> }
 
-export type ApiHandlerArgs = [NextRequest, Context]
-export const argsPattern = [P.instanceOf(Request), P.any] as const
+export type ApiHandlerArgs = [NextRequest | Request, Context]
```
arvinpoddar, line 17: App router context params must now be a `Promise`. Funny enough, our runtime appears to already have anticipated this, since we `await` params - see `validateApiRoute` in this file

packages/runtime/src/next/api-handler/index.ts
```diff
@@ -40,4 +40,6 @@ export function MakeswiftApiHandler(
 
     return sendResponse(await apiHandler(req, route))
   }
+
+  return handler as (...args: ApiHandlerArgs) => Promise<Response> | Promise<void>
```
arvinpoddar, line 56: Next.js now validates the type of the handler to be `void | Promise<void> | Response | Promise<Response>`. However, attempting to return the handler directly as `Promise<void> | Promise<Response>` from `handler` results in Typescript error:

```
The return type of an async function or method must be the global Promise<T> type. Did you mean to write 'Promise<void | Response>'?
```

 So we cast the function return type.

fikrikarim: Thanks for adding it as a comment

PR #1115, opened 2025-08-22
hotfix: API handler type compat with Next 15.5.0

arvinpoddar:


packages/runtime/src/next/api-handler/app-router-handler.ts
```diff
@@ -0,0 +1,11 @@
+import { NextRequest } from 'next/server'
+
+type Params = { [key: string]: string | string[] }
+
+export type Context = { params: Promise<Params> }
```
arvinpoddar, line 3: Updates app router context to support `Promise` of params. See: https://github.com/makeswift/makeswift/pull/1113/files#r2293896042

packages/runtime/src/next/api-handler/index.ts
```diff
@@ -128,6 +129,10 @@ export function MakeswiftApiHandler(
       .exhaustive()
   }
 
+  return handler as (
+    ...args: MakeswiftApiHandlerArgs
+  ) => Promise<NextResponse<MakeswiftApiHandlerResponse>> | Promise<void>
```
arvinpoddar, line 134: Casting to avoid restrictions with returning union of Promise types. See https://github.com/makeswift/makeswift/pull/1113/files#r2294093300

packages/runtime/src/next/api-handler/app-router-handler.ts
```diff
@@ -0,0 +1,9 @@
...
+
+// export type NextAppRouterRequest = NextRequest | Request
+
+// export function normalizeRequest(req: NextAppRouterRequest): NextRequest {
+//   return req instanceof NextRequest ? req : new NextRequest(req, req)
+// }
```
agurtovoy, line 9: @arvinpoddar Should we clean these up?

arvinpoddar: Left them in case I needed to revert.

arvinpoddar: I'll remove them

PR #1118, opened 2025-08-25
feat: alpha `express-react` package

agurtovoy:
`express-react` package, initial alpha release

### DIFF
--- a/packages/express-react/src/server/api-handler.ts
+++ b/packages/express-react/src/server/api-handler.ts
@@ -1,0 +1,47 @@
+import {
+  type Request as ExpressRequest,
+  type Response as ExpressResponse,
+  type RequestHandler,
+  type NextFunction,
+} from 'express'
+
+import {
+  type ApiHandlerUserConfig,
+  MAKESWIFT_SITE_VERSION_COOKIE,
+  createApiHandler as createMakeswiftApiHandler,
+  toApiRequest,
+  pipeResponseTo,
+} from '@makeswift/runtime/unstable-framework-support'
+
+import { Makeswift as MakeswiftClient } from '../client'
+
+const API_PATH_PREFIX = '/api/makeswift/'
+
+type UserConfig = ApiHandlerUserConfig & {
+  revalidationHandler?: (path?: string) => Promise<void>
+}
+
+export function createApiHandler(
+  apiKey: string,
+  { apiOrigin, runtime, revalidationHandler, ...userConfig }: UserConfig,
+): RequestHandler {
+  const client = new MakeswiftClient(apiKey, { apiOrigin, runtime })
+
+  return async function apiHandler(req: ExpressRequest, res: ExpressResponse, next: NextFunction) {
+    if (!req.path.startsWith(API_PATH_PREFIX)) {
+      return next()
+    }
+
+    const route = `/${req.path.replace(API_PATH_PREFIX, '')}`
+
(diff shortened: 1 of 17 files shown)


packages/express-react/src/server/preview.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    const siteVersion = await requestedSiteVersion({ searchParams, client })
+    if (siteVersion == null) return next()
+
+    const serializedSiteVersion = serializeSiteVersion(siteVersion)
+    const secondsUntilExpiration = secondsUntilSiteVersionExpiration(siteVersion)
+
+    res.cookie(MAKESWIFT_SITE_VERSION_COOKIE, serializedSiteVersion, {
+      ...cookieSettingOptions,
+      maxAge: secondsUntilExpiration,
```
fikrikarim: I think `res.cookie` `maxAge` is in [milliseconds](https://expressjs.com/en/api.html)?
image

packages/express-react/tsup.config.ts
```diff
@@ -0,0 +1,36 @@
...
+  const esmOptions: Options = {
+    ...commonOptions,
+    format: 'esm',
+  }
+
+  const cjsOptions: Options = {
+    ...commonOptions,
+    format: 'cjs',
+    outDir: 'dist/cjs',
+  }
```
fikrikarim, line 33: Do we need to also specify the `outDir` for `esmOptions` to `dist/esm` since it's specified on `package.json` as `dist/esm`?

packages/express-react/src/server/api-handler.ts
```diff
@@ -0,0 +1,51 @@
...
+  createApiHandler as createMakeswiftApiHandler,
+  toApiRequest,
+  pipeResponseTo,
+} from '@makeswift/runtime/unstable-framework-support'
+
+import { Makeswift as MakeswiftClient } from '../client'
+
+const API_PATH_PREFIX = '/api/makeswift/'
+
+async function revalidationHandler(path?: string): Promise<void> {
```
agurtovoy: I forgot to make this a part of the user config.

packages/express-react/tsup.config.ts
agurtovoy: Good question; it's not necessary, `tsup` will write to `dist/esm` by default; it might also be unnecessary for `cjs`, I just took our [`tsup` config from the `runtime` package](https://github.com/makeswift/makeswift/blame/main/packages/runtime/tsup.config.ts).

packages/express-react/src/server/preview.ts
agurtovoy: Wow, [you're right](https://github.com/expressjs/express/blob/ef7ad681b245fba023843ce94f6bcb8e275bbb8e/lib/response.js#L797), how bizarre, would have never thought to check since the `Set-Cookie` header's [`Max-Age` value is in seconds](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#max-agenumber). Good catch, thank you!

arvinpoddar: Did not expect this. Great find @fikrikarim

agurtovoy: It's extra funny because if both `Expires` and `Max-Age` are set, `Max-Age` has precedence, so the appearance of the greater precision that this interface gives is well, an illusion.

PR #1121, opened 2025-08-25
chore: export `BaseHeadSnippet` as `DefaultHeadSnippet`

agurtovoy:
Add (internal) `DefaultHeadSnippet` export for consistency with `DefaultHead`, `DefaultImage`, etc.


arvinpoddar (approved):
I wonder if we should rename this internally to `DefaultHeadSnippet`. Base implies the ability to extend its behavior, which I _guess_ we do for app router, but not by actually extending the snippet itself. I'll leave that up to you.

agurtovoy:
@arvinpoddar I like the current separation between the two symbols. `BaseHeadSnippet` is really a base implementation w/o any server-side support. Whether that is sufficient for a particular framework context, that's for the framework to decide. Re-exporting it as the default implementation in our default React framework context makes that decision (yes, it's sufficient) explicit.

PR #1125, opened 2025-08-28
chore: API handler type compatibility follow-up

agurtovoy:


arvinpoddar (approved):

arvinpoddar:
Confirmed that the snapshot here builds in an independent host running `0.15.5`

PR #1127, opened 2025-08-29
feat: expose experimental framework integration APIs

agurtovoy:
Prep for releasing alpha versions of `react-router`/`express-react` packages.


packages/runtime/package.json
```diff
@@ -111,6 +112,11 @@
       "require": "./dist/cjs/state/modules/breakpoints.js",
       "import": "./dist/esm/state/modules/breakpoints.js",
       "types": "./dist/types/state/modules/breakpoints.d.ts"
+    },
+    "./unstable-framework-support": {
+      "require": "./dist/cjs/unstable-framework-support/index.js",
+      "import": "./dist/esm/unstable-framework-support/index.js",
+      "types": "./dist/types/unstable-framework-support/index.d.ts"
```
agurtovoy, line 119: New entry point, explicitly marked as "unstable".

fikrikarim (commented):
So from what I understand, we need to expose this since `@makeswift/express-react` is a different package.

My only concern is that this will clutter or mess up the Intellisense/autocomplete in the host since we're exporting the same thing. For example, `MakeswiftComponent` is exported both on `@makeswift/runtime/next` and `@makeswift/runtime/unstable-framework-support`.

I did a bit of testing on the branch but somehow it's only showing `@makeswift/runtime/next`. I don't understand why the `@makeswift/runtime/unstable-framework-support` is not showed, even though I can import it if I manually type the package.

image

image

agurtovoy:
@fikrikarim Good point re: multiple entry point exporting the same thing. I don't think it's a problem in practice, but we should definitely keep an eye on it.

Ideally, we need to extract `@makeswift/next` from `@makeswift/runtime`, and that’s actually pretty trivial at this point. However, it would obviously be a major breaking change, requiring updates across Makeswift and Catalyst docs, tutorials, etc.

This PR is an incremental step toward that future. I'm taking this step now to enable building decoupled prototypes against a released version of the runtime.

fikrikarim (approved):

packages/runtime/src/unstable-framework-support/index.ts
arvinpoddar, line 1: Aren't some of these exports already exposed? Or are we okay with duplicate exports for the sake of them all being through the "unstable" entry point.

arvinpoddar (approved):

packages/runtime/src/unstable-framework-support/index.ts
agurtovoy: Most of them aren’t exported at the package level. The ones that are already exported are `type SiteVersion`, `MakeswiftComponent`, `Slot`, and `ReactRuntime`. I considered leaving these out of the new endpoint, but in the end decided that there is value in being explicit about what the “framework integration API” includes, exactly -- and, from the framework side, importing them from a single entry point is much cleaner.

agurtovoy: I should have commented on this, thank you both for the thoughtful review/discussion!

PR #1128, opened 2025-08-29
chore: unpin example apps next version

arvinpoddar:
After making the hotfix to 0.24.9 of the runtime for API handler type compatibility with Next.js 15.5.0, we can now unpin the version of Next.js being used by these example apps.


agurtovoy (approved):
🙏

PR #1135, opened 2025-09-04
fix: check null data at translation merge

arvinpoddar:
Fix a regression introduced in v0.20.0 of the runtime, where we no longer check for null data when merging translated data.

### DIFF
--- a/packages/controls/src/controls/group/group.ts
+++ b/packages/controls/src/controls/group/group.ts
@@ -248,11 +248,11 @@ class Definition<C extends Config> extends ControlDefinition<
   }
 
   mergeTranslatedData(
-    data: DataType<C>,
+    data: DataType<C> | undefined,
     translatedData: Record<string, DataType<C>>,
     context: MergeTranslatableDataContext,
   ): Data {
-    if (translatedData == null) return data
+    if (data == null || translatedData == null) return data
 
     const propsData = Definition.propsData(data)
 
--- a/packages/controls/src/controls/group/group.ts
+++ b/packages/controls/src/controls/group/group.ts
@@ -237,7 +237,7 @@ class Definition<C extends Config> extends ControlDefinition<
     )
   }
 
-  getTranslatableData(data: DataType<C>): Data {
+  getTranslatableData(data: DataType<C> | undefined): Data {
     if (data == null) return null
 
     const propsData = Definition.propsData(data)
--- a/.changeset/polite-dancers-teach.md
+++ b/.changeset/polite-dancers-teach.md
@@ -1,0 +1,6 @@
+---
+"@makeswift/controls": patch
+"@makeswift/runtime": patch
+---
+
+fix: check null data before attempting to merge translated data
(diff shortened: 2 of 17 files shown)


packages/controls/src/controls/group/group.ts
```diff
@@ -248,11 +248,11 @@ class Definition<C extends Config> extends ControlDefinition<
   }
 
   mergeTranslatedData(
-    data: DataType<C>,
+    data: DataType<C> | undefined,
     translatedData: Record<string, DataType<C>>,
     context: MergeTranslatableDataContext,
   ): Data {
-    if (translatedData == null) return data
+    if (data == null || translatedData == null) return data
```
agurtovoy, line 255: If `data` was not null when it was sent for translation, but is null now, should we discard the translation result? The answer might be yes, I just want us to consider it explicitly.

The argument for discarding is that the only way `data` would become null is if the user edited the prop, which means the translation is no longer relevant. The argument against discarding is that this would create a special case for just one type of edit, clearing the prop. In all other cases, the translated data will overwrite the edits, and perhaps it’s better to do so consistently and without exceptions.

agurtovoy: Independent of the above, should we update the `getTranslatableData` signature as well?

```ts
getTranslatableData(data: DataType<C> | undefined):
```

arvinpoddar: Having a consistent behavior makes sense. At the same time, I could see why it wouldn't be optimal. Consider you request translations on a large element tree, request translations, and delete a particularly large subsection of that tree. It would be weird if when the translation results returns, it restores a large block of content.

That being said, I do not have full context into why the current behavior was decided, so my gut reaction is uninformed. The goal of this PR is to restore the existing behavior that has been in place since [`0.11.3`](https://github.com/makeswift/makeswift/releases/tag/%40makeswift%2Fruntime%400.11.3) of the runtime and clear the regression.

agurtovoy: Could you link to the pre-regression code for the record? 🙏

arvinpoddar: Here's the original [`mergeTranslatedData` function](https://github.com/makeswift/makeswift/blob/1df1b83cb14cf55e4dc82dbd3ecebdf904aab58e/packages/runtime/src/controls/control.ts#L238).

This old implementation, like the [current](https://github.com/makeswift/makeswift/blob/f0c30f4404e8feacfcb5315290d455e1c2fc7c2d/packages/runtime/src/controls/control.ts#L78), has a top level `null` check on `data` that discards the translated data if true. The difference is that the previous implementation recurses back to this top-level function. The current implementation **does not**, so we don't perform this null check at each level of control data. 

For example, we see that the [`old mergeListTranslatedData function`](https://github.com/makeswift/makeswift/blob/1df1b83cb14cf55e4dc82dbd3ecebdf904aab58e/packages/runtime/src/controls/list.ts#L257) calls the top-level `mergeTranslatedData` function on each list item, so the null check is evaluated at each level.

arvinpoddar: And to reiterate, my intent here is solely to restore the original behavior. This should be one of those topics we revisit when we one day revamp our translation APIs.

PR #1138, opened 2025-09-04
fix: Cloudflare workers compat

agurtovoy:
The worker execution model [disallows asynchronous I/O, setting a timeout, and generating random values within the global scope](https://community.cloudflare.com/t/cant-issue-http-request-server-side-in-astro/760305).


packages/runtime/src/components/builtin/Carousel/Carousel.tsx
```diff
@@ -348,7 +348,6 @@ const Carousel = forwardRef(function Carousel(
             >
               {images.map(({ props: imageProps, key }) => (
                 <motion.div
-                  id={key}
```
agurtovoy, line 351: There is no good reason to set the image `div`'s HTML ID to its key here. It's a random ID that is never exposed to the user, and we never query or otherwise use it.

agurtovoy: Demos confirming that removing the ID doesn't break anything: 

https://github.com/user-attachments/assets/7213eb20-20a5-4e79-ad1f-b0ab0c164d2e

https://github.com/user-attachments/assets/d2e13831-68a6-4af7-8779-3cee2eb0ffbd

packages/runtime/src/components/builtin/Carousel/register.ts
```diff
@@ -39,9 +37,9 @@ export function registerComponent(runtime: ReactRuntime) {
         id: ElementID(),
         images: Images({
           preset: [
-            { key: uuid(), props: {} },
-            { key: uuid(), props: {} },
-            { key: uuid(), props: {} },
+            { key: 'image-1', props: {} },
+            { key: 'image-2', props: {} },
+            { key: 'image-3', props: {} },
```
agurtovoy, line 42: With the [change above](https://github.com/makeswift/makeswift/pull/1138/files#r2323345140), the keys don't have to be globally unique, just unique within the carousel's images.

packages/runtime/src/next/components/tests/controls/__snapshots__/list-control.test.tsx.snap
```diff
@@ -117,7 +117,7 @@ exports[`Page List of colors when value is {"value":[{"swatchId":"[swatch-test-i
               "props": {
                 "propKey": [
                   {
-                    "id": "xxxxxxxx-xxxx-xxxx-xxxx-100000000006",
+                    "id": "xxxxxxxx-xxxx-xxxx-xxxx-100000000003",
```
agurtovoy, line 120: Here and below, update test snapshots to reflect three fewer global `uuid` calls.

packages/runtime/src/components/builtin/Carousel/Carousel.tsx
arvinpoddar: Is it possible that this is required internally by `motion`?

arvinpoddar: Checked their docs to confirm, doesn't appear to be required.

agurtovoy: It's highly unlikely, but that's why I recorded the demos above to show that we aren't breaking anything.

`motion.div` is [essentially a `div`](https://github.com/motiondivision/motion/blob/c084ec07b483c00dce197e8958f9cf30bd493268/packages/framer-motion/src/render/components/motion/elements.ts#L37) enhanced with animation capabilities. The `id` here is just a [pass-through HTML ID attribute](https://github.com/motiondivision/motion/blob/c084ec07b483c00dce197e8958f9cf30bd493268/packages/framer-motion/src/motion/index.tsx#L64). I haven't seen a mention of it anywhere in the docs, and if `motion` did need an ID internally, they would just generate and assign it themselves using a `data-` attribute.

agurtovoy: E.g. they have this one: https://github.com/motiondivision/motion/blob/main/packages/framer-motion/src/animation/optimized-appear/data-id.ts

arvinpoddar (approved):

PR #1140, opened 2025-09-05
fix: match Request and NextRequest in API handler

arvinpoddar:
Fixes a regression in ba2e9de where the API handler does not match vanilla Requests. In older versions of Next (prior to 15.5.0), the inbound request is a Request. In newer versions, the request is a NextRequest. To maintain compatibility with all Next.js versions, we match against both.

### DIFF
--- a/packages/runtime/src/next/api-handler/config/app-router.ts
+++ b/packages/runtime/src/next/api-handler/config/app-router.ts
@@ -14,15 +14,23 @@ import { MakeswiftClient } from '../../../client'
 type Params = { [key: string]: string | string[] }
 type Context = { params: Promise<Params> }
 
-export type ApiHandlerArgs = [NextRequest, Context]
-export const argsPattern = [P.instanceOf(NextRequest), P.any] as const
+// In older versions of Next (prior to 15.5.0), the inbound request is typed as
+// a `NextRequest`, but requests received at runtime fail the `instanceof
+// NextRequest` check. In newer versions of Next, requests are correctly typed
+// as `NextRequest`. To maintain compatibility across all Next.js versions, we
+// match against both.
+export type ApiHandlerArgs = [NextRequest | Request, Context]
+export const argsPattern = [
+  P.union(P.instanceOf(Request), P.instanceOf(NextRequest)),
+  P.any,
+] as const
 
 export async function config({
   req,
   context,
   client,
 }: {
-  req: NextRequest
+  req: NextRequest | Request
   context: Context
   client: MakeswiftClient
 }): Promise<ApiHandlerConfig> {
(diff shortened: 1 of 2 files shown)


packages/runtime/src/next/api-handler/config/app-router.ts
```diff
@@ -14,15 +14,18 @@ import { MakeswiftClient } from '../../../client'
 type Params = { [key: string]: string | string[] }
 type Context = { params: Promise<Params> }
 
-export type ApiHandlerArgs = [NextRequest, Context]
-export const argsPattern = [P.instanceOf(NextRequest), P.any] as const
+export type ApiHandlerArgs = [NextRequest | Request, Context]
```
agurtovoy, line 22: I think we should add a comment here, something along the lines of

> In older versions of Next (prior to 15.5.0), the inbound request is typed as a `NextRequest`, but requests received at runtime fail the `instanceof NextRequest` check. In newer versions, runtime requests are correctly typed as `NextRequest`. To maintain compatibility across all Next.js versions, we match against both.

agurtovoy: ^^ I think that's what's happening here, based on the fact that `appRouterRedirectPreviewHandler` has code like this:

```ts
export async function appRouterRedirectPreviewHandler(
  request: NextRequest,
  _context: any,
  client: MakeswiftClient,
): Promise<NextResponse> {
  const previewToken = request.nextUrl.searchParams.get(SearchParams.PreviewToken)
```

and `request.nextUrl` has never failed us despite the runtime check always being `P.instanceOf(Request)`.

agurtovoy: Although, even though it's marked as a fix, one can argue that this change contradicts my statement about types: https://github.com/vercel/next.js/pull/82644/files.

Nevertheless, the requests received at runtime must have been at least duck-typed as `NextRequest`; otherwise, accessing `request.nextUrl` would not have been possible.

packages/runtime/src/next/api-handler/config/app-router.ts
```diff
@@ -40,7 +43,12 @@ export async function config({
     },
     customRoutes: async (route: string) => {
       if (route === '/redirect-preview') {
-        return { res: await appRouterRedirectPreviewHandler(req, context, client) }
+        // Convert any Request to NextRequest, if necessary. Due to an issue in
+        // how NextRequests are instantiated, we have to pass the original
+        // request to the constructor twice. See
+        // https://github.com/vercel/next.js/issues/52967
+        const request = req instanceof NextRequest ? req : new NextRequest(req, req)
```
agurtovoy, line 55: If my reasoning above is correct, this is unnecessary, but it might be less confusing to leave the conversion than to remove it.

arvinpoddar: I agree with your reasoning, but I'd also like to keep this given that it makes an implicit behavior (we think) explicit

packages/runtime/src/next/api-handler/config/app-router.ts
arvinpoddar: Who knows, maybe Next.js did something like patch `Request` if you're reading it in an API handler context.

PR #1147, opened 2025-09-12
Experimental React Server Components (RSCs) Support

fikrikarim:
## Demo

https://github.com/user-attachments/assets/ec1096eb-2bf6-40a7-909c-e53b1bb531e4

## Important Notes
- I try to minimize changes to the runtime core by isolating RSC-related code in its own entry points. This way, we can iterate on RSC quickly while ensuring the non-RSC path remains stable.
- Changes that affect runtime core are isolated on the first few commits.
- Currently the RSC support is tied to Next.js, this is why I put the entrypoint in `@makeswift/runtime/next/rsc`. Once we add RSC support to more framework, we should decouple the RSC core and framework-specific RSC implementation.
- The Next RSC package `packages/runtime/src/next/rsc` is still under heavy development and it's not production-ready yet. We want to merge this early so multiple people can collaborate on this and to unlock other initiative like Makeswift on Stencil.
- We’re planning another spike to explore what it would take to refactor the runtime from the ground up to be RSC-first. That said, this will be a longer journey. In the meantime, it’s nice to have incremental RSC support without having to refactor everything first.
- RSC-support can change everything, like how user register component, how user set up the `MakeswiftProvider`, and other things. This PR keeps most of the status quo, and focus on enabling users to register and edit RSCs.
- There are still quite a few bugs when using RSC. Again, the goal isn’t to deliver a completely bug-free experience, but to establish a foundation we can iterate on.

## RSC User Flow
- User imports experimental `ReactRuntime`, `MakeswiftPage`, `RuntimeProvider`, and `ServerProvider`, from `@makeswift/runtime/next/rsc`.
- User registers an RSC by adding `server: true` to `runtime.registerComponent`.
- If the registration includes a function, like `getOptions`, they need to add `use server` so we can serialize the functions from RSC to the client bundle.
- User creates two separate registrations files, `components.client` and `components.server`. User imports the clients and the server registration on the server provider, and import only the client registration on the client provider.
- Please refer to `next-rsc` to see an example on how to integrate RSCs.

## RSC Mechanism
- The runtime serializes server component registrations and pass that to the client. This is needed so we can show the correct overlays and controls when user is selecting an RSC.
- The runtime pre-renders all the RSCs before rendering any client components, then it'll pass the RSCs as a context to client components. If the client component has an RSC child, then it can return the RSC since the runtime already prerendered the RSC on the server. This relies on an important assumption that we can resolve RSC props ahead of time, before rendering any client components.
- The runtime has an RSC middleware that tracks all element tree changes, and calls `router.refresh()` when a user inserts an RSC.
- For `Style` control, on the server rendering, the runtime collects all the CSS, and inject them to the HTML. When it's on the builder, it'll listen to prop changes, and modify the `<style>` tag directly to avoid needing a server refresh when user is changing a style.
- For other controls, when user changes a property, the runtime will call `router.refresh()`.

### DIFF
--- a/packages/runtime/package.json
+++ b/packages/runtime/package.json
@@ -87,6 +87,16 @@
       "require": "./dist/cjs/next/plugin.js",
       "import": "./dist/cjs/next/plugin.js"
     },
+    "./next/rsc": {
+      "require": "./dist/cjs/next/rsc/index.js",
+      "import": "./dist/esm/next/rsc/index.js",
+      "types": "./dist/types/next/rsc/index.d.ts"
+    },
+    "./next/rsc/server": {
+      "require": "./dist/cjs/next/rsc/server/index.js",
+      "import": "./dist/esm/next/rsc/server/index.js",
+      "types": "./dist/types/next/rsc/server/index.d.ts"
+    },
     "./next/server": {
       "types": "./dist/types/next/server.d.ts",
       "module": "./dist/esm/next/server.js",
--- a/packages/runtime/src/next/rsc/client/refresh-middleware.ts
+++ b/packages/runtime/src/next/rsc/client/refresh-middleware.ts
@@ -1,0 +1,61 @@
+import { type Middleware } from '@reduxjs/toolkit'
+import { type useRouter } from 'next/navigation'
+
+import { type State, type Dispatch, Element } from '../../../state/react-page'
+import * as ComponentsMeta from '../../../state/modules/components-meta'
+import { ActionTypes, type Action, type DocumentPayload } from '../../../state/actions'
+import { actionMiddleware } from '../../../state/toolkit'
+import { traverseElementTree } from '../../../state/modules/element-trees'
+import { type DescriptorsByComponentType } from '../../../state/modules/prop-controllers'
+
+export function createRSCRefreshMiddleware(
+  router?: ReturnType<typeof useRouter>,
+): Middleware<Dispatch, State, Dispatch> {
+  return actionMiddleware<State, Action>(({ getState }) => next => {
+    return action => {
+      switch (action.type) {
+        // TODO: this can be much more efficient if we send a new operation from the builder when a new RSC element is added
+        case ActionTypes.CHANGE_ELEMENT_TREE: {
+          const { oldDocument, newDocument, descriptors } = action.payload
+          const componentsMeta = ComponentsMeta.getComponentsMeta(getState().componentsMeta)
+          const newlyAddedElements = getNewlyAddedElements(oldDocument, newDocument, descriptors)
+
+          for (const element of newlyAddedElements) {
+            const meta = componentsMeta.get(element.type)
+
+            if (meta?.server) {
+              console.log(`[RSC] element "${element.type}" added, refreshing page`)
+              router?.refresh()
+              break
+            }
+          }
+
+          break
+        }
+      }
+
+      return next(action)
+    }
+  })
+}
+
+function getNewlyAddedElements(
+  oldDoc: DocumentPayload,
+  newDoc: DocumentPayload,
+  descriptors: DescriptorsByComponentType,
+): Element[] {
+  const oldElementKeys = new Set<string>()
+  const newElements: Element[] = []
+
+  for (const element of traverseElementTree(oldDoc.rootElement, descriptors)) {
+    oldElementKeys.add(element.key)
+  }
+
+  for (const element of traverseElementTree(newDoc.rootElement, descriptors)) {
+    if (!oldElementKeys.has(element.key)) {
+      newElements.push(element)
+    }
+  }
+
+  return newElements
+}
--- a/packages/runtime/src/next/rsc/client/refresh-middleware.ts
+++ b/packages/runtime/src/next/rsc/client/refresh-middleware.ts
@@ -1,7 +1,7 @@
 import { type Middleware } from '@reduxjs/toolkit'
 import { type useRouter } from 'next/navigation'
 
-import { type State, type Dispatch, Element } from '../../../state/react-page'
+import { type State, type Dispatch, Element } from '../../../state/read-only-state'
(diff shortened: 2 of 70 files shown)


packages/controls/src/serialization/object-serialization.ts
```diff
@@ -41,6 +51,12 @@ export function serializeObject(object: unknown): [unknown, Transferable[]] {
 
   function serialize(value: unknown): unknown {
     return match(value)
+      // This serializer is reused in two flows:
+      //   • server components → client components (RSC/Flight)
+      //   • host → builder
+      // In the RSC flow, Flight/Next handles encoding/decoding server references across the boundary.
+      // In non-RSC flows (host → builder) server references shouldn’t appear.
+      .with(P.when(isServerReference), () => value)
```
fikrikarim: Example of a Combobox `getOptions` when it's serialized on RSC vs on the client:
Screenshot 2025-09-29 at 10 57 01

I tried to create a separate control serialization that's specific to RSC, but it ended up being too much work, while this patch is only several lines. We can look into having a separate control serialization for RSC again later.

As far as I know, this should be safe and shouldn't affect non-RSC code.

packages/runtime/src/next/rsc/server/resource-resolver.ts
```diff
@@ -0,0 +1 @@
+export const mockResourceResolver = {} as any
```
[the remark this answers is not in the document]
fikrikarim: This is intentional. We'll resolve this in future PR.

packages/runtime/src/next/rsc/server/server-element.tsx
```diff
@@ -0,0 +1,43 @@
...
+  if (!isRSC) {
+    return <Element key={element.key} element={element} />
+  }
+
+  if (isElementReference(element)) {
+    return <FallbackComponent text="Element reference is not supported on server yet" />
+  }
+
+  // TODO: Get this value from the context/props
+  const isPreview = true
```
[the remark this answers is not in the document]
fikrikarim: This is intentional. We'll resolve this in future PR.

packages/runtime/src/next/rsc/client/refresh-middleware.ts
```diff
@@ -0,0 +1,61 @@
...
+  router?: ReturnType<typeof useRouter>,
+): Middleware<Dispatch, State, Dispatch> {
+  return actionMiddleware<State, Action>(({ getState }) => next => {
+    return action => {
+      switch (action.type) {
+        // TODO: this can be much more efficient if we send a new operation from the builder when a new RSC element is added
+        case ActionTypes.CHANGE_ELEMENT_TREE: {
+          const { oldDocument, newDocument, descriptors } = action.payload
+          const componentsMeta = ComponentsMeta.getComponentsMeta(getState().componentsMeta)
+          const newlyAddedElements = getNewlyAddedElements(oldDocument, newDocument, descriptors)
```
arvinpoddar, line 21: I'm not sure what to make of this. I'm actually happy that this implementation does not rely on the builder being aware of whether it's editing an RSC or not, nor am I sure that it _should_ be aware. What do you think @agurtovoy?

packages/controls/src/serialization/object-serialization.ts
```diff
@@ -24,6 +24,16 @@ export abstract class Serializable {
...
 
+// https://github.com/facebook/react/blob/f739642745577a8e4dcb9753836ac3589b9c590a/packages/react-server-dom-webpack/src/ReactFlightWebpackReferences.js#L26-L35
+const SERVER_REFERENCE_TAG = Symbol.for('react.server.reference')
+function isServerReference(reference: unknown): boolean {
+  return (
+    typeof reference === 'function' &&
+    '$$typeof' in reference &&
+    reference.$$typeof === SERVER_REFERENCE_TAG
+  )
+}
```
arvinpoddar: Don't love that this made it's way down to the controls level. We went out of the way to ensure that this package has no coupling with React. Perhaps we need to redefine serialization at each layer of our package? cc: @agurtovoy

packages/runtime/package.json
```diff
@@ -77,6 +77,16 @@
       "require": "./dist/cjs/next/plugin.js",
       "import": "./dist/cjs/next/plugin.js"
     },
+    "./next/rsc": {
+      "require": "./dist/cjs/next/rsc/index.js",
+      "import": "./dist/esm/next/rsc/index.js",
+      "types": "./dist/types/next/rsc/index.d.ts"
+    },
+    "./next/rsc/server": {
```
arvinpoddar, line 95: TODO: Definitely throw in an `experimental` or `unstable` prefix on these entry points somewhere

fikrikarim: I was thinking that it might be unnecessary since we already have the prefix on the exported functions. But there's no harm on adding it here again.

packages/controls/src/serialization/object-serialization.ts
fikrikarim: I totally agree. It'll be much better if we don't have to touch this package. I spent some time experimenting on how we can approach this, and this is the best version I came up given the time-constraint. Another alternative is to duplicate the entire serialization code. Open to ideas how we can improve this.

packages/runtime/src/next/rsc/client/refresh-middleware.ts
fikrikarim: The action doesn't need to be specific to RSC. I was thinking more of adding a "metadata" of each operation since the op is too opaque.

packages/controls/src/serialization/object-serialization.ts
arvinpoddar: Addressed, was able to take advantage of the visitor pattern to create an RSC compatible serializer.

fikrikarim: 🐐

PR #1152, opened 2025-09-18
refactor: merge translatable data visitor

arvinpoddar:
Pulls out logic related to merging translatable data from the control interface in favor of consolidating this behavior in a visitor class.

### DIFF
--- a/packages/controls/src/controls/checkbox/checkbox.ts
+++ b/packages/controls/src/controls/checkbox/checkbox.ts
@@ -17,6 +17,7 @@ import {
   type SchemaType,
 } from '../definition'
 import { DefaultControlInstance, type SendMessage } from '../instance'
+import { ControlDefinitionVisitor } from '../visitor'
 
 type Config = z.infer<typeof Definition.schema.relaxed.config>
 
@@ -178,6 +179,10 @@ class Definition<C extends Config> extends ControlDefinition<
       type: Definition.type,
       version: this.version,
     })
+  }
+
+  accept<R>(visitor: ControlDefinitionVisitor<R>, ...args: unknown[]): R {
+    return visitor.visitCheckbox(this, ...args)
   }
 }
 
--- a/packages/controls/src/controls/definition.ts
+++ b/packages/controls/src/controls/definition.ts
@@ -4,11 +4,7 @@ import { type ValueSubscription } from '../lib/value-subscription'
 import { type ParseResult } from '../lib/zod'
 
 import { type Data } from '../common/types'
-import {
-  type CopyContext,
-  type MergeContext,
-  type MergeTranslatableDataContext,
-} from '../context'
+import { type CopyContext, type MergeContext } from '../context'
 import { type IntrospectionTarget } from '../introspection'
 import { type ResourceResolver } from '../resources/resolver'
 import {
@@ -19,6 +15,7 @@ import {
 import { type Stylesheet } from '../stylesheet'
 
 import { ControlInstance, type SendMessage } from './instance'
+import { ControlDefinitionVisitor } from './visitor'
 
 export type SchemaType<T> = z.ZodType<T>
 export type SchemaTypeAny = SchemaType<any> | z.ZodBranded<SchemaType<any>, any>
@@ -83,14 +80,6 @@ export abstract class ControlDefinition<
     return null
   }
 
-  mergeTranslatedData(
-    data: DataType | undefined,
-    _translatedData: Data,
-    _context: MergeTranslatableDataContext,
-  ): Data {
-    return data as Data
-  }
-
   abstract resolveValue(
     data: DataType | undefined,
     resolver: ResourceResolver,
@@ -101,6 +90,11 @@ export abstract class ControlDefinition<
   abstract createInstance(sendMessage: SendMessage<any>): InstanceType
 
   abstract serialize(): [SerializedRecord, Transferable[]]
+
+  abstract accept<R>(
+    visitor: ControlDefinitionVisitor<R>,
+    ...args: unknown[]
+  ): R
 
   introspect<R>(
     data: DataType | undefined,
--- a/packages/controls/src/controls/link/link.ts
+++ b/packages/controls/src/controls/link/link.ts
@@ -20,12 +20,24 @@ import {
   type SchemaType,
 } from '../definition'
 import { DefaultControlInstance, type SendMessage } from '../instance'
+import { ControlDefinitionVisitor } from '../visitor'
 
 import { type GenericMouseEvent } from './mouse-event'
 import { resolveLink } from './resolve-link'
 import * as LinkSchema from './schema'
 
-type Config = z.infer<typeof Definition.schema.config>
+/**
+ * Defining the config schema out of the Definition class to avoid a `Type alias
+ * 'Config' circularly references itself` error from TS. Given that we use this
+ * pattern of defining the config schema in the class itself in other controls,
+ * this may be a case of the compiler reaching some limit.
+ */
+const configSchema = z.object({
+  label: z.string().optional(),
+  description: z.string().optional(),
+})
+
+type Config = z.infer<typeof configSchema>
 
 type Schema<_C extends Config> = typeof Definition.schema
 type LinkTarget<C extends Config> = z.infer<Schema<C>['linkTarget']>
@@ -67,10 +79,7 @@ class Definition<
     const data = LinkSchema.data
     const value = data
 
-    const config = z.object({
-      label: z.string().optional(),
-      description: z.string().optional(),
-    })
+    const config = configSchema
 
     const definition = z.object({
       type,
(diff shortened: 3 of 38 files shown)


packages/controls/src/controls/checkbox/checkbox.ts
```diff
@@ -179,6 +180,10 @@ class Definition<C extends Config> extends ControlDefinition<
       version: this.version,
     })
   }
+
+  accept<T>(visitor: ControlDefinitionVisitor<T>, ...args: unknown[]): T {
```
arvinpoddar: Here and below (for a while), adding the `accept` method implementations to each control class. Occasionally exporting the config type if needed

packages/controls/src/controls/visitor/definition-visitor.ts
```diff
@@ -0,0 +1,93 @@
...
+import { RichTextDefinition, RichTextV1Definition } from '../rich-text'
+import { SelectConfig, SelectDefinition } from '../select'
+import { ShapeDefinition } from '../shape'
+import { SlotDefinition } from '../slot'
+import { StyleDefinition, StyleV2Definition } from '../style'
+import { TextAreaDefinition } from '../text-area'
+import { TextInputDefinition } from '../text-input'
+import { unstable_TypographyDefinition } from '../typography'
+
+abstract class ControlDefinitionVisitor<R> {
```
arvinpoddar, line 24: The base abstract `Visitor`, with methods for each concrete control

packages/controls/src/controls/visitor/merge-translations-visitor.ts
```diff
@@ -0,0 +1,223 @@
...
+import { ShapeDefinition } from '../shape'
+import { SlotDefinition } from '../slot'
+import { StyleDefinition, StyleV2Definition } from '../style'
+import { TextAreaDefinition } from '../text-area'
+import { TextInputDefinition } from '../text-input'
+import { unstable_TypographyDefinition } from '../typography'
+
+import { ControlDefinitionVisitor } from './definition-visitor'
+
+abstract class MergeTranslationsVisitor extends ControlDefinitionVisitor<Data> {
```
arvinpoddar, line 32: An abstract `Visitor` for merging translatable data. The only implementation that is left abstract is that for `rich-text-v2`, which we implement in the `runtime` package.

Most of these methods use either a no-op merge (just return back the data) or a default merge (accept the translated data, if present, or return back the original data)

packages/runtime/src/controls/rich-text-v2/translations/get-translations.ts
```diff
@@ -0,0 +1,83 @@
+import escapeHtml from 'escape-html'
```
arvinpoddar, line 1: Split the original `translations` file into three files:

1. Core types / utilities
2. Utils for extracting translatable data
3. Utils for merging translatable data

This split will aid in code-splitting and bundling later

packages/runtime/src/controls/visitors/merge-translations-visitor.ts
```diff
@@ -0,0 +1,26 @@
+import { Data, DataType, MergeTranslationsVisitor } from '@makeswift/controls'
+
+import { RichTextV2Definition } from '../rich-text-v2'
+import { RichTextTranslationDto } from '../rich-text-v2/translations/translations-core'
+import { mergeTranslatedNodes } from '../rich-text-v2/translations/merge-translations'
+
+export class ReactMergeTranslationsVisitor extends MergeTranslationsVisitor {
```
arvinpoddar, line 7: Here's the final concrete implementation of the visitor for merging translated data. We've managed to pull `mergeTranslatedNodes` out of the Rich Text control class itself.

packages/runtime/src/controls/control.ts
```diff
@@ -83,7 +84,8 @@ export function mergeTranslatedData(
 ): Data {
   if (data == null) return data
   if (!isLegacyDescriptor(definition)) {
-    return definition.mergeTranslatedData(data, translatedData, context)
+    const mergeTranslationsVisitor = new ReactMergeTranslationsVisitor(context)
+    return definition.accept(mergeTranslationsVisitor, data, translatedData)
```
arvinpoddar, line 88: Usage of the new visitor from https://github.com/makeswift/makeswift/pull/1152/files#r2361312802

packages/controls/src/controls/link/link.ts
```diff
@@ -20,12 +20,18 @@ import {
...
 
 import { type GenericMouseEvent } from './mouse-event'
 import { resolveLink } from './resolve-link'
 import * as LinkSchema from './schema'
 
-type Config = z.infer<typeof Definition.schema.config>
+const configSchema = z.object({
+  label: z.string().optional(),
+  description: z.string().optional(),
+})
```
arvinpoddar, line 38: Pulling the schema out of the definition itself, because otherwise TS throws a `Type alias 'Config' circularly references itself.` error.

I will add, this is not my first time seeing this error when messing around in the controls. I have a feeling this is just the TS compiler reaching its limit for some reason, since we do this pattern elsewhere.

packages/controls/src/controls/definition.ts
```diff
@@ -83,14 +80,6 @@ export abstract class ControlDefinition<
     return null
   }
 
-  mergeTranslatedData(
```
arvinpoddar, line 86: These implementations now live in the `Visitor` for merging translated data

packages/controls/src/controls/link/link.ts
agurtovoy: I'd add a comment to that effect for future selves.

packages/controls/src/controls/visitor/definition-visitor.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  abstract visitColor(def: ColorDefinition, ...args: unknown[]): R
+
+  abstract visitCombobox(def: ComboboxDefinition, ...args: unknown[]): R
+
+  abstract visitFont(def: FontDefinition, ...args: unknown[]): R
+
+  abstract visitGroup(def: GroupDefinition, ...args: unknown[]): R
+
+  abstract visitIconRadioGroup<C extends IconRadioGroupConfig>(
```
agurtovoy, line 30: This is going to sound like a nitpick, so please bear with me :). 

The newlines after each method make this class definition feel too spread out and hard to size up, which really affects comprehension for me. I’d keep the newlines after the multi-line methods, and collapse everything else:

```ts
abstract class ControlDefinitionVisitor<R> {
  abstract visitCheckbox(def: CheckboxDefinition, ...args: unknown[]): R
  abstract visitColor(def: ColorDefinition, ...args: unknown[]): R
  abstract visitCombobox(def: ComboboxDefinition, ...args: unknown[]): R
  abstract visitFont(def: FontDefinition, ...args: unknown[]): R
  abstract visitGroup(def: GroupDefinition, ...args: unknown[]): R
  abstract visitIconRadioGroup<C extends IconRadioGroupConfig>(
    def: IconRadioGroupDefinition<C>,
    ...args: unknown[]
  ): R

  abstract visitImage<C extends ImageConfig>(
    def: ImageDefinition<C>,
    ...args: unknown[]
  ): R

  abstract visitLink(def: LinkDefinition<any>, ...args: unknown[]): R
  abstract visitList(def: ListDefinition, ...args: unknown[]): R
  abstract visitNumber(def: NumberDefinition, ...args: unknown[]): R
  abstract visitRichTextV1(
    def: RichTextV1Definition<unknown, any>,
    ...args: unknown[]
  ): R

  abstract visitRichTextV2(
    def: RichTextDefinition<unknown, any>,
    ...args: unknown[]
  ): R

  abstract visitSelect<C extends SelectConfig>(
    def: SelectDefinition<C>,
    ...args: unknown[]
  ): R

  abstract visitShape(def: ShapeDefinition, ...args: unknown[]): R
  abstract visitSlot<RuntimeNode>(
    def: SlotDefinition<RuntimeNode>,
    ...args: unknown[]
  ): R
...
```

packages/controls/src/controls/checkbox/checkbox.ts
agurtovoy: I'd use `R` instead of `T` here as well.

agurtovoy (approved):
Love it, it never felt right to me to have translation-related methods in the `ControlDefinition`.

packages/controls/src/controls/visitor/definition-visitor.ts
arvinpoddar: I actually agree given that this class is a declaration of the interface, rather than any concrete behavior. I would have pushed back had this been the actual method implementations. Thanks for the feedback, addressed.

packages/controls/src/controls/link/link.ts
arvinpoddar: Done

packages/controls/src/controls/checkbox/checkbox.ts
arvinpoddar: Good catch. Addressed here and for all other controls.

PR #1155, opened 2025-09-24
feat: add warning when `previewMode` is disabled

arvinpoddar:
Add a warning message that appears during the build/start or dev process if the user has explicitly disabled the `previewMode`.

### DIFF
--- a/packages/makeswift-next-plugin/index.js
+++ b/packages/makeswift-next-plugin/index.js
@@ -1,9 +1,23 @@
 // @ts-check
 
 /** @typedef {import('next').NextConfig} NextConfig */
-/** @typedef {{ resolveSymlinks?: boolean; appOrigin?: string; previewMode?: boolean }} MakeswiftNextPluginOptions */
 /** @typedef {NonNullable<import('next').NextConfig['images']>} ImageConfig */
 /** @typedef {ImageConfig['remotePatterns']} ImageConfigRemotePatterns */
+
+/**
+ * Configuration options for the Makeswift Next.js plugin
+ * @typedef {Object} MakeswiftNextPluginOptions
+ * 
+ * @property {boolean} [resolveSymlinks] - Whether to resolve symlinks in
+ * webpack.
+ * 
+ * @property {string} [appOrigin] - The origin of the Makeswift builder. Used
+ * for CORS headers to allow the builder to communicate with your site
+ * 
+ * @property {boolean} [disableBuiltInPreview] - Whether to disable Makeswift's
+ * built-in preview mode. When true, you must implement a custom preview mode
+ * for visual editing to work
+ */
 
 const { satisfies } = require('semver')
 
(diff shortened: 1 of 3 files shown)


packages/makeswift-next-plugin/index.js
```diff
@@ -116,6 +116,12 @@ module.exports =
         throw new Error('Makeswift requires a minimum Next.js version of 13.4.0.')
       }
 
+      if (!previewMode) {
+        console.warn(
```
agurtovoy: Since there are legitimate uses for disabling this, issuing a warning doesn’t seem quite right. If the main problem we’re trying to solve is that Catalyst users are likely to keep `{ previewMode: false }` as they upgrade, a better solution might be to rename this parameter. That way, we both break existing code and have a chance of better communicating the purpose of this option. Perhaps `disableBuiltInPreview`, with a JSDoc comment explaining that it?

arvinpoddar: The name change is fine with me, but I would argue that there's still some value in adding some logging when this mode is set. Yes there's a legitimate use case, but they're rare and almost entirely internal. It'd be good to have some affordance to know that this is being set. It doesn't have to be a console.warn either.

agurtovoy: We could do a `console.log` but this would literally be the first "preemptive" info logging we do in the runtime. You can argue that this is a special case of a valid-but-rare configuration that justifies breaking the rule. I'm not opposed to it, but it still feels not quite deserving. Your call.

agurtovoy: If we do log anything, I think it should be something neutral along the lines "Makeswift's built-in preview mode is disabled. Check that a custom preview mode is implemented to enable visual editing.", rather than "are you sure this is what you want".

arvinpoddar: Yep, we're on the same page. I agree that the logging should be neutral. As long as we have some logging to signal this, I'm okay with it.

packages/makeswift-next-plugin/index.js
```diff
@@ -32,7 +32,7 @@ module.exports =
   ({
     resolveSymlinks,
     appOrigin = 'https://app.makeswift.com',
-    previewMode = true,
+    disableBuiltInPreview = false,
```
agurtovoy, line 49: Can we add a JSDoc comment explaining the options, including `disableBuiltInPreview`?

agurtovoy: My main thought here is to prevent someone who is migrating Catalyst and getting a compilation error on `previewMode: true` from jumping to the plugin definition to see what happened to the old param and fixing the error by migrating to `disableBuiltInPreview: true`.

Well, we can't really "prevent" it, but hopefully a comment will help.

agurtovoy: Maybe something like `@param {boolean} disableBuiltInPreview - disables Makeswift's built-in preview mode. Set this only if you are providing a custom preview mode implementation.`?

agurtovoy (approved):
🙏

PR #1156, opened 2025-09-24
refactor: decouple builtin registration from runtime

arvinpoddar:
Refactors the ReactRuntime such that builtin component registration is not
coupled to instantiation.Refactors the ReactRuntime such that builtin component registration is not
coupled to instantiation.


packages/runtime/package.json
```diff
@@ -34,9 +34,84 @@
...
+      "require": "./dist/cjs/react/index.js",
+      "import": "./dist/esm/react/index.js",
+      "types": "./dist/types/react/index.d.ts"
+    },
+    "./react/core": {
+      "require": "./dist/cjs/react/core/index.js",
+      "import": "./dist/esm/react/core/index.js",
+      "types": "./dist/types/react/core/index.d.ts"
+    },
+    "./react/builtins/carousel": {
```
arvinpoddar, line 50: Each builtin has a separate entrypoint for proper tree-shaking. If the registrations are exposed from a single entrypoint, then all the dependencies of the component are bundled together, bringing us back to where we started.

packages/runtime/src/api-handler/handlers/element-tree.ts
```diff
@@ -1,10 +1,10 @@
-import { ReactRuntime } from '../../react'
+import { ReactRuntimeCore } from '../../runtimes/react/react-runtime-core'
```
arvinpoddar, line 1: Internally, change all references to the `ReactRuntime` to instead refer to the `ReactRuntimeCore`, which does not include any registered components.

packages/runtime/src/runtimes/react/react-runtime.ts
```diff
@@ -1,59 +1,10 @@
...
-    )
-
-    return () => {
-      unregisterComponent()
-      unregisterReactComponent()
-    }
-  }
+import { ReactRuntimeCore } from './react-runtime-core'
 
+export class ReactRuntime extends ReactRuntimeCore {
```
arvinpoddar, line 5: Everything in this file is moved as-is to the `ReactRuntimeCore`. This class now only has to extend the core class to include all builtins on instantiation.

PR #1160, opened 2025-09-25
fix: type export declaration order

arvinpoddar:
Per https://publint.dev/rules#exports_types_should_be_first, "types" should be declared first.


agurtovoy (approved):
🙏

PR #1167, opened 2025-09-30
fix: readme to navigate to correct route / URL

drewpledger:
Redirects on the marketing site's next.config redirects to a 404 route


agurtovoy (approved):
🙏

PR #1170, opened 2025-09-30
feat: example inits now prompt user for environment variables

drewpledger:
https://github.com/user-attachments/assets/beee2f88-0ba6-4953-b3af-8232962c9642

I initially experimented with comments being included in the prompt, but realize it didn't look like a good experience. They can refer to the README for more information.

Forgot to include this in the video: the variables are indeed showing up in .env.local


packages/makeswift/src/utils/prompt-env-vars.ts
```diff
@@ -0,0 +1,120 @@
+import chalk from 'chalk'
+import inquirer from 'inquirer'
+import * as fs from 'fs'
+import * as path from 'path'
+
+/**
+ * Extracts environment variable names from a .env.example file.
+ * Looks for lines in the format: VARIABLE_NAME=value or VARIABLE_NAME=
+ */
+function parseEnvExampleFile(filePath: string): string[] {
```
arvinpoddar: Most of `parseEnvExampleFile` and `parseExistingEnvVars` is the same. These could be combined to a single function

packages/makeswift/src/utils/prompt-env-vars.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  const questions = missingVars.map(varName => ({
+    type: 'input',
+    name: varName,
+    message: `${varName}:`,
+    default: '',
+  }))
+
+  const answers = await inquirer.prompt(questions)
+  return answers as Record<string, string>
```
arvinpoddar: What's the actual return type of `inquirer.prompt`? Why do we need to cast?

packages/makeswift/src/utils/prompt-env-vars.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  const userProvidedVars = await promptForEnvVars(missingVars)
+
+  // Build the additional env vars string
+  let additionalEnvVars = ''
+  for (const [key, value] of Object.entries(userProvidedVars)) {
+    if (value.trim()) {
+      additionalEnvVars += `${key}=${value}\n`
+    }
+  }
```
arvinpoddar: Perhaps we do a declarative approach here? Something with a `.map(...).filter(...).join("\n")

arvinpoddar:
Also, I think this change needs a changeset. Not sure what qualifies a breaking change in our CLI, but I don't think this would be one. What do you think @agurtovoy

packages/makeswift/src/init.ts
```diff
@@ -71,7 +72,10 @@ async function init(
     useBun,
   })
 
-  fs.writeFileSync(`${nextAppDir}/.env.local`, envLocal)
+  // Prompt for any missing environment variables required by the example
+  const finalEnvLocal = await promptForMissingEnvVars(nextAppDir, envLocal)
```
agurtovoy: So `performHandshake` returning `envLocal` as a string rather than a record was a smell to me, and it turns out there are two reasons why it does that:
 - It [embeds a `buildLocalEnvFile` call](https://github.com/makeswift/makeswift/blob/fb291c5a25b956a551c9e4356a2d025604b839c8/packages/makeswift/src/utils/handshake.ts#L64) 
 - We (pretend to) receive user-provided env variables from the command line as a list of `VAR=value` pairs, and instead of converting it into a record, and operating on records, we flatten the whole thing into a string.

By "pretend to receive" I mean that the `env` argument in `init` is typed as `string[]`, but, in fact, the comma-separated vars are never spit up and so if user provides more than one env var using this flag, we end up writing it into the `.env.local` file as a single var.

I think we for sure want to:
 -  move the `buildLocalEnvFile` call from `performHandshake` to `init` to replace line 78 below
 -  fix the `--env` option parsing. 
 
As long as we're doing the latter, I'd use `dotenv` to parse the user-provided env into a record instead of an array (e.g. `dotenv.parse(opt.split(",").join("\n"))`).

With these two changes, `performHandshake` can return the combined env as a record, `promptForMissingEnvVars` can accept and return env as a record, and `buildLocalEnvFile` can simply take a record.

agurtovoy: Last but not least, I think this also highlights that we need tests.

packages/makeswift/src/utils/prompt-env-vars.ts
```diff
@@ -0,0 +1,108 @@
...
+import inquirer from 'inquirer'
+import * as fs from 'fs'
+import * as path from 'path'
+
+/**
+ * Parses environment variable content to extract variable names.
+ * Looks for lines in the format: VARIABLE_NAME=value or VARIABLE_NAME=
+ * Skips empty lines and comments.
+ */
+function parseEnvVarFromString(envContent: string): Set<string> {
```
agurtovoy: Since we're not parsing the comments, I'd just use `dotenv.parse` here.

packages/makeswift/src/utils/prompt-env-vars.ts
```diff
@@ -0,0 +1,108 @@
...
+  const missingVars = requiredVars.filter(varName => !existingVars.has(varName))
+
+  if (missingVars.length === 0) {
+    return existingEnvContent
+  }
+
+  const userProvidedVars = await promptForEnvVars(missingVars)
+
+  // Build the additional env vars string
+  let additionalEnvVars = Object.entries(userProvidedVars)
```
agurtovoy: I think we should simply return a combined env record from here and let the function that saves that object to a file deal with formatting.

packages/makeswift/src/utils/prompt-env-vars.ts
agurtovoy: I'd also use the default values in `.env.example` as suggested default values in the corresponding user prompts.

packages/makeswift/src/init.ts
```diff
@@ -71,7 +84,17 @@ async function init(
...
-  fs.writeFileSync(`${nextAppDir}/.env.local`, envLocal)
+  // Prompt for any missing environment variables required by the example
+  const finalEnvVars = await combineMissingEnvVars(nextAppDir, envVars)
+
+  // Build and write the .env.local file
+  const envFileContent = Object.entries(finalEnvVars)
+    .filter(([, value]) => value != null && value !== '')
+    .map(([key, value]) => `${key}=${value}`)
+    .join('\n')
+    .concat('\n')
```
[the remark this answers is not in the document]
drewpledger: POSIX defines a "line" as text ending with a newline character
Many linters/formatters (like Prettier, ESLint) require files to end with a newline
Git will show "No newline at end of file" warnings if it's missing

packages/makeswift/src/init.ts
```diff
@@ -54,11 +64,14 @@ async function init(
   const redirectUrl = new URL(`${MAKESWIFT_APP_ORIGIN}/cli/link-site`)
   redirectUrl.searchParams.set('host_url', nextAppUrl)
 
-  const { envLocal, example } = await performHandshake({
+  const parsedEnv: Record<string, string> =
```
agurtovoy, line 66: Following up on my earlier, perhaps not-too-clear comment:

> By "pretend to receive" I mean that the `env` argument in `init` is typed as `string[]`, but, in fact, the comma-separated vars are never spit up and so if user provides more than one env var using this flag, we end up writing it into the `.env.local` file as a single var.

The `env`/`passedInEnv` is still not populated correctly, at least if you provide the values as requested in the help prompt:

help prompt

env values

I guess it works if you provide the vars separated by spaces:

works with spaces

We should either correct the help prompt, or make it work with a comma-separated list. Don't have a strong opinion, as a format, comma-separated seems easier to explain, but maybe not.

packages/makeswift/src/utils/handshake.ts
```diff
@@ -35,8 +35,9 @@ export async function performHandshake({
   passedInExample && selectSiteUrl.searchParams.set('example', passedInExample)
   template && selectSiteUrl.searchParams.set('template', template)
 
-  if (env.length > 0) {
-    selectSiteUrl.searchParams.set('env_vars', env.join(','))
+  if (Object.keys(env).length > 0) {
+    // Encode env vars as JSON for safe URL transmission
+    selectSiteUrl.searchParams.set('env_vars', JSON.stringify(env))
```
agurtovoy, line 40: Are we making a corresponding change to the builder?

agurtovoy: Or does it just pass `env_vars` back to us without using it?

packages/makeswift/src/init.ts
```diff
@@ -71,7 +84,17 @@ async function init(
     useBun,
   })
 
-  fs.writeFileSync(`${nextAppDir}/.env.local`, envLocal)
+  // Prompt for any missing environment variables required by the example
+  const finalEnvVars = await combineMissingEnvVars(nextAppDir, envVars)
```
agurtovoy: I'd rename `combineMissingEnvVars` to something that indicates that we're prompting the user for the missing vars. Maybe `askForMissingEnvVars`?

packages/makeswift/src/utils/handshake.ts
drewpledger: Confirmed it passes the vars back to us

agurtovoy: It's a bit weird that we send it through to the server then, since we can just consume them below w/o doing that, and packing them into the URL increases the chance of running into a URL length limit.

packages/makeswift/src/init.ts
```diff
@@ -71,7 +83,17 @@ async function init(
...
+  const finalEnvVars = await promptForMissingEnvVars(nextAppDir, envVars)
+
+  // Build and write the .env.local file
+  const envFileContent = Object.entries(finalEnvVars)
+    .filter(([, value]) => value != null && value !== '')
+    .map(([key, value]) => `${key}=${value}`)
+    .join('\n')
+    .concat('\n')
+
+  fs.writeFileSync(`${nextAppDir}/.env.local`, envFileContent)
```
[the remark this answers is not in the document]
agurtovoy: @drewpledger ^^ that's a valid point.

PR #1171, opened 2025-10-01
chore: dissolve prop-controllers entry point

arvinpoddar:
Move the prop-controllers entry point into the builder entry point. Also removes some duplicate exports of core data types from the controls package.


packages/runtime/src/controls/control.ts
```diff
@@ -9,7 +9,7 @@ import { Types as PropControllerTypes } from '@makeswift/prop-controllers'
 
 import { Descriptor, isLegacyDescriptor } from '../prop-controllers/descriptors'
 import { copy as propControllerCopy } from '../prop-controllers/copy'
-import { DELETED_PROP_CONTROLLER_TYPES, PropControllerDescriptor } from '../prop-controllers'
+import { DELETED_PROP_CONTROLLER_TYPES } from '../prop-controllers/deleted'
```
arvinpoddar, line 12: The former `PropControllerDescriptor` import and the `Descriptor` import on L10 are the same.

packages/runtime/src/builder/index.ts
```diff
@@ -22,3 +22,17 @@ export {
 } from '../state/actions'
 
 export { createBaseDocument } from '../state/react-page'
+
+export type {
```
arvinpoddar, line 26: We now expose everything from the `prop-controllers` entry point via the `builder` entry point

packages/runtime/src/builder/serialization/control-serialization.ts
```diff
@@ -44,13 +46,11 @@ import {
 } from '../../controls'
 
 import {
-  Data,
```
arvinpoddar, line 47: These are re-exports from the `@makeswift/controls` package, which should be exposed from `/core`

PR #1246, opened 2026-02-13
feat(controls): add Slider control

drewpledger:
- Introduced a new Slider component that allows users to select single values.
- Implemented necessary props and UI for both single and range sliders in the SliderDemo component.
- Added tests for the Slider functionality to ensure correct behavior and value handling.
- Updated component registration in Makeswift and included the Slider in the controls package.

https://github.com/user-attachments/assets/2f731ab0-966b-4992-b4ec-ec596c6af9b6-


fikrikarim (approved):
Nice. Looks good! 

Update: I actually do have some comments.

packages/controls/src/controls/slider/slider.ts
```diff
@@ -0,0 +1,183 @@
...
+    return safeParse(this.dataSchema, data)
+  }
+
+  fromData(data: DataType | undefined): ValueType | undefined {
+    return match(data)
+      .with(Definition.dataSignature.v1, ({ value }) => value)
+      .otherwise((unversioned) => {
+        if (unversioned === undefined) return undefined
+        return typeof unversioned === 'number' ? unversioned : undefined
+      }) as ValueType | undefined
```
fikrikarim: Is it possible to not do type casting (`as`) here?

packages/controls/src/controls/slider/slider.ts
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    return match(this.version)
+      .with(1, () => ({
+        ...Definition.dataSignature.v1,
+        value: value as number,
+      }))
+      .with(undefined, () => value)
+      .exhaustive() as DataType
+  }
```
fikrikarim: Is it possible to not do type casting (`as`) here? It'd be great if we could remove all/most occurrences of `as` on this PR. We want to avoid using it because it can introduce subtle bugs and defeat the purpose of type checking. We can use the`Number` control as reference on some places that we still need to use the type casting.

fikrikarim (changes requested):

packages/controls/src/controls/slider/slider.ts
```diff
@@ -0,0 +1,183 @@
...
+type Config = z.infer<typeof Definition.schema.config>
+
+class Definition<C extends Config> extends ControlDefinition<
+  typeof Definition.type,
+  C,
+  DataType,
+  ValueType,
+  ResolvedValueType
+> {
+  private static readonly v1DataType = 'slider::v1' as const
```
agurtovoy: As per our recent conversation re: controls interoperability, I think this should be [`number::v1`](https://github.com/makeswift/makeswift/blob/001b965b5cbe32a4566d195f73fb5d5a7197a948/packages/controls/src/controls/number/number.ts#L45)

packages/controls/src/controls/slider/slider.ts
agurtovoy: `fromData` and `toData` -- in fact, everything in this control except for its name and control type (`static readonly type`) should be identical to `Number`.

In fact, it'd probably make sense to inherit from `Number`'s definition and override control type rather than copy & pasting everything.

agurtovoy: I guess there is also `showNumber` prop in the config (which IMHO should be `showInput`); if wrangling TypeScript to make inheritance work turns out to be too much of a hassle, we can leave this for later.

packages/controls/src/controls/slider/slider.ts
```diff
@@ -0,0 +1,183 @@
...
+    })
+
+    const config = z.object({
+      label: z.string().optional(),
+      description: z.string().optional(),
+      defaultValue: z.number().optional(),
+      min: z.number().optional(),
+      max: z.number().optional(),
+      step: z.number().optional(),
+      showNumber: z.boolean().optional(),
```
agurtovoy: I think `showInput` would be a better name here.

andrewreifman:
@drewpledger Small detail, but when hovering over the slider handle, do you mind using the left-right arrow cursor. Right now the arrow cursor appears on drag, but I think it should appear on hover as well like the input.

drewpledger:
> @drewpledger Small detail, but when hovering over the slider handle, do you mind using the left-right arrow cursor. Right now the arrow cursor appears on drag, but I think it should appear on hover as well like the input.

Yeah I can do that in the cosmos PR

packages/controls/src/controls/slider/slider.test.ts
```diff
@@ -0,0 +1,67 @@
...
+  })
+
+  describe('assignability', () => {
+    function assignTest(_def: SliderDefinition) {}
+
+    assignTest(Slider())
+    assignTest(Slider({ label: 'volume' }))
+    assignTest(Slider({ defaultValue: 50 }))
+    assignTest(Slider({ label: 'volume', defaultValue: 50 }))
+    assignTest(Slider({ defaultValue: 0 as number }))
```
agurtovoy: Is `as number` cast necessary here?

packages/controls/src/controls/slider/slider.test.ts
```diff
@@ -0,0 +1,67 @@
+import { testDefinition, testResolveValue } from '../../testing/test-definition'
+
+import { Slider, SliderDefinition } from './slider'
+
+describe('Slider', () => {
+  describe('constructor', () => {
+    test.each([0, 50, 100, undefined])(
+      'call with default value `%s` returns versioned definition',
```
agurtovoy: I'd just say "creates slider with default value `%s`" here; "returns versioned definition" highlights something that doesn't need to be highlighted, as it's true for all constructor tests.

packages/controls/src/controls/slider/slider.ts
```diff
@@ -0,0 +1,204 @@
...
+
+type ReturnedSchemaType<C extends Config> = {
+  definition: typeof Definition.schema.relaxed.definition
+  type: typeof Definition.schema.relaxed.type
+  data: SchemaType<DataType<C>>
+  value: SchemaType<ValueType<C>>
+  resolvedValue: SchemaType<ResolvedValueType<C>>
+}
+
+class Definition<C extends Config> extends ControlDefinition<
```
agurtovoy, line 42: Since we're not unifying this with the `Number` control, I'd leave a comment here along the lines of "Except for its name, control type, and the `showInput` config prop, this control is identical to `Number`. Consider consolidating their implementation in the future."

agurtovoy (dismissed):
Looks good, see inline for a couple of minor points.

fikrikarim (dismissed):

PR #1251, opened 2026-02-24
feat(controls): stabilize IconRadioGroup control

drewpledger:
## Summary

- Renames `unstable_IconRadioGroup` to `IconRadioGroup` and exports the stable API from `@makeswift/controls` and `@makeswift/runtime/controls`.
- Updates all internal usages in runtime slate plugins, tests, and snapshots.
- Adds a demo component in the Next.js app router example app.

https://github.com/user-attachments/assets/70f07aed-07cf-44fc-b80a-67ea60052459

**Companion PR:** [makeswift/cosmos#2391](https://github.com/makeswift/cosmos/pull/2391)


packages/controls/src/controls/icon-radio-group/icons.ts
```diff
@@ -0,0 +1,66 @@
+/**
+ * The full set of icons available to the `IconRadioGroup` control.
+ *
+ * Keys are the public names (e.g. `AlignLeft`); values are the builder's
+ * canonical icon identifiers — the icon component's filename stem, including
+ * the size suffix (`16` or `8`, e.g. `AlignLeft16`, `CaretDown8`).
+ */
```
drewpledger, line 8: Right now there's only 7 of these that are available in the builder side. We are going to need another PR for the rest.

packages/controls/src/controls/icon-radio-group/icons.ts
```diff
@@ -0,0 +1,66 @@
...
+  StarCircle: 'StarCircle16',
+  StarRoundedSquare: 'StarRoundedSquare16',
+  StarSquare: 'StarSquare16',
+  Subscript: 'Subscript16',
+  Sun: 'Sun16',
+  Superscript: 'Superscript16',
+  TextAlignCenter: 'TextAlignCenter16',
+  TextAlignJustify: 'TextAlignJustify16',
+  TextAlignLeft: 'TextAlignLeft16',
+  TextAlignRight: 'TextAlignRight16',
```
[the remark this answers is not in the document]
drewpledger: This is false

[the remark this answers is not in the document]
drewpledger: there are no persisted old icon strings in page data to fail parsing, and the deserialize path is builder-only on freshly-serialized current-code definitions. The transient host/builder runtime version-skew during rollout is inherent to any enum change and acceptable for an unstable_ API.

packages/controls/src/controls/icon-radio-group/icons.ts
arvinpoddar: This means we'd need to update the builder first before merging this PR, correct?

packages/prop-controllers/src/responsive-icon-radio-group/responsive-icon-radio-group.ts
arvinpoddar, line 1: Do we need to update this? This looks like a legacy prop controller, not the actual IconRadioGroup control

drewpledger: I opted to keep it consistent, I can remove this.

packages/controls/src/controls/icon-radio-group/icons.ts
drewpledger: Ideally, but not necessary

packages/runtime/src/state/__tests__/fixtures/serialized-descriptors-from-builder.ts
```diff
@@ -296,10 +296,10 @@ export const getMockedSerializedDescriptorsFromBuilder: (createMessagePort: () =
...
                       label: 'Alignment',
                       options: [
-                        { icon: 'TextAlignLeft', label: 'Left Align', value: 'left' },
-                        { icon: 'TextAlignCenter', label: 'Center Align', value: 'center' },
-                        { icon: 'TextAlignRight', label: 'Right Align', value: 'right' },
-                        { icon: 'TextAlignJustify', label: 'Justify', value: 'justify' },
+                        { icon: 'TextAlignLeft16', label: 'Left Align', value: 'left' },
+                        { icon: 'TextAlignCenter16', label: 'Center Align', value: 'center' },
+                        { icon: 'TextAlignRight16', label: 'Right Align', value: 'right' },
+                        { icon: 'TextAlignJustify16', label: 'Justify', value: 'justify' },
```
arvinpoddar: Why did these values change?

packages/controls/src/controls/icon-radio-group/icons.ts
drewpledger: It would be an additional PR because we need this prior to stabilizing the builder's RadioGroup

packages/controls/src/controls/icon-radio-group/icons.ts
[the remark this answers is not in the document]
arvinpoddar: I'd double check here. You're right that an enum change would be acceptable for an unstable API, _but_ we are using this unstable API inside of stable ones, like RichText. I'd check to see that those panels aren't broken by this change.

packages/runtime/src/state/__tests__/fixtures/serialized-descriptors-from-builder.ts
arvinpoddar: I think my concern here relates to https://github.com/makeswift/makeswift/pull/1251/changes#r3390073356

packages/controls/src/controls/icon-radio-group/icons.ts
arvinpoddar: Or I guess we don't need to block merging this PR, because we still wouldn't have released the runtime quite yet. So to confirm, I'm assuming the sequence of operations here would be:

1. Merge this PR
2. Update builder to use latest canary runtime, and update the icon support
3. Eventually release latest runtime version as stable.

Is that what you had in mind?

drewpledger: Precisely my thoughts

packages/controls/src/controls/icon-radio-group/icons.ts
```diff
@@ -0,0 +1,66 @@
...
+  HeightAuto: 'HeightAuto16',
+  HeightMatch: 'HeightMatch16',
+  Moon: 'Moon16',
+  Percent: 'Percent16',
+  Plus: 'Plus8',
+  Px: 'Px16',
+  RadioButton: 'RadioButton16',
+  RepeatNone: 'RepeatNone16',
+  RepeatX: 'RepeatX16',
+  RepeatXy: 'RepeatXy16',
```
arvinpoddar: I'm also not really sure I like the idea of us using the exact builder icon name in the runtime. Seems like an unnecessary coupling, I'd think we'd just have the names here:

```
{
 ...
ShadowInner: 'ShadowInner',
}
```

And then in the builder have a mapping from `ShadowInner` -> `ShadowInner16`. This is what we do with Component Icons.

fikrikarim: Agree with Arvin here

packages/runtime/src/state/__tests__/fixtures/serialized-descriptors-from-builder.ts
```diff
@@ -318,9 +318,9 @@ export const getMockedSerializedDescriptorsFromBuilder: (createMessagePort: () =
                 config: {
                   label: 'Inline',
                   options: [
-                    { icon: 'Superscript16', label: 'Superscript', value: 'superscript' },
-                    { icon: 'Subscript16', label: 'Subscript', value: 'subscript' },
-                    { icon: 'Code16', label: 'Code', value: 'code' },
```
arvinpoddar, line 323: Just a callout for the builder PR: we're going to need to continue supporting these mappings so we don't break the panels for the older versions of the runtime.

packages/controls/src/controls/icon-radio-group/icon-radio-group.test.ts
```diff
@@ -1,6 +1,14 @@
...
 
-import { unstable_IconRadioGroup as IconRadioGroup } from './icon-radio-group'
+import { IconRadioGroup, IconRadioGroupDefinition } from './icon-radio-group'
+
+function serializedDefinition(options: unknown): SerializedRecord {
+  return {
+    type: IconRadioGroupDefinition.type,
+    config: { options },
+  } as any as SerializedRecord
+}
```
arvinpoddar: Is there not a serialize function we can reuse in this package rather than having to use this?

arvinpoddar (dismissed):

PR #1252, opened 2026-02-25
feat(controls): add Code control

drewpledger:
Adds an `unstable_Code` control for authoring code snippets in Makeswift. It holds a single string value. The exports all carry the `unstable_` prefix since we're not committing to the API shape yet — in particular, the resolved shape is intentionally `{ value }` (not bare `string`) so we have room to append metadata like `language` (or `languages`) in a follow-up without a breaking change.

`resolveValue` returns `{ value }` and the result is memoized, so `readStable` gives back the same reference across calls and React doesn't see spurious changes.

Persisted data goes through a `code::v1` envelope (`{ '@@makeswift/type': 'code::v1', value }`) so we have room to migrate later. Since this is brand-new, there's no unversioned legacy to handle — the version is pinned to literal `1` and `toData` is just a conditional.

The control is wired into the visitor pattern (`DefinitionVisitor`, `ControlSerializationVisitor`, `MergeTranslationsVisitor`) and the deserialization registry, and there's a demo in the Next.js app router example with a realistic multi-line default so it's easy to poke at end-to-end.

https://github.com/user-attachments/assets/5982aa7e-89cb-4a78-9a07-46cbf9fdcb17

(Shoulda demo'd intellisense :C)

I have verified this works with both composite controls (List and Group)


packages/controls/src/controls/code/code.ts
```diff
@@ -0,0 +1,194 @@
...
+    })
+
+    const schemas = <V, D>(value: SchemaType<V>, data: SchemaType<D>) => {
+      const type = z.literal(this.type)
+
+      const config = z.object({
+        label: z.string().optional(),
+        description: z.string().optional(),
+        defaultValue: value,
+        languages: z.array(z.string()).optional(),
```
arvinpoddar: Shouldn't this be a `union` of known literals? This config would support any strings passed for `languages`, even if they're not real languages or not supported by our builder IDE.

packages/controls/src/controls/code/code.ts
```diff
@@ -0,0 +1,194 @@
...
+
+  toData(value: ValueType<C>): DataType<C> {
+    return match({ version: this.version, value })
+      .with({ version: 1, value: P.string }, ({ value }) => ({
+        ...Definition.dataSignature.v1,
+        value,
+      }))
+      .with({ version: 1, value: undefined }, () => undefined)
+      .with({ version: undefined }, ({ value }) => value)
+      .otherwise(() => value)
```
arvinpoddar: Do we need this? We only have 1 version, which we always define. This would only be necessary if we didn't have a version defined with an earlier version of this control.

packages/runtime/src/builder/serialization/control-serialization.ts
```diff
@@ -1000,33 +999,23 @@ export function deserializeLegacyControl<T extends Data>(
...
-    const record = options?.plugins ? deserializeObject(serializedControl, options?.plugins) : serializedControl
-    return deserializeLegacyControl(record as SerializedLegacyControl<T>)
+    return deserializeLegacyControl(serializedControl)
   }
 
-  const plugins = [functionDeserializationPlugin, ...(options?.plugins ?? [])]
-
   return deserializeUnifiedControlDef(
-    deserializeRecord(serializedControl, plugins)
+    deserializeRecord(serializedControl, [functionDeserializationPlugin]),
```
arvinpoddar: Why is this change here?

packages/runtime/src/controls/index.ts
```diff
@@ -4,6 +4,8 @@ export {
   ControlDefinition,
   Checkbox,
   CheckboxDefinition,
+  Code,
+  CodeDefinition,
```
arvinpoddar: Let's export these as `unstable_`

packages/controls/src/controls/code/code.ts
arvinpoddar: On this note, should this config even support multiple languages? Why not just support one?

packages/controls/src/controls/code/code.test.ts
```diff
@@ -0,0 +1,64 @@
...
+      })
+    })
+  })
+
+  describe('assignability', () => {
+    function assignTest(_def: CodeDefinition) {}
+    assignTest(Code())
+    assignTest(Code({ label: 'Code' }))
+    assignTest(Code({ defaultValue: '<div>hello</div>' }))
+    assignTest(Code({ defaultValue: 'text' as string }))
```
arvinpoddar: Why the cast?

drewpledger:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

fikrikarim:
Do we plan to squash the commits? (If yes, we can do it when we merge)

fikrikarim (dismissed):
Nice work! I think we should squash and merge so we can easily inspect the commit later.

packages/controls/src/testing/test-definition.ts
```diff
@@ -133,20 +134,22 @@ export function testDefinition<Def extends ControlDefinition>(
         },
       )
 
-      test.each(values)(
-        'returns v0 value for `%s` when definition is unversioned',
-        (value) => {
-          // Arrange
-          const v0Definition = toV0(definition)
+      if (!options.skipV0Definition) {
```
arvinpoddar, line 137: What's this and why do we need it?

arvinpoddar (commented):
I noticed we don't have type tests (`{control name}.test.types.ts`). Probably a good idea to add them so we can be more intentional about the API as we iterate on it.

packages/controls/src/testing/test-definition.ts
drewpledger: The shared helper has a v0 round-trip block that calls `deserialize` on a record without a `version` field. Now that Code's `version` is `z.literal(1)` (non-optional, per your earlier note that we always define it), that parse throws. Code is brand-new and never had unversioned data, so the v0 case isn't meaningful for it — the option lets us opt out of that block.

PR #1272, opened 2026-04-01
fix: global resource cache is shared across draft and live requests

agurtovoy:
Scope the runtime's store and host API client by site version + locale.

Demo/smoke test:

https://github.com/user-attachments/assets/49cfa37a-a791-4109-8dbe-e5fdb17653bb

(Correction for the video: "runtime version 0.17" -> "runtime version 0.27")

### DIFF
--- a/packages/hono-react/src/runtime.ts
+++ b/packages/hono-react/src/runtime.ts
@@ -1,7 +1,16 @@
-import { type Breakpoints, ReactRuntime } from '@makeswift/runtime/unstable-framework-support'
+import {
+  type Breakpoints,
+  type StoreKey,
+  ReactRuntime,
+} from '@makeswift/runtime/unstable-framework-support'
 
 export class HonoReactRuntime extends ReactRuntime {
-  constructor(args: { appOrigin?: string; apiOrigin?: string; breakpoints?: Breakpoints } = {}) {
+  constructor(args: {
+    requestKey?: StoreKey
+    appOrigin?: string
+    apiOrigin?: string
+    breakpoints?: Breakpoints
+  }) {
     super({
       ...args,
       fetch: (url, init) => fetch(url, init), // TODO: revalidation support
--- a/packages/runtime/src/api/client.ts
+++ b/packages/runtime/src/api/client.ts
@@ -61,16 +61,14 @@ export class MakeswiftHostApiClient {
   constructor({
     uri,
     fetch,
-    cacheData,
+    preloadedState,
   }: {
     uri: string
     fetch: MakeswiftApiClient.HttpFetch
-    cacheData?: CacheData
+    preloadedState: Partial<MakeswiftApiClient.State>
   }) {
     this.graphqlClient = new GraphQLClient(uri)
-    this.makeswiftApiClient = MakeswiftApiClient.configureStore({
-      serializedState: cacheData,
-    })
+    this.makeswiftApiClient = MakeswiftApiClient.configureStore({ preloadedState })
     this.fetch = fetch
     this.subscribe = this.makeswiftApiClient.subscribe
   }
--- a/packages/runtime/src/client/index.ts
+++ b/packages/runtime/src/client/index.ts
@@ -335,10 +335,12 @@ const getPageAPISchema = z.object({
 type GetPageAPI = z.infer<typeof getPageAPISchema>
 
 const getFontsAPISchema = z.object({
-  googleFonts: z.array(z.object({
-    family: z.string(),
-    variants: z.array(z.string()),
-  })),
+  googleFonts: z.array(
+    z.object({
+      family: z.string(),
+      variants: z.array(z.string()),
+    }),
+  ),
   siteId: z.string(),
 })
 
@@ -534,8 +536,7 @@ export class MakeswiftClient {
     siteVersion: SiteVersion | null,
     locale: string | null,
   ): Promise<CacheData> {
-    const runtime = this.runtime
-    const descriptors = getPropControllerDescriptors(runtime.store.getState())
+    const descriptors = this.getElementDescriptors()
     const swatchIds = new Set<string>()
     const fileIds = new Set<string>()
     const typographyIds = new Set<string>()
@@ -980,18 +981,11 @@ export class MakeswiftClient {
   }
 
   getTranslatableData(elementTree: ElementData): Record<string, Data> {
-    return getTranslatableContent(
-      getPropControllerDescriptors(this.runtime.store.getState()),
-      elementTree,
-    )
+    return getTranslatableContent(this.getElementDescriptors(), elementTree)
   }
 
   mergeTranslatedData(elementTree: ElementData, translatedData: Record<string, Data>): Element {
-    return mergeTranslatedContent(
-      getPropControllerDescriptors(this.runtime.store.getState()),
-      elementTree,
-      translatedData,
-    )
+    return mergeTranslatedContent(this.getElementDescriptors(), elementTree, translatedData)
   }
 
   async readPreviewToken(token: string): Promise<PreviewTokenPayload | null> {
@@ -1034,9 +1028,7 @@ export class MakeswiftClient {
     return parsed.data
   }
 
-  async unstable_getFonts(
-    siteVersion: SiteVersion | null = null,
-  ): Promise<GetFontsAPI | null> {
+  async unstable_getFonts(siteVersion: SiteVersion | null = null): Promise<GetFontsAPI | null> {
     const response = await this.fetch('v1_unstable/fonts', siteVersion)
 
     if (!response.ok) {
@@ -1061,5 +1053,9 @@ export class MakeswiftClient {
     }
 
     return parsed.data
+  }
+
+  private getElementDescriptors() {
+    return getPropControllerDescriptors(this.runtime.protoStore.getState())
   }
 }
--- a/packages/runtime/src/runtimes/react/components/RuntimeProvider.tsx
+++ b/packages/runtime/src/runtimes/react/components/RuntimeProvider.tsx
@@ -1,11 +1,12 @@
 'use client'
 
-import { ReactNode, useMemo } from 'react'
+import { ReactNode, useEffect, useMemo } from 'react'
 
 import { type SiteVersion } from '../../../api/site-version'
 
 import { ReactRuntimeContext } from '../hooks/use-react-runtime'
 import { useAsyncEffect } from '../hooks/use-async-effect'
+import { StoreContext } from '../hooks/use-store'
 import { type ReactRuntimeCore } from '../react-runtime-core'
 
 import { PreviewSwitcher } from './preview-switcher/preview-switcher'
@@ -23,25 +24,28 @@ export function RuntimeProvider({
 }) {
   const siteVersion = siteVersionProp ?? null
   const isPreview = siteVersion != null
-  // TODO: we need to decouple editability from the site version; specifically, previewing
-  // a draft version of the site should not lead to switching to the read-write state
-  const isReadOnly = siteVersion == null
 
-  // setting idempotent part of the state on render to have parity between server-side and
-  // client-side rendering depending on this state; `useMemo` here is purely a performance
-  // optimization; prior art for performing idempotent side effects on render:
-  // https://github.com/TanStack/query/blob/8f9f183f11df3709a1a38c4efce1452788041f88/packages/react-query/src/HydrationBoundary.tsx#L41
-  useMemo(
-    () => runtime.setIdempotent({ siteVersion, isReadOnly, locale }),
-    [runtime, siteVersion, isReadOnly, locale],
+  // see `getOrCreateStore` for the description of the stores' lifecycle
+  const store = useMemo(
+    () => runtime.getOrCreateStore({ siteVersion, locale }),
+    [runtime, siteVersion, locale],
   )
 
-  useAsyncEffect(() => runtime.setupStore({ isReadOnly }), [runtime, isReadOnly])
+  useEffect(() => {
+    runtime.retainStore({ siteVersion, locale }, store)
+    return () => runtime.releaseStore({ siteVersion, locale }, store)
+  }, [runtime, siteVersion, locale, store])
+
+  // if we're in the read-write mode, the reducers & middleware required for builder
+  // interactions are loaded only on client side, lazily and asynchronously
+  useAsyncEffect(() => store.loadReadWriteStateIfNeeded(), [store])
 
   return (
     <ReactRuntimeContext.Provider value={runtime}>
-      {children}
-      <PreviewSwitcher isPreview={isPreview} />
+      <StoreContext.Provider value={store}>
+        {children}
+        <PreviewSwitcher isPreview={isPreview} />
+      </StoreContext.Provider>
     </ReactRuntimeContext.Provider>
   )
 }
--- a/packages/runtime/src/runtimes/react/hooks/use-store.ts
+++ b/packages/runtime/src/runtimes/react/hooks/use-store.ts
@@ -1,8 +1,15 @@
 'use client'
 
+import { createContext, useContext } from 'react'
 import { type Store } from '../../../state/store'
-import { useReactRuntime } from './use-react-runtime'
+
+export const StoreContext = createContext<Store | null>(null)
 
 export function useStore(): Store {
-  return useReactRuntime().store
+  const store = useContext(StoreContext)
+  if (store == null) {
+    throw new Error('`useStore` must be used within a `StoreProvider`')
+  }
+
+  return store
 }
--- a/packages/runtime/src/runtimes/react/host-api-client.tsx
+++ b/packages/runtime/src/runtimes/react/host-api-client.tsx
@@ -2,8 +2,8 @@
 
 import { MakeswiftHostApiClient } from '../../api/client'
 
-import { useReactRuntime } from './hooks/use-react-runtime'
+import { useStore } from './hooks/use-store'
 
 export function useMakeswiftHostApiClient(): MakeswiftHostApiClient {
-  return useReactRuntime().hostApiClient
+  return useStore().hostApiClient
 }
--- a/packages/runtime/src/runtimes/react/react-runtime-core.ts
+++ b/packages/runtime/src/runtimes/react/react-runtime-core.ts
@@ -55,7 +55,7 @@ export class ReactRuntimeCore extends RuntimeCore {
   ): () => void {
     validateComponentType(type, component as unknown as ComponentType)
 
-    const unregisterComponent = this.store.dispatch(
+    const unregisterComponent = this.protoStore.dispatch(
       registerComponentEffect(
         type,
         { label, icon, hidden, description, builtinSuspense },
@@ -69,7 +69,7 @@ export class ReactRuntimeCore extends RuntimeCore {
       )
     }
 
-    const unregisterReactComponent = this.store.dispatch(
+    const unregisterReactComponent = this.protoStore.dispatch(
       registerReactComponentEffect(type, component as unknown as ComponentType),
     )
 
--- a/packages/runtime/src/runtimes/react/runtime-core.ts
+++ b/packages/runtime/src/runtimes/react/runtime-core.ts
@@ -13,78 +13,146 @@ import {
 import { copyElementTree } from '../../state/ops/copy-element-tree'
 
 import { getBreakpoints, type Element, type ElementData } from '../../state/read-only-state'
-import { configureStore, type Store } from '../../state/store'
-import { setLocale } from '../../builder'
 import {
-  resetLocaleState,
-  setIsReadOnly,
-  setSiteVersion,
-} from '../../state/actions/internal/read-only-actions'
+  configureProtoStore,
+  configureReadWriteStore,
+  type ProtoStore,
+  type Store,
+} from '../../state/store'
+
+import { RefCountedMap } from '../../utils/ref-counted-map'
+import { isServer } from '../../utils/is-server'
+
+export type StoreKey = {
+  siteVersion: SiteVersion | null
+  locale: string | undefined
+}
+
+const VERSION_TAG_LIVE = 'ref:live'
 
 export class RuntimeCore {
-  readonly store: Store
-  readonly hostApiClient: MakeswiftHostApiClient
+  // The unowned entry TTL affects performance, not correctness. The TTL controls how long an unretained store stays
+  // in the map. If an entry expires and is removed from the map before React commits, the store remains retained
+  // via `<StoreContext.Provider>`. The only impact is that future lookups will create a new store instance instead of
+  // reusing the existing one, reducing cache efficiency.
+  private readonly activeStores = new RefCountedMap<string | null, Store>({
+    unownedEntryTtlMs: 1000,
+    // Checking on retain/release is sufficient on the client, and we don't need to check on get on the server
+    // as the only scenario in which we add the store to the map is when the runtime is already bound to the
+    // requested site version, which should be the only site version for which the store is requested
+    ttlCheck: RefCountedMap.TTLCheck.ON_RETAIN | RefCountedMap.TTLCheck.ON_RELEASE,
+  })
+
+  readonly protoStore: ProtoStore
   readonly appOrigin: string
   readonly apiOrigin: string
+  readonly requestKey: StoreKey | undefined
+  readonly fetch: MakeswiftApiClient.HttpFetch
 
   constructor({
     appOrigin = 'https://app.makeswift.com',
     apiOrigin = 'https://api.makeswift.com',
     breakpoints,
+    requestKey,
     fetch,
   }: {
     appOrigin?: string
     apiOrigin?: string
     breakpoints?: BreakpointsInput
+    requestKey?: StoreKey
     fetch: MakeswiftApiClient.HttpFetch
   }) {
     this.appOrigin = validateOrigin(appOrigin, 'appOrigin')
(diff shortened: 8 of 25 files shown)


packages/runtime/src/runtimes/react/runtime-core.ts
```diff
@@ -13,79 +13,129 @@ import {
...
+    return copyElementTree(this.protoStore.getState(), elementTree, replacementContext)
   }
 
   getBreakpoints(): Breakpoints {
-    return getBreakpoints(this.store.getState())
+    return getBreakpoints(this.protoStore.getState())
+  }
+
+  private shouldUsePersistentStore(key: string | null): boolean {
+    return !isServer() || this.siteVersion?.version === key
```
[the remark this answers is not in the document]
agurtovoy: Fixed

packages/hono-react/src/runtime.ts
```diff
@@ -1,7 +1,16 @@
...
+import {
+  type Breakpoints,
+  type StoreKey,
+  ReactRuntime,
+} from '@makeswift/runtime/unstable-framework-support'
 
 export class HonoReactRuntime extends ReactRuntime {
-  constructor(args: { appOrigin?: string; apiOrigin?: string; breakpoints?: Breakpoints } = {}) {
+  constructor(args: {
+    requestKey?: StoreKey
```
agurtovoy, line 9: Add an option to scope the runtime state by request, see below.

packages/runtime/src/api/client.ts
```diff
@@ -61,16 +61,14 @@ export class MakeswiftHostApiClient {
   constructor({
     uri,
     fetch,
-    cacheData,
+    preloadedState,
   }: {
     uri: string
     fetch: MakeswiftApiClient.HttpFetch
-    cacheData?: CacheData
+    preloadedState: Partial<MakeswiftApiClient.State>
```
agurtovoy, line 68: Replace host API client's unused `cacheData` argument with `preloadedState` to enable proper initialization of the client's store.

packages/runtime/src/client/index.ts
```diff
@@ -335,10 +335,12 @@ const getPageAPISchema = z.object({
...
-  googleFonts: z.array(z.object({
-    family: z.string(),
-    variants: z.array(z.string()),
-  })),
+  googleFonts: z.array(
+    z.object({
+      family: z.string(),
+      variants: z.array(z.string()),
+    }),
+  ),
```
agurtovoy, line 343: Prettier reformat, no semantic changes.

packages/runtime/src/client/index.ts
```diff
@@ -534,8 +536,7 @@ export class MakeswiftClient {
     siteVersion: SiteVersion | null,
     locale: string | null,
   ): Promise<CacheData> {
-    const runtime = this.runtime
-    const descriptors = getPropControllerDescriptors(runtime.store.getState())
+    const descriptors = this.getElementDescriptors()
```
agurtovoy, line 539: Here and below, `getPropControllerDescriptors(runtime.store.getState())` -> `this.getElementDescriptors()`

packages/runtime/src/client/index.ts
```diff
@@ -1034,9 +1028,7 @@ export class MakeswiftClient {
     return parsed.data
   }
 
-  async unstable_getFonts(
-    siteVersion: SiteVersion | null = null,
-  ): Promise<GetFontsAPI | null> {
+  async unstable_getFonts(siteVersion: SiteVersion | null = null): Promise<GetFontsAPI | null> {
```
agurtovoy, line 1031: Prettier reformat, no semantic changes.

packages/runtime/src/client/index.ts
```diff
@@ -1062,4 +1054,8 @@ export class MakeswiftClient {
 
     return parsed.data
   }
+
+  private getElementDescriptors() {
+    return getPropControllerDescriptors(this.runtime.protoStore.getState())
+  }
```
agurtovoy, line 1060: Here and below,
`runtime.store` -> `runtime.protoStore`

packages/runtime/src/runtimes/react/components/RuntimeProvider.tsx
```diff
@@ -23,25 +24,28 @@ export function RuntimeProvider({
...
-  // client-side rendering depending on this state; `useMemo` here is purely a performance
-  // optimization; prior art for performing idempotent side effects on render:
-  // https://github.com/TanStack/query/blob/8f9f183f11df3709a1a38c4efce1452788041f88/packages/react-query/src/HydrationBoundary.tsx#L41
-  useMemo(
-    () => runtime.setIdempotent({ siteVersion, isReadOnly, locale }),
-    [runtime, siteVersion, isReadOnly, locale],
+
+  // see `getOrCreateStore` for the description of the stores' lifecycle
+  const store = useMemo(
+    () => runtime.getOrCreateStore({ siteVersion, locale }),
```
agurtovoy, line 30: Obtain an existing or create a new store for the provided site version and locale. See below for the full explanation of the store lifecycle.

packages/runtime/src/runtimes/react/components/RuntimeProvider.tsx
```diff
@@ -23,25 +24,28 @@ export function RuntimeProvider({
...
+    [runtime, siteVersion, locale],
   )
 
-  useAsyncEffect(() => runtime.setupStore({ isReadOnly }), [runtime, isReadOnly])
+  useEffect(() => {
+    runtime.retainStore({ siteVersion, locale }, store)
+    return () => runtime.releaseStore({ siteVersion, locale }, store)
```
agurtovoy, line 36: When running on the client, tie the store's lifetime to the provider's lifetime.

packages/runtime/src/runtimes/react/components/RuntimeProvider.tsx
```diff
@@ -23,25 +24,28 @@ export function RuntimeProvider({
...
+  }, [runtime, siteVersion, locale, store])
+
+  // if we're in the read-write mode, the reducers & middleware required for builder
+  // interactions are loaded only on client side, lazily and asynchronously
+  useAsyncEffect(() => store.loadReadWriteStateIfNeeded(), [store])
```
agurtovoy, line 41: NB

packages/runtime/src/runtimes/react/components/RuntimeProvider.tsx
```diff
@@ -23,25 +24,28 @@ export function RuntimeProvider({
...
 
   return (
     <ReactRuntimeContext.Provider value={runtime}>
-      {children}
-      <PreviewSwitcher isPreview={isPreview} />
+      <StoreContext.Provider value={store}>
```
agurtovoy, line 45: Since we don't have a singleton store anymore, we have to resurrect the store provider.

packages/runtime/src/runtimes/react/hooks/use-store.ts
```diff
@@ -1,8 +1,15 @@
...
-import { useReactRuntime } from './use-react-runtime'
+
+export const StoreContext = createContext<Store | null>(null)
 
 export function useStore(): Store {
-  return useReactRuntime().store
+  const store = useContext(StoreContext)
+  if (store == null) {
+    throw new Error('`useStore` must be used within a `StoreProvider`')
+  }
```
agurtovoy, line 12: See my comment above.

packages/runtime/src/runtimes/react/host-api-client.tsx
```diff
@@ -2,8 +2,8 @@
 
 import { MakeswiftHostApiClient } from '../../api/client'
 
-import { useReactRuntime } from './hooks/use-react-runtime'
+import { useStore } from './hooks/use-store'
 
 export function useMakeswiftHostApiClient(): MakeswiftHostApiClient {
-  return useReactRuntime().hostApiClient
+  return useStore().hostApiClient
```
agurtovoy, line 8: Host API client is tied to the current store.

packages/runtime/src/runtimes/react/runtime-core.ts
```diff
@@ -13,79 +13,144 @@ import {
...
+  // in the map. If an entry expires and is removed from the map before React commits, the store remains retained
+  // via `<StoreContext.Provider>`. The only impact is that future lookups will create a new store instance instead of
+  // reusing the existing one, reducing cache efficiency.
+  private readonly activeStores = new RefCountedMap<string | null, Store>({
+    unownedEntryTtlMs: 1000,
+    // Checking on retain/release is sufficient on the client, and we don't need to check on get on the server
+    // as the only scenario in which we add the store to the map is when the runtime is already bound to the
+    // requested site version, which should be the only site version for which the store is requested
+    ttlCheck: RefCountedMap.TTLCheck.ON_RETAIN | RefCountedMap.TTLCheck.ON_RELEASE,
+  })
```
agurtovoy, line 44: Replace a single store instance with a "proto" store that holds component registrations + active stores cache.

packages/runtime/src/runtimes/react/runtime-core.ts
```diff
@@ -13,79 +13,144 @@ import {
...
     appOrigin = 'https://app.makeswift.com',
     apiOrigin = 'https://api.makeswift.com',
     breakpoints,
+    requestKey,
     fetch,
   }: {
     appOrigin?: string
     apiOrigin?: string
     breakpoints?: BreakpointsInput
+    requestKey?: StoreKey
```
agurtovoy, line 62: Providing an optional `requestKey` when creating a runtime instance ties it to the specific request -- see below for more details.

packages/runtime/src/runtimes/react/runtime-core.ts
```diff
@@ -13,79 +13,144 @@ import {
...
+export type StoreKey = {
+  siteVersion: SiteVersion | null
+  locale: string | undefined
+}
+
+const VERSION_TAG_LIVE = 'ref:live'
 
 export class RuntimeCore {
-  readonly store: Store
-  readonly hostApiClient: MakeswiftHostApiClient
```
agurtovoy, line 26: Host API clients are also versioned and tied to a corresponding store now.

packages/runtime/src/runtimes/react/runtime-core.ts
```diff
@@ -13,79 +13,144 @@ import {
...
-    locale,
-  }: {
-    siteVersion: SiteVersion | null
-    isReadOnly: boolean
-    locale: string | undefined
-  }): void {
-    this.store.dispatch(setSiteVersion(siteVersion))
-    this.store.dispatch(setIsReadOnly(isReadOnly))
-    this.store.dispatch(locale != null ? setLocale(new Intl.Locale(locale)) : resetLocaleState())
+  getOrCreateStore({ siteVersion, locale }: StoreKey): Store {
```
agurtovoy, line 76: Stores are versioned on site version + locale

packages/runtime/src/utils/ref-counted-map.ts
```diff
@@ -0,0 +1,116 @@
...
+  ON_RETAIN: 1 << 1,
+  ON_RELEASE: 1 << 2,
+} as const
+
+/**
+ * Ref-counted map with explicit retention semantics. Entries are unowned by default and may be evicted after their
+ * TTL unless explicitly retained. This provides TTL-based caching for unowned entries, while allowing consumers to
+ * opt into stronger lifetime control via explicit retention.
+ */
+export class RefCountedMap<K, V> {
```
agurtovoy, line 25: ^^ NB

packages/runtime/src/runtimes/react/runtime-core.ts
agurtovoy: See also [`RefCountedMap`](https://github.com/makeswift/makeswift/pull/1272/changes#r3076794668)

packages/runtime/src/runtimes/react/runtime-core.ts
```diff
@@ -13,79 +13,144 @@ import {
...
+
+    // On the server, stores are ephemeral by default so SSR does not retain them in a long-lived runtime.
+    // The exception is a runtime that is already bound to the requested site version, where we can safely
+    // reuse the store across multiple root regions and benefit from the host API client's in-memory cache.
+    //
+    // On the client, stores are reference-counted by site version so multiple root regions can share a store
+    // instance, which preserves the pre-v0.17 per-site-version store behavior without requiring a separate
+    // runtime per root.
+    const usePersistentStore = this.shouldUsePersistentStore(key)
+    return usePersistentStore ? this.activeStores.getOrCreate(key, createStore) : createStore()
```
agurtovoy, line 108: NB ^^

packages/runtime/src/runtimes/react/runtime-core.ts
```diff
@@ -13,79 +13,147 @@ import {
...
 
   getBreakpoints(): Breakpoints {
-    return getBreakpoints(this.store.getState())
+    return getBreakpoints(this.protoStore.getState())
+  }
+
+  private shouldUsePersistentStore(key: string): boolean {
+    // don't persist stores on the server unless the runtime instance is bound to a request
+    // and the requested store matches the request's site version and locale
+    return !isServer() || (this.requestKey !== undefined && storeCacheKey(this.requestKey) === key)
```
agurtovoy, line 147: NB ^^

packages/runtime/src/state/__tests__/store.read-write-state.test.ts
```diff
@@ -72,29 +81,12 @@ describe('loadReadWriteState', () => {
...
-
-    expect(builderProxyMock).toHaveBeenCalledTimes(1)
-    expect(createReadWriteMiddlewareMock).toHaveBeenCalledTimes(1)
-    expect(teardownSpy).not.toHaveBeenCalled()
-
-    unloadB()
-    expect(teardownSpy).toHaveBeenCalledTimes(1)
-    expect(consoleError).not.toHaveBeenCalled()
-  })
-
```
agurtovoy, line 91: Remove obsolete test (there is no switching between read-write and read-only state anymore).

packages/runtime/src/runtimes/react/runtime-core.ts
```diff
@@ -13,79 +13,147 @@ import {
...
-    this.store.dispatch(locale != null ? setLocale(new Intl.Locale(locale)) : resetLocaleState())
+  getOrCreateStore({ siteVersion, locale }: StoreKey): Store {
+    const key = storeCacheKey({ siteVersion, locale })
+
+    const createStore = () => {
+      // host API client includes an in-memory cache of the site's resources and thus also has to be versioned
+      const hostApiClient = new MakeswiftHostApiClient({
+        uri: new URL('graphql', this.apiOrigin).href,
+        fetch: this.fetch,
+        preloadedState: { siteVersion, locale },
```
agurtovoy, line 84: Rather than being "initialized" by dispatching an action, both the host API store and the runtime stores are now directly populated with their initial site version and locale.

packages/runtime/src/runtimes/react/runtime-core.ts
```diff
@@ -13,79 +13,147 @@ import {
...
+
+      // TODO: we need to decouple editability from the site version; specifically, previewing
+      // a draft version of the site should not lead to switching to the read-write state
+      const isReadOnly = siteVersion == null
+
+      return configureReadWriteStore({
+        name: `Runtime read-write store (site version: ${key})`,
+        appOrigin: this.appOrigin,
+        hostApiClient,
+        preloadedState: { ...this.protoStore.getState(), siteVersion, isReadOnly, locale },
```
agurtovoy, line 95: Same as above.

packages/runtime/src/state/__tests__/store.read-write-state.test.ts
```diff
@@ -109,19 +101,4 @@ describe('loadReadWriteState', () => {
...
-    const unloadB = await store.loadReadWriteState({ isReadOnly: true })
-
-    expect(consoleError).toHaveBeenCalledWith('Read-write state mismatch', {
-      isReadOnly: true,
-      refCount: 1,
-    })
-
-    unloadA()
-    unloadB()
-  })
```
agurtovoy, line 126: Another obsolete test.

packages/runtime/src/state/__tests__/store.read-write-state.test.ts
```diff
@@ -25,19 +25,28 @@ jest.mock('../read-write-state', () => ({
...
     } as any,
-    preloadedState: null,
+    preloadedState: {
+      isReadOnly,
+      locale,
+    },
   })
 
-describe('loadReadWriteState', () => {
+describe('read-write store state', () => {
```
agurtovoy, line 49: Here and below, update tests to reflect the new store semantics (locale and read-only state are fixed at creation).

packages/runtime/src/state/middleware/read-write/builder-api/initialize-connection.ts
```diff
@@ -309,6 +309,10 @@ export function initializeBuilderConnection(
 
     const breakpoints = ReadOnlyState.getBreakpoints(getState())
     dispatch(Builder.setBreakpoints(breakpoints))
+
+    const locale = ReadOnlyState.getLocale(getState())
+    if (locale != null) dispatch(Builder.setLocale(new Intl.Locale(locale)))
```
agurtovoy, line 314: Move dispatching `setLocale` action for the builder to be with the rest of the builder connection initialization logic.

packages/runtime/src/state/makeswift-api-client.ts
```diff
@@ -261,15 +261,10 @@ function cacheVersioningMiddleware(): ThunkMiddleware<State, UnknownAction> {
   })
 }
 
-export function configureStore({ serializedState }: { serializedState?: SerializedState }) {
+export function configureStore({ preloadedState }: { preloadedState: Partial<State> }) {
```
agurtovoy, line 246: As per my earlier comment, replaced unused initialization from `serializedState` with the standard `preloadedState` init.

packages/runtime/src/state/read-only-state.ts
```diff
@@ -36,8 +37,9 @@ export type { ComponentType } from './modules/react-components'
 export type { ComponentMeta } from './modules/components-meta'
 
 export const reducers = {
-  isReadOnly: IsReadOnly.reducer,
   siteVersion: SiteVersion.reducer,
+  isReadOnly: IsReadOnly.reducer,
+  locale: Locale.reducer,
```
agurtovoy, line 42: Add `locale` slice to the main Redux state.

packages/runtime/src/state/store.ts
```diff
@@ -16,18 +17,74 @@ import { HostActionTypes } from './host-api'
...
-import { makeswiftApiClientSyncMiddleware } from './middleware/makeswift-api-client-sync'
+
+export { type State } from './unified-state'
+
+const configureStore = <Items extends readonly StoreEnhancer[] = []>({
+  name,
+  preloadedState,
+  enhancers,
+  middleware,
+}: {
```
agurtovoy, line 40: Local utility function for configuring a store

packages/runtime/src/state/store.ts
```diff
@@ -16,18 +17,74 @@ import { HostActionTypes } from './host-api'
...
+  return store
+}
+
+export function configureProtoStore({
+  name,
+  breakpoints,
+}: {
+  name: string
+  breakpoints: Breakpoints.State | undefined
+}) {
```
agurtovoy, line 77: Proto store creation

packages/runtime/src/state/store.ts
```diff
@@ -98,31 +155,29 @@ export function conditionalReadWriteMiddleware(
...
+function withMixin<M extends {}>(mixin: M): StoreEnhancer<M> {
   return next => (reducer, preloadedState?) => ({
     ...next(reducer, preloadedState),
-    loadReadWriteState,
+    ...mixin,
   })
 }
 
-export function configureStore({
+export function configureReadWriteStore({
```
agurtovoy, line 171: Read-write store creation

packages/runtime/src/state/store.ts
```diff
@@ -179,65 +234,52 @@ export function configureStore({
...
-        makeswiftApiClientSyncMiddleware(hostApiClient),
-        conditionalReadWriteMiddleware(readWriteMiddlewareRef),
-      ]),
+  const store = configureStore({
+    name,
+    preloadedState,
 
-    enhancers: getDefaultEnhancers =>
-      getDefaultEnhancers().concat(
-        withReadWriteState(async ({ isReadOnly }) => {
```
agurtovoy, line 199: Bad diffs here, there are no semantic changes to the dynamic read-write state loading other that instead of being passed as an argument, `isReadOnly` is now read from the state.

packages/runtime/src/utils/__tests__/ref-counted-map.test.ts
```diff
@@ -0,0 +1,187 @@
+import { RefCountedMap } from '../ref-counted-map'
+
+interface Cat {
+  name: string
+}
+
+describe('RefCountedMap', () => {
```
agurtovoy, line 7: `RefCountedMap` tests, see https://github.com/makeswift/makeswift/pull/1272/changes#r3076794668

fikrikarim (commented):
Do we want to support mixed locales or mixed site version in one page?

From what I understand, a lot of the code introduced here is so we could have an isolated store per locale and site version in the client and server.

Could we potentially simplify this if we don't want to support that? 

We could create one store per request on the server, and have one shared store in the client. Then we don't need the client-side store tracking/cleaning up.

Although from the correctness standpoint, supporting those is technically more correct if they indeed can stand by its own. I'm just not 100% sure that it's worth the additional complexity.

agurtovoy:
> Do we want to support mixed locales or mixed site version in one page?

Not really; in the past, we mused about supporting mixed locales for hard-coded elements, and you'll effectively get the "mixed locales" scenario when you client-side navigate between localized pages, but that could work with a single client-side store as long as all localized resources are partitioned (which they currently are). That said, after having lived with and debugged a single singleton store for a short while, I strongly prefer the suggested design of per version, per locale "disposable" stores.

As far as "lots of code"/complexity goes, the only new code introduced here is `RefCountedMap`, which is a well-defined, isolated, 120-lines-long abstraction, plus its tests, and I'd argue that by removing the scenario where a single store can potentially go back and forth between different site versions and read-only and read-write states actually removes a lot of complexity.

fikrikarim:
Oh yeah I totally forgot about the client-side navigation. Ok. Sounds good.

packages/runtime/src/runtimes/react/runtime-core.ts
```diff
@@ -13,79 +13,147 @@ import {
...
-  setIdempotent({
-    siteVersion,
-    isReadOnly,
-    locale,
-  }: {
-    siteVersion: SiteVersion | null
-    isReadOnly: boolean
-    locale: string | undefined
-  }): void {
-    this.store.dispatch(setSiteVersion(siteVersion))
```
fikrikarim, line 67: Do we still need the `setSiteVersion` and the `cacheVersioningMiddleware`?

packages/runtime/src/runtimes/react/react-runtime-core.ts
```diff
@@ -55,7 +55,7 @@ export class ReactRuntimeCore extends RuntimeCore {
   ): () => void {
     validateComponentType(type, component as unknown as ComponentType)
 
-    const unregisterComponent = this.store.dispatch(
```
fikrikarim, line 58: It's worth mentioning that this change means we can't dynamically call `registerComponent` in the client after the store is created since now it's dispatching to `protoStore` instead of to the `store`.

AFAIK we currently don't have any use case for this so this should be fine. 

And I think for widget transformer, all the registrations should be included on the first render, since the `import '@/makeswift/components'` is done before the first `RuntimeProvider` render, so this should be also fine.

packages/runtime/src/runtimes/react/runtime-core.ts
fikrikarim: This means that we need to use the `requestKey` in the MoS worker, and arguably all multi-root implementations, right?

fikrikarim: I wonder if it's possible to determine that automatically so users won't need to pass the `requestKey`. Probably not right now since we're the only one using multi-root, but something to consider if we support Astro for example.

packages/runtime/src/state/store.ts
fikrikarim: nit: I was confused when I first read this. I thought this was a special store that's only called when it's the builder. Then I realized that this is just the store.

fikrikarim (dismissed):
This is awesome. I agree with you that this approach is better than a store that can go back and forth between different site versions.

packages/runtime/src/runtimes/react/runtime-core.ts
agurtovoy: > This means that we need to use the requestKey in the MoS worker, and arguably all multi-root implementations, right?

Ideally, yes. I have the corresponding change queued up for MoS.

packages/runtime/src/runtimes/react/runtime-core.ts
agurtovoy: We don't! I was going to do a cleanup pass and forgot. Pushed a [cleanup commit](https://github.com/makeswift/makeswift/pull/1272/changes/7deb7a4c2e9f2ae7c8dabbef5026db9ca742de59) 🙏

packages/runtime/src/state/store.ts
agurtovoy: Yeah, I can see how it can be confusing. I considered labels like "main store" or "primary store", but they are IMHO worse.

fikrikarim (approved):

PR #1274, opened 2026-04-03
feat: communicate element visibility to builder

arvinpoddar:
Updates the box model data sent to the builder to include whether the [element is visible or not](https://developer.mozilla.org/en-US/docs/Web/API/Element/checkVisibility).

### DIFF
--- a/packages/controls/src/common/types.ts
+++ b/packages/controls/src/common/types.ts
@@ -1,8 +1,11 @@
+import { type BoxModel } from 'css-box-model'
 import { z } from 'zod'
 
 import * as Schema from './schema'
 
 export { type BoxModel } from 'css-box-model'
+
+export type BoxDisplayModel = BoxModel & { visible?: boolean }
 
 export type Data = z.infer<typeof Schema.data>
 export type Device = z.infer<typeof Schema.deviceId>
--- a/packages/controls/src/controls/slot/slot-control.ts
+++ b/packages/controls/src/controls/slot/slot-control.ts
@@ -1,15 +1,15 @@
-import { type BoxModel } from '../../common'
+import { type BoxDisplayModel } from '../../common'
 
 import { ControlInstance } from '../instance'
 
 type BoxModelChangeMessage = {
   type: typeof SlotControl.CONTAINER_BOX_MODEL_CHANGE
-  payload: { boxModel: BoxModel | null }
+  payload: { boxModel: BoxDisplayModel | null }
 }
 
 type ItemBoxModelChangeMessage = {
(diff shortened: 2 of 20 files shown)


packages/runtime/src/state/modules/read-write/box-models.ts
```diff
@@ -10,24 +10,36 @@ import { type Action, type UnknownAction, isKnownAction } from '../../actions'
...
+    opacityProperty: true,
+  })
+}
+
+export function getBox(element: Element): BoxDisplayModel {
+  const visible = isElementVisible(element)
+  const boxModel = withScroll(getBoxWithoutScroll(element))
+
+  return { ...boxModel, visible }
 }
```
arvinpoddar, line 46: Updating the box model measuring function to also measure visibility.

packages/runtime/src/components/builtin/Box/Box.tsx
```diff
@@ -122,7 +127,9 @@ const Box = forwardRef(function Box(
           left: parse(marginBoxComputedStyle.marginLeft),
         }
 
-        return borderBox ? createBox({ borderBox, padding, border, margin }) : null
+        const visible = marginBoxElement ? isElementVisible(marginBoxElement) : undefined
+
+        return borderBox ? createBox({ borderBox, padding, border, margin, visible }) : null
```
arvinpoddar, line 132: Unlike all other components, the builtin box component doesn't use `getBox` for measuring. Updating the custom measurement to account for visibility.

packages/controls/src/common/types.ts
```diff
@@ -1,9 +1,12 @@
+import { type BoxModel } from 'css-box-model'
 import { z } from 'zod'
 
 import * as Schema from './schema'
 
 export { type BoxModel } from 'css-box-model'
 
+export type BoxDisplayModel = BoxModel & { visible?: boolean }
```
arvinpoddar, line 8: Defining the new augmented type in the `controls` package.

packages/controls/src/controls/slot/slot-control.ts
```diff
@@ -1,15 +1,15 @@
-import { type BoxModel } from '../../common'
+import { type BoxDisplayModel } from '../../common'
```
arvinpoddar, line 1: Unless otherwise specified, most changes in this PR will be usage of the new augmented `BoxDisplayModel` type.

agurtovoy (dismissed):
Very nice ❤️

arvinpoddar:
Updated to add a check to make sure the `checkVisibility` method is available, and cleaned up some imports. No major semantic changes since last approval.

PR #1275, opened 2026-04-07
Remove rethrow of error in `renderToReadableStream`

Codeseph:
Part of [ARR-628](https://linear.app/makeswift/issue/ARR-628/sleep-country-specific-page-intermittently-fails-to-load), context is here: https://makeswifthq.slack.com/archives/C08JCF04397/p1775580422241279?thread_ts=1775258919.810929&cid=C08JCF04397

### DIFF
--- a/packages/hono-react/src/server/render.tsx
+++ b/packages/hono-react/src/server/render.tsx
@@ -23,6 +23,7 @@ type RenderOptions = RootStyleProps &
     | 'bootstrapScripts'
     | 'bootstrapModules'
     | 'progressiveChunkSize'
+    | 'onError'
   >
 
 export async function renderHtml(
@@ -36,12 +37,7 @@ export async function renderHtml(
     <RootStyleRegistry cache={cache} enableCssReset={enableCssReset}>
       {children}
     </RootStyleRegistry>,
-    {
-      ...renderOptions,
-      onError(error: unknown) {
-        throw Error(`Server-side rendering error: ${error}`)
-      },
-    },
+    { ...renderOptions },
   )
 
   return { html, getStyles }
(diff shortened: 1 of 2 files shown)


packages/hono-react/src/server/render.tsx
```diff
@@ -36,12 +36,7 @@ export async function renderHtml(
     <RootStyleRegistry cache={cache} enableCssReset={enableCssReset}>
       {children}
     </RootStyleRegistry>,
-    {
-      ...renderOptions,
-      onError(error: unknown) {
```
Codeseph, line 41: When a custom `onError` handler isn't provided, the default handler [includes an error log](https://react.dev/reference/react-dom/server/renderToReadableStream#parameters) which we should be able to see in the proxy worker server logs. For example I've been testing the worker with a component that errors during SSR, and after this change can see the log:

Image

packages/hono-react/src/server/render.tsx
```diff
@@ -23,6 +23,7 @@ type RenderOptions = RootStyleProps &
     | 'bootstrapScripts'
     | 'bootstrapModules'
     | 'progressiveChunkSize'
+    | 'onError'
```
Codeseph, line 26: I was just going to remove the custom handler and not add `onError` as an option here, but I'm thinking we might need this depending on the implementation we go with for [(2) here](https://makeswifthq.slack.com/archives/C08JCF04397/p1775676928268939?thread_ts=1775258919.810929&cid=C08JCF04397)

fikrikarim (approved):
Looks good. cc: @agurtovoy for visibility.

packages/hono-react/src/server/render.tsx
agurtovoy: Good thinking, I agree.

PR #1277, opened 2026-04-10
feat: add `unstable_getComponentSnapshots`

fikrikarim:
Currently the worker can only fetch 4 element trees at a time because of a [Cloudflare worker limitation](https://developers.cloudflare.com/workers/platform/limits/#simultaneous-open-connections).

Although the live page performance is not affected due to caching, the page takes significantly much longer to load in the builder. This method is needed so we can fetch all element trees in a single request.

For verifying the `introspectMany` implementation, I recommend reviewing it side by side with the existing `introspect` function.

### DIFF
--- a/packages/runtime/src/client/index.ts
+++ b/packages/runtime/src/client/index.ts
@@ -47,6 +47,7 @@ import { Schema } from '@makeswift/controls'
 import { EMBEDDED_DOCUMENT_TYPE, EmbeddedDocument } from '../state/modules/read-only-documents'
 import { mergeTranslatedContent } from '../state/translations/merge'
 import { getTranslatableContent } from '../state/translations/get'
+import { isNonNullable } from '../utils/isNonNullable'
 
 export { SnippetLocation } from '../api/graphql/generated/types'
 
@@ -373,7 +374,11 @@ export class MakeswiftClient {
     return this.runtime.apiOrigin
   }
 
-  private async fetch(path: string, siteVersion: SiteVersion | null): Promise<Response> {
+  private async fetch(
+    path: string,
+    siteVersion: SiteVersion | null,
+    init?: RequestInit,
+  ): Promise<Response> {
     const requestUrl = new URL(path, this.apiOrigin)
 
     const requestHeaders = new Headers({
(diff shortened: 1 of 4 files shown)


packages/runtime/src/client/index.ts
```diff
@@ -681,6 +728,237 @@ export class MakeswiftClient {
...
+
+      return {
+        apiResources,
+        localizedResourcesMap:
+          locale != null
+            ? { [locale]: Object.fromEntries(tracking.localizedResourcesMap.entries()) }
+            : {},
+      }
+    })
+  }
```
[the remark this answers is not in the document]
fikrikarim: Yeah this is not ideal, added a TODO to consolidate once we test introspectMany in production.

packages/runtime/src/client/index.ts
```diff
@@ -681,6 +732,237 @@ export class MakeswiftClient {
...
+    const perTree = trees.map(() => ({
+      swatchIds: new Set<string>(),
+      fileIds: new Set<string>(),
+      typographyIds: new Set<string>(),
+      tableIds: new Set<string>(),
+      pageIds: new Set<string>(),
+      globalElements: new Map<string, GlobalElement | null>(),
+      localizedGlobalElements: new Map<string, LocalizedGlobalElement | null>(),
+      localizedResourcesMap: new Map<string, string | null>(),
+    }))
```
arvinpoddar: Some general organizational thoughts:

- I'd feel better about this if the perTree state was doubly-keyed `treeId -> CacheData`. This would mean not doing CacheData access based on array order, but by element tree ID.
- Do we need shared global sets? Seems like we could just accumulate everything from the per tree state ahead of making the bulk calls. That's less state to manage.

packages/runtime/src/client/index.ts
```diff
@@ -681,6 +732,237 @@ export class MakeswiftClient {
...
+
+    // DFS traversal per tree
+    for (let treeIndex = 0; treeIndex < trees.length; treeIndex++) {
+      const tree = trees[treeIndex]
+      const tracking = perTree[treeIndex]
```
arvinpoddar: Maybe rename to `currentTreeCache` for clarity?

packages/runtime/src/client/index.ts
fikrikarim: Good points. I updated the commit so that it's not based on array order, and moved the shared sets to be after the tree traversal. It's not double-keyed though, it's just keyed by the `treeId`.

packages/runtime/src/client/index.ts
fikrikarim: Updated

arvinpoddar (dismissed):
Introspection is complicated, I'd like to see us get to a point where it's simplified — a lot of the bulk implementation is closely mirroring what we have for the existing `introspect` method, which I've never been a fan of haha. That being said, having it start as a mirror is fine as a starting point.

If we plan on stabilizing this, part of this should include stronger test cases (for each resource type, ensuring that 2+ trees have unique cache data, so on)

fikrikarim:
Merging this since it was approved earlier, and the last changes only affect tests and patches after rebase.

PR #1279, opened 2026-04-16
Fix incorrect Slot dropzones

fikrikarim:
Review per commit and see commit description.

See the [builder PR](https://github.com/makeswift/cosmos/pull/2635) for video explanation

### DIFF
--- a/.changeset/lemon-bees-melt.md
+++ b/.changeset/lemon-bees-melt.md
@@ -1,0 +1,5 @@
+---
+"@makeswift/controls": patch
+---
+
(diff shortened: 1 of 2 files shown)


drewpledger (approved):

arvinpoddar (commented):
Doing some digging online, I see that messages sent via MessageChannels should have a consistent send/receive order (i.e., messages that are sent first are received first). I wonder if we can avoid timestamping on the runtime altogether...

fikrikarim:
Closing this PR since we're adding the timestamp in the [builder](https://github.com/makeswift/cosmos/pull/2635)

PR #1284, opened 2026-04-30
fix: handle ops where deletion path is not in new element tree

arvinpoddar:
Fixes an issue where attempting to apply a deletion op can result in an error if the path of the deletion op is not present in the new element tree. Can organically occur when reparenting a deeply nested element to the top of the root, since the deletion path will be for the deeply nested path.

### DIFF
--- a/packages/runtime/src/state/modules/__tests__/element-trees.test.ts
+++ b/packages/runtime/src/state/modules/__tests__/element-trees.test.ts
@@ -3,6 +3,8 @@ import * as Fixtures from './fixtures/element-trees'
 import {
   ELEMENT_TREE_DEMO_COMPONENT_TYPE,
   ElementTreesDemo,
+  SLOT_DEMO_COMPONENT_TYPE,
+  SlotDemo,
 } from './fixtures/element-trees-demo-component'
 
 import { createReactRuntime } from '../../../runtimes/react/testing/react-runtime'
@@ -183,5 +185,63 @@ describe('resetting an element tree', () => {
     // they should share no common keys
     expect(initialElementKeys.some(key => elementKeysAfterReset.includes(key))).toBe(false)
     expect(elementKeysAfterReset.some(key => initialElementKeys.includes(key))).toBe(false)
+  })
+})
+
+describe('applyChanges', () => {
+  const runtime = createReactRuntime()
+  runtime.registerComponent(SlotDemo, {
+    type: SLOT_DEMO_COMPONENT_TYPE,
+    label: 'Slot Demo',
+    props: {
+      children: Slot(),
+    },
+  })
+
+  afterEach(() => {
+    jest.restoreAllMocks()
+  })
+
+  test('handle element reparenting op', () => {
+    const consoleError = jest.spyOn(console, 'error').mockImplementation(jest.fn())
+
+    const descriptors = getPropControllerDescriptors(runtime.protoStore.getState())
+    const documentKey = '11111111-1111-1111-111111111111'
+
+    const oldDocument = {
+      key: documentKey,
+      rootElement: Fixtures.reparentingElementTree.before,
+    }
+
+    const initialState = ElementTrees.reducer(
+      ElementTrees.getInitialState(),
+      createElementTree({
+        document: oldDocument,
+        descriptors,
+      }),
+    )
+
+    const newDocument = {
+      key: documentKey,
+      rootElement: Fixtures.reparentingElementTree.after,
+    }
+
+    const reparentOp = Fixtures.reparentingElementTree.op
+
+    const updatedState = ElementTrees.reducer(
+      initialState,
+      changeElementTree({
+        oldDocument,
+        newDocument,
+        descriptors,
+        operation: reparentOp,
+      }),
+    )
+
+    expect(consoleError).not.toHaveBeenCalled()
+
+    const elements = ElementTrees.getElements(updatedState, documentKey)
+    const rootKey = Fixtures.reparentingElementTree.before.key
+    expect(elements.get(rootKey)).toEqual(Fixtures.reparentingElementTree.after)
   })
 })
--- a/packages/runtime/src/state/modules/__tests__/fixtures/element-trees.ts
+++ b/packages/runtime/src/state/modules/__tests__/fixtures/element-trees.ts
@@ -1,3 +1,6 @@
+import { type Operation } from 'ot-json0'
+import { type ElementData } from '@makeswift/controls'
+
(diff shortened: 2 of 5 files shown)


packages/runtime/src/state/modules/element-trees.ts
```diff
@@ -355,21 +355,52 @@ function applyUpdate(
...
-    const hasDelete = 'ld' in op || 'od' in op
-    const hasInsert = 'li' in op || 'oi' in op
-    if (hasDelete && hasInsert) {
-      return applyUpdate(tree, descriptors, rootElements, op.p)
-    }
+  const result = operation.reduce(
+    (acc, op) => {
+      const rootBeforeOp = acc.root
+      const rootAfterOp = applyOpComponent(rootBeforeOp, op)
+      const opRoots = { old: rootBeforeOp, new: rootAfterOp }
```
arvinpoddar: The fix, in short: we now apply the ops to the "old" root element, and use that transformed root as the "new" root when applying changes. This means that the path read will be accurate.

packages/runtime/src/state/modules/__tests__/fixtures/element-trees.ts
```diff
@@ -4287,3 +4290,332 @@ export const resetElementTree = {
     },
   },
 }
+
+// Testing fixture that demonstrates reparenting an image from a deeply nested
+// position in the element tree to the top level of the tree.
+export const reparentingElementTree: {
```
arvinpoddar, line 4296: This fixture is captured from performing this sequence in the builder:

1. Start with an "Root" element tree containing three nested slots, and an image inside the inner-most slot. This is the "before" tree.
2. Move the image to be the first child in the Root element. This is the "after" tree. The op generated by the builder is the op in this fixture.

The issue here is that the builder op contains two operations: a deletion and a replacement. The deletion is on the inner most image, which has a deep path reading into the first child of the tree. When applying the delete, we attempt to update the parent elements in the "new" tree. However, this deep path on the first child isn't valid on the new tree, because the first child is an image.

packages/runtime/src/state/modules/__tests__/element-trees.test.ts
```diff
@@ -185,3 +187,61 @@ describe('resetting an element tree', () => {
...
+    props: {
+      children: Slot(),
+    },
+  })
+
+  afterEach(() => {
+    jest.restoreAllMocks()
+  })
+
+  test('handle deletion op when op path does not exist in new element tree', () => {
```
agurtovoy: "deletion op when op path does not exist in new element tree" sounds like an invalid screnario; maybe just "handle element reparenting op"?

packages/runtime/src/state/modules/element-trees.ts
```diff
@@ -334,42 +334,94 @@ function applyInsert(
...
+  const result = operation.reduce(
+    (acc, op) => {
+      const rootBefore = acc.rootBefore
+
+      // If there's only one component op, we can skip the application of the op
+      // and assume that the result will be the 'new' root element provided to
+      // us by the builder.
+      const rootAfter = operation.length > 1 ? applyOpComponent(rootBefore, op) : rootElements.new
+
+      const roots = { before: rootBefore, after: rootAfter }
```
agurtovoy, line 411: Maybe

```ts
      const roots = { before: rootBefore, after: rootAfter }
      const applyChange = selectTreeTransform(op)

      return {
        elementTree: applyChange(acc.elementTree, descriptors, roots, op.p),
        rootBefore: rootAfter,
      }
```

where `selectTreeTransform` is

```ts
function selectTreeTransform(op: Operation[number]): typeof applyDelete {
  const hasDelete = 'ld' in op || 'od' in op
  const hasInsert = 'li' in op || 'oi' in op

  if (hasDelete && hasInsert) return applyUpdate
  if (hasDelete) return applyDelete
  if (hasInsert) return applyInsert

  return elementTree => elementTree
}
```

?

agurtovoy (dismissed):
Great catch 🙏

packages/runtime/src/state/modules/element-trees.ts
```diff
@@ -269,17 +269,17 @@ function deleteElement(
...
   const [deleteElementPath, ...parentElementPaths] = getChangedElementsPaths(path)
-  const [deletedElement, propName] = getElementAndPropName(rootElements.old, deleteElementPath)
+  const [deletedElement, propName] = getElementAndPropName(rootElements.before, deleteElementPath)
 
   const elements = new Map(elementTree.elements)
   const elementIds = new Map(elementTree.elementIds)
 
   deleteElement({ elements, elementIds }, deletedElement, propName, descriptors)
-  updateParentElements(elements, parentElementPaths, rootElements.new)
+  updateParentElements(elements, parentElementPaths, rootElements.after)
```
agurtovoy, line 284: Maybe something like this for a comment here?

```ts
// Use the "after" state to efficiently update all of parent subtrees in the state
```

packages/runtime/src/state/modules/element-trees.ts
arvinpoddar: Good suggestion, addressed.

packages/runtime/src/state/modules/element-trees.ts
```diff
@@ -334,42 +338,86 @@ function applyInsert(
...
+
+  if ('li' in component) {
+    applied = setIn(applied, component.p, component.li)
+  }
+
+  if ('oi' in component) {
+    applied = setIn(applied, component.p, component.oi)
+  }
+
+  return applied
```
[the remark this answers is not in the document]
arvinpoddar: Ignoring

packages/runtime/src/state/modules/element-trees.ts
```diff
@@ -334,42 +338,86 @@ function applyInsert(
...
+  // to the 'old' root element to determine the changed elements and update the
+  // cache accordingly at each intermediate step.
+  const result = operation.reduce(
+    (acc, op) => {
+      const rootBefore = acc.rootBefore
+
+      // If there's only one component op, we can skip the application of the op
+      // and assume that the result will be the 'new' root element provided to
+      // us by the builder.
+      const rootAfter = operation.length > 1 ? applyOpComponent(rootBefore, op) : rootElements.new
```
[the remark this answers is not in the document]
arvinpoddar: Fair point. That being said, using `rootElements.new` for the final op is also an integrity problem if it's not the same outcomes as applying all the ops, so the suggestion doesn't really solve the underlying problem. Going to ignore for now.

PR #1295, opened 2026-05-13
fix(runtime): move current breakpoint calculation to the root

agurtovoy:
Moving current breakpoint calculation to the root reduces the number of `window.matchMedia` calls/event handlers, which significantly improves rendering performance on Safari.

Before:

https://github.com/user-attachments/assets/d6e65250-a6ae-4c30-9074-b58353eb7fbc

After:

https://github.com/user-attachments/assets/b5eb2927-72a7-4cdd-9b30-8d1bdc0b6431

### DIFF
--- a/packages/runtime/package.json
+++ b/packages/runtime/package.json
@@ -285,6 +285,7 @@
     "jest-environment-jsdom": "^29.7.0",
     "jsdom": "^26.0.0",
     "madge": "^6.0.0",
+    "mock-match-media": "^1.0.3",
     "msw": "^2.3.1",
     "next": "15.5.18",
     "node-mocks-http": "^1.16.2",
--- a/packages/runtime/src/components/hooks/__tests__/useMediaQuery.test.tsx
+++ b/packages/runtime/src/components/hooks/__tests__/useMediaQuery.test.tsx
@@ -1,0 +1,95 @@
+/** @jest-environment jsdom */
+
+import { act } from 'react'
+import { render, screen } from '@testing-library/react'
+
+import { matchMedia, setMedia } from 'mock-match-media'
+
+import { useMediaQuery } from '../useMediaQuery'
+
+import { createReactRuntime } from '../../../runtimes/react/testing/react-runtime'
+import { RuntimeProvider } from '../../../runtimes/react/components/RuntimeProvider'
+
+const breakpoints = {
+  mobile: { width: 575, viewport: 390, label: 'Mobile' },
+  tablet: { width: 768, viewport: 765, label: 'Tablet' },
+  laptop: { width: 1024, viewport: 1000, label: 'Laptop' },
+  external: { width: 1280, label: 'External' },
+}
+
+type DeviceId = keyof typeof breakpoints | 'desktop'
+
+const responsiveValue: Array<{ deviceId: DeviceId; value: string }> = [
+  {
+    deviceId: 'desktop',
+    value: 'none',
+  },
+  {
+    deviceId: 'external',
+    value: 'fadeTop',
+  },
+  {
+    deviceId: 'laptop',
+    value: 'fadeRight',
+  },
+  {
+    deviceId: 'tablet',
+    value: 'fadeLeft',
+  },
+  {
+    deviceId: 'mobile',
+    value: 'blurIn',
+  },
+]
+
+const testId = 'test-id'
+
+const TestComponent = () => {
+  const value = useMediaQuery(responsiveValue)
+  return <div data-testid={testId}>{value}</div>
+}
+
+let oldMatchMedia: any
+
+describe('useMediaQuery', () => {
+  beforeEach(() => {
+    oldMatchMedia = window.matchMedia
+    Object.defineProperty(window, 'matchMedia', {
+      writable: true,
+      value: matchMedia,
+    })
+  })
+
+  afterEach(() => {
+    Object.defineProperty(window, 'matchMedia', { writable: true, value: oldMatchMedia })
+  })
+
+  test.each([
+    { width: 1300, expectedValue: 'none' },
+    { width: 1280, expectedValue: 'fadeTop' },
+    { width: 1100, expectedValue: 'fadeTop' },
+    { width: 1024, expectedValue: 'fadeRight' },
+    { width: 800, expectedValue: 'fadeRight' },
+    { width: 768, expectedValue: 'fadeLeft' },
+    { width: 600, expectedValue: 'fadeLeft' },
+    { width: 575, expectedValue: 'blurIn' },
+    { width: 200, expectedValue: 'blurIn' },
+  ])(
+    'returns responsive value matching current media breakpoint ($width px)',
+    async ({ width, expectedValue }) => {
+      const runtime = createReactRuntime({ breakpoints })
+
+      setMedia({ width })
+
+      await act(async () =>
+        render(
+          <RuntimeProvider runtime={runtime} siteVersion={null}>
+            <TestComponent />
+          </RuntimeProvider>,
+        ),
+      )
+
+      expect(screen.getByTestId(testId)).toHaveTextContent(expectedValue)
+    },
+  )
+})
--- a/packages/runtime/src/components/hooks/useMediaQuery.ts
+++ b/packages/runtime/src/components/hooks/useMediaQuery.ts
@@ -1,49 +1,14 @@
-import { useCallback, useSyncExternalStore } from 'react'
-import {
-  type Breakpoints,
-  type DeviceOverride,
-  findBreakpointOverride,
-  getBaseBreakpoint,
-  getBreakpointMediaQuery,
-} from '@makeswift/controls'
+import { useMemo } from 'react'
+import { type DeviceOverride, findBreakpointOverride } from '@makeswift/controls'
 
 import { useBreakpoints } from '../../runtimes/react/hooks/use-breakpoints'
-
-const getDeviceQueries = (breakpoints: Breakpoints) =>
-  breakpoints.map(device => ({
-    id: device.id,
-    query: getBreakpointMediaQuery(device).replace('@media', ''),
-  }))
+import { useCurrentBreakpoint } from '../../runtimes/react/hooks/use-current-breakpoint'
 
 export function useMediaQuery<S>(responsiveValue?: Array<DeviceOverride<S>>): S | undefined {
   const breakpoints = useBreakpoints()
-  const baseBreakpointId = getBaseBreakpoint(breakpoints).id
-  const subscribe = useCallback(
-    (notify: () => void) => {
-      const cleanUpFns = getDeviceQueries(breakpoints).map(q => {
-        const mediaQueryList = window.matchMedia(q.query)
-        mediaQueryList.addEventListener('change', notify)
-        return () => mediaQueryList.removeEventListener('change', notify)
-      })
-      return () => cleanUpFns.forEach(fn => fn())
-    },
-    [breakpoints],
+  const breakpointId = useCurrentBreakpoint()
+  return useMemo(
+    () => findBreakpointOverride(breakpoints, responsiveValue, breakpointId)?.value,
+    [breakpoints, breakpointId, responsiveValue],
   )
-  const getServerSnapshot = () =>
-    findBreakpointOverride(breakpoints, responsiveValue, baseBreakpointId)?.value
-
-  function getSnapshot() {
-    const deviceId: string = getDeviceQueries(breakpoints).reduce(
-      (matchedDevice, deviceQueries) => {
-        if (window.matchMedia(deviceQueries.query).matches) {
-          return deviceQueries.id
-        }
-        return matchedDevice
-      },
-      baseBreakpointId,
-    )
-    return findBreakpointOverride(breakpoints, responsiveValue, deviceId)?.value
-  }
-
-  return useSyncExternalStore(subscribe, getSnapshot, getServerSnapshot)
 }
--- a/packages/runtime/src/runtimes/react/hooks/use-current-breakpoint.ts
+++ b/packages/runtime/src/runtimes/react/hooks/use-current-breakpoint.ts
@@ -1,0 +1,16 @@
+import { useSyncExternalStore } from 'react'
+
+import { type BreakpointId } from '../../../state/modules/breakpoints'
+import { getBaseBreakpoint, getClientBreakpoint } from '../../../state/read-only-state'
+
+import { useStore } from './use-store'
+
+export function useCurrentBreakpoint(): BreakpointId {
+  const store = useStore()
+
+  return useSyncExternalStore(
+    store.subscribe,
+    () => getClientBreakpoint(store.getState()),
+    () => getBaseBreakpoint(store.getState()),
+  )
+}
--- a/packages/runtime/src/runtimes/react/runtime-core.ts
+++ b/packages/runtime/src/runtimes/react/runtime-core.ts
@@ -116,7 +116,9 @@ export class RuntimeCore {
       return
     }
 
-    this.activeStores.retain(key, store)
+    if (this.activeStores.retain(key, store)) {
+      store.startBreakpointWatch()
+    }
   }
 
   releaseStore({ siteVersion, locale }: StoreKey, store: Store): void {
@@ -127,7 +129,9 @@ export class RuntimeCore {
       return
     }
 
-    this.activeStores.release(key, store)
+    if (this.activeStores.release(key, store)) {
+      store.stopBreakpointWatch()
+    }
   }
 
   copyElementTree(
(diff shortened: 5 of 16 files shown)


packages/runtime/src/components/hooks/useMediaQuery.ts
```diff
@@ -1,49 +1,14 @@
...
-    (notify: () => void) => {
-      const cleanUpFns = getDeviceQueries(breakpoints).map(q => {
-        const mediaQueryList = window.matchMedia(q.query)
-        mediaQueryList.addEventListener('change', notify)
-        return () => mediaQueryList.removeEventListener('change', notify)
-      })
-      return () => cleanUpFns.forEach(fn => fn())
-    },
-    [breakpoints],
+  const breakpointId = useCurrentBreakpoint()
```
agurtovoy, line 9: Moved current breakpoint into the Redux state.

packages/runtime/src/components/hooks/useMediaQuery.ts
```diff
@@ -1,49 +1,14 @@
...
 
 export function useMediaQuery<S>(responsiveValue?: Array<DeviceOverride<S>>): S | undefined {
   const breakpoints = useBreakpoints()
-  const baseBreakpointId = getBaseBreakpoint(breakpoints).id
-  const subscribe = useCallback(
-    (notify: () => void) => {
-      const cleanUpFns = getDeviceQueries(breakpoints).map(q => {
-        const mediaQueryList = window.matchMedia(q.query)
-        mediaQueryList.addEventListener('change', notify)
-        return () => mediaQueryList.removeEventListener('change', notify)
```
agurtovoy, line 26: Notifications about media changes moved to `BreakpointWatch` mixin: https://github.com/makeswift/makeswift/pull/1295/changes#r3253209414

packages/runtime/src/state/mixins/breakpoint-watch.ts
```diff
@@ -0,0 +1,70 @@
...
+export interface BreakpointWatch {
+  startBreakpointWatch(): void
+  stopBreakpointWatch(): void
+}
+
+export function breakpointWatchMixin({
+  dispatch,
+  getState,
+  subscribe: storeSubscribe,
+}: {
```
agurtovoy, line 16: Store mixin for watching for (client-side) breakpoint updates. 

I originally had this implemented as a React hook/component plugged into `RuntimeProvider`, but as I was putting it up for review, I realized that having it in the provider is still problematic, as we now can have dozens of React roots on a single page.

Having this as a store mixin also has a nice side-effect of moving more of the runtime logic out of React.

packages/runtime/src/runtimes/react/hooks/use-current-breakpoint.ts
```diff
@@ -0,0 +1,16 @@
...
+import { useStore } from './use-store'
+
+export function useCurrentBreakpoint(): BreakpointId {
+  const store = useStore()
+
+  return useSyncExternalStore(
+    store.subscribe,
+    () => getClientBreakpoint(store.getState()),
+    () => getBaseBreakpoint(store.getState()),
+  )
```
agurtovoy, line 15: This is a tricky one. My first naive approach was just 

```ts
return useSelector(state => getClientBreakpoint(state))
````

At first glance, this makes sense, until you remember that the server always renders using the base breakpoint, and thus we need to match that state on the client during the initial hydration. 

OK, you say, so let's just initialize `clientBreakpoint` in the state with the base breakpoint for the initial render and let `BreakpointWatch` or some other effect update it to the actual client breakpoint after mount. Well, it turns out that that doesn't work either, as with `<Suspense />` there is no guarantee that the React tree hydration is fully finished by the time a parent effect is run.

`useSyncExternalStore` solves this because it uses the server snapshot during server rendering *and* during client hydration, so even "late" hydrating components still read the base breakpoint during their initial client render.

File: packages/runtime/src/utils/ref-counted-map.ts [hunk 5]
```diff
@@ -65,38 +65,51 @@ export class RefCountedMap<K, V> {
     }
   }
 
-  retain(key: K, value: V): void {
+  /**
+   * Returns true on transition from unowned to owned
+   */
+  retain(key: K, value: V): boolean {
```
agurtovoy, line 71: Updated ref-counting map to return true when `retain` and `release` cause a change in ownership.

packages/runtime/src/runtimes/react/runtime-core.ts
```diff
@@ -116,7 +116,9 @@ export class RuntimeCore {
       return
     }
 
-    this.activeStores.retain(key, store)
+    if (this.activeStores.retain(key, store)) {
```
agurtovoy, line 119: If this is the first retain, start breakpoint watch. See https://github.com/makeswift/makeswift/pull/1295/changes#r3253247738

packages/runtime/src/runtimes/react/runtime-core.ts
```diff
@@ -127,7 +129,9 @@ export class RuntimeCore {
       return
     }
 
-    this.activeStores.release(key, store)
+    if (this.activeStores.release(key, store)) {
```
agurtovoy, line 132: If this is the last release, stop breakpoint watch.

packages/runtime/src/state/actions/internal/read-only-actions.ts
```diff
@@ -35,6 +35,8 @@ export const ReadOnlyActionTypes = {
   REGISTER_REACT_COMPONENT: 'REGISTER_REACT_COMPONENT',
   UNREGISTER_REACT_COMPONENT: 'UNREGISTER_REACT_COMPONENT',
 
+  UPDATE_CLIENT_BREAKPOINT: 'UPDATE_CLIENT_BREAKPOINT',
```
agurtovoy, line 38: New action to update the client breakpoint that has been moved to the state.

packages/runtime/src/state/mixins/breakpoint-watch.ts
agurtovoy: Called from here: https://github.com/makeswift/makeswift/pull/1295/changes#r3253249382

packages/runtime/src/state/modules/breakpoints.ts
```diff
@@ -10,7 +18,11 @@ export {
   type Breakpoints,
 } from '@makeswift/controls'
 
-export type State = Breakpoints
+export type State = {
+  breakpoints: Breakpoints
+  baseBreakpoint: BreakpointId
+  clientBreakpoint: BreakpointId
+}
```
agurtovoy, line 25: Update the state to include base and client breakpoints.

packages/runtime/src/components/hooks/useMediaQuery.ts
```diff
@@ -1,49 +1,14 @@
...
+import { useMemo } from 'react'
+import { type DeviceOverride, findBreakpointOverride } from '@makeswift/controls'
 
 import { useBreakpoints } from '../../runtimes/react/hooks/use-breakpoints'
-
-const getDeviceQueries = (breakpoints: Breakpoints) =>
-  breakpoints.map(device => ({
-    id: device.id,
-    query: getBreakpointMediaQuery(device).replace('@media', ''),
-  }))
```
agurtovoy, line 16: Moved to https://github.com/makeswift/makeswift/pull/1295/changes#r3253254087

packages/runtime/src/state/modules/breakpoints.ts
```diff
@@ -41,8 +53,33 @@ export const DEFAULT_BREAKPOINTS: Breakpoints = [
   },
 ]
 
+export const getDeviceQueries = (breakpoints: Breakpoints) =>
+  breakpoints.map(device => ({
+    id: device.id,
+    query: getBreakpointMediaQuery(device).replace('@media', ''),
+  }))
```
agurtovoy, line 60: Moved from https://github.com/makeswift/makeswift/pull/1295/changes#r3253253839

packages/runtime/src/components/hooks/useMediaQuery.ts
```diff
@@ -1,49 +1,14 @@
...
-  function getSnapshot() {
-    const deviceId: string = getDeviceQueries(breakpoints).reduce(
-      (matchedDevice, deviceQueries) => {
-        if (window.matchMedia(deviceQueries.query).matches) {
-          return deviceQueries.id
-        }
-        return matchedDevice
-      },
-      baseBreakpointId,
-    )
```
agurtovoy, line 44: Moved client breakpoint calculation to the state: https://github.com/makeswift/makeswift/pull/1295/changes#r3253255079

packages/runtime/src/state/modules/breakpoints.ts
```diff
@@ -41,8 +53,33 @@ export const DEFAULT_BREAKPOINTS: Breakpoints = [
...
+  const baseBreakpoint = getBaseBreakpointId(breakpoints)
+  if (isServer()) return baseBreakpoint
+
+  // using `findLast` to preserve the legacy behavior; might not be necessary
+  const matchingBreakpoint = getDeviceQueries(breakpoints).findLast(
+    ({ query }) => window.matchMedia(query).matches,
+  )
+
+  return matchingBreakpoint?.id ?? baseBreakpoint
+}
```
agurtovoy, line 93: Moved from https://github.com/makeswift/makeswift/pull/1295/changes#r3253254788

packages/runtime/src/state/read-only-state.ts
```diff
@@ -223,10 +223,22 @@ export function getBuilderEditMode(state: State): BuilderEditMode.State {
...
+  return getBreakpointsStateSlice(state).breakpoints
+}
+
+export function getBaseBreakpoint(state: State): Breakpoints.BreakpointId {
+  return getBreakpointsStateSlice(state).baseBreakpoint
+}
+
+export function getClientBreakpoint(state: State): Breakpoints.BreakpointId {
+  return getBreakpointsStateSlice(state).clientBreakpoint
+}
```
agurtovoy, line 240: Accessors for the new state.

packages/runtime/src/state/store.ts
```diff
@@ -245,6 +246,14 @@ export function configureReadWriteStore({
 
     enhancers: () =>
       new Tuple(
+        withMixin<BreakpointWatch>(
+          breakpointWatchMixin({
+            dispatch: (...args: Parameters<Dispatch>) => store.dispatch(...args),
+            getState: (): State => store.getState(),
+            subscribe: (listener): VoidFunction => store.subscribe(listener),
+          }),
+        ),
```
agurtovoy, line 255: Add `BreakpointWatch` mixin to the state.

packages/runtime/src/utils/__tests__/ref-counted-map.test.ts
```diff
@@ -52,7 +52,7 @@ describe('RefCountedMap', () => {
     const key = 'whiskers'
 
     const whiskers = map.getOrCreate(key, () => ({ name: 'Whiskers' }))
-    map.retain(key, whiskers)
+    expect(map.retain(key, whiskers)).toBe(true)
```
agurtovoy, line 55: Updated tests to reflect the enhanced behavior, see https://github.com/makeswift/makeswift/pull/1295/changes#r3253247738

packages/runtime/src/components/hooks/__tests__/useMediaQuery.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+    deviceId: 'mobile',
+    value: 'blurIn',
+  },
+]
+
+const testId = 'test-id'
+
+const TestComponent = () => {
+  const value = useMediaQuery(responsiveValue)
+  return <div data-testid={testId}>{value}</div>
```
agurtovoy, line 49: Make sure the `useMediaQuery` hook still works as expected after we've [disassembled and reassembled below](https://github.com/makeswift/makeswift/pull/1295/changes#r3251885043).

packages/runtime/package.json
```diff
@@ -285,6 +285,7 @@
     "jest-environment-jsdom": "^29.7.0",
     "jsdom": "^26.0.0",
     "madge": "^6.0.0",
+    "mock-match-media": "^1.0.3",
```
agurtovoy, line 288: New dev dependency for tests.

packages/runtime/src/state/mixins/breakpoint-watch.ts
```diff
@@ -0,0 +1,70 @@
...
+  const startWatch = (breakpoints: Breakpoints) => {
+    mediaQueryUnsubscribes.forEach(fn => fn())
+
+    const updateState = () => dispatch(updateClientBreakpoint())
+
+    mediaQueryUnsubscribes = getDeviceQueries(breakpoints).map(q => {
+      const mediaQueryList = window.matchMedia(q.query)
+      mediaQueryList.addEventListener('change', updateState)
+      return () => mediaQueryList.removeEventListener('change', updateState)
+    })
```
arvinpoddar, line 35: We _could_ flip this so that we determine the matching breakpoint here, and leave the reducer "dumb" by having the dispatch payload include the determined breakpoint. The benefit being that the reducer doesn't have `isServer` logic inside of it (maybe there's not a problem with this, I'm just not used to seeing it) and is simply a recipient of state updates.

But, I'm assuming there's a good reason we didn't go that route — could you explain why?

arvinpoddar (dismissed):
Looking good, just one question

packages/runtime/src/state/mixins/breakpoint-watch.ts
agurtovoy: That's a great question. I moved client breakpoint calculation to the reducer to avoid having an intermediate stale client breakpoint state after handling `SET_BREAKPOINTS`, and in the process of doing so realized that this also tightens `UPDATE_CLIENT_BREAKPOINT`, as we don't have to verify/assert that the value that would otherwise comes in the payload matches the available breakpoints. That said, you're right to question this as technically having this in reducer makes it "impure", not so much because of the `isServer` call since that's effectively a fixed environment guard for a given runtime, but because of the `window.matchMedia` calls, since they read mutable external browser state rather than deriving the next state purely from the current state and action.

That said, the impurity is narrowly scoped to a single state variable, and it does buys us concrete practical upsides: `SET_BREAKPOINTS` stays atomic and we avoid a potential extra rerender, and `UPDATE_CLIENT_BREAKPOINT` no longer has to trust or validate a payload against the current breakpoint set. Another way to achieve this would be to move the client breakpoint calculation into middleware, but then we'd have three separate pieces of logic handling a single concern, so I'm leaning towards keeping this as is. Let me know if you disagree now that you have more context.

arvinpoddar: Before I answer, can you clarify how including a payload with `UPDATE_CLIENT_BREAKPOINT` would impact the atomicity of `SET_BREAKPOINTS`?

I personally prefer the idea of performing payload validation in `UPDATE_CLIENT_BREAKPOINT` since that makes it easier to trace state changes compared to having the impurity (at least in theory; whether or not this is a practical benefit is a different question). But, the other benefits for `SET_BREAKPOINTS` are compelling, I'm just not seeing the relationship.

agurtovoy: > can you clarify how including a payload with UPDATE_CLIENT_BREAKPOINT would impact the atomicity of SET_BREAKPOINTS?

It wouldn't, these are two separate concerns. Right now `SET_BREAKPOINTS` is atomic because we recalculate the client breakpoint state (by reaching into the  borswer environment) when we update the set of available breakpoints. If we don't do that in the reducer, we'll have to do that from the outside, and the state will be holding a stale client breakpoint value in between.

agurtovoy: I should also add that moving the client breakpoint calculation out of the reducer will force us to have a stale initial state for that prop.

arvinpoddar: Got it. In that case, we can keep the impurity. Maybe we add a comment justifying why?

agurtovoy: Good idea, done: https://github.com/makeswift/makeswift/pull/1295/changes#r3260911783

packages/runtime/src/state/modules/breakpoints.ts
```diff
@@ -41,8 +53,51 @@ export const DEFAULT_BREAKPOINTS: Breakpoints = [
...
+// - We can have a valid, internally consistent *initial* state for the client
+//   breakpoint; if we move the client breakpoint calculation out of the reducer,
+//   we'll be forced to start with a stale initial state, which will introduce
+//   implicit coupling and potential timing issues between the
+//   `useCurrentBreakpoint` hook and the client breakpoint synchronization code
+//   in `BreakpointWatch`
+//
+// - `UPDATE_CLIENT_BREAKPOINT` can compute the client breakpoint from browser
+//    state using the current breakpoint set instead of having to validate and
+//    handle external payload
```
agurtovoy, line 81: Added a clarifying comment on the reducer impurity tradeoffs.

PR #1297, opened 2026-05-18
fix: correctly copy over table form field data

arvinpoddar:
Use the correct copy function for `PropControllerTypes.TableFormFields` data

### DIFF
--- a/packages/runtime/src/prop-controllers/copy.ts
+++ b/packages/runtime/src/prop-controllers/copy.ts
@@ -11,6 +11,7 @@ import {
   copyBackgroundsPropControllerData,
   copyResponsiveColorPropControllerData,
   copyTablePropControllerData,
+  copyTableFormFieldsPropControllerData,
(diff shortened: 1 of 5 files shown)


packages/runtime/src/prop-controllers/copy.ts
```diff
@@ -49,7 +50,7 @@ export function copy(descriptor: Descriptor, value: any, context: CopyContext) {
       return copyResponsiveColorPropControllerData(value, context)
 
     case PropControllerTypes.TableFormFields:
-      return copyTablePropControllerData(value, context)
+      return copyTableFormFieldsPropControllerData(value, context)
```
agurtovoy, line 53: Do we want to add a test for this?

arvinpoddar: Added to the top level React runtime copy test suite

agurtovoy (approved):
🙏

PR #1299, opened 2026-05-18
feat(controls): share data markers across primitive control groups

drewpledger:
## Summary

Swapping one primitive-typed control for another (e.g. `TextInput` → `TextArea`, `Number` → `Slider (wip)`) no longer drops the saved value.

Each primitive group now shares a `[@@makeswift/type]` marker — `'text'` for `TextInput`/`TextArea`/`Code`, `'number'` for `Number` — and every control accepts the full set of legacy markers (`text-input::v1`, `text-area::v1`, `code::v1`, `number::v1`) on read. Rewrite to the shared marker is lazy on next edit; no DB migration.

## Test

https://github.com/user-attachments/assets/d85a76b2-f901-4aa1-9935-98f12f8bade3

https://github.com/user-attachments/assets/3b94503e-2f1c-4186-b886-224b35fa5863

https://github.com/user-attachments/assets/4a108aca-eea9-4eca-817a-eac23b0ff413

- [x] `pnpm --filter @makeswift/controls test` — 871 tests pass (incl. 65 new interop assertions)
- [x] `pnpm --filter @makeswift/runtime test` — 788 tests pass
- [x] In cosmos: swap a `TextInput` prop to `TextArea` (and `Code`); value persists. Repeat `Number` ↔ `Slider`.
- [x] Open a page saved before this PR; legacy marker still parses, no warning.
- [x] Negative: swap `TextInput` → `Number` falls back to `defaultValue`.

### DIFF
--- a/.changeset/control-interop-shared-markers.md
+++ b/.changeset/control-interop-shared-markers.md
@@ -1,0 +1,6 @@
+---
+'@makeswift/controls': patch
+'@makeswift/prop-controllers': patch
+---
+
+Text and number primitives now read shared markers across `@makeswift/controls` and `@makeswift/prop-controllers`, so swapping a prop's control type — or overriding a built-in component and reverting — preserves the saved value. Modern controls write the canonical `'text'` / `'number'` marker on next edit.
--- a/.changeset/control-interop-shared-markers.md
+++ b/.changeset/control-interop-shared-markers.md
@@ -1,0 +1,5 @@
+---
+'@makeswift/controls': patch
+---
+
+`TextInput`, `TextArea`, and `Code` now share a `'text'` data marker; `Number` uses `'number'`. Swapping between controls in the same group preserves the saved value. The deprecated `@makeswift/prop-controllers` text and number markers are also accepted on read, so migrating a prop from the legacy package to the modern equivalent preserves its value too. Rewrite to the canonical marker happens lazily on next edit.
--- a/packages/controls/src/common/data-types.ts
+++ b/packages/controls/src/common/data-types.ts
@@ -1,0 +1,31 @@
+export const ControlDataTypeKey = '@@makeswift/type'
+
+export const DataType = {
+  Text: 'text',
+  Number: 'number',
+} as const
+
+const LegacyTextDataTypes = ['text-input::v1', 'text-area::v1'] as const
+
+const LegacyPropControllerTextDataTypes = [
+  'prop-controllers::text-input::v1',
+  'prop-controllers::text-area::v1',
+] as const
+
+export const AcceptedTextDataTypes = [
+  DataType.Text,
+  ...LegacyTextDataTypes,
+  ...LegacyPropControllerTextDataTypes,
+] as const
+
+const LegacyNumberDataTypes = ['number::v1'] as const
+
+const LegacyPropControllerNumberDataTypes = [
+  'prop-controllers::number::v1',
+] as const
+
+export const AcceptedNumberDataTypes = [
+  DataType.Number,
+  ...LegacyNumberDataTypes,
+  ...LegacyPropControllerNumberDataTypes,
+] as const
--- a/packages/controls/src/common/types.ts
+++ b/packages/controls/src/common/types.ts
@@ -25,3 +25,35 @@ export function isElementReference(
 }
 
 export const ControlDataTypeKey = '@@makeswift/type'
+
+export const TextDataType = 'text' as const
+export const NumberDataType = 'number' as const
+
+export const LegacyTextDataTypes = [
+  'text-input::v1',
+  'text-area::v1',
+  'code::v1',
+] as const
+
+export const LegacyPropControllerTextDataTypes = [
+  'prop-controllers::text-input::v1',
+  'prop-controllers::text-area::v1',
+] as const
+
+export const AcceptedTextDataTypes = [
+  TextDataType,
+  ...LegacyTextDataTypes,
+  ...LegacyPropControllerTextDataTypes,
+] as const
+
+export const LegacyNumberDataTypes = ['number::v1'] as const
+
+export const LegacyPropControllerNumberDataTypes = [
+  'prop-controllers::number::v1',
+] as const
+
+export const AcceptedNumberDataTypes = [
+  NumberDataType,
+  ...LegacyNumberDataTypes,
+  ...LegacyPropControllerNumberDataTypes,
+] as const
--- a/packages/controls/src/common/types.ts
+++ b/packages/controls/src/common/types.ts
@@ -23,5 +23,3 @@ export function isElementReference(
 ): element is ElementReference {
   return !('props' in element)
 }
-
-export const ControlDataTypeKey = '@@makeswift/type'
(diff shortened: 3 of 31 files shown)


packages/controls/src/common/types.ts
```diff
@@ -25,3 +25,24 @@ export function isElementReference(
...
 
 export const ControlDataTypeKey = '@@makeswift/type'
+
+export const TextDataType = 'text' as const
+export const NumberDataType = 'number' as const
+
+export const LegacyTextDataTypes = [
+  'text-input::v1',
+  'text-area::v1',
+  'code::v1',
```
drewpledger: Not sure how we feel about this being in the legacy. I'm sure nobody has used this yet

packages/controls/src/common/types.ts
```diff
@@ -25,3 +25,35 @@ export function isElementReference(
 }
 
 export const ControlDataTypeKey = '@@makeswift/type'
+
+export const TextDataType = 'text' as const
+export const NumberDataType = 'number' as const
+
+export const LegacyTextDataTypes = [
```
agurtovoy: I wouldn't export these if they are not used outside of this file.

packages/controls/src/common/types.ts
```diff
@@ -25,3 +25,35 @@ export function isElementReference(
...
+  'text-input::v1',
+  'text-area::v1',
+  'code::v1',
+] as const
+
+export const LegacyPropControllerTextDataTypes = [
```
agurtovoy: Ditto

packages/controls/src/common/types.ts
```diff
@@ -25,3 +25,35 @@ export function isElementReference(
...
+
+export const AcceptedTextDataTypes = [
+  TextDataType,
+  ...LegacyTextDataTypes,
+  ...LegacyPropControllerTextDataTypes,
+] as const
+
+export const LegacyNumberDataTypes = ['number::v1'] as const
+
+export const LegacyPropControllerNumberDataTypes = [
```
agurtovoy: Same for these two, no need to export.

packages/controls/src/common/types.ts
```diff
@@ -25,3 +25,35 @@ export function isElementReference(
 }
 
 export const ControlDataTypeKey = '@@makeswift/type'
+
+export const TextDataType = 'text' as const
+export const NumberDataType = 'number' as const
```
agurtovoy: I wonder if we should go for `DataType.Text` and `DataType.Number` instead.

packages/controls/src/common/types.ts
```diff
@@ -25,3 +25,35 @@ export function isElementReference(
 }
 
 export const ControlDataTypeKey = '@@makeswift/type'
```
agurtovoy: I know this file is already "impure" (`types.ts` supposed to be strictly TS types), but I'd move the `ControlDataTypeKey` constant and the new ones below to `data-types.ts`

packages/controls/src/controls/__tests__/interop.test.ts
```diff
@@ -0,0 +1,259 @@
...
+
+  describe('type smoke test', () => {
+    test('control factories yield the expected definition classes', () => {
+      expect(TextInput()).toBeInstanceOf(TextInputDefinition)
+      expect(TextArea()).toBeInstanceOf(TextAreaDefinition)
+      expect(Code()).toBeInstanceOf(CodeDefinition)
+      expect(Number()).toBeInstanceOf(NumberDefinition)
+    })
+  })
+})
```
agurtovoy, line 259: Nice tests!

packages/controls/src/common/types.ts
agurtovoy: I'd take it out

agurtovoy (dismissed):
Other than some file/symbol organization (see my inline comments), this looks great!

Since the reading path compatibility is pretty easy to implement, do we want to make a corresponding change in the prop controllers package so when, say, someone upgrades to a custom button and back, they are not faced with a compatibility issue with the built-in component?

packages/controls/src/common/data-types.ts
```diff
@@ -0,0 +1,31 @@
+export const ControlDataTypeKey = '@@makeswift/type'
+
+export const DataType = {
+  Text: 'text',
+  Number: 'number',
+} as const
+
+const LegacyTextDataTypes = ['text-input::v1', 'text-area::v1'] as const
```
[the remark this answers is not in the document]
drewpledger: Nobody is using code yet

.changeset/control-interop-shared-markers.md
drewpledger, line 1: Meaning to delete this

packages/controls/src/common/data-types.ts
```diff
@@ -0,0 +1,31 @@
+export const ControlDataTypeKey = '@@makeswift/type'
+
+export const DataMarker = {
```
agurtovoy: Since we're using "data type" throughout the rest of this file and the controls themselves (e.g. `private static readonly v1DataType = ...`), I think we should be consistent and name this enum `DataType`. It's a little generic, but it is an established term within this package.

agurtovoy: Or maybe `DataTypeTag`?

agurtovoy (dismissed):
See my note on naming, otherwise looks great

packages/prop-controllers/src/text-area/text-area.ts
```diff
@@ -69,8 +76,13 @@ export function getTextAreaPropControllerDataString(
 ): string | undefined {
   return match(data)
     .with(
-      { [ControlDataTypeKey]: TextAreaPropControllerDataV1Type },
-      (v1) => v1.value,
+      {
+        [ControlDataTypeKey]: P.union(
+          TextAreaPropControllerDataV1Type,
+          ...AcceptedTextDataTypes,
+        ),
```
agurtovoy: This should probably match what we do in controls, e.g.

```
{ [ControlDataTypeKey]: P.union(...AcceptedTextDataTypes) },
```

packages/prop-controllers/src/text-input/text-input.ts
```diff
@@ -71,8 +78,13 @@ export function getTextInputPropControllerDataString(
 ): string | undefined {
   return match(data)
     .with(
-      { [ControlDataTypeKey]: TextInputPropControllerDataV1Type },
-      (v1) => v1.value,
+      {
+        [ControlDataTypeKey]: P.union(
+          TextInputPropControllerDataV1Type,
+          ...AcceptedTextDataTypes,
```
agurtovoy: Same as https://github.com/makeswift/makeswift/pull/1299/changes/BASE..faf5f17171a5304cfd0497fc9aa40690ecc527d5#r3284268607

packages/prop-controllers/src/number.ts
```diff
@@ -70,8 +77,13 @@ export function getNumberPropControllerDataNumber(
 ): number {
   return match(data)
     .with(
-      { [ControlDataTypeKey]: NumberPropControllerDataV1Type },
-      (v1) => v1.value,
+      {
+        [ControlDataTypeKey]: P.union(
+          NumberPropControllerDataV1Type,
+          ...AcceptedNumberDataTypes,
+        ),
```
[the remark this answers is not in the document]
agurtovoy: I think the bot is right here.

fikrikarim (commented):
Nice. Glad that we're able to do this cleanly.

PR #1302, opened 2026-05-20
chore: stop exposing internal actions/state/reducer to the builder

agurtovoy:
See inline comments.

### DIFF
--- a/packages/runtime/src/builder/breakpoints/index.ts
+++ b/packages/runtime/src/builder/breakpoints/index.ts
@@ -1,8 +1,6 @@
 export {
-  reducer,
   DefaultBreakpointID,
   DEFAULT_BREAKPOINTS,
-  type State,
   type BreakpointId,
   type Breakpoint,
   type Breakpoints,
--- a/packages/runtime/src/builder/host-api.ts
+++ b/packages/runtime/src/builder/host-api.ts
@@ -1,11 +1,4 @@
-export {
-  type SetBreakpointsAction,
-  makeswiftConnectionInit,
-  registerDocument,
-  unregisterDocument,
-  setBreakpoints,
-  setLocale,
-} from '../state/shared-api'
+export { makeswiftConnectionInit, registerDocument, unregisterDocument } from '../state/shared-api'
 
 export {
   changeDocument,
--- a/packages/runtime/src/state/actions/internal/read-only-actions.ts
+++ b/packages/runtime/src/state/actions/internal/read-only-actions.ts
@@ -7,11 +7,13 @@ import { ElementImperativeHandle } from '../../../runtimes/react/element-imperat
 import { type APIResource, APIResourceType, APIResourceLocale } from '../../../api/types'
 import { type Descriptor as PropControllerDescriptor } from '../../../prop-controllers/descriptors'
 
+import { type Breakpoints } from '../../modules/breakpoints'
 import { type ComponentMeta } from '../../modules/components-meta'
 import { type PropControllersHandle } from '../../modules/prop-controller-handles'
 import { type ComponentType } from '../../modules/react-components'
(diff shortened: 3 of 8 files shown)


arvinpoddar (dismissed):
🙏

packages/runtime/src/builder/breakpoints/index.ts
```diff
@@ -1,8 +1,6 @@
 export {
-  reducer,
```
agurtovoy, line 2: We should never reuse a reducer or its state from the runtime in the builder, however convenient that seems at the moment, haha. The reuse boundary should be the message passing API and stateless utility functions. 

The degree of coupling here is unreasonable and violates all sorts of expectations one can reasonably have about maintaining the runtime's internal state, structure and dependencies.

packages/runtime/src/builder/host-api.ts
```diff
@@ -1,11 +1,4 @@
-export {
-  type SetBreakpointsAction,
-  makeswiftConnectionInit,
-  registerDocument,
-  unregisterDocument,
-  setBreakpoints,
-  setLocale,
-} from '../state/shared-api'
```
agurtovoy, line 8: `setBreakpoints` and `setLocale` aren't really part of the host API. The builder was/is dispatching `setBreakpoints`, but as its own internal action -- a terribly confusing form of reuse. The builder does not dispatch `setLocale` action.

packages/runtime/src/state/shared-api.ts
```diff
@@ -30,9 +27,6 @@ export const SharedActionTypes = {
 
   REGISTER_DOCUMENT: 'REGISTER_DOCUMENT',
   UNREGISTER_DOCUMENT: 'UNREGISTER_DOCUMENT',
-
-  SET_BREAKPOINTS: 'SET_BREAKPOINTS',
-  SET_LOCALE: 'SET_LOCALE',
```
agurtovoy, line 35: Contrary to the comment in line 27, these are not "bidirectional action types".

packages/runtime/src/state/actions/internal/read-only-actions.ts
```diff
@@ -35,8 +37,11 @@ export const ReadOnlyActionTypes = {
   REGISTER_REACT_COMPONENT: 'REGISTER_REACT_COMPONENT',
   UNREGISTER_REACT_COMPONENT: 'UNREGISTER_REACT_COMPONENT',
 
+  SET_BREAKPOINTS: 'SET_BREAKPOINTS',
   UPDATE_CLIENT_BREAKPOINT: 'UPDATE_CLIENT_BREAKPOINT',
 
+  SET_LOCALE: 'SET_LOCALE',
+
```
agurtovoy, line 44: Move `SET_BREAKPOINTS` and `SET_LOCALE` from shared to internal actions, see https://github.com/makeswift/makeswift/pull/1302/changes#r3277468226

packages/runtime/src/state/builder-api/actions.ts
```diff
@@ -11,15 +11,24 @@ import { type ElementSize } from '../middleware/read-write/builder-api/element-s
...
-// note that some of these actions, in addition to being dispatched to the builder,
-// might also be internally intercepted by the host as a part of maintaining the
-// runtime's "read-write" state
+// note that some of these actions (SET_BREAKPOINTS, SET_LOCALE), in addition to
+// being dispatched to the builder, are also internally intercepted by the host
+// as a part of maintaining the runtime's own state
 export const BuilderActionTypes = {
   ...SharedActionTypes,
+  SET_BREAKPOINTS: ReadOnlyActionTypes.SET_BREAKPOINTS,
+  SET_LOCALE: ReadOnlyActionTypes.SET_LOCALE,
```
agurtovoy, line 31: Per the edited comment above, also re-expose these two as builder actions.

PR #1303, opened 2026-05-21
fix: force cache all client fetches on live requests

arvinpoddar:
Note: this does not cache unsuccessful responses from the API (4XXs, 5XXs)

### DIFF
--- a/packages/runtime/src/client/index.ts
+++ b/packages/runtime/src/client/index.ts
@@ -401,7 +401,7 @@ export class MakeswiftClient {
     const response = await fetch(requestUrl.toString(), {
       ...init,
       headers: requestHeaders,
-      ...(siteVersion != null ? { cache: 'no-store' } : {}),
+      cache: siteVersion != null ? 'no-cache' : 'force-cache',
       ...this.fetchOptions(siteVersion),
     })
 
--- a/.changeset/rotten-dolls-tie.md
+++ b/.changeset/rotten-dolls-tie.md
@@ -1,0 +1,5 @@
+---
+"@makeswift/runtime": patch
+---
+
+Force cache Makeswift client requests on live page visits


fikrikarim (commented):
Does this properly invalidate the cache when users publish something?

Might need to test both the Page and region setup.

arvinpoddar:
@fikrikarim good callout, should've included this in the original PR.

Region:

https://github.com/user-attachments/assets/8ac7977b-7cfd-4c7e-b579-aa6b38bb7db6

Page:

https://github.com/user-attachments/assets/0c4c646a-bf69-4416-8b66-90d8021c5eac

packages/runtime/src/client/index.ts
```diff
@@ -401,7 +401,7 @@ export class MakeswiftClient {
     const response = await fetch(requestUrl.toString(), {
       ...init,
       headers: requestHeaders,
-      ...(siteVersion != null ? { cache: 'no-store' } : {}),
+      cache: siteVersion != null ? 'no-cache' : 'force-cache',
```
arvinpoddar, line 404: We can also implement this at the Next.js level by having framework specific `fetchOptions`. The reason I added it here is because `force-cache` is supported by `fetch` itself: https://fetch.spec.whatwg.org/#fetch-method, so it isn't inherently Next.js specific.

fikrikarim: Could you explain how the cache revalidation works? I'm familiar with Next.js caching and revalidation, but not when it's on the `fetch` level. Is Next.js also revalidating this fetch request after [revalidateTag(MAKESWIFT_CACHE_TAG)](https://github.com/makeswift/makeswift/blob/4332f17ac91671555a57d48caa8b135c17b6c1f8/packages/runtime/src/next/api-handler/config/app-router.ts#L46) since we mark the fetch request with `MAKESWIFT_CACHE_TAG`?

Wait, I just realized that it's how [fetch caching works in Next.js](https://nextjs.org/docs/app/api-reference/functions/fetch#optionscache). They're extending `fetch`. I can't find anything related to `fetchOptions`. In that case, it's revalidated because of the tags.

arvinpoddar: `fetchOptions` is something we have in the runtime so that higher-level framework extensions (like for [Next.js](https://github.com/makeswift/makeswift/blob/fb8ec2a256708f4a7e8bb5697705e22e9479add2/packages/runtime/src/next/client.ts#L25)) can define their own properties to spread to the actual `fetch` call here.

Basically, the alternative I'm bringing up is that we can define this fetch policy at the higher level of the runtime, or at the core (like we are currently)

agurtovoy: @arvinpoddar if we force cache here, and a different framework (say, React Router) doesn't natively tie the `fetch` `cache` option with their invalidation infra (which they don't have), would this force the client to "forever" cache these requests?

agurtovoy: On a separate but related note, this made me realize that `MakeswiftClient` isn't using `runtime.fetch`: https://github.com/makeswift/makeswift/pull/1307

arvinpoddar: @agurtovoy I don't think I can give a blanket answer here: I believe this is up to the server-side of the implementing framework to decide how to deal with `force-cache`. That being said, I'm going to close this PR for two reasons:

1. The more impactful change to solve the problem at hand is to make the 404ing endpoints return a 200
2. As you said, MakeswiftClient should probably use `runtime.fetch`, which means we can pass these options from the outside.

PR #1307, opened 2026-05-22
refactor: switch `MakeswiftClient` to using `runtime.fetch`

agurtovoy:
Following up on https://github.com/makeswift/makeswift/pull/1219, where we moved `fetch` parameterization from the framework context to the runtime instance, but didn't unify it with `MakeswiftClient`'s usage.

### DIFF
--- a/packages/runtime/src/next/client.ts
+++ b/packages/runtime/src/next/client.ts
@@ -4,7 +4,6 @@ import { z } from 'zod'
 import { deserializeSiteVersion, type SiteVersion } from '../api/site-version'
 
 import { MakeswiftClient } from '../client'
-import { MAKESWIFT_CACHE_TAG } from './cache'
 
 const previewDataSchema = z.object({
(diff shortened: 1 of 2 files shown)


packages/runtime/src/next/client.ts
```diff
@@ -23,10 +22,6 @@ export class Makeswift extends MakeswiftClient {
   }
 
   fetchOptions(_siteVersion: SiteVersion): Record<string, unknown> {
-    return {
-      next: {
-        tags: [MAKESWIFT_CACHE_TAG],
-      },
```
arvinpoddar, line 29: For posterity, these values are already defined in https://github.com/makeswift/makeswift/blob/45dd9c62c3d63fde562583ce5c4013c14e641e8b/packages/runtime/src/next/fetch.ts#L6

agurtovoy:
Superseded by https://github.com/makeswift/makeswift/pull/1336

PR #1308, opened 2026-05-26
feat: support default opacity value in Color control

andrewreifman:
This update allows users to set a default color opacity value along with the default color value. With changes included in `cosmos`, this default opacity value will also be displayed in the Color control.

https://github.com/user-attachments/assets/caa3daef-5c00-4e7e-a106-11309192755d

### DIFF
--- a/packages/controls/src/controls/color/color.test.ts
+++ b/packages/controls/src/controls/color/color.test.ts
@@ -132,6 +132,7 @@ describe('Color', () => {
     function assignTest(_def: ColorDefinition) {}
     assignTest(Color({ label: 'background' }))
     assignTest(Color({ defaultValue: 'red' }))
+    assignTest(Color({ defaultValue: { color: '#4f46e5', opacity: 0.25 } }))
     assignTest(Color({ label: 'background', labelOrientation: 'horizontal' }))
     assignTest(
       Color({
@@ -176,6 +177,25 @@ describe('Color', () => {
           .resolveValue(data, colorResolver)
           .readStable(),
       ).toBe('rgba(0, 0, 0, 0.5)')
+    })
+
+    test('resolves object defaultValue with opacity when no data', () => {
+      expect(
+        Color({
+          label: 'Color',
+          defaultValue: { color: '#4f46e5', opacity: 0.25 },
+        })
+          .resolveValue(undefined, noOpResourceResolver)
+          .readStable(),
+      ).toBe('rgba(79, 70, 229, 0.25)')
+    })
+
+    test('resolves string defaultValue when no data', () => {
+      expect(
+        Color({ label: 'Color', defaultValue: '#4f46e5' })
+          .resolveValue(undefined, noOpResourceResolver)
+          .readStable(),
+      ).toBe('rgb(79, 70, 229)')
     })
   })
 
--- a/packages/controls/src/controls/color/color.ts
+++ b/packages/controls/src/controls/color/color.ts
@@ -269,38 +269,48 @@ class Definition<C extends Config> extends ControlDefinition<
 
 export class ColorDefinition<C extends Config = Config> extends Definition<C> {}
 
-type ColorDefaultValueInput = string | { color: string; opacity: number }
+type ColorDefaultValueInput =
+  | string
+  | { color: string; opacity: number }
+  | undefined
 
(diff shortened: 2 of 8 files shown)


fikrikarim:
Is this still in draft, or is this ready for a review @andrewreifman?

andrewreifman:
@fikrikarim I requested a review via GH a little too soon. Making some adjustments, so not ready yet.

packages/controls/src/controls/color/color.ts
```diff
@@ -269,16 +269,38 @@ class Definition<C extends Config> extends ControlDefinition<
 
 export class ColorDefinition<C extends Config = Config> extends Definition<C> {}
 
-type UserConfig<D extends Config['defaultValue']> = Config & {
-  defaultValue?: D
+type ColorDefaultValueInput = string | { color: string; opacity: number }
```
agurtovoy: I think it should be `type ColorDefaultValueInput = string | { color: string; opacity: number } | undefined` to reflect the full range of inputs and simplify the type usage below.

packages/controls/src/controls/color/color.ts
```diff
@@ -269,16 +269,38 @@ class Definition<C extends Config> extends ControlDefinition<
...
+
+type NormalizedDefaultValue<D extends ColorDefaultValueInput | undefined> =
+  D extends undefined ? undefined : string
+
+type UserConfig = Omit<Config, 'defaultValue'> & {
+  defaultValue?: ColorDefaultValueInput
```
agurtovoy: There is some weirdness happening here (which was also present in the original code): we define `UserConfig` with the "right" default value, and then override it again and use `UserConfig & { defaultValue?: D }` in the `Color` constructor function below anyway.

packages/controls/src/controls/color/color.ts
```diff
@@ -269,16 +269,38 @@ class Definition<C extends Config> extends ControlDefinition<
...
+  if (typeof defaultValue === 'string') return defaultValue
+  return safeColorString(defaultValue.color, defaultValue.opacity)
+}
+
+export function Color<D extends ColorDefaultValueInput | undefined = undefined>(
+  config?: UserConfig & { defaultValue?: D },
 ): ColorDefinition<NormedConfig<D>> {
-  return new ColorDefinition<NormedConfig<D>>(config ?? {}, 1)
+  if (!config) {
+    return new ColorDefinition<NormedConfig<D>>({} as any, 1)
```
agurtovoy: We try to avoid using `any` and type cast if we can; both are unnecessary here and below.

packages/controls/src/controls/color/color.ts
```diff
@@ -269,16 +269,38 @@ class Definition<C extends Config> extends ControlDefinition<
...
+  }
+
+  const { defaultValue, ...rest } = config
+  const normalized = normalizeDefaultValue(defaultValue)
+  const normalizedConfig =
+    normalized !== undefined ? { ...rest, defaultValue: normalized } : rest
+
```
agurtovoy, line 308: Inlining this logic drops the level of abstraction too much for my taste.

packages/controls/src/controls/color/color.ts
agurtovoy: @andrewreifman I addressed this and other comments below in https://github.com/makeswift/makeswift/pull/1310

packages/controls/src/controls/color/color.test.ts
```diff
@@ -132,6 +132,7 @@ describe('Color', () => {
     function assignTest(_def: ColorDefinition) {}
     assignTest(Color({ label: 'background' }))
     assignTest(Color({ defaultValue: 'red' }))
+    assignTest(Color({ defaultValue: { color: '#4f46e5', opacity: 0.25 } }))
```
agurtovoy, line 135: Also need to update type tests. Addressed in https://github.com/makeswift/makeswift/pull/1310

packages/controls/src/controls/color/color.ts
andrewreifman: Appreciate it!

agurtovoy (approved):
@andrewreifman nice work! Please make sure to squash and merge so that this ends up being a single atomic change in `main`

fikrikarim (approved):
nice!

PR #1315, opened 2026-05-29
Stable class CSS runtime

Codeseph:
The PR description is in: [ARR-1034: Linear PR Description: Stable Class CSS Runtime](https://linear.app/commerce/issue/ARR-1034/linear-pr-description-stable-class-css-runtime)

The new css runtime is in `packages/runtime/src/runtimes/react/css-runtime`


packages/runtime/src/runtimes/react/css-runtime/styles-registry.ts
```diff
@@ -0,0 +1,290 @@
...
+
+    const perPrecedenceProps = [
+      resetPrecedenceStylesProps,
+      mediumPrecedenceStylesProps,
+      highPrecedenceStylesProps,
+    ]
+    return includeEmpty
+      ? perPrecedenceProps
+      : perPrecedenceProps.filter(props => props.href.length > 0)
+  }
```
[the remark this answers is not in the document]
Codeseph: The duplication brought up here does not occur

> "Client MakeswiftStyle emits one tag per class with that class as href"

While the JSX reads as though there's a single `<style>` tag per `href`, React's hoisting logic will often create a combined `<style>` tag (see [(3a) here](https://linear.app/commerce/issue/ARR-801/probe-react-19-hoisting-behavior#comment-28354685)) similarly to what I'm manually doing in the code above. It might be the case that React's logic for combining tags only happens on the server, and that client-side React always hoists tags individually. I don't know. But what matters is the implications for potential duplication of `<style>`s.

The meaningful question is whether a `<style>` that's **manually** created during SSR and combines many different `href` values (and corresponding css) can actually be discovered by client-side React such that it doesn't render new (duplicate) `<style>`s. That question should be asked for all of the cases where we _don't_ rely on React's native hoisting behavior to move `<style>`s from the `<body>` to the `<head>` during SSR. This means probing the behavior in pages router and in frameworks that use our wrappers around `renderToReadableStream` and `renderToPipeableStream`. 

In the screenshot below I'm testing against a pages router host. The top portion is showing the SSR payload, and the bottom is showing what's in the DOM some time after page load. You can see that
(A) the SSR'd `<style>` elements have **not** been duplicated by client-side React
(B) `<style>`s that are newly added by client-side React (i.e., from dragging and dropping in a new component via the Makeswift builder) are hoisted in a way that "recognizes" the `<style>`s that were manually created and placed in the `<head>`. This is demonstrated most directly by the placement of the `<style>` element outlined in green to be below the SSR'd `<style>` of matching `precedence' (medium), and above the SSR'd `<style>` with high `precedence`.

image

packages/runtime/src/next/document.tsx
```diff
@@ -10,24 +8,50 @@ import NextDocument, {
...
+      pull them from the styles registry and place them in the `<head>` manually.
+      This is a special case for pages router, where React hoisting will not work during
+      SSR.
+    */
+    ctx.renderPage = () => {
+      return originalRenderPage({
+        enhanceComponent: Component => props => (
+          <RootStyleRegistry stylesRegistry={stylesRegistry} shouldRenderStyleElements={false}>
+            <Component {...props}></Component>
+          </RootStyleRegistry>
```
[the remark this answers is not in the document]
Codeseph: This occurred to me earlier in the week and I meant to address it. A good catch

packages/runtime/src/runtimes/react/css-runtime/styles-registry.ts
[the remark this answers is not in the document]
Codeseph: Regarding `href` deduplication and how React knows which `href`s have already been hoisted:

image

packages/runtime/src/next/document.tsx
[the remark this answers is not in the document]
Codeseph: Fixed

image

packages/runtime/src/next/document.tsx
```diff
@@ -10,31 +8,50 @@ import NextDocument, {
...
-      ),
+      styles: ssrStyleElements,
     }
   }
 
   render() {
     return (
       <Html>
-        <Head />
+        <Head>{this.props.styles}</Head>
```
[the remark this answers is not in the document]
Codeseph: Oversight from rushing to fix pages router this last week. Fixed.

packages/runtime/src/runtimes/react/css-runtime/styles-registry.ts
```diff
@@ -0,0 +1,290 @@
...
+    const highPrecedenceStylesProps = this.getPropsForHighPrecedenceStyles()
+
+    const perPrecedenceProps = [
+      resetPrecedenceStylesProps,
+      mediumPrecedenceStylesProps,
+      highPrecedenceStylesProps,
+    ]
+    return includeEmpty
+      ? perPrecedenceProps
+      : perPrecedenceProps.filter(props => props.href.length > 0)
```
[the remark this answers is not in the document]
Codeseph: This is true, and is called out in my comment:

image


Right now I'm intentionally omitting the empty tags because, in multi-root scenarios (such as MoS), the consequences are fairly obnoxious. See the empty tags coming from each region below, and note how they "break up" the `precedence` blocks:
image

Something I briefly considered would be to have `renderToReadableStream` and `renderToPipeableStream` return `<style>` attributes/content as an array of objects so that users who are rendering multiple React roots can combine all of them into a sequence of `<style>` elements that is properly ordered by precedence. This of course complicates our APIs in `packages/hono-react`, `packages/react-router`, and `packages/express-react`. But it's worth considering.

packages/runtime/src/runtimes/react/css-runtime/browser-dynamic-styles-bridge.ts
```diff
@@ -0,0 +1,66 @@
...
+  private handleWrite({
+    className,
+    currentData,
+    initialData,
+  }: {
+    className: string
+    currentData: ControlledStyleData
+    initialData: ControlledStyleData
+  }) {
+    if (currentData === initialData || this.isDisposed) return
```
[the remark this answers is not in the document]
Codeseph: A regression introduced yesterday, fixed

packages/runtime/src/next/document.tsx
```diff
@@ -1,7 +1,5 @@
 'use client'
 
-import { cache } from '@emotion/css'
```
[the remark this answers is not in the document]
Codeseph: Unrelated to this PR

packages/runtime/src/runtimes/react/css-runtime/components/controlled-styles.tsx
```diff
@@ -0,0 +1,43 @@
...
+  className: string
+  styleData: ControlledStyleData
+  options?: { precedence?: string }
+}): ReactNode {
+  useEffect(() => {
+    const onBoxModelChange = styleData.onBoxModelChange
+    if (onBoxModelChange == null) return
+    const element = document.querySelector(`.${className}`)
+    return pollBoxModel({ element, onBoxModelChange })
+  }, [className])
```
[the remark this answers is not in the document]
Codeseph: Updated the effect dependency

packages/runtime/src/runtimes/react/controls/control.tsx
```diff
@@ -9,33 +9,51 @@ import {
...
+    [stylesheet.key()],
   )
 
-  stylesheetFactory.useDefinedStyles()
-
-  return children(value)
+  return (
+    <>
+      {children(value)}
+      {styleElements}
```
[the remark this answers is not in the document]
Codeseph: It's fair to say that the namespace here should be sectioned off similarly to `useLegacyControlledStyle`. I'll change the namespace to include the prop path components through the Slate leaf, which will look like

image

and prevent each `ControlValue` (per leaf/element, see screenshot at bottom of comment) from trying to render every `<style>` element used across the Slate editor. In practice the actual `<style>` elements get deduplicated by React (by `href`), but this is still a good catch

----------------------------------------

image

packages/runtime/src/runtimes/react/css-runtime/components/controlled-styles.tsx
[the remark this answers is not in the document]
Codeseph: update: removed it because it led to this flashing of the margin/padding areas

https://github.com/user-attachments/assets/4ff61ccb-21f2-4e9e-bba2-bd6d3fffc327

packages/runtime/src/runtimes/react/css-runtime/hooks/use-css-reset.ts
```diff
@@ -0,0 +1,30 @@
...
+  }
+  stylesRegistry.setCssReset(styleData)
+  const href = `makeswift-css-reset-${contentHash}`
+
+  const styleElement = React.createElement(MakeswiftStyle, {
+    key: contentHash,
+    href: href,
+    css: content,
+    precedence: MakeswiftStylePrecedence.RESET,
+  })
```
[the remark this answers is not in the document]
Codeseph: For the so-called "base styles"

The old CSS runtime did not run them through the Emotion cache, which means that there was no CSS preprocessor being used on them (recall that the Emotion serializer itself does not perform preprocessing: [ARR-780: Document layers involved in Emotion's conversion of a JS styles object to a css string](https://linear.app/commerce/issue/ARR-780/document-layers-involved-in-emotions-conversion-of-a-js-styles-object))

I tested what difference would be made by utilizing the preprocessor versus not, and there was no difference, so I went ahead and added the usage just to avoid confusion, and in case the contents of our "base styles" ever change such that this becomes consequential... 

image

Probably a waste of time, but oh well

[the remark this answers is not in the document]
Codeseph: For CSS resets - 
these were in fact being run through a CSS preprocessor, by way of the Emotion cache
 
image

So I've also updated CSS resets to flow through our css preprocessor

packages/runtime/package.json
```diff
@@ -201,11 +201,11 @@
...
   "peerDependencies": {
-    "@types/react": "^18.0.0 || ^19.0.0",
-    "@types/react-dom": "^18.0.0 || ^19.0.0",
+    "@types/react": "^19.0.0",
+    "@types/react-dom": "^19.0.0",
     "next": "^13.4.0 || ^14.0.0 || ^15.0.0",
-    "react": "^18.0.0 || ^19.0.0",
-    "react-dom": "^18.0.0 || ^19.0.0"
+    "react": "^19.0.0",
+    "react-dom": "^19.0.0"
```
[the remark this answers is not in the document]
Codeseph: changed 'next' peer dependency to be  >= 15

packages/runtime/src/next/components/tests/components/__snapshots__/button-component-rendering.test.tsx.snap
```diff
@@ -194,120 +221,112 @@ exports[`correctly renders button component with link 1`] = `
...
+      box-shadow 0.15s ease-in-out;
   }
-
-  .emotion-3:hover {
+  .ms-wg7e91 :hover {
     background: #000000;
   }
-
-  .emotion-3:active {
+  .ms-wg7e91 :active {
```
Codeseph: Looking into what's going on here (why there's a space between the class and the pseudoselector, which changes the behavior)

Codeseph: This was a place where we were relying on one of Emotion's stylis plugins to assume an intended `&` that isn't explicitly written in the JS styles object, for example:

image


The result of assuming an intended `&` before a pseudo class is that the resulting preprocessed css would be `.my-class:active` instead of `.my-class :active`. I intentionally did not recreate this plugin behavior in the new css runtime. More on this later, the issue where we originally discovered this is: https://linear.app/commerce/issue/ARR-780/document-layers-involved-in-emotions-conversion-of-a-js-styles-object#comment-2ffdf228.

Codeseph: Anyway, to deal with this I've gone back and updated builtin components to be explicit by including the nesting selector (`&`) before pseudo classes

packages/runtime/src/next/components/tests/components/__snapshots__/box-component-rendering.test.tsx.snap
```diff
@@ -82,55 +145,57 @@ exports[`correctly renders box component with background image 1`] = `
...
     align-self: auto;
   }
 }
 
 @media only screen and (max-width: 575px) {
-  .emotion-5 {
+  .ms-7cpjp3 {
     -webkit-align-self: auto;
     -ms-flex-item-align: auto;
+    -ms-grid-row-align: auto;
```
Codeseph: Our css preprocessor (`stylis`) is configured to use a vendor prefixer, and that prefixer is adding the `ms-grid-row-align` property that wasn't in the snapshot output previously. 

Image

The vendor prefixer is responsible for adding `-webkit-align-self`, `-ms-flex-item-align`, and `-ms-grid-row-align` declarations in this case. You can see this by removing the `prefixer` from the middleware config shown in the screenshot above, and regenerating this snapshot:

Image

So I think the "before" snapshot output here (for the `emotion-5` class) can be explained by Emotion using a custom vendor prefixer that produced `-webkit-align-self` and `-ms-flex-item-align` property declarations, but **not** `ms-grid-row-align`. 

And the "after" snapshot output here (for `ms-7cpjp3`) can be explained by our use of the out-of-the-box vendor prefixer from `stylis`, rather than a custom prefixer

packages/runtime/src/next/components/tests/components/__snapshots__/box-component-rendering.test.tsx.snap
```diff
@@ -82,55 +145,57 @@ exports[`correctly renders box component with background image 1`] = `
...
     align-self: auto;
   }
 }
 
 @media only screen and (min-width: 576px) and (max-width: 768px) {
-  .emotion-5 {
+  .ms-7cpjp3 {
     -webkit-align-self: auto;
     -ms-flex-item-align: auto;
+    -ms-grid-row-align: auto;
```
Codeseph: see my [comment](https://github.com/makeswift/makeswift/pull/1315/changes#r3807396299) on where this new declaration comes from

packages/runtime/src/next/components/tests/components/__snapshots__/box-component-rendering.test.tsx.snap
```diff
@@ -82,55 +145,57 @@ exports[`correctly renders box component with background image 1`] = `
...
+.ms-1vg8v23 {
   display: flex;
 }
 
 @media only screen and (min-width: 769px) {
-  .emotion-5 {
+  .ms-7cpjp3 {
     -webkit-align-self: auto;
     -ms-flex-item-align: auto;
+    -ms-grid-row-align: auto;
```
Codeseph: see my [comment](https://github.com/makeswift/makeswift/pull/1315/changes#r3807396299) on where this new declaration comes from

packages/runtime/src/next/components/tests/components/__snapshots__/box-component-rendering.test.tsx.snap
```diff
@@ -64,7 +73,61 @@ exports[`correctly renders box component with background image 1`] = `
...
+}
+
+@media only screen and (max-width: 575px) {
+  .ms-1ypjwdd {
+    padding-top: 10px;
+    padding-right: 10px;
+    padding-bottom: 470px;
+    padding-left: 10px;
+  }
+}
```
Codeseph, line 201: It's not lined up in a way that's easy to see in the diff, but this big block is not new. The class `ms-1ypjwdd` is replacing `emotion-8` in the old snapshot, and `emotion-8` is further down in the 'before' 

Image

packages/runtime/src/next/components/tests/components/__snapshots__/box-component-rendering.test.tsx.snap
```diff
@@ -1,43 +1,52 @@
...
 exports[`correctly renders box component with background image 1`] = `
-.emotion-0 {
+.ms-12226a4 {
+  display: flex;
+  -webkit-box-flex-wrap: wrap;
+  -webkit-flex-wrap: wrap;
+  -ms-flex-wrap: wrap;
+  flex-wrap: wrap;
+  width: 100%;
+}
```
Codeseph, line 11: Not new

Image

packages/runtime/src/next/components/tests/components/__snapshots__/box-component-rendering.test.tsx.snap
```diff
@@ -64,7 +73,61 @@ exports[`correctly renders box component with background image 1`] = `
...
+
+@media only screen and (max-width: 575px) {
+  .ms-be1rlk {
+    border-top: 0px solid black;
+    border-right: 0px solid black;
+    border-bottom: 0px solid black;
+    border-left: 0px solid black;
+  }
+}
```
Codeseph, line 228: Image

packages/runtime/src/next/components/tests/components/__snapshots__/box-component-rendering.test.tsx.snap
```diff
@@ -141,123 +206,61 @@ exports[`correctly renders box component with background image 1`] = `
...
     align-content: flex-start;
   }
 }
 
 @media only screen and (max-width: 575px) {
-  .emotion-12 {
+  .ms-m001dn {
     -webkit-align-content: flex-start;
-    -ms-flex-line-pack: flex-start;
+    -ms-flex-line-pack: start;
```
Codeseph: I believe the change from `flex-start` to `start` comes from our vendor prefixer, which is the default from `stylis` and behaves slightly differently than Emotion's prefixer. 

`ms-flex-line-pack` appears to be ancient, so I'm hoping we can not worry about this

packages/runtime/src/next/components/tests/components/__snapshots__/button-component-rendering.test.tsx.snap
```diff
@@ -179,7 +179,34 @@ exports[`correctly renders button component with SEND_EMAIL link 1`] = `
...
+}
+
+@media only screen and (max-width: 575px) {
+  .ms-scsqh5 {
+    margin-top: 60px;
+    margin-right: auto;
+    margin-bottom: 80px;
+    margin-left: auto;
+  }
+}
```
Codeseph: replacing `emotion-2`, no actual diff here

packages/runtime/src/next/components/tests/components/__snapshots__/image-component-rendering.test.tsx.snap
```diff
@@ -146,16 +101,62 @@ exports[`correctly renders image component 1`] = `
...
+  .ms-k7d2qc {
+    opacity: 1;
+  }
+}
+
+@media only screen and (max-width: 575px) {
+  .ms-k7d2qc {
+    opacity: 1;
+  }
+}
```
Codeseph: Not new

Image

packages/runtime/src/next/components/tests/controls/__snapshots__/slot-control.test.tsx.snap
```diff
@@ -719,96 +574,44 @@ NodeList [
...
-}
-
-.emotion-0 {
-  display: flex;
-  -webkit-box-flex-wrap: wrap;
-  -webkit-flex-wrap: wrap;
-  -ms-flex-wrap: wrap;
-  flex-wrap: wrap;
-  width: 100%;
-}
```
Codeseph, line 774: Lots of duplicates of the original `emotion-0` block here. Not sure if this was an issue with the old jest snapshot serializer for Emotion, or something else.

Anyway, `emotion-0` is replaced with `ms-12226a4`

packages/runtime/src/next/components/tests/controls/__snapshots__/slot-control.test.tsx.snap
```diff
@@ -719,96 +574,44 @@ NodeList [
...
-    -ms-flex-preferred-size: calc(100% * 1.00000);
-    flex-basis: calc(100% * 1.00000);
-    min-width: calc(100% * 1.00000);
+  .ms-v6iyjh {
+    -webkit-flex-basis: calc(100% * 1);
+    -ms-flex-preferred-size: calc(100% * 1);
+    flex-basis: calc(100% * 1);
+    min-width: calc(100% * 1);
   }
 }
```
Codeseph, line 606: The difference in precision here,

before:
`-webkit-flex-basis: calc(100% * 1.00000);`

after:
`-webkit-flex-basis: calc(100% * 1);`

is a result of using prettier to format content for the jest snapshot:

Image

You can see this yourself by commenting out the prettier usage in the serializer:

Image

packages/runtime/src/next/components/tests/controls/rich-text-v2-control/__snapshots__/client.test.tsx.snap
```diff
@@ -11,20 +10,13 @@ NodeList [
...
 }
 
-.emotion-0 {
-  display: inline-block;
-  width: 0;
-  max-width: 100%;
-  white-space: nowrap;
-  opacity: 0.333;
-  vertical-align: text-top;
-}
```
Codeseph, line 695: Duplicate of the `emotion-0` block above

packages/runtime/src/next/components/tests/controls/rich-text-v2-control/__snapshots__/client.test.tsx.snap
Codeseph, line 1: For this particular test, the snapshot output after migrating to the new css runtime has significantly more css content than prior to the migration. This is expected in cases where we're dealing with "controlled" styles that share the exact same css.

In the old css runtime, hash-based class names would ensure that two distinct Slate leaves which shared the exact same style data would share the same class name. This plays out for several class names within this file. For example, in the screenshot below you can see that `emotion-4` is used as the class name on two separate Slate leaves. 

Image

But the new css runtime only bases class name on a content hash when dealing with "uncontrolled" styles. 

Because the style data for Slate leaves is editable in Makeswift, the class names generated are not based on a content hash. Instead, they are based on a "prop path". In the case of the Slate leaves to which the `emotion-4` class was being applied, the "prop path" differs. Because the "prop path" differs, we get two distinct css classes for what was previously `emotion-4`.

Image

packages/runtime/src/next/components/tests/controls/rich-text-v2-control/__snapshots__/server.test.tsx.snap
```diff
@@ -42,7 +42,7 @@ NodeList [
...
...media only screen and (max-width: 575px){.ms-fvnubh{color:hsla(238,87%,49%,1);font-family:Lato;font-size:16px;}}@media only screen and (min-width: 769px){.ms-xf7dky{color:hsla(238,87%,49%,1);font-family:Lato;font-size:16px;}}@media only screen and (min-width: 576px) and (max-width: 768px){.ms-xf7dky{color:hsla(238,87%,49%,1);font-family:Lato;font-size:16px;}}@media only screen and (max-width: 575px){.ms-xf7dky{color:hsla(238,87%,49%,1);font-family:Lato;font-size:16px;}}@media only screen and (min-width: 769px){.ms-ws78w0{color:hsla(238,87%,49%,1);font-family:Lato;font-size:16px;}}@media only screen and (min-width: 576px) and (max-width: 768px){.ms-ws78w0{color:hsla(238,87%,49%,1);font-family:Lato;font-size:16px;}}@media only screen and (max-width: 575px){.ms-ws78w0{color:hsla(238,87%,49%,1);font-family:Lato;font-size:16px;}}@media only screen and (min-width: 769px){.ms-od8iyo{color:hsla(238,87%,49%,1);font-family:Lato;font-size:16px;}}@media only screen and (min-width: 576px) and (max-width: 768px){.ms-od8iyo{color:hsla(238,87%,49%,1);font-family:Lato;font-size:16px;}}@media only screen and (max-width: 575px){.ms-od8iyo{color:hsla(238,87%,49%,1);font-family:Lato;font-size:16px;}}",
```
Codeseph: The list styles are missing here, I'm looking into why

For `ms-nqi6pk`, as an example, we should be seeing

`.ms-nqi6pk{list-style-position:inside;-webkit-padding-start:20px;padding-inline-start:20px;list-style-type:disc;}.ms-nqi6pk ul{list-style-type:circle;}.ms-nqi6pk ul ul{list-style-type:square;}`

Codeseph: So far: they're present in the server-rendered payload in actual testing

Image

and when running this test, they're present in the styles registry

packages/runtime/src/next/components/tests/controls/rich-text-v2-control/__snapshots__/server.test.tsx.snap
```diff
@@ -42,7 +42,7 @@ NodeList [
 
 exports[`renders provided text content: component styles 1`] = `
 [
-  ".mswft-ytumd6{-webkit-text-decoration:none;text-decoration:none;}.mswft-1uk1gs8{margin:0;}.mswft-vz2etd{padding:0.5em 10px;font-size:1.25em;font-weight:300;border-left:5px solid rgba(0, 0, 0, 0.1);}.mswft-99zmsg{list-style-position:inside;-webkit-padding-start:20px;padding-inline-start:20px;list-style-type:disc;}.mswft-99zmsg ul{list-style-type:circle;}.mswft-99zmsg ul ul{list-style-type:square;}.mswft-dgqoak{list-style-position:inside;-webkit-padding-start:20px;padding-inline-start:20px;list-style-type:decimal;}@media only screen and (min-width: 769px){.mswft-144fjko{color:hsla(238,87%,49%,1);font-family:Lato;font-size:16px;}}@media only screen and (min-width: 576px) and (max-width: 768px){.mswft-144fjko{color:hsla(238,87%,49%,1);font-family:Lato;font-size:16px;}}@media only screen and (max-width: 575px){.mswft-144fjko{color:hsla(238,87%,49%,1);font-family:Lato;font-size:16px;}}",
```
Codeseph, line 45: The unordered list styles here are not present in the snapshot after migration.

This is expected, since the element data that backs this test (at `packages/runtime/src/next/components/tests/controls/fixtures/rich-text-v2.ts`) does not have `"type": "unordered-list"`. 

In `packages/runtime/src/slate/BlockPlugin/index.tsx` we have a switch statement based on `props.element.type`. In the new css runtime, where we render style elements alongside the content that depends on them, we don't render the style element for unordered list styles unless we're actually entering that case (which we don't, in this test).

Image

Codeseph: For sanity, I manually tested unordered lists, and they work:

Image

packages/runtime/src/next/components/tests/controls/rich-text-v2-control/__snapshots__/server.test.tsx.snap
Codeseph: Comparing the before and after here, what we're seeing is that

- `mswft-ytumd6` has been replaced with `ms-1fa0n4z`, which is 1:1 
- `mswft-1uk1gs8` has been replaced with `ms-1azh6ix`, which is 1:1
- `mswft-vz2etd` has no replacement in the "after" snapshot
	- this is expected, since this class is for a "quote style", and we do not render a quote style in this test (see [my comment](https://github.com/makeswift/makeswift/pull/1315/changes#r3815122104))
- `mswft-99zmsg` and `mswft-dgqoak` have no replacement in the "after" snapshot
	- this is expected, since these classes are for list styles, and we do not render list styles in this test (see [my comment](https://github.com/makeswift/makeswift/pull/1315/changes#r3815122104))
- `mswft-144fjko` (spanning 3 media queries), has been replaced with `ms-1lqw28m`, `ms-fvnubh`, `ms-xf7dky`, `ms-ws78w0`, and `ms-od8iyo` in the "after" snapshot
	- this is expected, because the original class `mswft-144fjko` was shared across 5 Slate leaves. Since styles for Slate leaves are considered "controlled" styles by the new css runtime, their class names are calculated using a prop path rather than a content hash. Because the prop path for each Slate leaf differs, the class name for each Slate leaf now differs (see [my comment from earlier](https://github.com/makeswift/makeswift/pull/1315/changes#r3814245822))

packages/runtime/src/next/components/tests/prop-controllers/__snapshots__/page-border-prop-controller.test.tsx.snap
Codeseph, line 1: This new snapshot file exists because the corresponding test was previously testing styles using [toHaveStyleRule](https://emotion.sh/docs/@emotion/jest#tohavestylerule) from `@emotion/jest`, and we've removed that dependency.

packages/runtime/src/next/components/tests/prop-controllers/__snapshots__/page-margin-prop-controller.test.tsx.snap
Codeseph, line 1: See my comment [here](https://github.com/makeswift/makeswift/pull/1315/changes#r3815563088)

packages/runtime/src/next/components/tests/prop-controllers/__snapshots__/page-padding-prop-controller.test.tsx.snap
Codeseph, line 1: see my comment [here](https://github.com/makeswift/makeswift/pull/1315/changes#r3815563088)

packages/runtime/src/next/components/tests/prop-controllers/__snapshots__/page-width-prop-controller.test.tsx.snap
Codeseph, line 1: see my comment [here](https://github.com/makeswift/makeswift/pull/1315/changes#r3815563088)

packages/runtime/src/next/components/tests/prop-controllers/page-border-prop-controller.test.tsx
```diff
@@ -140,8 +138,7 @@ describe('Page', () => {
       ),
     )
 
-    expect(screen.getByTestId(testId)).toHaveStyleRule('border-top', '1px solid black', {
```
Codeseph, line 143: [toHaveStyleRule](https://emotion.sh/docs/@emotion/jest#tohavestylerule) is from `@emotion/jest`, which has been removed

packages/runtime/src/next/components/tests/prop-controllers/page-margin-prop-controller.test.tsx
```diff
@@ -133,8 +131,7 @@ describe('Page', () => {
       ),
     )
 
-    expect(screen.getByTestId(testId)).toHaveStyleRule('margin-top', 'auto', {
```
Codeseph, line 136: See my comment [here](https://github.com/makeswift/makeswift/pull/1315/changes#r3815592010)

packages/runtime/src/next/components/tests/prop-controllers/page-padding-prop-controller.test.tsx
```diff
@@ -133,8 +131,7 @@ describe('Page', () => {
       ),
     )
 
-    expect(screen.getByTestId(testId)).toHaveStyleRule('padding-top', '17px', {
```
Codeseph, line 136: See my comment [here](https://github.com/makeswift/makeswift/pull/1315/changes#r3815592010)

packages/runtime/src/next/components/tests/prop-controllers/page-width-prop-controller.test.tsx
```diff
@@ -123,8 +121,7 @@ describe('Page', () => {
       ),
     )
 
-    expect(screen.getByTestId(testId)).toHaveStyleRule('width', '17px', {
```
Codeseph, line 126: See my comment [here](https://github.com/makeswift/makeswift/pull/1315/changes#r3815592010)

packages/runtime/src/next/components/tests/__snapshots__/global-element-rendering.test.tsx.snap
```diff
@@ -105,52 +106,30 @@ exports[`Page correctly renders a global element 1`] = `
...
-  .emotion-0 {
-    width: 700px;
-  }
-}
-
-@media only screen and (max-width: 575px) {
-  .emotion-0 {
-    width: 700px;
-  }
-}
```
Codeseph, line 128: `emotion-0` replaced by `ms-1sr6xt0`

The `emotion-0` statements here did not disappear. These were duplicated in the original snapshot (see the lines below, which exactly match the selection).

Statements for `ms-1sr6xt0` appear only once in the new snapshot.

packages/runtime/src/testing/jest-snapshot-serializer.ts
Codeseph, line 1: ** This file is (I believe) the only functionality in the PR that I used an LLM to write - sorry if it's ugly. **

If you look at the top of various jest snapshots, you will see css that was collected, formatted, and dumped into the snapshot by this serializer.

Image

packages/runtime/src/jest-setup.ts
```diff
@@ -41,6 +47,81 @@ if (typeof window !== 'undefined') {
...
+    parsedRules.reverse().forEach(rule => {
+      this.insertRule(rule.cssText, 0)
+    })
+
+    return this
+  }
+  window.CSSStyleSheet.prototype.replaceSync = function (cssText) {
+    return Promise.resolve(this.replace(cssText))
+  }
+}
```
Codeseph: I'm sorry that I'm adding this cruft instead of upgrading testing libraries so that we don't need it. It looked like the upgrade sequence was going to be extensive, and I'm so far behind already

https://linear.app/commerce/issue/ARR-1022/handle-jsdom-missing-apis-for-adopted-stylesheets-and-cssstylesheet#comment-087d4577

packages/runtime/package.json
```diff
@@ -216,17 +216,13 @@
     }
   },
   "dependencies": {
-    "@emotion/cache": "^11.14.0",
-    "@emotion/css": "^11.13.5",
     "@emotion/serialize": "^1.3.3",
```
Codeseph, line 224: See the `CSS Transformation Pipeline` section of the [PR description](https://linear.app/commerce/issue/ARR-1034/linear-pr-description-stable-class-css-runtime#comment-d5fb3bb3) if you're curious why this library was retained

Codeseph: TLDR: this does a lot more than crawl a styles object for declarations and shove them into a string

packages/runtime/package.json
```diff
@@ -254,17 +251,18 @@
     "slate": "^0.91.4",
     "slate-hyperscript": "^0.77.0",
     "slate-react": "^0.91.7",
+    "stylis": "^4.4.0",
```
Codeseph, line 259: This is the css preprocessor that was used by the Emotion cache (see [ARR-780: Document layers involved in Emotion's conversion of a JS styles object to a css string](https://linear.app/commerce/issue/ARR-780/document-layers-involved-in-emotions-conversion-of-a-js-styles-object))

While we've retained Emotion's `serializeStyles`, we've gotten rid of the Emotion cache. So the preprocessor itself had to be brought in to fill the hole there.

packages/runtime/package.json
```diff
@@ -243,6 +239,7 @@
     "js-base64": "^3.7.7",
     "jwt-decode": "^4.0.0",
     "ky": "^1.14.3",
+    "murmurhash-js": "^1.0.0",
```
Codeseph, line 247: Used in class name generation

packages/runtime/package.json
```diff
@@ -216,17 +216,13 @@
...
     "@emotion/serialize": "^1.3.3",
-    "@emotion/server": "^11.11.0",
-    "@emotion/sheet": "^1.4.0",
-    "@emotion/utils": "^1.4.2",
     "@makeswift/controls": "workspace:*",
     "@makeswift/next-plugin": "workspace:*",
     "@makeswift/prop-controllers": "workspace:*",
     "@reduxjs/toolkit": "^2.8.2",
     "@use-gesture/react": "^10.2.24",
+    "clsx": "^2.1.1",
```
Codeseph, line 230: See usages in the builtin components, as well as by the `composeStyles` utility.

This is (partially) filling a hole from the loss of Emotion's `cx` utility

packages/runtime/src/jest-setup.ts
```diff
@@ -41,6 +47,81 @@ if (typeof window !== 'undefined') {
...
+ *
+ * See `__react` keys in: `react/packages/react-dom-bindings/src/client/ReactDOMComponentTree.js`
+ */
+function clearDocumentCachedHoistedResources() {
+  for (const key of Object.getOwnPropertyNames(document)) {
+    if (key.startsWith('__reactResources$')) {
+      delete (document as unknown as Record<string, unknown>)[key]
+    }
+  }
+}
```
Codeseph, line 65: For example:

Image

packages/runtime/src/unstable-framework-support/index.ts
```diff
@@ -33,12 +33,12 @@ export { GoogleFontLink } from '../runtimes/react/components/GoogleFontLink'
...
-  createRootStyleCache,
   RootStyleRegistry,
   DefaultRootStyleRegistry,
-  styleTagHtml,
-  StyleTagSSR,
   type RootStyleProps,
 } from '../runtimes/react/root-style-registry'
 
+export { type StylesRegistry } from '../runtimes/react/css-runtime/styles-registry'
+export { styleTagHtml, createMakeswiftStylesRegistry } from '../runtimes/react/css-runtime/utils'
```
Codeseph, line 44: `createMakeswiftStylesRegistry` is needed by the wrappers around streaming APIs (`renderToReadableStream`, `renderToPipeableStream`) that live in `packages/hono-react`, `packages/react-router`, and `packages/express-react`, and also by Pages router (see `packages/runtime/src/next/document.tsx`).

They create a Makeswift styles registry (rather than relying on the default `RootStyleRegistry` to create one for them) so that they can serialize its content into style elements and manually add them to the page `<head>`.

In the case of the streaming APIs, this is done because they can be called without passing a full document (including a `<head>`), and React's native hoisting mechanism requires access to a full document to work. More on this buried within [ARR-924: Explore React 19 hoisting behavior for `renderToPipeableStream` and `renderToReadableStream`](https://linear.app/commerce/issue/ARR-924/explore-react-19-hoisting-behavior-for-rendertopipeablestream-and), and in the `Applying Styles to the Browser` section of the [Linear PR description](https://linear.app/commerce/issue/ARR-1034/linear-pr-description-stable-class-css-runtime).

In the case of Pages router, the need is driven by the same underlying problem - React not having access to a `<head>` when rendering style elements.

apps/nextjs-pages-router/src/pages/_app.tsx
```diff
@@ -30,7 +33,9 @@ export default function App({
         siteVersion={siteVersion}
         locale={locale}
       >
-        <Component {...pageProps} />
+        <RootStyleRegistry>
```
Codeseph, line 36: The `RootStyleRegistry` should now always be included under the runtime provider.

- All of the hooks that are used to render Makeswift styles (`useStyle`, `useControlledStyles`, `useBaseStyles`, etc.) store their resulting style data in a styles registry that is provided by the `RootStyleRegistry` 
- The `RootStyleRegistry` is responsible for setup that's necessary to handle dynamic style updates
- Context provided by the `RootStyleRegistry` also includes settings such as the custom class prefix (previously configurable by users via the Emotion cache key), a boolean controlling whether css resets are enabled, a boolean controlling whether style elements should be rendered alongside components, etc.).

packages/runtime/src/runtimes/react/root-style-registry.tsx
```diff
@@ -84,39 +26,43 @@ export type RootStyleProps = {
...
-    <CacheContext.Provider value={cache}>
-      <CSSResetEnabledContext.Provider value={enableCssReset ?? DEFAULT_CSS_RESET_ENABLED}>
-        {children}
-      </CSSResetEnabledContext.Provider>
-    </CacheContext.Provider>
+    <StylesContextProvider
+      classNamePrefix={classNamePrefix}
+      enableCssReset={enableCssReset}
+      stylesRegistry={stylesRegistry}
+      shouldRenderStyleElements={shouldRenderStyleElements}
```
Codeseph, line 46: Because we utilize React's hoisting mechanism we are able to write style elements in JSX alongside the components that depend on them (see any of the builtin components that are modified by this PR).

But there are certain cases where we can't depend on hoisting during SSR. In these cases, we intentionally disable rendering the style elements, opting instead to serialize the content of the Makeswift styles registry into an html string (containing `<style>`s) that can be manually added to a document.

packages/runtime/src/runtimes/react/root-style-registry.tsx
```diff
@@ -84,39 +26,43 @@ export type RootStyleProps = {
...
+  useDynamicBrowserStyleUpdates(stylesRegistry)
+
   return (
-    <CacheContext.Provider value={cache}>
-      <CSSResetEnabledContext.Provider value={enableCssReset ?? DEFAULT_CSS_RESET_ENABLED}>
-        {children}
-      </CSSResetEnabledContext.Provider>
-    </CacheContext.Provider>
+    <StylesContextProvider
+      classNamePrefix={classNamePrefix}
```
Codeseph, line 43: Class name prefix was previously configurable, albeit in a less obvious way. The old `cacheKey` prop from `RootStyleProps` was used by the Emotion cache to prefix all generated class names. `createRootStyleCache` (called by the `DefaultRootStyleRegistry` would set this to `mswft` if a value was not provided by the user.

In our Pages router example app, where no `DefaultRootStyleRegistry` was used, there was no `createRootStyleCache` call to set this default. That's why css class names generated in our Pages router example had a prefix of `css` instead of `mswft` - the default prefix value of `css` was coming from Emotion, rather than us.

Image

packages/runtime/src/runtimes/react/root-style-registry.tsx
```diff
@@ -84,39 +26,43 @@ export type RootStyleProps = {
...
+      enableCssReset={enableCssReset}
+      stylesRegistry={stylesRegistry}
```
Codeseph, line 45: All of the hooks which serve as entrypoints for the css runtime's functionality (`useStyle`, `useControlledStyles`, etc.) save the data that they produce to the `stylesRegistry` provided here.

The styles registry:
- is needed by the current mechanism for handling dynamic style updates, which works by subscribing to the registry
- is critically important for cases where we need to serialize the registry's content and handle creating style elements ourselves, rather than through React
- is especially convenient for tests, since we can spy on the registry

packages/runtime/src/runtimes/react/root-style-registry.tsx
```diff
@@ -84,39 +26,43 @@ export type RootStyleProps = {
...
+  enableCssReset = DEFAULT_CSS_RESET_ENABLED,
+  shouldRenderStyleElements = DEFAULT_SHOULD_RENDER_STYLE_ELEMENTS,
 }: PropsWithChildren<{
-  cache: StyleCache
+  stylesRegistry: StylesRegistry
+  classNamePrefix?: string
   enableCssReset?: boolean
+  shouldRenderStyleElements?: boolean
 }>) {
+  useDynamicBrowserStyleUpdates(stylesRegistry)
```
Codeseph, line 39: Kicks off the mechanism for handling runtime style edits

packages/runtime/src/next/root-style-registry.tsx
Codeseph, line 1: This is now essentially a passthrough to the `DefaultRootStyleRegistry`.

Note that in the case of Pages router, we have special setup (not user-facing, in `packages/runtime/src/next/document.tsx`) that uses the non-default `RootStyleRegistry` for handling styles during SSR.

packages/runtime/src/runtimes/react/controls/rich-text-v2/EditableTextV2/editable-text-v2.tsx
```diff
@@ -116,18 +119,59 @@ export function EditableTextV2({ text, definition, control }: Props) {
...
+  */
+  const decorate = useCallback(
+    ([node, path]: NodeEntry): Range[] => {
+      if (!SlateText.isText(node)) return []
+      return [
+        {
+          anchor: { path, offset: 0 },
+          focus: { path, offset: node.text.length },
+          // @ts-expect-error - this is a custom property that is used by the css runtime
+          _makeswiftLeafPath: `slate-editor-path-[${path.join(',')}]`,
```
Codeseph, line 153: While this gives us what we need for Slate leaves, we need an analogous "prop path"  for Slate **elements**, and I didn't find a way to make that available such that I could avoid recalculating it. The calculation of a "prop path" for Slate elements is in `packages/runtime/src/runtimes/react/controls/rich-text-v2/EditableTextV2/render-element.tsx`

packages/runtime/src/runtimes/react/controls/rich-text-v2/EditableTextV2/render-element.tsx
```diff
@@ -1,31 +1,65 @@
...
+  pathComponents,
+  parentStylesheet,
+  ...props
+}: RichTextV2ElementProps) {
+  const slateEditor = useSlateStatic()
   function initialRenderElement(props: RenderElementProps) {
     return props.children
   }
 
+  const slatePath = () => {
```
Codeseph, line 24: A separate Slate path is also necessary for leaves, see `decorate` in `packages/runtime/src/runtimes/react/controls/rich-text-v2/EditableTextV2/editable-text-v2.tsx`

packages/runtime/src/runtimes/react/controls/rich-text-v2/EditableTextV2/render-element.tsx
```diff
@@ -4,28 +4,51 @@ import { RichTextV2Definition } from '../../../../../controls/rich-text-v2'
...
+        `element`,
+      ]
+
       return (
-        <ControlValue definition={control.definition} data={control.getElementValue(props.element)}>
+        <ControlValue
+          definition={control.definition}
+          data={control.getElementValue(props.element)}
+          elementKey={pseudoElementKey}
+          propPathComponents={elementPathComponents}
```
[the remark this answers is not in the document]
Codeseph: (fixed earlier)

packages/runtime/src/testing/jest-snapshot-serializer.ts
```diff
@@ -0,0 +1,116 @@
...
+  }
+  return rules
+}
+
+function isElementArray(val: any): boolean {
+  return val && Array.isArray(val) && val.every(v => v instanceof Element)
+}
+
+function isMakeswiftClassName(className: string): boolean {
+  return className.startsWith(`${DEFAULT_CSS_CLASS_NAME_PREFIX}-`)
```
Codeseph, line 66: This creates a problem if you write a test with a `RootStyleRegistry` that's configured to use a custom class name prefix. Which isn't great...

packages/runtime/src/runtimes/react/css-runtime/components/makeswift-style.tsx
```diff
@@ -0,0 +1,20 @@
...
+  href,
+  css,
+  precedence,
+}: {
+  href: string
+  css: string
+  precedence?: string
+}) {
+  const { shouldRenderStyleElements } = useStylesContext()
+  if (!shouldRenderStyleElements) return null
```
Codeseph, line 16: See my [comment](https://github.com/makeswift/makeswift/pull/1315/changes#r3816204426) regarding this configuration

packages/runtime/src/runtimes/react/css-runtime/components/makeswift-style.tsx
```diff
@@ -0,0 +1,20 @@
...
+  return (
+    <style href={href} precedence={precedence ?? DEFAULT_CSS_PRECEDENCE}>
```
Codeseph, line 18: In the hoisting-managed style element that React creates, `href` becomes `data-href`, and `precedence` becomes `data-precedence`

packages/runtime/src/runtimes/react/css-runtime/__tests__/runtime-integration/controlled-styles-integration.test.tsx
```diff
@@ -0,0 +1,100 @@
...
+      expect(registryNotifyListenersSpy).toHaveBeenCalledTimes(2)
+      expect(adoptedStylesheetApplierSpy).toHaveBeenCalled()
+      expect(document.adoptedStyleSheets).toHaveLength(1)
+      const adoptedStylesheet = document.adoptedStyleSheets[0]
+
+      jestSnapshotSource = {
+        styleElementCss: formatStyleElementContent(styleElement),
+        adoptedStylesheetCss: formatStylesheetContent(adoptedStylesheet),
+      }
+    }
```
Codeseph, line 71: Relying on an uncached API resource to trigger the dynamic style update pathway is bad. We should be explicit and dispatch an element tree data change. I'd need to do some digging to figure out how to assemble the op correctly. It's probably worth doing soon

packages/runtime/src/runtimes/react/css-runtime/stylesheet-engine.ts
```diff
@@ -0,0 +1,80 @@
...
+    return this.breakpointsData
+  }
+
+  defineStyle(
+    resolvedStyle: ResolvedStyle,
+    onBoxModelChange?: (boxModel: BoxDisplayModel | null) => void,
+  ): string {
+    const joinedPropPath = this.propPathComponents.join('.')
+    const className = generateClassName({
+      data: `${this.documentKey}-${this.elementKey}-${joinedPropPath}`,
```
Codeseph: Including the document key in class names is consequential in certain situations where it prevents us from depending on a fragile ordering of css.

For example, consider this sequence involving multiple React roots and a global element:

(1) `Root A` mounts. It contains `global element X`.
(2) I edit the color of some text in `Root A`'s `global element X` from black to blue.
(3) `Root B` mounts. It also contains `global element X`. The font color is correct (blue).
(4) In `Root B` I edit the color of the same text element in `global element X` from blue to red. The change takes effect visually in both `Root A`'s `global element X` and `Root B`'s `global element X`.
(5) I delete the instance of `global element X` in `Root B`.
(6) In `Root A` I edit the color of the text element in `global element X` from red to purple. But nothing happens visually.

What went wrong?

In step (2):
the edit in this step leads to the creation of an adopted stylesheet for `Root A`'s `global element X`. The last content in this stylesheet includes the declaration setting color to blue. Everything is fine so far.

In step (4):
the edit in this step leads to the creation of an adopted stylesheet for `Root A`'s `global element X`. The document's `adoptedStylesheets` property would now show two adopted stylesheets. One created by `Root A`, and one created by `Root B`.

In step (6): 
the edit in this step leads to a [replaceSync](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet/replaceSync) of the content of the adopted stylesheet for `Root A`'s `global element X`, rather than the creation of a new stylesheet. But this adopted stylesheet was created before the adopted stylesheet from `Root B`. Because (in this example) our class names are _not_ based on a document ID, the declarations in the second adopted stylesheet (from `Root B`) target the same class name, and their latter position means that they win out over the same-specificity declarations from the first adopted stylesheet (from `Root A`). So the declaration setting color to red wins out over the declaration setting color to purple.

packages/runtime/src/runtimes/react/components/page/Page.tsx
```diff
@@ -8,23 +8,36 @@ import { usePageSnippets } from '../hooks/use-page-snippets'
...
+import { CSSObject } from '@emotion/serialize'
+import { useBaseStyles } from '../../css-runtime/hooks/use-base-styles'
 
 type Props = {
   page: MakeswiftPageDocument
   rootDocument: Document
   metadata?: boolean | PageMetadataSettings
 }
 
+const pageBaseStyles: CSSObject = {
```
Codeseph: This was previously written in the PageHead component (`packages/runtime/src/runtimes/react/components/page/PageHead.tsx`) which handled moving style elements into the page `<head>` via wrapping them in the [FrameworkContext](https://github.com/makeswift/makeswift/blob/d2f7e6b96c0bcebfbbff7b08b28cf1508f87313b/packages/runtime/src/runtimes/react/components/framework-context.tsx#L45)'s `Head`. For pages router, [this `Head` is the next/head component](https://github.com/makeswift/makeswift/blob/d2f7e6b96c0bcebfbbff7b08b28cf1508f87313b/packages/runtime/src/next/components/framework-provider/pages-router/index.tsx#L1). 

Because pages router handles `next/head` content separately from the main React tree, attempting to render the style element returned from the css runtime's `useBaseStyles` hook (or, any other hook that returns a `MakeswiftStyle`):

`const { styleElement: baseStylesElement } = useBaseStyles({ styles: pageBaseStyles })`

would lead to a problem. `MakeswiftStyle` reads from context (`StylesContext`) to check whether style elements are intended to be rendered alongside content. But this context is provided in the main React tree (by `StylesContextProvider` within the `RootStyleRegistry`), which the `next/head` component does not have access to. So attempting to render the  `baseStylesElement` would throw an error for context being unavailable.

This was the issue that we worked out in[ this thread](https://bigcommerce.slack.com/archives/C0ASKGAMQKU/p1785882403335279)

packages/runtime/src/runtimes/react/components/page/PageHead.tsx
```diff
@@ -76,16 +76,6 @@ export function PageHead({ document: page, metadata = {} }: Props): ReactNode {
...
-      <PageStyle precedence="high" href="makeswift-base-styles">
-        {`
-        html {
-            font-family: sans-serif;
-        }
-        div#__next {
-            overflow: hidden;
-        }
-        `}
-      </PageStyle>
```
Codeseph, line 88: See [this comment](https://github.com/makeswift/makeswift/pull/1315/changes#r3822087924) on base styles moving to the `Page` component

packages/runtime/src/runtimes/react/components/resolve-props.tsx
```diff
@@ -19,22 +21,48 @@ export function ResolveProps({
   const [legacyDescriptors, definitions] = useControlDefs({ elementType: element.type })
 
   const elementKey = element.key
-  const stylesheetFactory = useStylesheetFactory()
+  const { getStylesheet, styleElements } = useControlledStyles({ namespace: elementKey })
```
Codeseph, line 24: This is the main point of access for the "controlled styles pathway".

The other points to be aware of are:
- usages of `useLegacyControlledStyle` in `resolveLegacyDescriptorProp`
- usages of `useControlledStyles` by RichTextV2 plugin controls in the `ControlValue` component.

Codeseph: These other points use a separate `namespace` to avoid attempting to render the same style elements

packages/runtime/src/runtimes/react/controls/control.tsx
```diff
@@ -9,33 +9,51 @@ import {
...
   definition,
   children,
+  elementKey,
+  propPathComponents,
   control,
 }: ControlValueProps): ReactNode {
-  const stylesheetFactory = useStylesheetFactory()
-  const id = `cv-${useCssId()}`
+  const breakpoints = useBreakpoints()
+  const styleNamespace = `${elementKey}-controlValue-${propPathComponents.join('-')}`
```
Codeseph, line 33: Intentionally namespaced separately from styles that will be rendered in `ResolveProps`, see [here](https://github.com/makeswift/makeswift/pull/1315/changes#r3822159554)

packages/runtime/src/runtimes/react/hooks/use-css-id.ts
```diff
@@ -1,5 +0,0 @@
-import { useId } from 'react'
-
-export function useCssId(): string {
-  return useId().replaceAll(':', '') // CSS class names prohibit colons
```
Codeseph, line 4: We ran into multiple problems from attempting to use this. Some of that saga is recorded [here](https://linear.app/commerce/issue/ARR-938/investigate-useid-hydration-mismatches-leading-to-unresponsive-styles). There was also the matter of avoiding [configuration that would be required](https://react.dev/reference/react/useId#specifying-a-shared-prefix-for-all-generated-ids) to use this hook in multi-root contexts.

packages/runtime/src/runtimes/react/css-runtime/stylesheet-engine.ts
fikrikarim: Is it intentional that key() is using `:` while the className data use `-`?

```
  key(): string {
    return `${this.documentKey}-${this.elementKey}:${this.propPathComponents.join('.')}`
  }
```

I was thinking if it's possible to just use `this.key()` on the data:

```
const className = generateClassName({ data: this.key() })
```

But perhaps they're not related? The AI is against using it here:

```
The two have different invariants, and they're about to diverge:

- key() must change when memoized resolution goes stale. It's the dep array at control.tsx:50.
- The class name must never change, or the whole PR is pointless.
```

packages/runtime/src/runtimes/react/css-runtime/styles-registry.ts
```diff
@@ -0,0 +1,296 @@
...
+    const combinedHref = hrefValues.join(' ')
+    const combinedCss = allCss.join('')
+    return {
+      href: combinedHref,
+      css: combinedCss,
+      precedence: MakeswiftStylePrecedence.RESET,
+    }
+  }
+
+  private getPropsForMediumPrecedenceStyles(): ComponentProps<typeof MakeswiftStyle> {
```
fikrikarim, line 248: nit: i wonder if it's worth having one helper for going through the same loop. something like (from AI):

```
private static combine(
  entries: Array<{ href: string; css: string }>,
  precedence: MakeswiftStylePrecedence,
): ComponentProps<typeof MakeswiftStyle> {
  return {
    href: entries.map(e => e.href).join(' '),
    css: entries.map(e => e.css).join(''),
    precedence,
  }
}

private getPropsForResetPrecedenceStyles() {
  return StylesRegistry.combine(
    [...this.cssResets.values()].map(d => ({
      href: `makeswift-css-reset-${d.contentHash}`,
      css: d.css,
    })),
    MakeswiftStylePrecedence.RESET,
  )
}

private getPropsForMediumPrecedenceStyles() {
  return StylesRegistry.combine(
    [
      ...[...this.uncontrolledClassStyles.values()].map(d => ({ href: d.className, css: d.css })),
      ...[...this.getAllControlledStyles().values()].map(d => ({ href: d.className, css: d.css })),
      ...[...this.keyframes.values()].map(d => ({ href: d.keyframesName, css: d.css })),
    ],
    MakeswiftStylePrecedence.MEDIUM,
  )
}

private getPropsForHighPrecedenceStyles() {
  return StylesRegistry.combine(
    [...this.baseStyles.values()].map(d => ({
      href: `makeswift-base-styles-${d.contentHash}`,
      css: d.css,
    })),
    MakeswiftStylePrecedence.HIGH,
  )
}
```

packages/runtime/src/runtimes/react/css-runtime/components/controlled-styles.tsx
```diff
@@ -0,0 +1,45 @@
...
+  options?: { precedence?: string }
+}): ReactNode {
+  // Do not list the style data's box model callback in dependencies, as this will lead
+  // to visually jarring overlay redraws in the builder
+  useEffect(() => {
+    const onBoxModelChange = styleData.onBoxModelChange
+    if (onBoxModelChange == null) return
+    const element = document.querySelector(`.${className}`)
+    return pollBoxModel({ element, onBoxModelChange })
+  }, [className])
```
fikrikarim, line 41: Is this calling `pollBoxModel` on live pages too? Should we not call it when it's rendered on live pages?

packages/runtime/src/runtimes/react/root-style-registry.tsx
fikrikarim, line 39: no need to do this right now. but i wonder if it's worth lazy loading it and only initialize it if it's in the builder.

packages/runtime/src/runtimes/react/css-runtime/hooks/use-style.ts
```diff
@@ -0,0 +1,98 @@
...
+ * This is intended to be used for "uncontrolled" (not editable in Makeswift) styles
+ * such as those defined inline in builtin components.
+ */
+export function useStyle(style: CSSObject) {
+  const { classNamePrefix, stylesRegistry } = useStylesContext()
+  const className = generateClassName({
+    data: JSON.stringify(style),
+    classNamePrefix,
+  })
+  const { css } = toCssStatements(style, className)
```
fikrikarim: Thoughts on this? Seems right to me. From AI:

> No memoization on the uncontrolled path. useStyle has no useMemo. Every render, for every call site, it does `JSON.stringify(style) → murmur3 → Emotion serializeStyles → stylis compile + serialize`. The style objects are static literals — the output is byte-identical every time. Carousel has 22 useStyle calls; that's 22 full CSS serializations per render, forever.
>
> Emotion's cache made this cheap: hash, look up, skip insertion on a hit. Here the registry dedupes storage but nothing dedupes computation. And the fix is nearly free, because the class name is derived from `JSON.stringify(style)` alone and is computed before the CSS (use-style.ts:19-23):

```
const className = generateClassName({ data: JSON.stringify(style), classNamePrefix })
const existing = stylesRegistry.getUncontrolledClassStyles().get(className)
const { css } = existing ?? toCssStatements(style, className)
```

packages/runtime/src/runtimes/react/components/resolve-props.tsx
```diff
@@ -19,22 +21,48 @@ export function ResolveProps({
...
+      {styleElements}
+    </>
+  )
 }
+
+/**
+ * This level of memoization exists so that changes to element data do not lead to re-rendering
+ * a component unless they actually result in changes to that component's resolved prop values.
+ */
+const MemoizedComponent = memo(function MemoizedComponent({
```
fikrikarim, line 60: Is this true? Is the `renderFn` stable across render?

packages/runtime/src/runtimes/react/css-runtime/hooks/use-legacy-controlled-style.ts
```diff
@@ -0,0 +1,53 @@
+import { CSSObject } from '@emotion/serialize'
+import { generateClassName } from '../utils'
+import React from 'react'
+import { useStylesContext } from './use-styles-context'
+import { ControlledStyles } from '../components/controlled-styles'
+import { toCssStatements } from '../serialize-css'
+import { ControlledStyleData } from '../types'
+import { useDocumentKey } from '../../hooks/use-document-context'
+
+export function useLegacyControlledStyle(style: CSSObject, elementKey: string, propName: string) {
```
fikrikarim, line 10: do we need to/is it worth memoize this so it doesn't need to process the CSS on every render?

packages/controls/src/stylesheet.ts
```diff
@@ -25,4 +25,5 @@ export interface Stylesheet {
   ): string
 
   child(id: string): Stylesheet
+  key(): string
```
fikrikarim, line 28: do we need to add a changeset for `@makeswift/controls`?

packages/runtime/src/next/components/tests/components/__snapshots__/button-component-rendering.test.tsx.snap
fikrikarim: Do we need to add this behavior change to the changelog? I think it's affecting `unstable_StyleV2`?

packages/runtime/src/runtimes/react/components/page/Page.tsx
```diff
@@ -8,23 +8,36 @@ import { usePageSnippets } from '../hooks/use-page-snippets'
...
+  html: {
+    fontFamily: 'sans-serif',
+  },
+  'div#__next': {
+    overflow: 'hidden',
+  },
+}
```
agurtovoy: Oh, wow, I didn't realize we still had these in `PageHead`! `div#__next` [has been dead code for a while](https://github.com/vercel/next.js/discussions/45541), but I think we're also overreaching with the `fontFamily` style here. 

There is obviously a backward compatibility concern so we can't outright remove these, but I'd definitely move these out of our generic React runtime subtree (`src/runtimes/react`) and into the Next-specific version in `src/next/components/page.tsx`.

packages/controls/src/stylesheet.ts
agurtovoy: Good point, I think we need to in order to procure a new package version.

Codeseph:
cursor review

packages/runtime/src/runtimes/react/css-runtime/styles-registry.ts
```diff
@@ -0,0 +1,320 @@
...
+      allCss.push(keyframesData.css)
+    }
+    const combinedHref = hrefValues.join(' ')
+    const combinedCss = allCss.join('')
+    return {
+      href: combinedHref,
+      css: combinedCss,
+      precedence: MakeswiftStylePrecedence.MEDIUM,
+    }
+  }
```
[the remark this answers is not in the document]
Codeseph: > serializeToHtmlStyleTags merges every medium-precedence rule into one <style> whose data-href is a space-joined list of class names

> StyleElementUpdater also queries style[data-href="${className}"], which cannot find the combined server tag, so builder updates can lose to stale SSR CSS.

The first quote is true, but this does not include controlled styles, and controlled styles are the only styles for which the `StyleElementUpdater` needs to be able to find an isolated style element in the DOM (re: the second quote). See that `getPropsForMediumPrecedenceStyles` (which feeds `serializeToHtmlStyleTags`) handles only uncontrolled class styles and keyframes. Controlled styles are fed to `serializeToHtmlStyleTags` by way of `getPropsForCustomPrecedenceStyles`. 

> React 19 matches hoisted styles by the full href string, so hydration will not reuse the server tags and will insert duplicates

We know from experimentation that this is not correct - the hoisting mechanism is fairly smart here, and will not create the duplicates mentioned in the quote. See the second paragraph of [this comment](https://github.com/makeswift/makeswift/pull/1315#discussion_r3739557079).

packages/runtime/src/runtimes/react/css-runtime/stylesheet-engine.ts
Codeseph: I agree with you (Fikri > LLM) and don't see any reason why `generateClassName` doesn't pass `this.key()` as the `data` value.

I think this escaped notice because I added `key` to the `Stylesheet` interface some time after implementing the class name generation logic, but I'm not sure. Good catch

Codeseph: updated

packages/runtime/src/runtimes/react/css-runtime/components/controlled-styles.tsx
Codeseph: I think you're probably right, although for now I was just matching the old functionality of `useDefinedStyles` (previously returned from `useStylesheetFactory`, which is removed by this PR)

packages/runtime/src/runtimes/react/root-style-registry.tsx
Codeseph: What scares me is that we depend on the mechanism here any time style prop data re-resolves, rather than depending on it only for explicit user edits to style data made in the builder. The relationship feels kind of fuzzy in my head in a way that makes me not want to use the "are we in the builder" check to gate what's here. Maybe I shouldn't feel that hesitation, though, and only feel that way because I don't have a good grip on the runtime client's fetching/caching logic

packages/runtime/src/runtimes/react/css-runtime/hooks/use-style.ts
Codeseph: Yeah I don't see any harm. I've now added something similar to the suggestion above to the `useStyle`, `useCssReset`, and `useBaseStyles` hooks

packages/runtime/src/runtimes/react/components/resolve-props.tsx
Codeseph: Not when the relevant element type has legacy descriptors, but in other cases yes. Added a note with the stipulation.

I recall having difficulties handling this (including what you asked about in this [comment](https://github.com/makeswift/makeswift/pull/1315#discussion_r3831330579)) for legacy descriptors early in the project. Perhaps it's worth another look now, although my main goal here was to avoid re-rendering beefy custom components when their controlled class's style data changes

Codeseph:
cursor review

packages/controls/src/stylesheet.ts
Codeseph: Moved this change into a separate commit + added the changeset

packages/runtime/src/runtimes/react/components/page/Page.tsx
Codeseph: Just moved these into `packages/runtime/src/next/components/page.tsx` in a separate commit

Codeseph:
Moving back to draft while I look into [ARR-1169: RSC margin/padding overlays disappear after re-render request to `/__rsc-element`](https://linear.app/commerce/issue/ARR-1169/rsc-marginpadding-overlays-disappear-after-re-render-request-to-rsc)

Codeseph:
cursor review

Codeseph:
cursor review

apps/vite-rsc-hono-react-cloudflare/src/lib/makeswift/styles.ts
Codeseph, line 1: This isn't intended to be cruft (I do not actually care about using a custom class name prefix in this example app). `rootStyleProps` is exported to point out the importance of passing the same custom class name prefix (if a custom prefix is actually being used) to all the different locations that will generate class names. These locations will include `RootStyleRegistry` usages **as well as the RSC render context**. 

Image

packages/runtime/src/runtimes/react/css-runtime/components/controlled-styles.tsx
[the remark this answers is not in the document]
agurtovoy: We should disable the corresponding ESLint warning:

deps warning

Also, I'd expand on the comment below, as it currently explains only the user-visible symptom ("visually jarring overlay redraws in the builder"), but not the reason for it (why does the callback identity change when user updates the style?) nor why is it safe to continue using the old callback.


>   // Do not list the style data's box model callback in dependencies, as this will lead
  // to visually jarring overlay redraws in the builder

packages/runtime/src/runtimes/react/css-runtime/components/controlled-styles.tsx
Codeseph: I ended up already needing the suggested check for whether we're in the builder in order to resolve [ARR-1174: RSC box model slightly offset on page load, corrects on first remeasure?](https://linear.app/commerce/issue/ARR-1174/rsc-box-model-slightly-offset-on-page-load-corrects-on-first-remeasure)

apps/vite-rsc-hono-react-cloudflare/src/lib/makeswift/styles.ts
agurtovoy: Could we add this as a comment in the source code itself? 🙏

agurtovoy (dismissed):
Looking great ❤️

packages/runtime/src/runtimes/react/poll-box-model.ts
```diff
@@ -2,15 +2,16 @@ import { BoxDisplayModel, getBox } from '../../state/modules/read-write/box-mode
...
   onBoxModelChange,
 }: {
-  element: Element | null
+  getElement: () => Element | null
   onBoxModelChange(boxModel: BoxDisplayModel | null): void
 }): () => void {
   let currentBoxModel: BoxDisplayModel | null = null
 
   const handleAnimationFrameRequest = () => {
+    const element = getElement()
```
fikrikarim, line 14: I assume `document.querySelector` is fast enough to do in every `requestAnimationFrame` and there's no significant performance hit?

packages/runtime/src/runtimes/react/css-runtime/components/controlled-styles.tsx
```diff
@@ -36,8 +36,20 @@ export function ControlledStyle({
...
-    const element = document.querySelector(`.${className}`)
-    return pollBoxModel({ element, onBoxModelChange })
+
+    const findElement = () => document.querySelector(`.${className}`)
+    let element = findElement()
+
+    return pollBoxModel({
+      getElement: () => {
+        // The 'isConnected' check is important to RSC re-renders, where the DOM node bearing the controlled class gets replaced
+        if (element == null || !element.isConnected) {
```
fikrikarim: oh nice I didn't know `isConnected` existed

fikrikarim (dismissed):
nice work

packages/runtime/src/runtimes/react/poll-box-model.ts
Codeseph: A good consideration. Fortunately it doesn't run on every animation frame -

For `ControlledStyle`, the `getElement` passed to `pollBoxModel` actually only calls `document.querySelector` conditionally, when `(element == null || !element.isConnected)`. So `document.querySelector` should run quite infrequently.

And for other callers of `pollBoxModel`, the `getElement` passed in simply returns an element, so the concern doesn't apply:
image

packages/runtime/src/runtimes/react/css-runtime/components/controlled-styles.tsx
Codeseph: me neither, until the other day

apps/vite-rsc-hono-react-cloudflare/src/lib/makeswift/styles.ts
Codeseph: Done

packages/runtime/src/runtimes/react/css-runtime/components/controlled-styles.tsx
```diff
@@ -0,0 +1,71 @@
...
+  }, [className, isInBuilder])
+
+  return (
+    <MakeswiftStyle
+      href={className}
+      css={styleData.css}
+      precedence={getControlledStylePrecedence(styleData)}
+    />
+  )
+}
```
Codeseph, line 71: I can't find the feedback item after my rebase, but I've updated the comment above the `useEffect` here and disabled the ESLint warning

fikrikarim (approved):

agurtovoy (approved):

PR #1316, opened 2026-05-30
refactor(runtime): add `details` text to the fallback error component

agurtovoy:
See inline comments.

### DIFF
--- a/packages/runtime/src/components/shared/FallbackComponent/FallbackComponent.tsx
+++ b/packages/runtime/src/components/shared/FallbackComponent/FallbackComponent.tsx
@@ -4,10 +4,11 @@ import { Warning20 } from '../../icons/Warning20'
 
 type Props = {
   text: string
+  details?: string
 }
 
 export const FallbackComponent = forwardRef(function FallbackComponent(
-  { text }: Props,
+  { text, details }: Props,
   ref: Ref<HTMLDivElement>,
 ) {
   return (
@@ -30,6 +31,7 @@ export const FallbackComponent = forwardRef(function FallbackComponent(
     >
       <Warning20 />
       <span>{text}</span>
+      <span className={useStyle({ display: 'none' })}>{details}</span>
     </div>
   )
 })


packages/runtime/src/components/shared/FallbackComponent/FallbackComponent.tsx
```diff
@@ -4,10 +4,11 @@ import { Warning20 } from '../../icons/Warning20'
 
 type Props = {
   text: string
+  details?: string
 }
 
 export const FallbackComponent = forwardRef(function FallbackComponent(
-  { text }: Props,
+  { text, details }: Props,
```
agurtovoy, line 11: Traditionally, since this is a component that can show up on your live website, we kept the text message that is passed here short and generic (e.g. "Error rendering component"), and would log a more detailed message to the error console.

Now that we're rendering some of the custom components server-side, logging to the console (especially in production) is no longer adequate, but we still want the ability to surface more detailed diagnostics to our developer users. 

Adding a new `details` prop to enable that.

packages/runtime/src/components/shared/FallbackComponent/FallbackComponent.tsx
```diff
@@ -30,6 +31,7 @@ export const FallbackComponent = forwardRef(function FallbackComponent(
     >
       <Warning20 />
       <span>{text}</span>
+      <span className={useStyle({ display: 'none' })}>{details}</span>
```
agurtovoy, line 34: Currently rendered as hidden with an expectation that you'd examine the DOM to get the details. Might be okay to render as a tooltip (`role="tooltip"`), but I think this is adequate for now.

arvinpoddar: What if we didn't use `useStyle` and just defined the style itself inline, so we're not introducing a reliance on our CSS runtime functioning correctly?

arvinpoddar (approved):
Clever idea.

packages/runtime/src/components/shared/FallbackComponent/FallbackComponent.tsx
agurtovoy: Not a bad idea; we already use`useStyle` on the parent div (lines 15-30), though, and I didn't feel like touching it.

PR #1321, opened 2026-06-05
fix(controls): write per-control legacy data-type markers

drewpledger:
## Summary

Unifying the text and number controls onto shared `text` / `number` data-type markers caused a regression when upgrading the builder's runtime. This PR restores the per-control legacy markers on write while keeping read-side interop intact.

https://github.com/user-attachments/assets/22d46366-022e-403e-8353-79b33cc0c818

### DIFF
--- a/packages/controls/src/common/data-types.ts
+++ b/packages/controls/src/common/data-types.ts
@@ -1,31 +1,39 @@
 export const ControlDataTypeKey = '@@makeswift/type'
 
-export const DataType = {
+// TODO: Unify tag serialization for controls that share the same underlying
+// data type. For example, all plain-text controls should write the same tag
+// (e.g. `text`). This requires refactoring `ControlDefinition` so prop editing
+// remains backward-compatible when the host is running an older runtime.
+const DataType = {
   Text: 'text',
   Number: 'number',
 } as const
 
-const LegacyTextDataTypes = ['text-input::v1', 'text-area::v1'] as const
+export const TextDataTypes = {
+  textInput: 'text-input::v1',
+  textArea: 'text-area::v1',
+  code: 'code::v1',
+} as const
 
-const LegacyPropControllerTextDataTypes = [
+const PropControllerTextDataTypes = [
   'prop-controllers::text-input::v1',
   'prop-controllers::text-area::v1',
 ] as const
 
 export const AcceptedTextDataTypes = [
   DataType.Text,
-  ...LegacyTextDataTypes,
-  ...LegacyPropControllerTextDataTypes,
+  TextDataTypes.textInput,
+  TextDataTypes.textArea,
+  TextDataTypes.code,
+  ...PropControllerTextDataTypes,
 ] as const
 
-const LegacyNumberDataTypes = ['number::v1'] as const
+export const NumberDataTypes = { number: 'number::v1' } as const
 
-const LegacyPropControllerNumberDataTypes = [
-  'prop-controllers::number::v1',
-] as const
+const PropControllerNumberDataTypes = ['prop-controllers::number::v1'] as const
 
 export const AcceptedNumberDataTypes = [
   DataType.Number,
-  ...LegacyNumberDataTypes,
-  ...LegacyPropControllerNumberDataTypes,
+  NumberDataTypes.number,
+  ...PropControllerNumberDataTypes,
 ] as const
--- a/packages/prop-controllers/src/__tests__/interop.test.ts
+++ b/packages/prop-controllers/src/__tests__/interop.test.ts
@@ -4,7 +4,9 @@ import {
   Code,
   ControlDataTypeKey,
   Number as ModernNumber,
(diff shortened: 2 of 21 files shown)


packages/controls/src/common/data-types.ts
```diff
@@ -1,11 +1,16 @@
 export const ControlDataTypeKey = '@@makeswift/type'
 
+// This was an attempt to unify the data types. It caused a regression when upgrading the builder's runtime.
 export const DataType = {
```
fikrikarim: Should we remove the export here? Is this still used anywhere?

packages/prop-controllers/src/__tests__/interop.test.ts
```diff
@@ -112,12 +112,13 @@ describe('@makeswift/controls -> @makeswift/prop-controllers interop', () => {
...
       // 2. Host overrides with custom component using modern controls; on next
-      //    edit, modern toData rewrites the marker to canonical 'text'.
+      //    edit, modern toData rewrites the marker to its own legacy marker
+      //    'text-input::v1'.
       const afterModernRead = ModernTextInput().fromData(persisted as never)
       const afterModernEdit = ModernTextInput().toData(afterModernRead!)
 
       expect(afterModernEdit).toEqual({
-        [ControlDataTypeKey]: 'text',
+        [ControlDataTypeKey]: 'text-input::v1',
```
fikrikarim: Should we import the variable here instead of hardcoding the string?

packages/prop-controllers/src/__tests__/interop.test.ts
```diff
@@ -171,7 +172,7 @@ describe('@makeswift/controls -> @makeswift/prop-controllers interop', () => {
       const afterModernEdit = ModernNumber().toData(afterModernRead!)
 
       expect(afterModernEdit).toEqual({
-        [ControlDataTypeKey]: 'number',
+        [ControlDataTypeKey]: 'number::v1',
```
fikrikarim: Same as the previous comment

fikrikarim (commented):
Overall it's looking good. I have two questions:

1. Is it only a temporary solution to write the `LegacyTextDataTypes`, or is this a long term solution?

If this is a long-term solution, should we drop the `Legacy` from the naming, since it's not a `Legacy` anymore, and actually the most up-to-date data type?

2. And, the `DataType` is now technically the Legacy, since no control write that data type anymore.
```
export const DataType = {
  Text: 'text',
  Number: 'number',
} as const
```
Should we bite the bullet and just drop support for that data type? Otherwise, it's gonna be confusing if someone read the code 6 months from now and see that there's `DataType` that's not being used at all. 

AFAIK, most people that accidentally had that data type on their site already had seen the broken component on the builder anyway. If we want to play safe, perhaps we rename that one to be the `Legacy` ones?

packages/controls/src/common/data-types.ts
```diff
@@ -1,11 +1,16 @@
...
   Text: 'text',
   Number: 'number',
 } as const
 
-const LegacyTextDataTypes = ['text-input::v1', 'text-area::v1'] as const
+export const LegacyTextDataTypes = {
```
agurtovoy: I'd just call this `TextDataTypes` to avoid tainting the current code with the "legacy" prefix.

packages/controls/src/common/data-types.ts
```diff
@@ -14,18 +19,20 @@ const LegacyPropControllerTextDataTypes = [
...
+  LegacyTextDataTypes.textInput,
+  LegacyTextDataTypes.textArea,
+  LegacyTextDataTypes.code,
   ...LegacyPropControllerTextDataTypes,
 ] as const
 
-const LegacyNumberDataTypes = ['number::v1'] as const
+export const LegacyNumberDataTypes = { number: 'number::v1' } as const
 
 const LegacyPropControllerNumberDataTypes = [
```
agurtovoy: Same as above, strike "legacy" from `LegacyPropControllerNumberDataTypes` and  `LegacyNumberDataTypes`

packages/controls/src/common/data-types.ts
```diff
@@ -1,11 +1,16 @@
...
   Text: 'text',
   Number: 'number',
 } as const
 
-const LegacyTextDataTypes = ['text-input::v1', 'text-area::v1'] as const
+export const LegacyTextDataTypes = {
+  textInput: 'text-input::v1',
+  textArea: 'text-area::v1',
+  code: 'code::v1',
+} as const
```
agurtovoy, line 16: I'd add a comment here along the lines of 

> TODO: Unify tag serialization for controls that share the same underlying data type. For example, all plain-text controls should write the same tag (e.g. `text`). This requires refactoring `ControlDefinition` so prop editing remains backward-compatible when the host is running an older runtime.

agurtovoy (dismissed):
See inline comments.

PR #1322, opened 2026-06-08
fix: receive 200 response from element tree API

arvinpoddar:
Updates the getComponentSnapshot to pass a QSP that forces the endpoint to return a 200 response when the element tree is not found.

Will make it easier for certain clients to cache responses.


arvinpoddar:
Closing now that the client has been refactored in https://github.com/makeswift/makeswift/pull/1334

PR #1323, opened 2026-06-10
fix(hono-react): move `renderHtml` to a separate entry point

agurtovoy:
Move `renderHtml` to a separate entry point so `hono-react/server` can be imported in environments where `react-dom/server` does not provide `renderToReadableStream` (e.g. Node.js). Breaking change.


fikrikarim (approved):

fikrikarim:
I also had to remove that export to make RSC work in Vite.

arvinpoddar (approved):

PR #1329, opened 2026-06-15
feat(controls): add Date control

drewpledger:
## ARR-575 — Date control (runtime)

https://github.com/user-attachments/assets/f8336644-f495-4c54-9637-f015dbc95b66

Utilizes the Calendar rendered in the scheduled publish dialog for a first-class control in `@makeswift/controls`.

- Adds `DateDefinition` (`makeswift::controls::date`) following the Slider control pattern, with a versioned ISO-string data schema and an optional `includeTime` config flag.
- Wires the new control into the definition / serialization / merge-translations visitors.
- Exports a stable `Date()` factory from `@makeswift/controls`.
- Adds unit tests and a demo component.

Companion cosmos (builder) PR: https://github.com/makeswift/cosmos/pull/2835

Linear: https://linear.app/commerce/issue/ARR-575/date-control


packages/controls/src/common/data-types.ts
```diff
@@ -37,3 +37,12 @@ export const AcceptedNumberDataTypes = [
   NumberDataTypes.number,
   ...PropControllerNumberDataTypes,
 ] as const
+
+export const DateDataTypes = { date: 'date::v1' } as const
+
+const PropControllerDateDataTypes = ['prop-controllers::date::v1'] as const
```
fikrikarim, line 43: Why do we need this?

fikrikarim: I totally forgot that we have Date PropController 🤦

fikrikarim (commented):
Do we need to add a changeset?

apps/nextjs-app-router/src/components/date-demo/date-demo.makeswift.ts
fikrikarim, line 1: Should we add a changeset?

apps/nextjs-app-router/src/components/date-demo/date-demo.tsx
arvinpoddar, line 1: Can we merge these demo components _after_ the runtime PR is merged and upgraded in the builder. I've noticed that when these components get merged in now, there's a period where pulling `main` on the runtime won't work with the builder until the builder is upgraded.

fikrikarim:
I ran an AI review locally and it caught one interesting problem:

```
  The gap is the shape of the string, which the PR leaves implicit. The schema is a bare z.string(), and the demo/defaults always write a full UTC datetime
  (2024-01-01T00:00:00.000Z) even for the date-only eventDate. For a date-only value that's a classic timezone footgun:

  new Date('2024-01-01T00:00:00.000Z').toLocaleDateString()
  // "12/31/2023" anywhere west of UTC

  So the "correct form" actually depends on includeTime:
  - includeTime: false → a date-only string YYYY-MM-DD (no time, no zone), interpreted in local time.
  - includeTime: true → full ISO 8601 with offset/Z.
 ```

I wonder if we can validate the string depending on the config?

PR #1337, opened 2026-06-18
feat(controls): lowercase kebab-case icon names for IconRadioGroup

drewpledger:
## ARR-556 — IconRadioGroup kebab-case icon names

Aligns `IconRadioGroup` icon names with the `registerComponent` component-icon convention (lowercase kebab-case).

- `IconRadioGroup.Icon.Sun` etc. now resolve to kebab-case values (`'sun'`, `'align-left'`), and bare strings work: `icon: 'align-left'`.
- Expanded the icon set to the full builder icon list.
- Clean break: dropped legacy `*16`/PascalCase acceptance (those only ever shipped in canary, never `latest`).
- Demo updated to bare kebab strings; tests, snapshots, and the builder descriptors fixture updated.

Verification: `@makeswift/controls` 1005 tests pass; `@makeswift/runtime` 799 pass; both typecheck + build.

Companion builder PR: https://github.com/makeswift/cosmos/pull/2845


packages/controls/src/controls/icon-radio-group/icons.ts
```diff
@@ -1,76 +1,187 @@
...
-  VerticalAlignEnd: 'VerticalAlignEnd',
-  VerticalAlignMiddle: 'VerticalAlignMiddle',
-  VerticalAlignSpaceBetween: 'VerticalAlignSpaceBetween',
-  VerticalAlignStart: 'VerticalAlignStart',
+  Activity: 'activity',
+  AlignCenter: 'align-center',
+  AlignLeft: 'align-left',
+  AlignRight: 'align-right',
+  Angellist: 'angellist',
+  Apple: 'apple',
```
agurtovoy: I wonder if we want to do something like `logo-apple`, `logo-angellist`, `logo-bc` etc. for all the company logos 

cc: @andrewreifman

andrewreifman: Makes sense to me. And for specifics, we should name the BC logo `logo-bigcommerce`. But yes, totally fine prepending `logo-` to all brand icons.

agurtovoy: @drewpledger We probably want to prefix the keys as well for consistency (e.g. `LogoApple`), which will also have a nice side effect of having all the logos grouped together in a single block.

PR #1339, opened 2026-06-19
fix: introspect component element trees with result locale

arvinpoddar:
Fixes a bug where element trees for builtin elements are introspected using the locale provided by the user, rather than the locale actually corresponding to the result. The mismatch can occur when the user requests a localized tree that does not exist, and the method falls back to getting the base locale tree. With the bug, the base locale tree will be introspected using the originally requested locale.

### DIFF
--- a/packages/runtime/src/client/index.ts
+++ b/packages/runtime/src/client/index.ts
@@ -817,7 +817,7 @@ export class MakeswiftClient extends MakeswiftRestAPIClient {
     }
 
     const document = Schema.componentDocument.parse(await response.json())
-    const cacheData = await this.introspect(document.data, siteVersion, locale ?? null)
+    const cacheData = await this.introspect(document.data, siteVersion, document.locale)
 
     return {
       document,
--- a/packages/runtime/src/client/tests/client.get-component-snapshot.test.ts
+++ b/packages/runtime/src/client/tests/client.get-component-snapshot.test.ts
@@ -4,7 +4,8 @@ import { http, HttpResponse, graphql } from 'msw'
 import { createReactRuntime } from '../../runtimes/react/testing/react-runtime'
 
 import { server } from '../../mocks/server'
-import { TestWorkingSiteVersion } from '../../testing/fixtures'
+import { makePagePathnameSlice, TestWorkingSiteVersion } from '../../testing/fixtures'
+import { Link } from '../../controls'
 
 const TEST_API_KEY = 'myApiKey'
 const runtime = createReactRuntime()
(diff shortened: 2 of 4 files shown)


packages/runtime/src/client/index.ts
```diff
@@ -817,7 +817,7 @@ export class MakeswiftClient extends MakeswiftRestAPIClient {
     }
 
     const document = Schema.componentDocument.parse(await response.json())
-    const cacheData = await this.introspect(document.data, siteVersion, locale ?? null)
+    const cacheData = await this.introspect(document.data, siteVersion, document.locale)
```
arvinpoddar, line 820: The actual fix: using the retrieved document locale to introspect, rather than the one provided by the user.

agurtovoy: Funny enough, we had to make a similar fix in `getPageSnapshot` a while ago: https://github.com/makeswift/makeswift/pull/950

packages/runtime/src/client/tests/client.get-component-snapshot.test.ts
```diff
@@ -272,4 +273,135 @@ describe('getComponentSnapshot using v2 element tree endpoint', () => {
...
+    function dummyPathnameSlice(pageId: string, locale: string | null) {
+      return {
+        id: pageId,
+        value: {
+          __typename: 'PagePathnameSlice',
+          id: pageId,
+          basePageId: pageId,
+          pathname: `/${pageId}`,
+          localizedPathname: null,
+        },
```
agurtovoy: I'd use the newly minted [`makePagePathnameSlice` fixture](https://github.com/makeswift/makeswift/blob/da62b4e5cc4080f78efdcfe67c52fc6e34117de6/packages/runtime/src/testing/fixtures/resources.ts#L77) here.

packages/runtime/src/client/tests/client.get-component-snapshot.test.ts
```diff
@@ -272,4 +273,135 @@ describe('getComponentSnapshot using v2 element tree endpoint', () => {
...
+
+    function makeSlicesHandler() {
+      return jest.fn(({ request }: { request: Request }) => {
+        const ids = new URL(request.url).searchParams.getAll('ids')
+        return HttpResponse.json(
+          ids.map(id => ({
+            __typename: 'PagePathnameSlice' as const,
+            id: `slice-${id}`,
+            basePageId: id,
+            pathname: `/${id}`,
```
agurtovoy: ... and here as well, but you'll have to add a `basePageId` option.

agurtovoy (dismissed):
🙏

packages/runtime/src/client/tests/client.get-component-snapshot.test.ts
arvinpoddar: Addressed!

PR #1341, opened 2026-06-19
feat(runtime): experimental flag to enforce 200 responses

arvinpoddar:
Introduces a new option for the `getComponentSnapshot` that causes the endpoint to return a 200 response, making the fetched results more easily cacheable by higher-level frameworks.

### DIFF
--- a/packages/runtime/src/client/index.ts
+++ b/packages/runtime/src/client/index.ts
@@ -760,63 +760,90 @@ export class MakeswiftClient extends MakeswiftRestAPIClient {
       siteVersion: siteVersionPromise,
       locale,
       allowLocaleFallback = true,
+      unstable_enforceSuccess = false,
     }: {
       siteVersion: SiteVersion | null | Promise<SiteVersion | null>
       locale?: string
       allowLocaleFallback?: boolean
+      unstable_enforceSuccess?: boolean
     },
   ): Promise<MakeswiftComponentSnapshot> {
-    const searchParams = new URLSearchParams()
-    if (locale) searchParams.set('locale', locale)
-
     const siteVersion = await siteVersionPromise
     const key = deterministicUUID({ id, locale, seed: this.apiKey.split('-').at(0) })
     const baseLocaleWasRequested = locale == null
     const canAttemptLocaleFallback = !baseLocaleWasRequested && allowLocaleFallback
 
-    let response
-    const responseForRequestedLocale = await this.fetch(
-      `v2/element-trees/${encodeURIComponent(id)}?${searchParams.toString()}`,
-      siteVersion,
-    )
-
-    if (responseForRequestedLocale.status === 404 && canAttemptLocaleFallback) {
-      await failedResponseBody(responseForRequestedLocale)
-      response = await this.fetch(`v2/element-trees/${encodeURIComponent(id)}`, siteVersion)
-    } else {
-      response = responseForRequestedLocale
-    }
-
-    if (!response.ok) {
-      // See comment on `failedResponseBody` for why we always consume the
-      // response body of failed responses.
-      const failedBody = await failedResponseBody(response)
-      if (response.status === 404) {
-        return {
-          document: {
-            id,
-            locale: locale ?? null,
-            data: null,
-          },
-          key,
-          cacheData: CacheData.empty(),
-          meta: {
-            allowLocaleFallback,
-            requestedLocale: locale ?? null,
-          },
+    const parseElementTreeResponse = async (
+      response: Response,
+    ): Promise<MakeswiftComponentDocument | null> => {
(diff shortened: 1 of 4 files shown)


packages/runtime/src/client/schema/component-document.ts
```diff
@@ -16,3 +16,12 @@ export const componentDocumentFallback = z.object({
   locale: z.string().nullable(),
   data: z.null(),
 })
+
+const notFoundComponentDocumentResponse = z.object({
```
arvinpoddar: New schema for the "empty" response from the endpoint

packages/runtime/src/client/index.ts
```diff
@@ -760,63 +760,90 @@ export class MakeswiftClient extends MakeswiftRestAPIClient {
...
     const key = deterministicUUID({ id, locale, seed: this.apiKey.split('-').at(0) })
     const baseLocaleWasRequested = locale == null
     const canAttemptLocaleFallback = !baseLocaleWasRequested && allowLocaleFallback
 
-    let response
-    const responseForRequestedLocale = await this.fetch(
-      `v2/element-trees/${encodeURIComponent(id)}?${searchParams.toString()}`,
-      siteVersion,
-    )
+    const parseElementTreeResponse = async (
```
arvinpoddar, line 776: I refactored this method to (hopefully) be a bit clearer to read by creating some internal functions.

`parseElementTreeResponse` abstracts handling the result from the API call. As part of the new flag being introduced, the only semantic change here is the schema parsing accounts for the `{ "notFound": true }` response.

Then, we have two internal functions to retrieve the "primary" and "fallback" trees. The latter is only called if the result from the first is empty (and if the appropriate options have been fetched by the user to fallback).

packages/runtime/src/client/index.ts
```diff
@@ -760,63 +760,90 @@ export class MakeswiftClient extends MakeswiftRestAPIClient {
...
+      if (unstable_enforceSuccess) searchParams.set('unstable_enforceSuccess', 'true')
+      return await getElementTree(searchParams)
+    }
 
-      throw new Error(`Failed to get component snapshot for '${id}': ${responseError(response)}`)
+    const primaryTree = await getPrimaryTree()
+
+    const result =
+      primaryTree == null && canAttemptLocaleFallback ? await getFallbackTree() : primaryTree
+
```
arvinpoddar, line 829: Now, the behavior of this method is condensed down to this line. Hopefully this makes it easier to understand.

arvinpoddar:
cursor review

arvinpoddar:
ez no-scope

arvinpoddar:
cursor review

packages/runtime/src/client/index.ts
```diff
@@ -760,63 +760,90 @@ export class MakeswiftClient extends MakeswiftRestAPIClient {
       siteVersion: siteVersionPromise,
       locale,
       allowLocaleFallback = true,
+      unstable_enforceSuccess = false,
```
agurtovoy, line 763: I know we're using the same name here as the endpoint's query param, and it's unstable, so we can merge as-is, but IMHO `enforceSuccess` is both too vague (what's this about?) and too broad (always succeed?). A better name for the current endpoint semantic would be something like `successIfNotFound`, but also see [below](https://github.com/makeswift/makeswift/pull/1341/changes#r3445275577).

packages/runtime/src/client/index.ts
agurtovoy: Thank you for refactoring this, it's a lot clearer!

packages/runtime/src/client/index.ts
```diff
@@ -760,63 +760,90 @@ export class MakeswiftClient extends MakeswiftRestAPIClient {
...
+      return await getElementTree(searchParams)
+    }
 
-      throw new Error(`Failed to get component snapshot for '${id}': ${responseError(response)}`)
+    const primaryTree = await getPrimaryTree()
+
+    const result =
+      primaryTree == null && canAttemptLocaleFallback ? await getFallbackTree() : primaryTree
+
+    if (result == null) {
```
agurtovoy, line 830: Ideally, I think we should do two things here:

- Introduce a `v3` version of the `element-trees` endpoint that natively handles `allowLocaleFallback` query param the same way it's handled in the pages documents endpoint ( =v3+). This would significantly cut down on the amount of code here.
- When requested "success on not found", instead of returning `null`, the new endpoint should simply return a component document fallback that we're currently constructing manually; this would eliminate the one-off `notFound: true` result and unify the return paths here. The query param requesting this behavior could then be called `fallbackIfNotFound` or `blankIfNotFound`.

agurtovoy (dismissed):

PR #1343, opened 2026-06-24
refactor(runtime): factor out `useControlDefs` hook

agurtovoy:
Factor out `useControlDefs` hook, move `ResolveProps` component to `components/resolve-props.tsx`. No semantic changes.


packages/runtime/src/runtimes/react/components/resolve-props.tsx
```diff
@@ -0,0 +1,29 @@
+import { type ReactNode } from 'react'
+
+import { type ElementData } from '../../../state/read-only-state'
+
+import { useControlDefs } from '../hooks/use-control-defs'
+import { useResolvedProps } from '../hooks/use-resolved-props'
+
+import { resolveLegacyDescriptorProp } from '../legacy-controls'
+
+export function ResolveProps({
```
agurtovoy, line 10: Moved from `controls.tsx`; no semantic changes.

packages/runtime/src/runtimes/react/hooks/use-control-defs.ts
```diff
@@ -0,0 +1,20 @@
...
+import { getComponentPropControllerDescriptors } from '../../../state/read-only-state'
+import { type LegacyDescriptor, isLegacyDescriptor } from '../../../prop-controllers/descriptors'
+
+import { useStore } from './use-store'
+
+export function useControlDefs({
+  elementType,
+}: {
+  elementType: string
+}): readonly [Record<string, LegacyDescriptor>, Record<string, ControlDefinition>] {
```
agurtovoy, line 16: Extracted from `controls.tsx`; no semantic changes.

arvinpoddar (dismissed):
Do these need to be `.tsx` files? I'm not seeing any jsx syntax

agurtovoy:
@arvinpoddar Good catch! `ResolveProps` is a React component, so I'm going to leave it as `.tsx` in case it evolves in the future, but the hook definitely doesn't have any reason to be a `.tsx`.

arvinpoddar (approved):

PR #1328, opened 2026-06-14
refactor(runtime): move GraphQL ops into a separate client

agurtovoy:
Decouple `MakeswiftHostApiClient`'s API resource caching/fetching from a GraphQL mutation used in form submissions. 

Form submissions smoke test:

https://github.com/user-attachments/assets/a86ab6c3-6fba-4028-b416-522caec61809

PR #1344, opened 2026-06-25
feat(runtime): use `MakeswiftApiResourcesClient` on the server if runtime instance was provided with an API key

agurtovoy:
Non-breaking. No behavior change unless user provided an API key when constructing the runtime.

### DIFF
--- a/packages/runtime/src/api/makeswift-api-resources-client.ts
+++ b/packages/runtime/src/api/makeswift-api-resources-client.ts
@@ -13,9 +13,11 @@ import {
 } from './types'
 
 import { ApiResourcesClient } from './api-resources-client'
-import { MakeswiftGraphQLApiClient } from './graphql-api-client'
-import { MakeswiftRestAPIClient } from './rest-api-client'
 import { type SiteVersion } from './site-version'
+
+// see the comment in `MakeswiftApiResourcesClient` constructor below
+import { type MakeswiftGraphQLApiClient } from './graphql-api-client'
+import { type MakeswiftRestAPIClient } from './rest-api-client'
 
 /**
  * Server-side implementation that makes direct authenticated requests to the Makeswift API.
@@ -25,8 +27,8 @@ import { type SiteVersion } from './site-version'
  * `/api/makeswift/...` endpoints.
  */
 export class MakeswiftApiResourcesClient extends ApiResourcesClient {
-  private restApiClient: MakeswiftRestAPIClient
-  private graphQlClient: MakeswiftGraphQLApiClient
+  private restApiClient: Promise<MakeswiftRestAPIClient>
+  private graphQlClient: Promise<MakeswiftGraphQLApiClient>
 
   constructor({
     fetch,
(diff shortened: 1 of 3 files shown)


packages/runtime/src/api/makeswift-api-resources-client.ts
```diff
@@ -13,10 +13,12 @@ import {
 } from './types'
 
 import { ApiResourcesClient } from './api-resources-client'
-import { MakeswiftGraphQLApiClient } from './graphql-api-client'
-import { MakeswiftRestAPIClient } from './rest-api-client'
 import { type SiteVersion } from './site-version'
 
+// see the comment in `MakeswiftApiResourcesClient` constructor below
+import { type MakeswiftGraphQLApiClient } from './graphql-api-client'
+import { type MakeswiftRestAPIClient } from './rest-api-client'
```
agurtovoy, line 20: Updated `MakeswiftApiResourcesClient` implementation to not import server-only dependencies unless we're on the server.

packages/runtime/src/runtimes/react/runtime-core.ts
```diff
@@ -43,6 +44,8 @@ export class RuntimeCore {
     ttlCheck: RefCountedMap.TTLCheck.ON_RETAIN | RefCountedMap.TTLCheck.ON_RELEASE,
   })
 
+  private readonly apiKey: string | undefined
+
```
agurtovoy, line 49: When running on the server, passing the API key is what switches the runtime from using `HostApiResourcesClient` to `MakeswiftApiResourcesClient`.

packages/runtime/src/runtimes/react/runtime-core.ts
```diff
@@ -149,6 +155,16 @@ export class RuntimeCore {
     siteVersion: SiteVersion | null
     locale: string | undefined
   }) {
```
arvinpoddar: Maybe we specify the return type as `ApiResourcesClient` now that there's a clear interface?

packages/runtime/src/runtimes/react/runtime-core.ts
```diff
@@ -52,18 +55,21 @@ export class RuntimeCore {
...
     appOrigin?: string
     apiOrigin?: string
+    apiKey?: string
     breakpoints?: BreakpointsInput
     requestKey?: StoreKey
     fetch: HttpFetch
   }) {
     this.appOrigin = validateOrigin(appOrigin, 'appOrigin')
     this.apiOrigin = validateOrigin(apiOrigin, 'apiOrigin')
+    this.apiKey = apiKey
```
arvinpoddar, line 73: This feels like a return of the sprawl of all the various values you have to pass around in order to set up Makeswift. Not a big deal, just something that crossed my mind.

Since both the API Handler and client take in a runtime instance, maybe there's a way (in the future) to avoid having to pass the API key during their construction, so now all configuration comes from the runtime?

packages/runtime/src/api/makeswift-api-resources-client.ts
```diff
@@ -45,51 +47,65 @@ export class MakeswiftApiResourcesClient extends ApiResourcesClient {
...
+    // `HostApiResourcesClient`, the overhead of this module on its own is negligible)
+    this.restApiClient = (async () => {
+      const { MakeswiftRestAPIClient } = await import('./rest-api-client')
+      return new MakeswiftRestAPIClient({ fetch, apiKey, apiOrigin })
+    })()
+
+    this.graphQlClient = (async () => {
+      const { MakeswiftGraphQLApiClient } = await import('./graphql-api-client')
+      return new MakeswiftGraphQLApiClient({ endpoint: graphqlApiEndpoint })
+    })()
```
arvinpoddar, line 65: Is it worth lazy loading this since we need the GQL client on the server and client? 

Or is the intent of this optimization specifically to be even more conservative and only pull in this client when it's actually needed (i.e., when a file/table need to be retrieved)? I'd imagine this is effectively every time.

arvinpoddar (dismissed):

packages/runtime/src/api/makeswift-api-resources-client.ts
agurtovoy: Since [this refactoring](https://github.com/makeswift/makeswift/pull/1328), the GraphQL client is only imported in the browser if you're using the `Form` component (and we should really replace that with a call to the host API). The file/table retrieval goes through `HostApiResourcesClient`, which simply fetches from the host API endpoints.

packages/runtime/src/runtimes/react/runtime-core.ts
agurtovoy: Yes, if we formalize this pattern (constructing server instance of the runtime with an API key) for all configurations, we can definitely update the client and the API handler constructors to not require separate passing of the API key.

packages/runtime/src/runtimes/react/runtime-core.ts
agurtovoy: Good idea, done.

PR #1345, opened 2026-06-25
feat(runtime): stackable `FrameworkContextProvider`

agurtovoy:
Stackable `FrameworkContextProvider` with support for partial overrides.

As the framework context expands to cover more orthogonal concerns, we need to support selective, partial overrides without replacing unrelated defaults.

### DIFF
--- a/packages/react-router/src/components/framework-provider/index.tsx
+++ b/packages/react-router/src/components/framework-provider/index.tsx
@@ -1,21 +1,11 @@
-import { type PropsWithChildren, useState } from 'react'
+import { type PropsWithChildren } from 'react'
 
-import {
-  FrameworkContext,
-  DefaultHead,
-  DefaultHeadSnippet,
-  DefaultImage,
-} from '@makeswift/runtime/unstable-framework-support'
+import { FrameworkContextProvider } from '@makeswift/runtime/unstable-framework-support'
 
 import { Link } from './link'
 
-export function FrameworkProvider({ children }: PropsWithChildren) {
-  const [context] = useState<FrameworkContext>({
-    Head: DefaultHead,
-    HeadSnippet: DefaultHeadSnippet,
-    Image: DefaultImage,
-    Link,
-  })
+const context = { Link }
 
-  return <FrameworkContext.Provider value={context}>{children}</FrameworkContext.Provider>
+export function FrameworkProvider({ children }: PropsWithChildren) {
+  return <FrameworkContextProvider value={context}>{children}</FrameworkContextProvider>
 }
--- a/packages/runtime/src/next/components/framework-provider/app-router/index.ts
+++ b/packages/runtime/src/next/components/framework-provider/app-router/index.ts
@@ -1,13 +1,9 @@
-import {
-  type FrameworkContext,
-  DefaultHead,
-} from '../../../../runtimes/react/components/framework-context'
+import { type FrameworkContext } from '../../../../runtimes/react/components/framework-context'
 
 import { HeadSnippet } from './HeadSnippet'
 
-export const context: Pick<FrameworkContext, 'Head' | 'HeadSnippet'> = {
+export const context: Pick<FrameworkContext, 'HeadSnippet'> = {
   // The App Router uses built-in React Canary releases, which include all stable
   // React 19 features, so we can simply use our default head implementation
-  Head: DefaultHead,
   HeadSnippet,
 }
--- a/packages/runtime/src/next/components/tests/makeswift-page-snippets-rendering.test.tsx
+++ b/packages/runtime/src/next/components/tests/makeswift-page-snippets-rendering.test.tsx
@@ -1,9 +1,10 @@
 /** @jest-environment jsdom */
 
 import '@testing-library/jest-dom'
-import { type Snippet } from '../../../client'
 
 import { createMakeswiftPageSnapshot, testMakeswiftPageHeadRendering } from '../../testing'
+
+import { createSnippet } from '../../../testing/fixtures'
 
 import { pageDocument } from './__fixtures__/page-document'
 
@@ -20,15 +21,6 @@ jest.mock('next/script', () => ({
     </script>
   )),
 }))
-
-const createSnippet = ({ id, code }: { id: string; code: string }): Snippet => ({
-  id,
-  code,
-  location: 'HEAD',
-  liveEnabled: true,
-  builderEnabled: true,
-  cleanup: null,
-})
 
 const getPageSnippetTags = (container: HTMLElement) =>
   Array.from(container.querySelectorAll('*[data-makeswift-snippet-id]'))
--- a/packages/runtime/src/runtimes/react/components/__tests__/__snapshots__/framework-context.test.tsx.snap
+++ b/packages/runtime/src/runtimes/react/components/__tests__/__snapshots__/framework-context.test.tsx.snap
@@ -1,0 +1,96 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Provider values override default context 1`] = `
+<div
+  data-testid="test-id"
+>
+  <img
+    alt="example image"
+    loading="lazy"
+    src="https://example.com/image.jpg"
+  />
+  <a
+    data-test-id="custom link"
+    href="https://example.com"
+  >
+    Click here
+  </a>
+</div>
+`;
+
+exports[`Provider values override default context: head 1`] = `
+<head>
+  <title>
+    Document title
+  </title>
+  <script
+    data-makeswift-snippet-id="test-snippet"
+  >
+    console.log("hello")
+  </script>
+</head>
+`;
+
+exports[`Providers are stackable, support partial overrides 1`] = `
+<div
+  data-testid="test-id"
+>
+  <img
+    alt="example image"
+    data-test-id="custom image 2"
+    src="https://example.com/image.jpg"
+  />
+  <a
+    data-test-id="custom link"
+    href="https://example.com"
+  >
+    Click here
+  </a>
+</div>
+`;
+
+exports[`Providers are stackable, support partial overrides: head 1`] = `
+<head>
+  <meta
+    name="extra"
+  />
+  <title>
+    Document title
+  </title>
+  <script
+    data-makeswift-snippet-id="test-snippet"
+  >
+    console.log("hello")
+  </script>
+</head>
+`;
+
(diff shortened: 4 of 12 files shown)


packages/react-router/src/components/framework-provider/index.tsx
```diff
@@ -1,21 +1,11 @@
-import { type PropsWithChildren, useState } from 'react'
+import { type PropsWithChildren, useMemo } from 'react'
 
-import {
-  FrameworkContext,
-  DefaultHead,
-  DefaultHeadSnippet,
-  DefaultImage,
-} from '@makeswift/runtime/unstable-framework-support'
+import { FrameworkContextProvider } from '@makeswift/runtime/unstable-framework-support'
```
agurtovoy, line 3: The new API allows us to provide only properties that are being overwritten, which means we don't need to import the defaults (and generally simplifies user code).

packages/react-router/src/components/framework-provider/index.tsx
```diff
@@ -1,21 +1,11 @@
...
 import { Link } from './link'
 
 export function FrameworkProvider({ children }: PropsWithChildren) {
-  const [context] = useState<FrameworkContext>({
-    Head: DefaultHead,
-    HeadSnippet: DefaultHeadSnippet,
-    Image: DefaultImage,
-    Link,
-  })
+  const context = useMemo(() => ({ Link }), [])
```
agurtovoy: See above; also switched from `useState` to `useMemo` as being more idiomatic here.

packages/runtime/src/next/components/framework-provider/app-router/index.ts
```diff
@@ -1,13 +1,9 @@
...
-} from '../../../../runtimes/react/components/framework-context'
+import { type FrameworkContext } from '../../../../runtimes/react/components/framework-context'
 
 import { HeadSnippet } from './HeadSnippet'
 
-export const context: Pick<FrameworkContext, 'Head' | 'HeadSnippet'> = {
+export const context: Pick<FrameworkContext, 'HeadSnippet'> = {
   // The App Router uses built-in React Canary releases, which include all stable
   // React 19 features, so we can simply use our default head implementation
-  Head: DefaultHead,
```
agurtovoy, line 11: No need to explicitly fill in the default here.

packages/runtime/src/next/components/tests/makeswift-page-snippets-rendering.test.tsx
```diff
@@ -21,15 +22,6 @@ jest.mock('next/script', () => ({
...
 }))
 
-const createSnippet = ({ id, code }: { id: string; code: string }): Snippet => ({
-  id,
-  code,
-  location: 'HEAD',
-  liveEnabled: true,
-  builderEnabled: true,
-  cleanup: null,
-})
```
agurtovoy, line 31: Factor out existing snippet fixture to `testing/fixtures`

packages/runtime/src/runtimes/react/components/__tests__/__snapshots__/framework-context.test.tsx.snap
```diff
@@ -0,0 +1,96 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Provider values override default context 1`] = `
```
agurtovoy, line 3: Snapshots for the newly added tests, see below.

packages/runtime/src/runtimes/react/components/__tests__/framework-context.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+          code: '<script>console.log("hello")</script>',
+        })}
+      />
+      <Image src="https://example.com/image.jpg" alt="example image" />
+      <Link href="https://example.com">Click here</Link>
+    </div>
+  )
+}
+
+test('useFrameworkContext without a provider returns default context', async () => {
```
agurtovoy, line 56: Tests for the new API.

packages/runtime/src/runtimes/react/components/hooks/use-framework-context.tsx
```diff
@@ -1,7 +1 @@
-import { useContext } from 'react'
-
-import { FrameworkContext } from '../framework-context'
-
-export function useFrameworkContext() {
-  return useContext(FrameworkContext)
-}
```
agurtovoy, line 7: Moved to the main file to avoid exporting too many implementation details.

packages/runtime/src/runtimes/react/components/framework-context.tsx
```diff
@@ -44,11 +46,11 @@ export type FrameworkContext = {
 }
 
 // React 19 automatically hoists metadata tags to the <head>
-export const DefaultHead = ({ children }: PropsWithChildren) => <>{children}</>
```
agurtovoy, line 47: Remove `export`s from default components as they shouldn't be used directly anymore.

packages/runtime/src/runtimes/react/components/framework-context.tsx
```diff
@@ -65,13 +67,29 @@ export const DefaultImage: ImageComponent = ({ priority, fill, style, ...props }
   />
 )
 
-export const DefaultLink: LinkComponent = forwardRef<HTMLAnchorElement, LinkProps>(
+const DefaultLink: LinkComponent = forwardRef<HTMLAnchorElement, LinkProps>(
   ({ linkType, ...props }, ref) => <a {...props} ref={ref} />,
 )
 
-export const FrameworkContext = createContext<FrameworkContext>({
+const FrameworkContext = createContext<Partial<FrameworkContext>>({})
```
agurtovoy, line 74: Empty root context.

packages/runtime/src/runtimes/react/components/framework-context.tsx
```diff
@@ -65,13 +67,29 @@ export const DefaultImage: ImageComponent = ({ priority, fill, style, ...props }
...
+const defaultContext: FrameworkContext = {
   Head: DefaultHead,
   HeadSnippet: DefaultHeadSnippet,
   Image: DefaultImage,
   Link: DefaultLink,
-})
+}
+
+export function useFrameworkContext(): FrameworkContext {
+  return { ...defaultContext, ...useContext(FrameworkContext) }
```
agurtovoy, line 94: `useFrameworkContext` merges default props with the layered value from the parent context.

packages/runtime/src/testing/fixtures/snippet.ts
```diff
@@ -0,0 +1,10 @@
+import { type Snippet } from '../../client/page-snapshot'
+
+export const createSnippet = ({ id, code }: { id: string; code: string }): Snippet => ({
+  id,
+  code,
+  location: 'HEAD',
+  liveEnabled: true,
+  builderEnabled: true,
+  cleanup: null,
+})
```
agurtovoy, line 10: Test fixture extracted from https://github.com/makeswift/makeswift/pull/1345/changes#r3478257973

packages/runtime/src/unstable-framework-support/index.ts
```diff
@@ -20,10 +20,8 @@ export { MakeswiftClient } from '../client'
 export { type BreakpointsInput as Breakpoints } from '../state/modules/breakpoints'
 
 export {
-  FrameworkContext,
-  DefaultHead,
-  DefaultHeadSnippet,
-  DefaultImage,
+  type FrameworkContext,
+  FrameworkContextProvider,
 } from '../runtimes/react/components/framework-context'
```
agurtovoy, line 25: Clean up `unstable-framework-support` exports.

packages/runtime/src/runtimes/react/components/framework-context.tsx
```diff
@@ -65,13 +67,29 @@ export const DefaultImage: ImageComponent = ({ priority, fill, style, ...props }
...
 
-export const FrameworkContext = createContext<FrameworkContext>({
+const FrameworkContext = createContext<Partial<FrameworkContext>>({})
+
+export const FrameworkContextProvider = ({
+  value,
+  children,
+}: PropsWithChildren<{ value: Partial<FrameworkContext> }>) => {
+  const parentContext = useContext(FrameworkContext)
+  const mergedContext = useMemo(() => ({ ...parentContext, ...value }), [parentContext])
```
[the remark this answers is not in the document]
agurtovoy: Fixed

packages/runtime/src/runtimes/react/components/framework-context.tsx
```diff
@@ -65,13 +67,29 @@ export const DefaultImage: ImageComponent = ({ priority, fill, style, ...props }
...
 
-export const FrameworkContext = createContext<FrameworkContext>({
+const FrameworkContext = createContext<Partial<FrameworkContext>>({})
+
+export const FrameworkContextProvider = ({
+  value,
+  children,
+}: PropsWithChildren<{ value: Partial<FrameworkContext> }>) => {
+  const parentContext = useContext(FrameworkContext)
+  const mergedContext = useMemo(() => ({ ...parentContext, ...value }), [parentContext, value])
```
agurtovoy, line 81: Each framework provider merges its props with the parent props (if any).

Prior art: [Emotion's `ThemeProvider`](https://github.com/emotion-js/emotion/blob/b882bcba85132554992e4bd49e94c95939bbf810/packages/react/src/theming.tsx#L71)

packages/react-router/src/components/framework-provider/index.tsx
arvinpoddar: Two things: 
1. I recall Miguel remarking how `useMemo` doesn't actually guarantee a stable identity
2. Maybe a silly question: could we just construct `{ Link }` outside of React and pass it as the value to the provider so we avoid the stability problem?

arvinpoddar (dismissed):
Very nice!

packages/react-router/src/components/framework-provider/index.tsx
agurtovoy: Good reminder on the  `useMemo`; while it's true that React can, [under certain circumstances](https://react.dev/reference/react/useMemo#caveats), throw away the cached value, that doesn't affect **correctness** of this code, only its performance -- but you could argue that since `useState` or `useRef` don't have this caveat, they are a better fit here despite the naming mismatch. 

That said, since `Link` is completely static, moving the context construction to the global scope is completely appropriate here -- thanks for the idea!

PR #1347, opened 2026-06-30
feat: add `instanceKey` (`elementKey` + `propPath`) to control instances

agurtovoy:
This allows us to properly identify & look up control instances at runtime.

### DIFF
--- a/packages/controls/src/controls/checkbox/checkbox.ts
+++ b/packages/controls/src/controls/checkbox/checkbox.ts
@@ -12,7 +12,7 @@ import {
   type Resolvable,
   type SchemaType,
 } from '../definition'
-import { DefaultControlInstance, type SendMessage } from '../instance'
+import { DefaultControlInstance, type ControlInstanceArgs } from '../instance'
 import { ControlDefinitionVisitor } from '../visitor'
 
 type Config = z.infer<typeof Definition.schema.relaxed.config>
@@ -166,8 +166,8 @@ class Definition<C extends Config> extends ControlDefinition<
     }
   }
 
-  createInstance(sendMessage: SendMessage<any>) {
-    return new DefaultControlInstance(sendMessage)
+  createInstance(args: ControlInstanceArgs) {
+    return new DefaultControlInstance(args)
   }
 
   accept<R>(visitor: ControlDefinitionVisitor<R>, ...args: unknown[]): R {
--- a/packages/controls/src/controls/group/group-control.ts
+++ b/packages/controls/src/controls/group/group-control.ts
@@ -1,8 +1,8 @@
 import { ControlDefinition } from '../definition'
 import {
   ControlInstance,
+  type ControlInstanceArgs,
   type ControlMessage,
-  type SendMessage,
 } from '../instance'
 
 import { type GroupDefinition } from './group'
@@ -22,9 +22,9 @@ export class GroupControl<
 
   constructor(
     private readonly definition: Def,
-    sendMessage: SendMessage<Message>,
+    args: ControlInstanceArgs,
   ) {
-    super(sendMessage)
+    super(args)
     this.childControls = new Map(
       Object.entries(this.definition.propDefs).map(([key, def]) => [
         key,
@@ -44,11 +44,17 @@ export class GroupControl<
   child = (key: string) => this.childControls.get(key)
 
   createChildControl = (def: ControlDefinition, key: string) => {
-    return def.createInstance((message) =>
-      this.sendMessage({
-        type: GroupControl.CHILD_CONTROL_MESSAGE,
-        payload: { message, key },
-      }),
-    )
+    const { elementKey, propPath } = this.instanceKey
+    return def.createInstance({
+      instanceKey: {
+        elementKey,
+        propPath: `${propPath}.${key}`,
+      },
+      sendMessage: (message) =>
+        this.sendMessage({
+          type: GroupControl.CHILD_CONTROL_MESSAGE,
+          payload: { message, key },
+        }),
+    })
   }
 }
--- a/packages/controls/src/controls/group/group.test.ts
+++ b/packages/controls/src/controls/group/group.test.ts
@@ -177,6 +177,45 @@ describe('Group', () => {
     })
   })
 
+  describe('prop resolution', () => {
+    const group = Group({
+      props: {
+        color: Color({ defaultValue: 'red' }),
+        checkbox: Checkbox(),
+        text: TextInput(),
+      },
+    })
+
+    const groupKeys = Object.keys(group.propDefs)
+    const instanceKey = { elementKey: '[test-element]', propPath: 'group-prop' }
+
+    const fixtures = () => {
+      const groupInstance = group.createInstance({
+        instanceKey,
+        sendMessage: () => {},
+      })
+
+      return { groupInstance }
+    }
+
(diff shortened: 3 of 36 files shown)


packages/controls/src/controls/checkbox/checkbox.ts
```diff
@@ -166,8 +166,8 @@ class Definition<C extends Config> extends ControlDefinition<
     }
   }
 
-  createInstance(sendMessage: SendMessage<any>) {
-    return new DefaultControlInstance(sendMessage)
+  createInstance(args: ControlInstanceArgs) {
```
agurtovoy, line 169: Here and below, switched from passing individual args required for constructing a control instance to passing an args object. This drastically reduces the number of files one needs to touch when updating/iterating on instance construction.

packages/controls/src/controls/group/group-control.ts
```diff
@@ -44,11 +44,17 @@ export class GroupControl<
...
-        type: GroupControl.CHILD_CONTROL_MESSAGE,
-        payload: { message, key },
-      }),
-    )
+    const { elementKey, propPath } = this.instanceKey
+    return def.createInstance({
+      instanceKey: {
+        elementKey,
+        propPath: `${propPath}.${key}`,
+      },
```
agurtovoy, line 52: Constructing a control instance now requires an instance key. `propPath` of a nested prop's instance is a composite path of a parent prop name and the child's own prop name.

packages/controls/src/controls/group/group.test.ts
```diff
@@ -177,6 +177,45 @@ describe('Group', () => {
     })
   })
 
+  describe('prop resolution', () => {
```
agurtovoy, line 180: Added a test to verify the new behavior.

packages/controls/src/controls/list/list-control.ts
```diff
@@ -38,26 +38,36 @@ export class ListControl<
 
   child = (key: string) => this.itemControls.get(key)
 
-  childControls = (ids: string[] | undefined) => {
-    if (ids == null || hasAllKeys(this.itemControls, ids)) {
```
agurtovoy, line 42: Previously, we did not recreate child controls for list items when user reordered a list. Now that each item control has an instance key, its identity is tied to a specific prop path, so the control must be recreated whenever that identity changes, including on reorder.

packages/controls/src/controls/list/list-control.ts
```diff
@@ -38,26 +38,36 @@ export class ListControl<
...
+  childControls = (ids: string[] | undefined): Map<string, ControlInstance> => {
+    const orderedIds: string[] = [...this.itemControls.keys()]
```
agurtovoy, line 42: `Map` (`itemControls`) [guarantees to preserve its insertion order](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map).

packages/controls/src/controls/list/list-control.ts
```diff
@@ -38,26 +38,36 @@ export class ListControl<
...
-        payload: { message, itemId: id },
-      }),
-    )
+  createItemControl = (index: number, itemId: string): ControlInstance => {
+    const { elementKey, propPath } = this.instanceKey
+    return this.definition.itemDef.createInstance({
+      instanceKey: {
+        elementKey,
+        propPath: `${propPath}.${index}`,
+      },
```
agurtovoy, line 68: Prop path of a list item prop is `parentProp.<index>`, e.g. `parentProp.0`.

packages/controls/src/controls/list/list.test.ts
```diff
@@ -126,6 +128,96 @@ describe('List', () => {
     })
   })
 
+  describe('prop resolution', () => {
```
agurtovoy, line 131: Added tests to verify the new behavior.

packages/controls/src/controls/shape/v1/shape-control.ts
```diff
@@ -44,11 +44,17 @@ export class ShapeControl<
...
-        type: ShapeControl.CHILD_CONTROL_MESSAGE,
-        payload: { message, key },
-      }),
-    )
+    const { elementKey, propPath } = this.instanceKey
+    return def.createInstance({
+      instanceKey: {
+        elementKey,
+        propPath: `${propPath}.${key}`,
+      },
```
agurtovoy, line 52: Same as `Group`.

packages/controls/src/controls/style/v2/style-v2-control.ts
```diff
@@ -22,14 +26,16 @@ export class StyleV2Control extends ControlInstance<Message> {
...
+  constructor(propDef: ControlDefinition, args: ControlInstanceArgs<Message>) {
+    super(args)
+    this.control = propDef.createInstance({
+      ...args,
+      sendMessage: (message) =>
+        this.sendMessage({
+          type: StyleV2Control.CHILD_CONTROL_MESSAGE,
+          payload: { message },
+        }),
+    })
```
agurtovoy, line 38: Pass through instance args, no semantic changes.

packages/runtime/src/state/middleware/read-write/prop-controller-handles.ts
```diff
@@ -34,11 +34,14 @@ function createAndRegisterPropControllers(
...
     const propControllers = Object.entries(descriptors).reduce(
       (acc, [propName, descriptor]) => {
-        const propController = createPropController(descriptor, message =>
-          dispatch(
-            Builder.messageBuilderPropController(documentKey, elementKey, propName, message),
-          ),
-        ) as ControlInstance
+        const propController = createPropController({
+          descriptor,
+          instanceKey: { elementKey, propPath: propName },
```
agurtovoy, line 39: `propPath` of a top-level prop is simply its prop name.

packages/controls/src/controls/list/list-control.ts
```diff
@@ -38,26 +38,36 @@ export class ListControl<
...
+    // return existing controls if we have a full set of them and they are
+    // already correctly ordered
+    if (ids == null || shallowEqual(ids, orderedIds)) {
       return this.itemControls
     }
 
+    // otherwise reorder
     return new Map(
-      ids.map((id) => [id, this.child(id) ?? this.createItemControl(id)]),
+      ids.map((id, index) => [id, this.createItemControl(index, id)]),
```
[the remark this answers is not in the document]
agurtovoy: Nothing is "dropped", but we could do better. Fixed.

packages/controls/src/controls/group/group-control.ts
Codeseph: What I ended up doing for the css runtime turned out to be similar, just done through a `Stylesheet` hierarchy so that I'd have the keys available server-side 

Image

bc-drew-cutchins (approved):
This all looks good to me 👍

Codeseph (approved):

packages/controls/src/controls/group/group-control.ts
agurtovoy: It's a good question whether we should store prop path as a string vs a (non-empty) array. I went for a string primarily out of convenience since I'm using it as a key and comparing one `propPath` to another in a couple of places, but that also means I have to parse it in a couple of other places. I'm going to keep it as a string for now, but it might be worth taking a second look once we have all of the use cases in.

arvinpoddar (approved):
Looking good!

PR #1350, opened 2026-07-02
fix: warn when resolving image without dimensions

arvinpoddar:
Show a warning in the Image control when the following are both true:

1. The control format is `Image.Format.WithDimensions`
2. The image being resolved does not have both dimensions defined.


agurtovoy (approved):
🙏

PR #1351, opened 2026-07-03
feat(controls): unstable_Cascade control + PokéAPI demo (ARR-769)

drewpledger:
First spin at implementing the Cascade control. Details can be found in this [linear doc](https://linear.app/commerce/document/cascade-control-d7b7eaa3b4e1).


arvinpoddar:
Ended up going with a different approach to solve our use case.

PR #1352, opened 2026-07-10
fix(runtime): Fix accidental deletion of elements from element tree state when deleting props

bc-drew-cutchins:
Currently, the reducer managing the state of the element tree deletes elements whenever they have a prop deleted. This PR fixes the reducer to handle the following two cases:
1. A prop is deleted on an element (previously causing the element to be deleted)
2. A prop referencing children is deleted on an element (previously causing the element to be deleted alongside its children)

This first commit creates failing unit tests for both of these cases, the following commit fixes the tested behavior. 

I recommend reviewing this PR by stepping through the individual commits to view the fix in isolation from the test updates.


packages/runtime/src/state/__tests__/__snapshots__/react-builder-preview.test.ts.snap
```diff
@@ -2,8 +2,8 @@
 
 exports[`elementTreeMiddleware correctly tracks document changes: final element ids 1`] = `
 Map {
-  "c9e7d4ba-9511-4d70-bda1-9e5804fe6f27" => "cta",
```
bc-drew-cutchins, line 5: These snapshots are equivalent, only the ordering of the map elements has changed.

packages/runtime/src/state/modules/element-trees.ts
```diff
@@ -272,13 +281,18 @@ function applyDelete(
   rootElements: { before: Element; after: Element },
   path: OperationPath,
 ): ElementTree {
-  const [deleteElementPath, ...parentElementPaths] = getChangedElementsPaths(path)
-  const [deletedElement, propName] = getElementAndPropName(rootElements.before, deleteElementPath)
```
bc-drew-cutchins, line 276: `deletedElement` is misleading since the element is only deleted if `propName` is `null`.

packages/runtime/src/state/modules/element-trees.ts
```diff
@@ -272,13 +281,18 @@ function applyDelete(
...
+  const [targetElementPath, ...parentElementPaths] = getChangedElementsPaths(path)
+  const [targetElement, propName] = getElementAndPropName(rootElements.before, targetElementPath)
 
   const elements = new Map(elementTree.elements)
   const elementIds = new Map(elementTree.elementIds)
 
-  deleteElement({ elements, elementIds }, deletedElement, propName, descriptors)
+  if(propName == null) {
```
bc-drew-cutchins: I personally think the added verbosity/duplication of `deleteElement` and `deleteChildren` is preferable to the previous function that hides the relationship between the propName argument and which elements should be deleted. 

This makes it clear the relationship between the `propName` value and what elements should be deleted.

I'm open to other opinions here!

packages/runtime/src/state/modules/element-trees.ts
```diff
@@ -251,18 +251,27 @@ function hasChildren(
...
 ) {
-  if (propName == null || hasChildren(deletedElement, propName, descriptors)) {
-    for (const element of traverseElementTree(deletedElement, descriptors)) {
-      elements.delete(element.key)
-      elementIds.delete(element.key)
+  for (const element of traverseElementTree(parentElement, descriptors)) {
+    if(element === parentElement) {
+      continue
     }
-  } else {
```
bc-drew-cutchins, line 263: This `else` led to the deletion of the element when we wished to only delete a prop from it.

packages/runtime/src/state/modules/element-trees.ts
```diff
@@ -251,18 +251,27 @@ function hasChildren(
...
+  }
+}
+
+function deleteChildren(
+  { elements, elementIds }: ElementTree,
+  parentElement: Element,
   descriptors: DescriptorsByComponentType,
 ) {
-  if (propName == null || hasChildren(deletedElement, propName, descriptors)) {
-    for (const element of traverseElementTree(deletedElement, descriptors)) {
```
bc-drew-cutchins, line 259: In this case we are requesting to delete the children elements referenced by a prop, but `traverseElementTree(deletedElement, descriptors)` also included the element itself, causing it to be deleted when we only wanted to delete the children.

packages/runtime/src/state/modules/element-trees.ts
arvinpoddar: I noticed this block is repeated twice, is it worth abstracting this?

bc-drew-cutchins: I couldn't come up with a name that makes it obvious what the extracted function here would do, if you have a suggestion I'm happy to factor it out. I'm probably more afraid of the abstraction than the duplication here since the abstraction is what made this bug so difficult to track down.

arvinpoddar: I'm not hard pressed to abstract it, curious what @agurtovoy thinks

packages/runtime/src/state/modules/__tests__/element-trees.test.ts
```diff
@@ -198,50 +200,96 @@ describe('applyChanges', () => {
...
-    const updatedState = ElementTrees.reducer(
-      initialState,
-      changeElementTree({
-        oldDocument,
-        newDocument,
-        descriptors,
-        operation: reparentOp,
-      }),
-    )
+    const updatedState = reduceChangeElementTree(oldDocument, newDocument, reparentOp)
```
agurtovoy, line 253: @bc-drew-cutchins thanks for factoring this out!

packages/runtime/src/state/modules/element-trees.ts
```diff
@@ -251,18 +251,27 @@ function hasChildren(
...
-      elements.delete(element.key)
-      elementIds.delete(element.key)
+  for (const element of traverseElementTree(parentElement, descriptors)) {
+    if(element === parentElement) {
```
agurtovoy: Here and below, there are some missing spaces between the `if` and the paren.

bc-drew-cutchins: Fixed this and other prettier fixes in file 👍

packages/runtime/src/state/modules/element-trees.ts
```diff
@@ -272,13 +281,18 @@ function applyDelete(
...
   const elements = new Map(elementTree.elements)
   const elementIds = new Map(elementTree.elementIds)
 
-  deleteElement({ elements, elementIds }, deletedElement, propName, descriptors)
+  if(propName == null) {
+    deleteElement({ elements, elementIds }, targetElement, descriptors)
+  }
+  else if(hasChildren(targetElement, propName, descriptors)) {
+    deleteChildren({ elements, elementIds }, targetElement, descriptors)
+  }
```
agurtovoy: It looks like if the prop doesn't children, we never delete it. I know the tests are passing, and I don't have a good explanation why, but printing `updatedBoxElement` to the console in line 291 clearly shows that `width` is still there:

```
updatedBoxElement {
      "key": "ecb2a405-fac7-4b99-a0e0-d7ecb361d949",
      "props": {
        "padding": [
         ...
        ],
        "width": {
          "@@makeswift/type": "prop-controllers::width::v1",
          "value": [
            {
              "deviceId": "desktop",
              "value": {
                "value": 50,
                "unit": "%"
              }
            }
          ]
        }
      },
      "type": "./components/Box/index.js"
    }
```

This check also fails:

```
    expect((updatedBoxElement as any)?.props?.width).toBeUndefined()
```

agurtovoy: Oh, I know why this check passes, we never assert anything! What a foot gun.

```
    expect(
      updatedBoxElement && 'props' in updatedBoxElement && !('width' in updatedBoxElement.props),
    )
```

bc-drew-cutchins: Oops 🤦. Good catch, will fix. 

I think the cursor comment is getting at the same thing: the op is not actually applied to the element. 

On first glance I don't think this is a regression and that the op was never actually being applied. I guess this is of course true because we were simply deleting the element that we should have been deleting the prop from. 

The plot thickens...

packages/runtime/src/state/modules/element-trees.ts
```diff
@@ -272,13 +281,17 @@ function applyDelete(
...
+  const [targetElementPath, ...parentElementPaths] = getChangedElementsPaths(path)
+  const [targetElement, propName] = getElementAndPropName(rootElements.before, targetElementPath)
 
   const elements = new Map(elementTree.elements)
   const elementIds = new Map(elementTree.elementIds)
 
-  deleteElement({ elements, elementIds }, deletedElement, propName, descriptors)
+  if (propName == null) {
+    deleteElement({ elements, elementIds }, targetElement, descriptors)
+  } else if (hasChildren(targetElement, propName, descriptors)) {
```
agurtovoy: We should probably rename `hasChildren` to `isChildrenProp`

packages/runtime/src/state/modules/element-trees.ts
```diff
@@ -272,13 +281,17 @@ function applyDelete(
...
+    deleteChildren({ elements, elementIds }, targetElement, descriptors)
```
agurtovoy: This deletes *all* element's children rather than just the children under the deleted prop.

packages/runtime/src/state/modules/element-trees.ts
agurtovoy: I'd abstract it into something like `handleDeleteProp`. I'd say the main thing that was obscuring the bug here was not the abstraction, but (really) poor naming.

packages/runtime/src/state/modules/element-trees.ts
agurtovoy: I'd add a test that catches that.

bc-drew-cutchins: Riiight... another good catch. And I believe something that was already in the implementation here? I'd imagine the fix is something like this?

```ts
function deleteChildrenInProp(
  { elements, elementIds }: ElementTree,
  parentElement: Element,
  propName: string,
  descriptors: DescriptorsByComponentType,
) {
  if (isElementReference(parentElement)) return

  const propDescriptor = descriptors.get(parentElement.type)?.[propName]
  if (propDescriptor == null) return

  const childrenInProp = Introspection.getElementChildren(
    propDescriptor,
    parentElement.props[propName],
  )

  for (const child of childrenInProp) {
    elements.delete(child.key)
    elementIds.delete(child.key)
  }
}
```

agurtovoy: Children could have other children, so it should be

```ts
for (const child of childrenInProp) {
    deleteElement({ elements, elementIds }, child, descriptors)
}
```

I'd also extract getting the children into something like 

```ts
function getPropChildren(
  element: Element,
  propName: string,
  descriptors: DescriptorsByComponentType,
): Element[] | null {
  if (isElementReference(element)) return null

  const propDescriptor = descriptors.get(element.type)?.[propName]
  if (propDescriptor == null) return null

  return Introspection.getElementChildren(propDescriptor, element.props[propName])
}
```

and then `isChildrenProp` can become

```ts
function isChildrenProp(
  element: Element,
  propName: string,
  descriptors: DescriptorsByComponentType,
): boolean {
   const children = getPropChildren(element, propName, descriptors)
   return children && children.length > 0
}
```

agurtovoy: ... or, since we're calling it `isChildrenProp` rather than `doesPropHaveChildren`, just 

```ts
function isChildrenProp(
  element: Element,
  propName: string,
  descriptors: DescriptorsByComponentType,
): boolean {
    const children = getPropChildren(element, propName, descriptors)
    return children != null
}
```

packages/runtime/src/state/modules/element-trees.ts
bc-drew-cutchins: Fixed in [this commit](https://github.com/makeswift/makeswift/pull/1352/changes/ae24a00205b5e62969ddacb30e0b011306d7caec)

packages/runtime/src/state/modules/element-trees.ts
```diff
@@ -272,13 +281,17 @@ function applyDelete(
...
+  }
```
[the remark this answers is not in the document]
bc-drew-cutchins: Fixed in [this commit](https://github.com/makeswift/makeswift/pull/1352/changes/ae24a00205b5e62969ddacb30e0b011306d7caec)

packages/runtime/src/state/modules/__tests__/fixtures/element-trees.ts
```diff
@@ -4619,3 +4619,353 @@ export const reparentingElementTree: {
     },
   ],
 }
+
+export const deletingImageInBox = {
```
agurtovoy, line 4623: Can we also add a test fixture for deleting an element from an array (`List` control)? 🙏

packages/runtime/src/state/modules/__tests__/element-trees.test.ts
```diff
@@ -198,50 +200,103 @@ describe('applyChanges', () => {
...
 
     expect(consoleError).not.toHaveBeenCalled()
 
     const elements = ElementTrees.getElements(updatedState, documentKey)
     const rootKey = Fixtures.reparentingElementTree.before.key
     expect(elements.get(rootKey)).toEqual(Fixtures.reparentingElementTree.after)
   })
+
+  test('delete sole child of element', () => {
```
agurtovoy, line 262: I think we should also add a test for "replacing" a prop (delete + insert)

packages/runtime/src/state/modules/element-trees.ts
bc-drew-cutchins: Fixed and applied suggested refactors in [this commit](https://github.com/makeswift/makeswift/pull/1352/changes/fb5524369dd80dc00ddbeeed997d37261b6c0af9)

packages/runtime/src/state/modules/element-trees.ts
```diff
@@ -272,13 +291,21 @@ function applyDelete(
...
-  deleteElement({ elements, elementIds }, deletedElement, propName, descriptors)
+  if (propName == null) {
+    deleteElement({ elements, elementIds }, targetElement, descriptors)
+  } else if (isChildrenProp(targetElement, propName, descriptors)) {
+    deleteChildrenInProp({ elements, elementIds }, targetElement, propName, descriptors)
+  }
+
+  if (propName !== null && !isElementReference(targetElement)) {
+    delete targetElement.props[propName]
+  }
```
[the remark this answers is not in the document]
bc-drew-cutchins: Ah, this was only working because the unit test constructed the element tree from the old document, meaning the objects in the element tree were the same as those in the old document. I have updated the unit test setup to create a structured clone of the old document to use for creating the element tree.

I have updated the code to mutate only the element from the ElementTree, not the document.

Fixed in [this commit](https://github.com/makeswift/makeswift/pull/1352/changes/352457d1a416209850c8c63ef38931916b1daafd).

packages/runtime/src/state/modules/element-trees.ts
bc-drew-cutchins: Ended up factoring this out into `deleteReferencedElements` in [this commit](https://github.com/makeswift/makeswift/pull/1352/changes/585a9e552229becb4b2b5abcfa3725b2609d3225). 

I feel like `handleDeleteProp` doesn't make it clear that the element itself would be deleted if the prop is null, but also can defer to your judgement here?

packages/runtime/src/state/modules/__tests__/fixtures/element-trees.ts
bc-drew-cutchins: Added in [this commit](https://github.com/makeswift/makeswift/pull/1352/changes/0c6e529e776f7f6ed4db898b0a5053d0fd9283c8)

packages/runtime/src/state/modules/__tests__/element-trees.test.ts
bc-drew-cutchins: Added in [this commit](https://github.com/makeswift/makeswift/pull/1352/changes/0c6e529e776f7f6ed4db898b0a5053d0fd9283c8)

packages/runtime/src/state/modules/element-trees.ts
```diff
@@ -272,13 +304,18 @@ function applyDelete(
...
   const elements = new Map(elementTree.elements)
   const elementIds = new Map(elementTree.elementIds)
 
-  deleteElement({ elements, elementIds }, deletedElement, propName, descriptors)
+  deleteReferencedElements({ elements, elementIds }, targetElement, propName, descriptors)
+
+  const elementFromTree = elementTree.elements.get(targetElement.key)
+  if (propName !== null && elementFromTree && !isElementReference(elementFromTree)) {
+    delete elementFromTree.props[propName]
+  }
```
agurtovoy: This is still not entirely kosher, as we're directly mutating the current state (`elementTree.elements`) rather than building a new state that fully encapsulates our changes (`elements` and `elementIds` maps). This happens to work because while the maps themselves are new, the elements they are referencing are shared with the original state, so mutating that element in the original state also mutates this in the new state.

A more correct way to handle a prop deletion here would be to copy the updated element or its props, delete the prop, and place the updated copy in the `elements` map, e.g.

```ts
  const element = elements.get(targetElement.key)
  if (propName !== null && element && !isElementReference(element)) {
    const elementProps = { ...element.props }
    delete elementProps[propName]
    elements.set(targetElement.key, { ...element, props: elementProps })
  }
```

agurtovoy: On a related note, I think this logic should be a part of what's currently called `deleteReferencedElements`. While the update (delete + insert) technically works as-is because the insert will replace the "stale" prop with the new value, it feels unclean/raises a question that would at the very least warrant a comment, and personally I'd prefer the "clean" version over a version with a comment that forces you to calculate whether that comment always holds.

agurtovoy: Moving the prop deletion into that function would also allow make it a bit cleaner:

```ts
function deletePropOrElement(
  { elements, elementIds }: ElementTree,
  targetElement: Element,
  propName: string | null,
  descriptors: DescriptorsByComponentType,
) {
  if (propName == null) {
    deleteElement({ elements, elementIds }, targetElement, descriptors)
    return
  }

  if (isChildrenProp(targetElement, propName, descriptors)) {
    deleteChildrenInProp({ elements, elementIds }, targetElement, propName, descriptors)
  }

  const element = elements.get(targetElement.key)
  if (element && !isElementReference(element)) {
    const elementProps = { ...element.props }
    delete elementProps[propName]
    elements.set(targetElement.key, { ...element, props: elementProps })
  }
}
```

packages/runtime/src/state/modules/element-trees.ts
agurtovoy: I think `deleteReferencedElements` is a good name for the logic that's currently in this function, but see my comment here: https://github.com/makeswift/makeswift/pull/1352#discussion_r3573435820

packages/runtime/src/state/modules/element-trees.ts
bc-drew-cutchins: Ah right, I mistakenly though we would need to mutate the existing object in order to correctly update the objects referenced by other elements throughout the tree. However it seems like `updateParentElements(elements, parentElementPaths, rootElements.after)` addresses this. 

Implemented this function, renaming it to `deleteElementOrProp` to match the parameter order and added a console error whenever `elements.get(targetElement.key)` returns null. Can revert these if prefer your original suggestion.

Fixed in [this commit](https://github.com/makeswift/makeswift/pull/1352/changes/1a125b4bc1835b0837818217e2abc6b944eced4d).

agurtovoy (approved):
Looks great, thanks for adding the tests that should have been there from the start!

PR #1354, opened 2026-07-10
feat(ci): add linting configuration

arvinpoddar:
Define a base linting configuration used by all library packages. To start, rules that cause errors are being replaced with warnings so we have time to gradually fix them piecewise.

Also introduces a Github CI workflow to ensure lint checks pass.


packages/eslint-config/index.mjs
arvinpoddar, line 1: Defining a base eslint config in a package that can be extended by the other library packages.

agurtovoy (approved):
🙏

PR #1359, opened 2026-07-13
feat(controls): add unstable_Gallery control (ARR-977)

drewpledger:
New Gallery control to display a list of selectable images through the callback `getOptions` similar to Combobox:

Post-polished after Andrew's review:

https://github.com/user-attachments/assets/4f7e762a-848c-4171-b537-6b3fd8d89e68


arvinpoddar:
Could you also add some type tests? Also, is there anything majorly different about how this control is constructed vs. the combobox (aside from the `thumbnailUrl` field) worth calling out?

drewpledger:
How would you like me to call these differences out? Is this for documentation or for your review?

arvinpoddar:
Just for my review

drewpledger:
> Just for my review

Here's what I gathered:

There's no argument for `getOptions`, whereas Combobox takes in `query` to filter.
The return shape is a bit different, Combobox returns a bare array whereas Gallery wraps it in an object, allowing for more metadata in the future.

Other than that they are nearly identical besides the `thumbnailUrl` field

arvinpoddar (dismissed):

arvinpoddar (approved):

PR #1361, opened 2026-07-15
feat(runtime): unstable API to migrate components

arvinpoddar:
Introduce unstable API as part of component registrations that allows users to point to a "successor" component registration.

### DIFF
--- a/packages/runtime/src/runtimes/react/react-runtime-core.ts
+++ b/packages/runtime/src/runtimes/react/react-runtime-core.ts
@@ -37,6 +37,7 @@ export class ReactRuntimeCore extends RuntimeCore {
       hidden = false,
       description,
       builtinSuspense,
+      unstable_migration,
       props,
     }: {
       type: string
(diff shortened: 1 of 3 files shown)


packages/runtime/src/runtimes/react/react-runtime-core.ts
```diff
@@ -50,6 +51,12 @@ export class ReactRuntimeCore extends RuntimeCore {
        * Defaults to `true`.
        */
       builtinSuspense?: boolean
+      /**
+       * Marks this registration as superseded by `latestType`. When
+       * `latestType` differs from `type`, the builder offers a way to update
+       * existing instances.
+       */
+      unstable_migration?: { latestType: string }
```
agurtovoy: `latestType` sounds a little awkward to me and may not be semantically precise; generally speaking, a component can have multiple generations, and the migration may target any one of them, not necessarily the latest.

Maybe `targetType` or `replacementType`? Or just `componentType`?

arvinpoddar: I like `replacementType`, I think it communicates the clearest intent.

agurtovoy (approved):

PR #1363, opened 2026-07-16
fix: builtin embed resetting DOM mutations in builder

arvinpoddar:
A new behavior in React 19 wipes any changes to the DOM that were introduced via `dangerouslySetInnerHtml` if the prop identity changes.

For the builtin embed component, this would manifest in the builder for components that contained a script that modified the  DOM.

1. Page initially renders in builder
2. Embed renders, script runs, modifying the DOM created by the Embed.
3. Builder connection initializes, re-render occurs. The `dangerouslySetInnerHTML` inline object prop has a new identity, causing the Embed's div innerHtml to be reset.
4. The Embed scripts don't re-execute since the HTML string itself didn't change, so the DOM remains stale.

https://github.com/user-attachments/assets/cd03ac02-3c1e-4b29-8c47-9ab8a387e9c8

### DIFF
--- a/.changeset/puny-monkeys-talk.md
+++ b/.changeset/puny-monkeys-talk.md
@@ -1,0 +1,5 @@
+---
+'@makeswift/runtime': patch
+---
+
+Fixes an issue where scripts within the builtin `Embed` component may not have properly persisted DOM manipulations in the builder.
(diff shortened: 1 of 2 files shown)


agurtovoy (approved):
🙏

PR #1364, opened 2026-07-16
fix: explicitly include Jest types in package globals

agurtovoy:
Make sure the IDE can find type definitions for the `describe`/`test`/`expect` et al. globals coming from Jest. These used to be discovered automatically, but are no longer resolved consistently; adding them to `tsconfig` to restore consistency.

### DIFF
--- a/packages/controls/tsconfig.json
+++ b/packages/controls/tsconfig.json
@@ -6,6 +6,7 @@
     "module": "ESNext",
     "lib": ["ESNext", "DOM"],
     "moduleResolution": "Node",
+    "types": ["jest"],
     "strict": true,
     "sourceMap": true,
     "resolveJsonModule": true,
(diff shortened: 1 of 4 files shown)


packages/controls/tsconfig.json
```diff
@@ -6,6 +6,7 @@
     "module": "ESNext",
     "lib": ["ESNext", "DOM"],
     "moduleResolution": "Node",
+    "types": ["jest"],
```
agurtovoy, line 9: The [`types` field](https://www.typescriptlang.org/tsconfig/#types) regulates only types that are implicitly injected in the global scope. It does not affect regular type resolution/visibility of explicitly imported symbols.

It also does *not* affect visibility of browser's globals, which are regulated by the [`lib` field](https://www.typescriptlang.org/tsconfig/#lib).

arvinpoddar (approved):
🙏

PR #1366, opened 2026-07-17
feat(controls): add `resolvesToRenderableNode` method to control instances

agurtovoy:
Pure addition. Returns true for `Slot` and `RichText`, false for all other controls.

### DIFF
--- a/packages/controls/src/controls/group/group-control.ts
+++ b/packages/controls/src/controls/group/group-control.ts
@@ -43,6 +43,8 @@ export class GroupControl<
 
   child = (key: string) => this.childControls.get(key)
 
+  resolvesToRenderableNode = () => false
+
   createChildControl = (def: ControlDefinition, key: string) => {
     const { elementKey, propPath } = this.instanceKey
     return def.createInstance({
--- a/packages/controls/src/controls/list/list-control.ts
+++ b/packages/controls/src/controls/list/list-control.ts
@@ -38,6 +38,8 @@ export class ListControl<
 
   child = (key: string) => this.itemControls.get(key)
 
+  resolvesToRenderableNode = () => false
+
   childControls = (ids: string[] | undefined): Map<string, ControlInstance> => {
     const orderedIds: string[] = [...this.itemControls.keys()]
     // return existing controls if we have a full set of them and they are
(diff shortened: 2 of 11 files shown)


packages/controls/src/controls/group/group-control.ts
```diff
@@ -43,6 +43,8 @@ export class GroupControl<
 
   child = (key: string) => this.childControls.get(key)
 
+  resolvesToRenderableNode = () => false
```
agurtovoy, line 46: Keeping the instance property syntax for controls that use it for the rest of their methods.

packages/controls/src/controls/list/list-control.ts
```diff
@@ -38,6 +38,8 @@ export class ListControl<
 
   child = (key: string) => this.itemControls.get(key)
 
+  resolvesToRenderableNode = () => false
```
bc-drew-cutchins, line 41: Just double checking: This (or group) doesn't need to recursively check whether any of the child controls resolve to a renderable node?

bc-drew-cutchins (dismissed):
LGTM assuming the answer to my question is no

packages/controls/src/controls/group/group-control.ts
arvinpoddar: So this method will return `false` even if its children return `true`, like a `List` of `Slot`?

packages/controls/src/controls/list/list-control.ts
arvinpoddar: Lol same question at the same time: https://github.com/makeswift/makeswift/pull/1366/changes#r3606212889

packages/controls/src/controls/group/group-control.ts
agurtovoy: Yes, both `List` and `Group` returns false here because they don't render to a node. They could potentially render to an array or a record of nodes (like in your example), but this isn't what we're after here.

packages/controls/src/controls/list/list-control.ts
agurtovoy: Great question, I should have commented on this. Please see https://github.com/makeswift/makeswift/pull/1366/changes#r3606230195

arvinpoddar (dismissed):

PR #1367, opened 2026-07-18
feat(controls): allow to access `List` item controls by index

agurtovoy:
Follow up to https://github.com/makeswift/makeswift/pull/1347.

### DIFF
--- a/packages/controls/src/controls/instance.ts
+++ b/packages/controls/src/controls/instance.ts
@@ -41,6 +41,13 @@ export abstract class ControlInstance<
   }
 
   abstract recv(message: M): void
+
+  /**
+   *
+   * Returns the control instance for the nested prop identified by `key`.
+   *
+   * For list controls, `key` is a stringified item index.
+   */
   abstract child(key: string): ControlInstance | undefined
 
   /**
--- a/packages/controls/src/controls/list/list-control.ts
+++ b/packages/controls/src/controls/list/list-control.ts
@@ -19,7 +19,8 @@ export class ListControl<
   static readonly ITEM_CONTROL_MESSAGE =
     'makeswift::controls::list::message::item-control-message'
 
-  private itemControls: Map<string, ControlInstance> = new Map()
+  private itemControlsById: Map<string, ControlInstance> = new Map()
+  private itemControlsList: Array<ControlInstance> = []
 
   constructor(
     private readonly definition: Def,
@@ -31,34 +32,54 @@ export class ListControl<
   recv = (message: Message) => {
     switch (message.type) {
       case ListControl.ITEM_CONTROL_MESSAGE: {
-        this.child(message.payload.itemId)?.recv(message.payload.message)
+        const { itemId, message: itemMessage } = message.payload
+        this.childByItemId(itemId)?.recv(itemMessage)
       }
     }
   }
 
-  child = (key: string) => this.itemControls.get(key)
+  /**
+   *  Returns the control instance for the nested prop identified by `key`,
+   *  where `key` is a stringified list item index.
+   */
+  child = (key: string): ControlInstance | undefined => {
+    const index = parseIndex(key)
+
+    if (index == null) {
+      console.error(
+        `Expected a stringified list item index, received '${key}'`,
+        this.instanceKey,
+      )
+
+      return undefined
+    }
+
+    return this.itemControlsList[index]
+  }
 
   resolvesToRenderableNode = () => false
 
-  childControls = (ids: string[] | undefined): Map<string, ControlInstance> => {
-    const orderedIds: string[] = [...this.itemControls.keys()]
+  childControls = (ids?: string[]): Map<string, ControlInstance> => {
+    const orderedIds: string[] = [...this.itemControlsById.keys()]
     // return existing controls if we have a full set of them and they are
     // already correctly ordered
     if (ids == null || shallowEqual(ids, orderedIds)) {
-      return this.itemControls
+      return this.itemControlsById
     }
 
     // otherwise recreate controls for new/updated items as needed
     return new Map(
       ids.map((id, index) => {
-        const matchingChild = id === orderedIds[index] ? this.child(id) : null
+        const matchingChild =
+          id === orderedIds[index] ? this.childByItemId(id) : null
         return [id, matchingChild ?? this.createItemControl(index, id)]
       }),
     )
   }
 
   setChildControls = (children: Map<string, ControlInstance>) => {
-    this.itemControls = children
+    this.itemControlsById = children
+    this.itemControlsList = [...children.values()]
   }
 
   createItemControl = (index: number, itemId: string): ControlInstance => {
@@ -75,4 +96,14 @@ export class ListControl<
         }),
     })
   }
+
+  private childByItemId = (itemId: string): ControlInstance | undefined =>
+    this.itemControlsById.get(itemId)
+}
+
+const parseIndex = (key: string): number | undefined => {
+  const parsedNum = Number.parseInt(key, 10)
+  const isValidIndex =
+    !Number.isNaN(parsedNum) && parsedNum >= 0 && String(parsedNum) === key
+  return isValidIndex ? parsedNum : undefined
 }
--- a/packages/controls/src/controls/list/list.test.ts
+++ b/packages/controls/src/controls/list/list.test.ts
@@ -154,16 +154,12 @@ describe('List', () => {
     test('resolveValue creates a list of child controls with correct instance keys', async () => {
       const { listInstance, resolveValueContext } = fixtures()
 
-      expect(
-        listData
-          .map(({ id }) => listInstance.child(id))
-          .filter((c) => c != null),
-      ).toEqual([])
+      expect(listInstance.childControls().size).toBe(0)
 
       const resolved = list.resolveValue(listData, ...resolveValueContext)
       await resolved.triggerResolve()
 
-      const childControls = listData.map(({ id }) => listInstance.child(id))
+      const childControls = [...listInstance.childControls().values()]
 
       expect(
         childControls
@@ -184,11 +180,11 @@ describe('List', () => {
 
       const resolved = list.resolveValue(listData, ...resolveValueContext)
       await resolved.triggerResolve()
-      const childControls = listData.map(({ id }) => listInstance.child(id))
+      const childControls = [...listInstance.childControls().values()]
 
       const resolved2 = list.resolveValue(listData, ...resolveValueContext)
       await resolved2.triggerResolve()
-      const childControls2 = listData.map(({ id }) => listInstance.child(id))
+      const childControls2 = [...listInstance.childControls().values()]
 
       childControls2.forEach((item, i) => expect(item).toBe(childControls[i]))
     })
@@ -198,12 +194,12 @@ describe('List', () => {
 
       const resolved = list.resolveValue(listData, ...resolveValueContext)
       await resolved.triggerResolve()
-      const childControls = listData.map(({ id }) => listInstance.child(id))
+      const childControls = [...listInstance.childControls().values()]
 
       const listData2 = [...listData].sort((a, b) => b.id.localeCompare(a.id))
       const resolved2 = list.resolveValue(listData2, ...resolveValueContext)
       await resolved2.triggerResolve()
-      const childControls2 = listData2.map(({ id }) => listInstance.child(id))
+      const childControls2 = [...listInstance.childControls().values()]
 
       childControls2.forEach((item, i) =>
         expect(item).not.toBe(childControls[i]),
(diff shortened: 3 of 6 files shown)


packages/controls/src/controls/list/list-control.ts
```diff
@@ -19,7 +19,8 @@ export class ListControl<
   static readonly ITEM_CONTROL_MESSAGE =
     'makeswift::controls::list::message::item-control-message'
 
-  private itemControls: Map<string, ControlInstance> = new Map()
+  private itemControlsById: Map<string, ControlInstance> = new Map()
+  private itemControlsList: Array<ControlInstance> = []
```
agurtovoy, line 23: Here and below, renamed `itemControls` to `itemControlsById` for clarity. 

The map gives us efficient lookup by item ID, the array gives us efficient indexing.

packages/controls/src/controls/list/list-control.ts
```diff
@@ -36,16 +37,21 @@ export class ListControl<
     }
   }
 
-  child = (key: string) => this.itemControls.get(key)
+  child = (key: string) => {
+    const [keyOrIndex, isIndex] = childKeyOrIndex(key)
+    return isIndex
+      ? this.itemControlsList[keyOrIndex]
+      : this.itemControlsById.get(key)
+  }
```
agurtovoy, line 58: Given that we [fully control](https://github.com/makeswift/makeswift/blob/main/packages/controls/src/controls/list/list.ts#L153) list ID generation and the fact that in JS `array[0]` and `array["0"]` mean the same thing, technically this could be just `return this.itemControlsList[key] ?? this.itemControlsById.get(key)`, but that felt too dirty/fragile to me.

packages/controls/src/controls/list/list-control.ts
```diff
@@ -58,7 +64,8 @@ export class ListControl<
   }
 
   setChildControls = (children: Map<string, ControlInstance>) => {
-    this.itemControls = children
+    this.itemControlsById = children
+    this.itemControlsList = [...children.values()]
```
agurtovoy, line 82: Set both states from a single source.

packages/controls/src/controls/list/list.test.ts
```diff
@@ -243,6 +243,21 @@ describe('List', () => {
         'list-prop.3',
       ])
     })
+
+    test('child controls can be accessed through `child(index)`', async () => {
```
agurtovoy, line 243: Added a test to verify the new behavior.

packages/controls/src/controls/list/list-control.ts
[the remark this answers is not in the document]
agurtovoy: Treating numeric IDs as indexes is the whole point of this PR. Item IDs are not numeric, and we have tests that will break if that ever changes. See also https://github.com/makeswift/makeswift/pull/1367/changes#r3607363427.

packages/controls/src/controls/list/list-control.ts
```diff
@@ -36,16 +37,21 @@ export class ListControl<
     }
   }
 
-  child = (key: string) => this.itemControls.get(key)
+  child = (key: string) => {
```
bc-drew-cutchins: I tend to feel that sending an integer index as a string is a bit surprising. I also shared Cursor's concern about numerical ids being interpreted as indices on first read (though fair enough that a unit test would catch numerical ids).

Are we having to conform to an interface where `child` must be used to access by index or would it be possible to create a new method `childAtIndex` or type `key` as `string | number`?

agurtovoy: First of all, I want to explicitly acknowledge that this is somewhat of a code smell, but the roots of it are less in the functionality that's being added in this PR and more in maintaining backward compatibility with the initial misguided design. 

The root problem here is that `child(key)` was always intended to be a nested prop accessor; for a list, nested property keys correspond to indexes in the list, and that's how `ListControl.child(key)` should have interpreted the key from the start. However, we ended up not having a use case for uniform nested prop access, and at the same time the list control implementation needed an association between a list item ID (a completely separate concept from prop key) and the corresponding list item control, and the `child(key)` method semantics morphed into "access a nested control using a key, whatever the key means for this particular parent". 

In practice, we are not even using `ListControl.child(key)` to access list item controls by list item ID, but I felt cautious over breaking this original albeit misguided contract, and so instead of removing the item-by-ID behavior, I opted for the dual semantics -- which is IMHO the most problematic aspect here.

Note that the above is separate from the question of whether it's okay for `child(key: string)` to coerce the key to an array index.

The latter is a typical tradeoff of the Composite design pattern: we want to be able to iterate over/access nested prop controls uniformly, without having to do a check for specific instance types. The price for doing so is two-fold: first, the base class interface (in our case, `ControlInstance`) has to become somewhat overgeneralized to include methods for accessing nested controls even though the majority of derived classes don't have them; secondly, that interface has to balance the differing needs of all of the derived classes. 

In this specific case, adding `childAtIndex` either to a list instance or to the base `ControlInstance` interface doesn't solve the problem of uniformly accessing a child at a specific property path. A property path is a composite string along the lines of `foo.items.0.bar`; this is a common convention in libraries and aligns with JS object model, in which `[17, 31, 42]["0"]` is a perfectly valid expression. While you could write code along the lines of "if a prop path component is a numeric index, use `childAtIndex` instead of `child`", this simply moves index parsing to the call site and makes client code more verbose. 

Changing the `child` method to have `key` be typed as `string | number` is similarly not helpful, for as long as property path is represented as a composite string, the index will be coming in as a string. Additionally, allowing the key to be a number would force us to add a check for a numeric key to the group control, so we'd effectively be trading one check for another. It's also worth noting that parsing the index in the list control is necessary only to satisfy TS. At runtime, the array converts back to a string anyway.

That said, there is an alternative design here that is worth considering: have `ControlInstance` define both `child(key: string)` and `childAtIndex(key: number)` and change the property path representation from a string to `[string, ...(string|number)[]]`. Client code would become more complicated and would still have to check the type of each path component and dispatch to either `child` or `childAtIndex` accordingly, but we would no longer have to parse list indexes. The group control's `childAtIndex` would always return `undefined`, and the list control's `child` would always return `undefined`. The property path would also cease to be a scalar, and would not longer be easily comparable or serializable, so we'd have to add utility functions for that. I don't hate this, but the number of tradeoffs/concessions here is non-trivial (an extra child accessor in the base interface, more awkward client code, property paths that are harder to work with), and so at the moment it feels like a wash to me.

After writing all of this down, I *am* inclined to remove the backward compatibility item IDs path from list control's `child` and have it access the nested properties by index only.

bc-drew-cutchins: Thanks for taking the time to write all this out, that all makes sense! You definitely have more context for what's the right call here so I'm happy with whatever approach you think is best.

bc-drew-cutchins (dismissed):

packages/controls/src/controls/instance.ts
```diff
@@ -41,6 +41,13 @@ export abstract class ControlInstance<
...
 
   abstract recv(message: M): void
+
+  /**
+   *
+   * Returns the control instance for the nested prop identified by `key`.
+   *
+   * For list controls, `key` is a stringified item index.
+   */
   abstract child(key: string): ControlInstance | undefined
```
agurtovoy, line 51: Added a comment clarifying `child` method's semantics. See the corresponding discussion here: https://github.com/makeswift/makeswift/pull/1367#discussion_r3614701339

packages/controls/src/controls/list/list-control.ts
```diff
@@ -31,34 +32,54 @@ export class ListControl<
   recv = (message: Message) => {
     switch (message.type) {
       case ListControl.ITEM_CONTROL_MESSAGE: {
-        this.child(message.payload.itemId)?.recv(message.payload.message)
+        const { itemId, message: itemMessage } = message.payload
+        this.childByItemId(itemId)?.recv(itemMessage)
```
agurtovoy, line 36: `child` -> `childByItemId`

`child` is now officially reserved specifically for accessing control instances of nested props; see https://github.com/makeswift/makeswift/pull/1367/changes#r3617238340.

packages/controls/src/controls/list/list-control.ts
```diff
@@ -31,34 +32,54 @@ export class ListControl<
...
+    if (index == null) {
+      console.error(
+        `Expected a stringified list item index, received '${key}'`,
+        this.instanceKey,
+      )
+
+      return undefined
+    }
+
+    return this.itemControlsList[index]
```
agurtovoy, line 57: Redefine `child` to satisfy the new semantics.

packages/controls/src/controls/list/list-control.ts
```diff
@@ -31,34 +32,54 @@ export class ListControl<
...
-      return this.itemControls
+      return this.itemControlsById
     }
 
     // otherwise recreate controls for new/updated items as needed
     return new Map(
       ids.map((id, index) => {
-        const matchingChild = id === orderedIds[index] ? this.child(id) : null
+        const matchingChild =
+          id === orderedIds[index] ? this.childByItemId(id) : null
```
agurtovoy, line 74: `child` -> `childByItemId`

packages/controls/src/controls/list/list-control.ts
```diff
@@ -31,34 +32,54 @@ export class ListControl<
...
+    }
+
+    return this.itemControlsList[index]
+  }
 
   resolvesToRenderableNode = () => false
 
-  childControls = (ids: string[] | undefined): Map<string, ControlInstance> => {
-    const orderedIds: string[] = [...this.itemControls.keys()]
+  childControls = (ids?: string[]): Map<string, ControlInstance> => {
```
agurtovoy, line 62: Made `ids` optional so we can use `childControls()` as a map accessor.

packages/controls/src/controls/list/list.test.ts
```diff
@@ -154,16 +154,12 @@ describe('List', () => {
     test('resolveValue creates a list of child controls with correct instance keys', async () => {
       const { listInstance, resolveValueContext } = fixtures()
 
-      expect(
-        listData
-          .map(({ id }) => listInstance.child(id))
-          .filter((c) => c != null),
-      ).toEqual([])
+      expect(listInstance.childControls().size).toBe(0)
```
agurtovoy, line 157: Here and below, reworked tests to not rely on child-by-ID lookup.

packages/controls/src/controls/style/v2/style-v2-control.ts
```diff
@@ -11,42 +11,42 @@ type ItemBoxModelChangeMessage = {
   payload: { boxModel: BoxDisplayModel | null }
 }
 
-type ChildControlMessage = {
-  type: typeof StyleV2Control.CHILD_CONTROL_MESSAGE
+type InnerControlMessage = {
+  type: typeof StyleV2Control.INNER_CONTROL_MESSAGE
   payload: { message: ControlMessage }
 }
```
agurtovoy, line 17: Stop calling the inner `StyleV2` control a "child" to avoid conflating it with nested props.

packages/controls/src/controls/style/v2/style-v2-control.ts
```diff
@@ -11,42 +11,42 @@ type ItemBoxModelChangeMessage = {
...
-type Message = ItemBoxModelChangeMessage | ChildControlMessage
+type Message = ItemBoxModelChangeMessage | InnerControlMessage
 
 export class StyleV2Control extends ControlInstance<Message> {
   static readonly CHANGE_BOX_MODEL =
     'makeswift::controls::style::message::change-box-model'
-  static readonly CHILD_CONTROL_MESSAGE =
-    'makeswift::controls::style-v2::message::child-control-message'
+  static readonly INNER_CONTROL_MESSAGE =
+    'makeswift::controls::style-v2::message::inner-control-message'
```
agurtovoy, line 25: This is internal message that's only dispatched and intercepted by for this control, so it is safe for us to change this literal to reflect the larger rename.

packages/controls/src/controls/style/v2/style-v2-control.ts
```diff
@@ -11,42 +11,42 @@ type ItemBoxModelChangeMessage = {
...
+          type: StyleV2Control.INNER_CONTROL_MESSAGE,
           payload: { message },
         }),
     })
   }
 
-  child(_key?: string): ControlInstance | undefined {
-    return this.control
+  child(_key: string): ControlInstance | undefined {
+    return undefined
```
agurtovoy, line 42: `StyleV2` is not a composite control and doesn't have nested props.

packages/controls/src/controls/list/list-control.ts
agurtovoy: I went with strengthening the semantics of the `child` method as a nested prop accessor and reworking a couple of places in the code that conflicted with that definition, including the list control's `child`. Thanks for raising this question, answering it helped me clarify my own thinking, and both this and the [follow-up PR](https://github.com/makeswift/makeswift/pull/1368) are/will be better for it.

bc-drew-cutchins: Awesome, looks great! Thank you!

PR #1368, opened 2026-07-18
feat(controls): add `children`, `isCompositeProp` methods

agurtovoy:
`children` gives us a generic way to iterate over all control instances corresponding to the props of a given element; `isCompositeProp` gives us the ability to differentiate between "scalar" and "composite" props, including cases where a control is composite but exposes no child control instances.

See inline comments for details.

### DIFF
--- a/packages/controls/src/controls/group/group-control.ts
+++ b/packages/controls/src/controls/group/group-control.ts
@@ -41,6 +41,10 @@ export class GroupControl<
     }
   }
 
+  isCompositeProp = () => true
+
+  children = () => [...this.childControls.values()]
+
   child = (key: string) => this.childControls.get(key)
 
   resolvesToRenderableNode = () => false
--- a/packages/controls/src/controls/instance.ts
+++ b/packages/controls/src/controls/instance.ts
@@ -43,7 +43,18 @@ export abstract class ControlInstance<
   abstract recv(message: M): void
 
   /**
+   * Returns true if the control represents a composite prop value (a prop that contains other props).
    *
+   * In contrast with `children()`, this return value doesn't change over the control lifetime.
+   */
+  abstract isCompositeProp(): boolean
+
+  /**
+   * Returns a list of child control instances for this control's immediate nested props.
+   */
+  abstract children(): ControlInstance[]
+
+  /**
    * Returns the control instance for the nested prop identified by `key`.
    *
    * For list controls, `key` is a stringified item index.
@@ -62,6 +73,14 @@ export abstract class ControlInstance<
 export class DefaultControlInstance extends ControlInstance {
   recv(_message: ControlMessage) {
     // Do nothing
+  }
+
+  isCompositeProp(): boolean {
+    return false
+  }
+
+  children(): ControlInstance[] {
+    return []
   }
 
   child(_key: string) {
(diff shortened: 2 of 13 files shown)


packages/controls/src/controls/instance.ts
```diff
@@ -41,6 +41,17 @@ export abstract class ControlInstance<
...
   abstract recv(message: M): void
+
+  /**
+   * Returns true if the control represents a composite prop value (a prop that contains other props).
+   *
+   * Note that this isn't the same as having child controls; for example, `StyleV2` control instance
+   * has a child control, but `StyleV2` value resolves to a single scalar prop, and therefore its
+   * `isCompositeProp` implementation returns false.
+   */
+  abstract isCompositeProp(): boolean
```
agurtovoy, line 50: I considered naming this `resolvesToCompositeProp` to match `resolvesToRenderableNode`, but the it didn't quite feel right given its usage context.

Perhaps `resolvesToRenderableNode` should have been named `isRenderableNodeProp` 🤔 .

packages/controls/src/controls/group/group-control.ts
```diff
@@ -41,6 +41,10 @@ export class GroupControl<
     }
   }
 
+  isCompositeProp = () => true
```
agurtovoy, line 44: See https://github.com/makeswift/makeswift/pull/1368/changes#r3607434651

packages/controls/src/controls/instance.ts
```diff
@@ -41,6 +41,17 @@ export abstract class ControlInstance<
...
+
+  abstract children(): ControlInstance[]
```
agurtovoy, line 55: `children` gives us a generic way to iterate over all control instances corresponding to the props of a given element; `isCompositeProp` gives us the ability to differentiate between "scalar" and "composite" props.

packages/controls/src/controls/group/group-control.ts
agurtovoy: Returns `true` for groups and lists, false for all other controls.

packages/controls/src/controls/list/list-control.ts
```diff
@@ -37,6 +37,10 @@ export class ListControl<
     }
   }
 
+  isCompositeProp = () => true
+
+  children = () => [...this.itemControlsList]
```
agurtovoy, line 43: Return a copy of the list so that it can't be accidentally mutated from the outside.

packages/controls/src/controls/shape/v1/shape-control.ts
```diff
@@ -41,6 +41,10 @@ export class ShapeControl<
     }
   }
 
+  isCompositeProp = () => true
+
+  children = () => [...this.childControls.values()]
```
agurtovoy, line 46: Same a `Group`.

bc-drew-cutchins (dismissed):

packages/controls/src/controls/list/list.test.ts
```diff
@@ -180,11 +180,11 @@ describe('List', () => {
 
       const resolved = list.resolveValue(listData, ...resolveValueContext)
       await resolved.triggerResolve()
-      const childControls = [...listInstance.childControls().values()]
+      const childControls = listInstance.children()
```
agurtovoy, line 183: Here and below, reformulate some of the list tests using the new `children()` method. No semantic changes.

PR #1357, opened 2026-07-12
getOptions control-context seam + unstable_Gallery

drewpledger:
Adds a fresh "unstable_Gallery" control, and seams the context from the builder to the messageChannel port.

PR #1369, opened 2026-07-20
feat(controls): unstable_Cascade composable spike (ARR-970)

drewpledger:
One of three parallel explorations of the same product goal (product → image picker) — siblings are #1351 (ARR-769) and #1357 (ARR-969). This one keeps the cascade as the container but makes `unstable_Cascade` **composable**: steps hold real controls (`Checkbox`, `Combobox`, ...) instead of fixed display types, piping each step's resolved value into the next step's factory.

Also cleaned up the message-port serialization it relies on: swapped a hand-rolled `MessageChannel` RPC for the shared `serializeFunction` helper, and made that helper propagate errors as rejections (previously a failure would just hang forever) and auto-forward nested `MessagePort`s instead of throwing `DataCloneError`.

**Known limitations — flagged for discussion, not fixed here:**
- Steps needing a real resolver (Image, Color) will throw; only a no-op stand-in resolver is available where the chain gets materialized.
- Only step 0's child control is wired up — step ≥1 interactions silently no-op until the builder-side panel exists.
- Translated content in a step control isn't merged yet.
- `deserialize` only accepts the live-connection shape, not persisted data (fine today — it's never called any other way in practice).

Draft, opened to align on direction before going further

### DIFF
--- a/packages/runtime/src/controls/serialization/message-port/function-serialization.ts
+++ b/packages/runtime/src/controls/serialization/message-port/function-serialization.ts
@@ -52,7 +52,10 @@ function collectTransferables(value: unknown, seen: Set<unknown> = new Set()): T
   return []
 }
 
-export function serializeFunction<T extends AnyFunction>(func: T): SerializedFunction<T> {
+export function serializeFunction<T extends AnyFunction>(
+  func: T,
+  hostPortRegistry?: MessagePort[],
+): SerializedFunction<T> {
   type CallMessageEvent = MessageEvent<[CallID, Parameters<T>]>
 
   const messageChannel = new MessageChannel()
@@ -71,6 +74,8 @@ export function serializeFunction<T extends AnyFunction>(func: T): SerializedFun
         },
       )
   }
+
+  hostPortRegistry?.push(messageChannel.port1)
 
   return messageChannel.port2 as SerializedFunction<T>
 }
--- a/packages/runtime/src/controls/serialization/message-port/function-serialization.ts
+++ b/packages/runtime/src/controls/serialization/message-port/function-serialization.ts
@@ -10,9 +10,10 @@ export { type DeserializedFunction } from '@makeswift/controls'
 declare const SerializedFunctionTag: unique symbol
 
 type ResolveCallPromise<T extends AnyFunction> = (value: SerializedFunctionReturnType<T>) => void
+type RejectCallPromise = (error: unknown) => void
 
 type OnMessageHandler<T extends AnyFunction> = MessagePort['onmessage'] & {
-  newCall?(resolve: ResolveCallPromise<T>): number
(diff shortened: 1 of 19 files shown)


packages/runtime/src/controls/serialization/message-port/function-serialization.ts
```diff
@@ -25,34 +26,83 @@ export function isSerializedFunction(value: any): value is SerializedFunction<An
...
   const messageChannel = new MessageChannel()
 
   messageChannel.port1.onmessage = ({ data: [callId, args] }: CallMessageEvent) => {
     Promise.resolve()
       .then(() => func.apply(null, args))
-      .then(result => messageChannel.port1.postMessage([callId, result]))
+      .then(
+        result => {
+          const reply: ReplyMessage<T> = [callId, 'resolve', result]
+          messageChannel.port1.postMessage(reply, collectTransferables(result))
```
drewpledger, line 69: From what I can smell, this won't affect any existing controls and will enable multi-staged transferrables.

packages/runtime/src/controls/serialization/message-port/function-serialization.ts
```diff
@@ -25,34 +26,83 @@ export function isSerializedFunction(value: any): value is SerializedFunction<An
...
+function toErrorDescriptor(error: unknown): ErrorDescriptor {
+  if (error instanceof Error) return { message: error.message, name: error.name }
+  return { message: String(error) }
+}
 
-export function serializeFunction<T extends AnyFunction>(func: T): SerializedFunction<T> {
+// Walks a plain object/array tree collecting any MessagePort found within it,
+// since a reply carrying one (e.g. a nested serialized function/control) must
+// list it as a transferable or postMessage throws DataCloneError.
+function collectTransferables(value: unknown, seen: Set<unknown> = new Set()): Transferable[] {
```
drewpledger, line 42: Open to suggestions, this smells fine to me.

packages/runtime/src/controls/serialization/message-port/function-serialization.ts
```diff
@@ -25,34 +26,83 @@ export function isSerializedFunction(value: any): value is SerializedFunction<An
...
+  if (Array.isArray(value)) return value.flatMap(item => collectTransferables(item, seen))
+  if (value.constructor === Object) {
+    return Object.values(value).flatMap(item => collectTransferables(item, seen))
+  }
+  return []
+}
+
+export function serializeFunction<T extends AnyFunction>(
+  func: T,
+  hostPortRegistry?: MessagePort[],
```
drewpledger, line 57: This is really only going to be used for cascade-chain-serialization. I cannot think of another way to clear the stale ports.

arvinpoddar:
Spike completed, closing this PR.

PR #1370, opened 2026-07-21
refactor: `useResolvedProps` take control instances + related cleanup

agurtovoy:
Decouple `useResolvedProps` from the `useControlInstances` hook.

### DIFF
--- a/packages/runtime/src/runtimes/react/hooks/use-resolved-props.ts
+++ b/packages/runtime/src/runtimes/react/hooks/use-resolved-props.ts
@@ -12,7 +12,6 @@ import { useResourceResolver } from './use-resource-resolver'
 import { type StylesheetFactory } from '../stylesheet-factory'
 import { propErrorHandlingProxy } from '../utils/prop-error-handling-proxy'
 
-import { useControlInstances } from './use-control-instances'
 import { useResolvableRecord } from './use-resolvable-record'
 
 type CacheItem = {
@@ -22,24 +21,25 @@ type CacheItem = {
 }
 
 export function useResolvedProps({
+  elementKey,
   propDefs,
   propData,
-  elementKey,
+  controlInstances,
   stylesheetFactory,
 }: {
+  elementKey: string
   propDefs: Record<string, ControlDefinition>
   propData: Record<string, Data>
-  elementKey: string
+  controlInstances: Record<string, ControlInstance> | null
   stylesheetFactory: StylesheetFactory
 }): Record<string, unknown> {
   const resourceResolver = useResourceResolver()
-  const controls = useControlInstances(elementKey)
 
   const cache = useRef<Record<string, CacheItem>>({}).current
   const resolveProp = useCallback(
     (def: ControlDefinition, propName: string) => {
       const data = propData[propName]
-      const control = controls?.[propName]
+      const control = controlInstances?.[propName]
 
       if (
         cache[propName] != null &&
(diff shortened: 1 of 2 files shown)


packages/runtime/src/runtimes/react/hooks/use-resolved-props.ts
```diff
@@ -22,24 +21,25 @@ type CacheItem = {
...
   propData,
-  elementKey,
+  controlInstances,
   stylesheetFactory,
 }: {
+  elementKey: string
   propDefs: Record<string, ControlDefinition>
   propData: Record<string, Data>
-  elementKey: string
+  controlInstances: Record<string, ControlInstance> | null
```
agurtovoy, line 33: Added `controlInstances` param, moved `elementKey` to the top.

packages/runtime/src/runtimes/react/hooks/use-resolved-props.ts
```diff
@@ -59,7 +59,7 @@ export function useResolvedProps({
       cache[propName] = { data, control, resolvedValue }
       return resolvedValue
     },
-    [controls, propData, resourceResolver, stylesheetFactory],
+    [propData, cache, resourceResolver, controlInstances, stylesheetFactory],
```
agurtovoy, line 62: Add `cache` to this hook's deps.

packages/runtime/src/runtimes/react/hooks/use-resolved-props.ts
```diff
@@ -68,21 +68,23 @@ export function useResolvedProps({
         const defaultValue = (def.config as any)?.defaultValue
         return propErrorHandlingProxy(resolveProp(def, propName), defaultValue, error => {
           console.warn(
-            `Error reading value for prop "${propName}", falling back to \`${defaultValue}\`.`,
+            `Error reading value for prop "${propName}" of element ${elementKey}, falling back to \`${defaultValue}\`.`,
```
agurtovoy, line 71: Add element key to the prop resolution error message.

packages/runtime/src/runtimes/react/hooks/use-resolved-props.ts
```diff
@@ -68,21 +68,23 @@ export function useResolvedProps({
...
   const props = useResolvableRecord(resolvables)
 
   // no need to call `triggerResolve` on the server, all the resources should already be in
   // the host API client's cache (populated from the snapshot's cache)
   useEffect(() => {
-    props.triggerResolve()
-  }, [props])
+    props.triggerResolve().catch(err => {
+      console.log(`Error while resolving props for element ${elementKey}`, err)
+    })
```
agurtovoy, line 86: `props.triggerResolve()` is async, handle any potential errors.

bc-drew-cutchins (approved):
For my own learning: why was this coupling undesired in this case?

agurtovoy:
@bc-drew-cutchins Great question. I'm using `useResolvedProps` to determine which element props have changed while user is editing an RSC component. Most of the prop resolution logic resides in a control instance's `resolveValue` method, and I was playing with the idea of proxying that call, and the implicit dependency on `useControlInstances` was getting in the way. I ultimately took a different approach to the larger problem, but decided to keep the refactoring as it makes the code more amendable to future experiments along these lines.

bc-drew-cutchins:
> @bc-drew-cutchins Great question. I'm using `useResolvedProps` to determine which element props have changed while user is editing an RSC component. Most of the prop resolution logic resides in a control instance's `resolveValue` method, and I was playing with the idea of proxying that call, and the implicit dependency on `useControlInstances` was getting in the way. I ultimately took a different approach to the larger problem, but decided to keep the refactoring as it makes the code more amendable to future experiments along these lines.

That makes sense, thanks!

PR #1372, opened 2026-07-21
fix(runtime): resolve cyclic module dependencies in state action modules

agurtovoy:
When we [split](https://github.com/makeswift/makeswift/pull/1181) the monolithic `actions.ts` file into separate shared, internal, builder, and host API modules, we inadvertently introduced cyclic dependencies among the latter (e.g. `read-write-actions.ts` imports `actions.ts` to get the combined action type, which in turn imports all of the above individual modules, one of which imports `read-write-actions.ts`). So far, this has not caused us issues, but [it's starting to](https://github.com/makeswift/makeswift/actions/runs/29868413520/job/88762275758?pr=1371).

This PR breaks the cycles by moving all action types into their own modules. No semantic changes.


packages/runtime/src/builder/host-to-builder-actions.ts
```diff
@@ -4,13 +4,14 @@ import { isAction } from '@reduxjs/toolkit'
 // and in the builder; they have been since split into two separate action types,
 // REGISTER_COMPONENT and REGISTER_BUILDER_COMPONENT, but we still have to expose what
 // are now internal actions to the builder for backward compatibility with the older runtimes
+import { ReadOnlyActionTypes } from '../state/actions/internal/read-only-action-types'
```
agurtovoy, line 7: Here and below, move action types constants into their own modules, adjust the imports correspondingly.

No semantic changes.

packages/runtime/src/state/actions/internal/index.ts
```diff
@@ -1,9 +1,4 @@
-import { type ReadOnlyAction, ReadOnlyActionTypes } from './read-only-actions'
-import { type ReadWriteAction, ReadWriteActionTypes } from './read-write-actions'
-
-export const InternalActionTypes = {
-  ...ReadOnlyActionTypes,
-  ...ReadWriteActionTypes,
-} as const
```
agurtovoy, line 7: Remove the action types export from the index file.

arvinpoddar (approved):

PR #1373, opened 2026-07-22
fix(runtime): register documents on render

agurtovoy:


agurtovoy:
This is ill-conceived in the current architecture. Dispatching a Redux action during render updates the store and notifies subscribed components while React may still be rendering the tree, which in turn triggers an error about updating one component while rendering another.

PR #1374, opened 2026-07-23
chore: upgrade Next 15 dependencies to LTS

arvinpoddar:
Upgrade Next.js dependencies to July Security Releases. See here: https://nextjs.org/blog/july-2026-security-release


examples/editable-regions/package-lock.json
arvinpoddar, line 1: Our examples aren't supposed to include a `package-lock.json`, so I dropped it from this one.

Codeseph (approved):

PR #1376, opened 2026-07-24
refactor(runtime): move prop controllers to read-write state

agurtovoy:
The prop controller reducer and actions were misclassified due to my incomplete understanding of their registration mechanism when the read-only/read-write state/actions/reducers distinction was [introduced](https://github.com/makeswift/makeswift/pull/1218): prop controllers are registered in the state as a part of the `REGISTER_COMPONENT_HANDLE` action logic in `propControllerHandlesMiddleware`, which is read-write middleware and is itself [only registered](https://github.com/makeswift/makeswift/blob/f0ce0a4a5238d83e227c900c880dd2ca822a4ffe/packages/runtime/src/state/middleware/read-write/index.ts#L13) when we [switch the store to the read-write state](https://github.com/makeswift/makeswift/blob/f0ce0a4a5238d83e227c900c880dd2ca822a4ffe/packages/runtime/src/state/store.ts#L220).


packages/runtime/src/components/builtin/Form/Form.tsx
```diff
@@ -33,7 +33,7 @@ import Button from '../Button'
 import type { TableColumn } from './types'
 import { Link } from '../../shared/Link'
 import { getBox, type BoxModelHandle } from '../../../state/modules/read-write/box-models'
-import { PropControllersHandle } from '../../../state/modules/prop-controller-handles'
+import { PropControllersHandle } from '../../../state/modules/read-write/prop-controllers'
```
agurtovoy, line 36: Here and below, adjust imports to the new module location and/or file name.

packages/runtime/src/runtimes/react/hooks/use-control-instances.ts
```diff
@@ -1,14 +1,15 @@
...
+import { isReadWriteState } from '../../../state/unified-state'
 import { useDocumentKey } from './use-document-context'
 import { useSelector } from './use-selector'
 
 export function useControlInstances(elementKey: string): Record<string, ControlInstance> | null {
   const documentKey = useDocumentKey()
 
   return useSelector(state => {
-    if (documentKey == null) return null
+    if (documentKey == null || !isReadWriteState(state)) return null
```
agurtovoy, line 12: Added explicit "write state" guard to prevent a `getPropControllers` call on the read-only state below.

packages/runtime/src/state/__tests__/get-translatable-content.test.ts
```diff
@@ -6,7 +6,10 @@ import {
 } from '@makeswift/controls'
 
 import { getTranslatableContent } from '../translations/get'
-import { DescriptorsByComponentType, DescriptorsByProp } from '../modules/prop-controllers'
+import {
+  DescriptorsByComponentType,
+  DescriptorsByProp,
+} from '../modules/prop-controller-descriptors'
```
agurtovoy, line 12: Renamed the module containing the (read-only) "descriptors" state to have `descriptors` in its name; at some point we should unify our terminology here (descriptors vs control definitions), leaving this as-is for the time being.

packages/runtime/src/state/actions/internal/read-only-action-types.ts
```diff
@@ -8,13 +8,6 @@ export const ReadOnlyActionTypes = {
 
   REGISTER_COMPONENT: 'REGISTER_COMPONENT',
   UNREGISTER_COMPONENT: 'UNREGISTER_COMPONENT',
-  REGISTER_COMPONENT_HANDLE: 'REGISTER_COMPONENT_HANDLE',
-  UNREGISTER_COMPONENT_HANDLE: 'UNREGISTER_COMPONENT_HANDLE',
-
-  REGISTER_PROP_CONTROLLERS: 'REGISTER_PROP_CONTROLLERS',
-  UNREGISTER_PROP_CONTROLLERS: 'UNREGISTER_PROP_CONTROLLERS',
-  REGISTER_PROP_CONTROLLERS_HANDLE: 'REGISTER_PROP_CONTROLLERS_HANDLE',
-  UNREGISTER_PROP_CONTROLLERS_HANDLE: 'UNREGISTER_PROP_CONTROLLERS_HANDLE',
```
agurtovoy, line 11: Here and below, moved the corresponding type and action definitions to read-write actions.

packages/runtime/src/state/read-only-state.ts
```diff
@@ -44,8 +43,7 @@ export const reducers = {
   elementTrees: ElementTrees.reducer,
   reactComponents: ReactComponents.reducer,
   componentsMeta: ComponentsMeta.reducer,
-  propControllers: PropControllers.reducer,
-  propControllerHandles: PropControllerHandles.reducer,
```
agurtovoy, line 48: Removed the reducers for prop controllers and their handles from the read-only state.

packages/runtime/src/state/read-only-state.ts
```diff
@@ -117,66 +114,26 @@ export function getComponentMeta(state: State, type: string): ComponentsMeta.Com
...
-  }: { documentKey: string; elementKey: string; propName: string },
-) {
-  return PropControllerHandles.getPropController(
-    getPropControllerHandlesStateSlice(state),
-    documentKey,
-    elementKey,
-    propName,
-  )
-}
-
```
agurtovoy, line 179: Moved state accessors to the read-write state.

arvinpoddar (approved):
Looks good 🙏. At some point we should document (probably in code, as comments) what each of these pieces of state/actions are needed for

agurtovoy:
> At some point we should document (probably in code, as comments) what each of these pieces of state/actions are needed for

@arvinpoddar Agreed 100%.

PR #1377, opened 2026-07-24
fix(runtime): don't register elements when we're not in the builder

agurtovoy:
Register elements in the read-write state only if we're in the read-write state. Sister PR to https://github.com/makeswift/makeswift/pull/1376. 

Tested performance impact on the `makeswift-nextjs-app-router` deployment.

Before:
before

After:
after

### DIFF
--- a/packages/runtime/src/runtimes/react/components/Element.tsx
+++ b/packages/runtime/src/runtimes/react/components/Element.tsx
@@ -1,17 +1,33 @@
 'use client'
 
-import { forwardRef, memo, Ref, useCallback, useImperativeHandle, useRef, ReactNode } from 'react'
+import {
+  type ReactNode,
+  type Ref,
+  type PropsWithChildren,
+  forwardRef,
+  memo,
+  useCallback,
+  useImperativeHandle,
+  useRef,
+  lazy,
+} from 'react'
+
 import {
   isElementReference,
   type Element as ElementDataOrRef,
 } from '../../../state/read-only-state'
-import { ElementRegistration } from './ElementRegistration'
-import { ElementReference } from './ElementReference'
-import { ElementData } from './ElementData'
-import { ElementImperativeHandle } from '../element-imperative-handle'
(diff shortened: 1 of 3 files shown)


packages/runtime/src/runtimes/react/components/Element.tsx
```diff
@@ -1,18 +1,34 @@
...
+
 import { FallbackComponent } from '../../../components/shared/FallbackComponent'
 import { ErrorBoundary } from '../../../components/shared/ErrorBoundary'
 
+import { useIsReadOnly } from '../hooks/use-is-read-only'
+import { ElementImperativeHandle } from '../element-imperative-handle'
+import { FindDomNode } from '../find-dom-node'
+
+import { ElementReference } from './ElementReference'
+import { ElementData } from './ElementData'
```
agurtovoy, line 28: Imports reorg.

packages/runtime/src/runtimes/react/components/Element.tsx
```diff
@@ -1,18 +1,34 @@
...
+
+const BuilderElementRegistration = lazy(() => import('./ElementRegistration'))
```
agurtovoy, line 30: Lazy import `ElementRegistration`

packages/runtime/src/runtimes/react/components/Element.tsx
```diff
@@ -39,6 +55,8 @@ export const Element = memo(
 
     useImperativeHandle(ref, () => imperativeHandleRef.current, [])
 
+    const ElementRegistration = useIsReadOnly() ? NoOp : BuilderElementRegistration
```
agurtovoy, line 58: Don't import or run element registration effects unless we're in the read-write state. The effects will effectively be no-ops in the read-only state anyway, but there is no reason to load their code or spent cycles dispatching them, and drawing a clear boundary between read-only and read-write states is always a plus.

bc-drew-cutchins (approved):
Seems like a big win! Great stuff!

PR #1371, opened 2026-07-21
refactor: make element prop controllers available on first render

agurtovoy:
Element prop controller instances were previously both created and registered in the Redux state in an effect, and therefore were unavailable during an element's initial render. 

Practically, this wasn't felt because other related state such as prop controller descriptors was (and is) also not available during the initial render, and we didn't have code like the following that would exercise this particular path:

```ts
const prevInstances = useRef(instances)
```

That said, this did lead to an additional unnecessary rerender before React settled on the final props.

Fixed by separating prop controller creation from prop controller registration. Controllers are now created on first render, registered and unregistered in an effect, and accessed through context rather than Redux state when resolving props during render.

Before:
before

After:
after

Best reviewed commit-by-commit.

PR #1380, opened 2026-07-29
test: add tests for all aspects of element registration

agurtovoy:
Includes a fix for a theoretical read-write actions dispatch bug that was making the tests flaky. Prep for https://github.com/makeswift/makeswift/pull/1371.


packages/runtime/src/runtimes/react/components/__tests__/element-registration.test.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+function TestComponent({ className }: { className: string }) {
+  return (
+    <div data-testid={TEST_ID} className={className}>
+      This is a test
+    </div>
+  )
+}
+
+describe('Element registration', () => {
```
agurtovoy, line 46: Comprehensive tests for [element registration](https://github.com/makeswift/makeswift/blob/1e6c66c5b660d7a1f115bf7431202f45cca58aa5/packages/runtime/src/runtimes/react/components/ElementRegistration.tsx#L17) covering all four aspects:
1. Creating and registering element prop controllers
2. Registering the element prop controllers handle
3. Dispatching the `MOUNT_COMPONENT` action to the builder
4. Not registering prop controllers or a handle for an element reference, and not registering or dispatching anything for the referenced global component.

packages/runtime/src/runtimes/react/components/__tests__/element-registration.test.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+    // make sure we actually rendered the referenced global component
+    const renderedGlobalElement = screen.getByTestId(TEST_ID)
+    expect(renderedGlobalElement.innerHTML).toEqual('This is a test')
+
+    const state = runtime.getOrCreateStore({ siteVersion, locale: undefined }).getState()
+    expectReadWriteState(state)
+
+    expect(getPropControllers(state, fullElementKey)).toBe(null)
+    expect(getPropControllersHandle(state, fullElementKey)).toBe(null)
```
agurtovoy, line 156: Note that these are `null` not because of the [`isRegisterElementDisabled` check](https://github.com/makeswift/makeswift/blob/1e6c66c5b660d7a1f115bf7431202f45cca58aa5/packages/runtime/src/runtimes/react/components/ElementRegistration.tsx#L28), which will return `false` for the element registration of the element reference itself, but because `propControllerHandlesMiddleware` includes an explicit check for element references: https://github.com/makeswift/makeswift/blob/1e6c66c5b660d7a1f115bf7431202f45cca58aa5/packages/runtime/src/state/middleware/read-write/prop-controller-handles.ts#L81

packages/runtime/src/runtimes/react/components/__tests__/element-registration.test.tsx
```diff
@@ -0,0 +1,193 @@
...
+      getPropControllersHandle(state, {
+        ...fullElementKey,
+        elementKey: globalElement.data.key,
+      }),
+    ).toBe(null)
+  })
+
+  test.each([false, true])(
+    'dispatches MOUNT_COMPONENT to the builder when editing (element reference: %s)',
+    async elementReference => {
```
agurtovoy, line 167: `MOUNT_COMPONENT` is dispatched regardless of whether an element is a reference (but is not dispatched for the nested elements of the global component tree)

packages/runtime/src/state/actions/internal/read-write-action-types.ts
```diff
@@ -1,4 +1,6 @@
 export const ReadWriteActionTypes = {
+  FLUSH_ACTION_BUFFER: 'FLUSH_ACTION_BUFFER',
```
agurtovoy, line 2: Part of the bug fix for `conditionalReadWriteMiddleware` below: https://github.com/makeswift/makeswift/pull/1380/changes#diff-9482ad43d318df9564624a2db405fb25650252a5cb8b85497f3f5397318f9530R231

packages/runtime/src/state/store.ts
```diff
@@ -118,6 +119,7 @@ export function conditionalReadWriteMiddleware(
       }
 
       if (state.isReadOnly) {
+        // `middlewareRef` is set, but we're still somehow in a read-only state
```
agurtovoy, line 122: Add a clarifying comment.

packages/runtime/src/state/store.ts
```diff
@@ -162,7 +164,7 @@ export interface ReadWriteStateMixin {
   loadReadWriteStateIfNeeded(): Promise<() => void>
 }
 
-function withMixin<M extends {}>(mixin: M): StoreEnhancer<M> {
+function withMixin<M extends object>(mixin: M): StoreEnhancer<M> {
```
agurtovoy, line 167: Fix ESLint warning.

packages/runtime/src/state/store.ts
```diff
@@ -222,6 +224,12 @@ export function configureReadWriteStore({
       const dispatch = store.dispatch as ReadWriteDispatch
       const builderProxyCleanup = dispatch(setupBuilderProxy(builderProxy))
 
+      // trigger a flush of buffered actions in `conditionalReadWriteMiddleware`;
+      // the `dispatch(setupBuilderProxy(builderProxy))` call above is not sufficient
+      // as it dispatches a thunk action, which is intercepted and handled by the
+      // default thunk middleware and never reaches `conditionalReadWriteMiddleware`
+      dispatch(flushActionBuffer())
```
agurtovoy, line 231: Theoretically, if no other actions are dispatched after switching the store to the read-write state, the actions that `conditionalReadWriteMiddleware` buffered while we were lazy-loading the read-write reducers and middleware will never be dispatched. It's never an issue in practice because even the simplest user action such moving the cursor over the page triggers a flood of actions, but this *is* a bug and it was causing flakiness in the tests introduced in this PR.

Technically dispatching any no-op action here, e.g. `dispatch(setIsReadOnly(false))`, would suffice, but for clarity I opted to introduce a new action specifically for this purpose.

packages/runtime/src/testing/element-data.ts
```diff
@@ -6,10 +6,10 @@ import {
   type MakeswiftComponentSnapshot,
 } from '../client'
 import { CacheData } from '../api/api-resources-client'
-import { type ElementData } from '../state/read-only-state'
+import { type Element, type ElementData } from '../state/read-only-state'
 import { MakeswiftComponentType } from '../components/builtin/constants'
 
-export function createRootComponent(elements: ElementData[], rootId?: string) {
+export function createRootComponent(elements: Element[], rootId?: string) {
```
agurtovoy, line 12: Here and below, adjust test helpers to accept element references.

packages/runtime/src/testing/read-write-state.ts
```diff
@@ -0,0 +1,5 @@
+import { type State, type ReadWriteState, isReadWriteState } from '../state/unified-state'
+
+export function expectReadWriteState(state: State): asserts state is ReadWriteState {
+  expect(isReadWriteState(state)).toBe(true)
+}
```
agurtovoy, line 5: Added a new helper for read-write state tests.

arvinpoddar (dismissed):
Looks good, quite the find

packages/runtime/src/runtimes/react/components/__tests__/element-registration.test.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+    const state = runtime.getOrCreateStore({ siteVersion, locale: undefined }).getState()
+    expect(isReadWriteState(state) ? getPropControllers(state, fullElementKey) : null).toBe(null)
+    expect(isReadWriteState(state) ? getPropControllersHandle(state, fullElementKey) : null).toBe(
+      null,
+    )
+    expect(BuilderAPIProxyExecute).not.toHaveBeenCalled()
+  })
+
+  test('registers element prop controllers and handles when editing', async () => {
```
Codeseph: I was slightly confused by the names of the tests, which seem to use "element" to refer strictly to "element data", whereas [elsewhere](https://github.com/makeswift/makeswift/blob/1e6c66c5b660d7a1f115bf7431202f45cca58aa5/packages/controls/src/common/schema.ts#L40) (including in the [fixture](https://github.com/makeswift/makeswift/pull/1380/changes#diff-039d58e6c7477e622778b37a645873bd49d5e2f5c3ee84ab2c1ca84cd721cf2eR76)) "element" means "either element data or an element reference". In this case I understood the intent of the test by contrasting it with the one below, and by noticing the lack of a parameter being passed to `createFixtures`.

I guess the "elements versus references" wording makes sense, it's just not what I'm used to

packages/runtime/src/runtimes/react/components/__tests__/element-registration.test.tsx
Codeseph: I'm trying to understand (3) and (4).

I imagine these are related to how global element editing works (requires clicking to "focus" a global element, which leads to registering a new document in which the global element's actual data is the root element, I believe).

So:

Is the behavior in (4) intended because the global element is not "focused" for editing? Which we know because it appears as a reference to begin with?

What should I know about (3) in the case of element references? In the builder, the preview frame's handling of `HostToBuilderActionTypes.MOUNT_COMPONENT` is [here](https://github.com/makeswift/cosmos/blob/c829b652cd40cc55eee98ab9e9f7433f1c6f67e0/packages/leo/src/legacy/scenes/Builder/components/Canvas/components/PreviewFrame/index.tsx#L385) and seems to lead to some sort of prop controller registration on that side of the house. I would guess that's related to required setup for being able to focus the global element, but I don't know.

I don't mind looking into these on my own, just pointing out what I don't know here

Codeseph (dismissed):

packages/runtime/src/runtimes/react/components/__tests__/element-registration.test.tsx
agurtovoy: @Codeseph Good callout! I personally find using the term "element data" for the primary case of a directly rendered element awkward, and so in my head it's "elements versus references"; that said, the former is our official terminology right now, so we should keep the test names consistent, especially if my shorthand naming is causing confusion. Fixed.

packages/runtime/src/runtimes/react/components/__tests__/element-registration.test.tsx
agurtovoy: > I imagine these are related to how global element editing works (requires clicking to "focus" a global element, which leads to registering a new document in which the global element's actual data is the root element, I believe).

Yes.

> Is the behavior in (4) intended because the global element is not "focused" for editing? Which we know because it appears as a reference to begin with?

Correct. It's worth noting though that before writing these tests/refactoring the corresponding code, I was confused about the mechanism of how this happens. I thought prop controller registration for the element references themselves was suppressed by the [`isRegisterElementDisabled` checks in `ElementRegistration`](https://github.com/makeswift/makeswift/blob/1e6c66c5b660d7a1f115bf7431202f45cca58aa5/packages/runtime/src/runtimes/react/components/ElementRegistration.tsx#L25), but that's not the case because the `ElementRegistration` instance for the reference is not wrapped in `DisableRegisterElement.Provider`; only the nested elements under the reference are. Instead, reference registration is suppressed by this explicit check `propControllerHandlesMiddleware`: https://github.com/makeswift/makeswift/blob/1e6c66c5b660d7a1f115bf7431202f45cca58aa5/packages/runtime/src/state/middleware/read-write/prop-controller-handles.ts#L81

> What should I know about (3) in the case of element references? In the builder, the preview frame's handling of HostToBuilderActionTypes.MOUNT_COMPONENT is [here](https://github.com/makeswift/cosmos/blob/c829b652cd40cc55eee98ab9e9f7433f1c6f67e0/packages/leo/src/legacy/scenes/Builder/components/Canvas/components/PreviewFrame/index.tsx#L385) and seems to lead to some sort of prop controller registration on that side of the house. I would guess that's related to required setup for being able to focus the global element, but I don't know.

Similarly to my misconceptions about reference registration, I actually didn't know that `MOUNT_COMPONENT` for references was dispatched to the builder until I started writing these tests; I thought it was suppressed by the `isRegisterElementDisabled` check. It's not though, and it is being dispatched, which makes sense since the builder does need to know about the component existence in order to show any sort of UX for it. 

The `builder.registerPropControllers` call that you're referring to is a red herring though; if you look inside, you'll see that it has [the same check as `propControllerHandlesMiddleware`](https://github.com/makeswift/cosmos/blob/a7d6672213bd2cfb92b8f83c8deeb599e1b3a2a7/packages/leo/src/legacy/builder/actions.ts#L627)

Codeseph: Thanks 🙏

PR #1382, opened 2026-07-31
refactor: formalize `ControlInstance<any>` as `AnyControlInstance`

agurtovoy:
Formalize `ControlInstance<any>` as `AnyControlInstance` to avoid having to suppress warnings about `any` throughout the codebase.

### DIFF
--- a/.changeset/free-camels-lead.md
+++ b/.changeset/free-camels-lead.md
@@ -1,0 +1,5 @@
+---
+'@makeswift/controls': patch
+---
+
+Formalize `ControlInstance<any>` as `AnyControlInstance` to avoid having to suppress warnings about `any` throughout the codebase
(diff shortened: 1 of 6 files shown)


arvinpoddar (approved):

bc-drew-cutchins (commented):
Would `type AnyControlInstance = ControlInstance<unknown>` be a safer alternative? Not sure how these typically get consumed.

agurtovoy:
@bc-drew-cutchins No, unfortunately `unknown` doesn't cut here. Control instance is defined as `ControlInstance<M extends ControlMessage>`, and `unknown` doesn't satisfy this constraint. At some point we need to take a second look at how we type control messages and whether there is a better way to achieve the desired type safety w/o paying the price in complexity we're paying now, but that's a journey for some other time. This is just a chore PR formalizing our existing practice and following up on https://github.com/makeswift/makeswift/pull/1354.

bc-drew-cutchins:
@agurtovoy Ah, gotcha, that makes sense 👍

PR #1383, opened 2026-07-31
fix(runtime): Buffer unmount and mount actions within a commit

bc-drew-cutchins:
This PR contains a potential fix for the flicker in the elements panel occurring whenever an element is moved.

## Behavior
Note below how the Lime element label in the panel appears, disappears, then reappears at its new position after being dragged.

https://github.com/user-attachments/assets/74e13d6a-699d-47e7-a2cd-322392de7ad2

Note that the CPU is throttled with a 4x slowdown to make the flicker obvious in the above video.

## Root Cause
The element tree updates optimistically when you drop, so getElementChildren in the builder reflects the new position instantly.

However, the appearance of these elements in the elements panel is controlled by a map tracking which elements are actually rendered in the DOM. This is updated in response to the iframe sending mount and unmount actions to the builder, [triggered by react effects in the mounted/unmounted elements](https://github.com/makeswift/makeswift/blob/2532e7af97d0cce22c7082521ff44909cc1905f6/packages/runtime/src/runtimes/react/components/ElementRegistration.tsx#L72).

After dropping, the following sequence occurs:
1. The element label appears at the new position after the element tree is updated, while its mount flag from before the move is still true.
2. The element label disappears as the element is unmounted from its old position in the DOM, causing the react cleanup effect to run, triggering an UNMOUNT_COMPONENT event. This removes the element from the mounted element set in the builder, causing the element to be filtered from the panel.
3. The element reappears as the element is mounted at its new position, causing the react render effect to run, triggering a MOUNT_COMPONENT event. This re-adds the element to the mounted element set in the builder, causing the element to appear in the panel again.

## Solution
This PR implements a `BuilderMountChangeActionBuffer`, that queues a batch of MOUNT and UNMOUNT actions from a single react commit and removes pairs of these events before sending the actions to the builder. This ensures that whenever a component is mounted and unmounted within the same commit, the builder will not receive MOUNT and UNMOUNT events that result in no net state change.

See below how the elements panel is stable after moving the element in the canvas.

https://github.com/user-attachments/assets/57ab22c4-0211-4865-b53a-b397dc64222b

## Limitations
The main purpose of this PR is to prove the root cause of this bug, but I'm not sure if this solution should actually be used as it comes with a few drawbacks:
1. Given this fix is implemented in the runtime, it will not address the bug in older versions of the runtime. A fix in the builder would instead give all users a fix without requiring a runtime upgrade. At this time though, I cannot figure out how this might be addressed in the builder without some sort of temporal batching of MOUNT/UNMOUNT events that would cause a lag in processing non paired mount changes.
2. MOUNT/UNMOUNT pairs from a move operation can only be correlated if their actions are dispatched within the same commit. Any deferred rendering caused by Suspense boundaries or transitions would cause the actions to be emitted during separate commits. I believe this could happen with a custom component that wraps its children inside a `<Suspense>`.
3. This potentially breaks the semantics of MOUNT/UNMOUNT events, as we can no longer depend on them to be fired whenever an element is mounted in a new location. The existing logic in the builder does not use these events for any sort of logic relative to the elements position in the element tree, but this could be a potential footgun in the future.


packages/runtime/src/runtimes/react/components/RuntimeProvider.tsx
```diff
@@ -36,6 +37,13 @@ export function RuntimeProvider({
     return () => runtime.releaseStore({ siteVersion, locale }, store)
   }, [runtime, siteVersion, locale, store])
 
+  // Flushing after every commit ensures mount/unmount actions dispatched by effects
+  // within this commit are processed together.
+  // See BuilderMountChangeActionBuffer for more context.
+  useEffect(() => {
+    store.dispatch(flushMountChangeActions())
+  })
```
agurtovoy: Directionally, I like the idea of cancelling out matching mount/unmount pairs as they are effectively noise and carry no useful information for the builder. That said, flushing on a parent's commit does not cover every situation that can lead to a component mount, such as an update triggered by a click or `setTimeout` in a deeper ancestor component. This may work accidentally due to unrelated actions being dispatched and triggering `RuntimeProvider` rerendering, but it's not something we want to rely on.

IMHO this is the main issue we need to figure out before we can proceed with this approach.

packages/runtime/src/state/middleware/read-write/builder-api/index.ts
```diff
@@ -11,19 +11,30 @@ import { HostActionTypes } from '../../../host-api'
...
   return actionMiddleware(({ dispatch }) => next => {
     if (typeof window === 'undefined') return (action: Action) => next(action)
 
     let cleanUp = () => {}
     return (action: Action) => {
       switch (action.type) {
-        case BuilderActionTypes.CHANGE_ELEMENT_BOX_MODELS:
         case BuilderActionTypes.MOUNT_COMPONENT:
         case BuilderActionTypes.UNMOUNT_COMPONENT:
+          mountChangeBuffer.enqueue(action)
```
agurtovoy: What if we flushed in a [microtask](https://developer.mozilla.org/en-US/docs/Web/API/HTML_DOM_API/Microtask_guide) that we scheduled on `MOUNT_COMPONENT`?

packages/runtime/src/state/middleware/read-write/builder-api/builder-mount-change-action-buffer.ts
```diff
@@ -0,0 +1,93 @@
...
+}
+
+/**
+ * Removes pairs of MOUNT_COMPONENT and UNMOUNT_COMPONENT actions from a batch of MountChangeActions.
+ */
+function removeMountUnmountPairs(batch: MountChangeAction[]): MountChangeAction[] {
+  // A positive number of mountsPerElement[key] indicates a net amount of MOUNT_COMPONENTs
+  // A negative number of mountsPerElement[key] indicates a net amount of UNMOUNT_COMPONENTs
+  const mountsPerElement = batch.reduce((mounts, action) => {
+    if (action.type == BuilderActionTypes.MOUNT_COMPONENT) {
```
agurtovoy: Nit: we use `===` for everything but `null` comparisons; should probably have a lint rule for this.

packages/runtime/src/state/middleware/read-write/builder-api/builder-mount-change-action-buffer.ts
```diff
@@ -0,0 +1,93 @@
...
+  return `${a.payload.documentKey}:${a.payload.elementKey}`
+}
+
+/**
+ * Removes pairs of MOUNT_COMPONENT and UNMOUNT_COMPONENT actions from a batch of MountChangeActions.
+ */
+function removeMountUnmountPairs(batch: MountChangeAction[]): MountChangeAction[] {
+  // A positive number of mountsPerElement[key] indicates a net amount of MOUNT_COMPONENTs
+  // A negative number of mountsPerElement[key] indicates a net amount of UNMOUNT_COMPONENTs
+  const mountsPerElement = batch.reduce((mounts, action) => {
```
agurtovoy: Use of `reduce` here is IMHO misleading: it suggests an accumulated result, but the callback mutates and returns the same `Map` on every iteration. `forEach` would express what's actually happening here more accurately.

agurtovoy: Maybe something like this?

```ts
const mountsPerElement = new Map<string, number>()
batch.forEach(action => {
  const actionKey = key(action)
  const current = mountsPerElement.get(actionKey) ?? 0
  mountsPerElement.set(
    actionKey,
    action.type === BuilderActionTypes.MOUNT_COMPONENT ? current + 1 : current - 1,
  )
})
```

packages/runtime/src/state/middleware/read-write/builder-api/mount-change-action-buffer.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    if (action.type == BuilderActionTypes.UNMOUNT_COMPONENT && mountsRemaining < 0) {
+      coalescedBatch.push(action)
+      mountsPerElement.set(key(action), mountsRemaining + 1)
+      continue
+    }
+  }
+
+  return coalescedBatch
+}
```
agurtovoy, line 87: After looking at this part, I'd rewrite the whole thing as this:

```ts
  const mountsPerElement = new Map<
    string,
    { count: number; payload: MountChangeAction['payload'] }
  >()

  batch.forEach(action => {
    const actionKey = key(action)
    const currentCount = mountsPerElement.get(actionKey)?.count ?? 0
    mountsPerElement.set(actionKey, {
      count:
        action.type === BuilderActionTypes.MOUNT_COMPONENT ? currentCount + 1 : currentCount - 1,
      payload: action.payload,
    })
  })

  return [...mountsPerElement.values()]
    .filter(({ count }) => count != 0)
    .map(({ count, payload }) => ({
      type: count > 0 ? BuilderActionTypes.MOUNT_COMPONENT : BuilderActionTypes.UNMOUNT_COMPONENT,
      payload,
    }))
```

We're basing this whole approach on the idea that extra mount/unmount events carry no useful information for the builder, so there is no reason to replay more than one unmatched event per element.

packages/runtime/src/state/middleware/read-write/builder-api/builder-mount-change-action-buffer.ts
bc-drew-cutchins: Ah, fixed!

packages/runtime/src/state/middleware/read-write/builder-api/builder-mount-change-action-buffer.ts
bc-drew-cutchins: Interesting, I was under the impression that this was an idiomatic use of `reduce` since this is an accumulated result (right?)  and simply viewed the mutation/return of the map as an optimization over creating a copy of the map.

I'm happy to defer to your judgement here, updated.

packages/runtime/src/state/middleware/read-write/builder-api/mount-change-action-buffer.ts
bc-drew-cutchins: Ah yep that's fair, fixed!

packages/runtime/src/runtimes/react/components/RuntimeProvider.tsx
bc-drew-cutchins: Ah right, not sure how I missed this 🤦. I actually used queueMicrotask in my first pass at this and then went with the top level useEffect thinking it was sufficient. 

Refactored to use `queueMicrotask` and confirmed it also fixes the issue!

packages/runtime/src/state/middleware/read-write/builder-api/index.ts
bc-drew-cutchins: Updated.

packages/runtime/src/state/middleware/read-write/builder-api/builder-mount-change-action-buffer.ts
agurtovoy: > Interesting, I was under the impression that this was an idiomatic use of reduce since this is an accumulated result (right?) and simply viewed the mutation/return of the map as an optimization over creating a copy of the map.

People definitely do that, but I personally don't consider it to be idiomatic; for me, `reduce` is first and foremost a functional algorithm, and the mix of `mounts.set` calls with `return mounts` at the end of the callback triggers me, haha. If I modify state in place, I'd rather be explicit about it and use a `forEach` or a plain loop.

packages/runtime/src/state/middleware/read-write/builder-api/mount-change-action-buffer.ts
```diff
@@ -0,0 +1,93 @@
...
+  return `${a.payload.documentKey}:${a.payload.elementKey}`
+}
+
+/**
+ * Coalesces a batch of MountChangeActions into an array of either 0 or 1 MountChangeActions corresponding
+ * to the net result of the batch.
+ *
+ * Removes pairs of MOUNT_COMPONENT and UNMOUNT_COMPONENT actions, if there are actions leftover it returns a single
+ * action corresponding to the leftover type.
+ */
```
agurtovoy, line 64: I'd shorten the comment to just the last sentence; it explains what this function does perfectly.

packages/runtime/src/state/middleware/read-write/builder-api/builder-mount-change-action-buffer.ts
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * Note that it is safe to remove these actions pairs as these actions do not communicate
+ * the location of a component's mount but simply whether it is currently mounted.
+ */
+export class BuilderMountChangeActionBuffer {
+  private readonly builderProxy: BuilderAPIProxy
+  private scheduledActions: MountChangeAction[]
+  private flushScheduled: boolean
+
+  constructor(builderProxy: BuilderAPIProxy) {
```
agurtovoy: I'd decouple this class from knowing about `BuilderAPIProxy` and just accept an `execute` callback:

```ts
  constructor(private readonly execute: (action: MountChangeAction) => void) {
    this.queuedActions = []
  }

```

packages/runtime/src/state/middleware/read-write/builder-api/builder-mount-change-action-buffer.ts
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * Without this buffer, moved elements will result in separate UNMOUNT and MOUNT
+ * actions being sent to the builder at the time of the elements unmount at its original
+ * location and remount at its new location. This causes a flicker in the builder as
+ * the element is considered unmounted in the time between its UNMOUNT and MOUNT actions.
+ *
+ * Note that it is safe to remove these actions pairs as these actions do not communicate
+ * the location of a component's mount but simply whether it is currently mounted.
+ */
+export class BuilderMountChangeActionBuffer {
```
agurtovoy: Related to the comment below, I'd drop `Builder` from the name here, and rename the file correspondingly. This file already lives under `builder-api`, and we try to balance the length and explicitness of the name with the context of the file location.

packages/runtime/src/state/middleware/read-write/builder-api/index.ts
```diff
@@ -11,19 +11,25 @@ import { HostActionTypes } from '../../../host-api'
...
   return actionMiddleware(({ dispatch }) => next => {
     if (typeof window === 'undefined') return (action: Action) => next(action)
 
     let cleanUp = () => {}
     return (action: Action) => {
       switch (action.type) {
-        case BuilderActionTypes.CHANGE_ELEMENT_BOX_MODELS:
         case BuilderActionTypes.MOUNT_COMPONENT:
         case BuilderActionTypes.UNMOUNT_COMPONENT:
+          mountChangeBuffer.schedule(action)
```
agurtovoy, line 29: We should add a test either for (this aspect of) the middleware or the buffer itself.

packages/runtime/src/state/middleware/read-write/builder-api/mount-change-action-buffer.ts
bc-drew-cutchins: Done!

packages/runtime/src/state/middleware/read-write/builder-api/builder-mount-change-action-buffer.ts
bc-drew-cutchins: Good call, done.

packages/runtime/src/state/middleware/read-write/builder-api/builder-mount-change-action-buffer.ts
bc-drew-cutchins: That's fair, done!

packages/runtime/src/state/middleware/read-write/builder-api/index.ts
bc-drew-cutchins: Added

packages/runtime/src/state/middleware/read-write/builder-api/__tests__/index.test.ts
agurtovoy, line 1: I'd rename the original file to `middleware.ts` and have `index.ts` re-export from it, and rename the test to `middleware.test.ts`.

packages/runtime/src/state/middleware/read-write/builder-api/__tests__/mount-change-action-buffer.test.ts
```diff
@@ -0,0 +1,42 @@
...
+    ]
+
+    actions.forEach(action => buffer.schedule(action))
+
+    await Promise.resolve()
+
+    expect(mockExecute).toHaveBeenCalledTimes(2)
+    expect(mockExecute).toHaveBeenCalledWith(mountComponent('a', '1'))
+    expect(mockExecute).toHaveBeenCalledWith(mountComponent('a', '2'))
+  })
```
agurtovoy, line 41: I'd add a test that asserts the fact that we don't call `execute` if the buffer contains only the matching pairs of mount/unmount events.

agurtovoy (dismissed):
Looks good overall, just a couple of minor comments on tests.

packages/runtime/src/state/middleware/read-write/builder-api/__tests__/middleware.test.ts
bc-drew-cutchins: Done

packages/runtime/src/state/middleware/read-write/builder-api/__tests__/mount-change-action-buffer.test.ts
bc-drew-cutchins: Added

agurtovoy (approved):

PR #1381, opened 2026-07-30
feat: RSC-enabled `Slot` prop rendering

agurtovoy:
Reworked `Slot` prop rendering to be compatible with React Server Components. See inline comments for details.

Smoke test:

https://github.com/user-attachments/assets/e99255ad-dcab-49ec-98f6-d74b09fbde7f

Best reviewed commit-by-commit and with the "Hide whitespace" toggle on.

PR #1384, opened 2026-07-31
feat: `useControlInstance` hook + tests

agurtovoy:
New hook for looking up a control instance for a specific prop from the context based on the instance key and control instance class. Following up https://github.com/makeswift/makeswift/pull/1371, prep for [RSC-enabled `Slot` prop rendering](https://github.com/makeswift/makeswift/pull/1381).

### DIFF
--- a/packages/runtime/src/runtimes/react/components/control-instances-context.tsx
+++ b/packages/runtime/src/runtimes/react/components/control-instances-context.tsx
@@ -2,18 +2,18 @@
 
 import { createContext, useContext } from 'react'
 
-import { ControlInstance } from '@makeswift/controls'
+import { type AnyControlInstance } from '@makeswift/controls'
 
 type Context = {
   elementKey: string
-  instances: Record<string, ControlInstance> | null
+  instances: Record<string, AnyControlInstance> | null
 }
 
 const ControlInstancesContext = createContext<Context | null>(null)
 
 export const ControlInstancesProvider = ControlInstancesContext.Provider
 
-export function useControlInstances(elementKey: string): Context['instances'] | null {
+export function useControlInstances(elementKey: string | undefined): Context['instances'] | null {
   const cx = useContext(ControlInstancesContext)
   return cx && cx.elementKey === elementKey ? cx.instances : null
 }
--- a/packages/runtime/src/runtimes/react/hooks/__tests__/use-control-instance.test.tsx
+++ b/packages/runtime/src/runtimes/react/hooks/__tests__/use-control-instance.test.tsx
@@ -1,0 +1,270 @@
+/** @jest-environment jsdom */
+
+import { type PropsWithChildren, act } from 'react'
+
+import '@testing-library/jest-dom'
+import { renderHook } from '@testing-library/react'
+
+import {
+  ControlInstance,
+  ResourceResolver,
+  Stylesheet,
+  type AnyControlInstance,
+} from '@makeswift/controls'
+
+import { ControlInstancesProvider } from '../../components/control-instances-context'
+import { SlotControl, Slot, GroupControl, Group, TextInput, List } from '../../../../controls'
+
+import { useControlInstance, getInstanceChainByPropPath } from '../use-control-instance'
+
+const createFixtures = ({
+  createInstances,
+}: { createInstances?: () => Record<string, AnyControlInstance> } = {}) => {
+  const elementKey = 'test-element-key'
+  const instanceKey = { elementKey, propPath: 'group.nestedGroup.slot' }
+
+  const groupProp = Group({
+    props: {
+      nestedGroup: Group({
+        props: {
+          title: TextInput(),
+          slot: Slot(),
+        },
+      }),
+    },
+  })
+
+  const instances = {
+    group: groupProp.createInstance({
+      instanceKey: { elementKey, propPath: 'group' },
+      sendMessage: () => {},
+    }),
+  }
+
+  const wrapper = ({ children }: PropsWithChildren) => {
+    return (
+      <ControlInstancesProvider
+        value={{
+          elementKey,
+          instances: createInstances ? createInstances() : instances,
+        }}
+      >
+        {children}
+      </ControlInstancesProvider>
+    )
+  }
+
+  return { elementKey, instanceKey, wrapper, instances }
+}
+
+describe('useControlInstance', () => {
+  let consoleError: jest.SpyInstance
+  beforeEach(() => {
+    consoleError = jest.spyOn(console, 'error').mockImplementation(jest.fn())
+  })
+
+  afterEach(() => {
+    consoleError.mockRestore()
+  })
+
+  describe("returns 'undefined' when no instance is found", () => {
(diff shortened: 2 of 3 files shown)


packages/runtime/src/runtimes/react/components/control-instances-context.tsx
```diff
@@ -2,18 +2,18 @@
 
 import { createContext, useContext } from 'react'
 
-import { ControlInstance } from '@makeswift/controls'
+import { type AnyControlInstance } from '@makeswift/controls'
```
agurtovoy, line 5: Updated `instances` type to be compatible with all control instances.

packages/runtime/src/runtimes/react/components/control-instances-context.tsx
```diff
@@ -2,18 +2,18 @@
...
-  instances: Record<string, ControlInstance> | null
+  instances: Record<string, AnyControlInstance> | null
 }
 
 const ControlInstancesContext = createContext<Context | null>(null)
 
 export const ControlInstancesProvider = ControlInstancesContext.Provider
 
-export function useControlInstances(elementKey: string): Context['instances'] | null {
+export function useControlInstances(elementKey: string | undefined): Context['instances'] | null {
```
agurtovoy, line 16: Allow `elementKey` to be undefined.

packages/runtime/src/runtimes/react/hooks/__tests__/use-control-instance.test.tsx
```diff
@@ -0,0 +1,145 @@
...
+      >
+        {children}
+      </ControlInstancesProvider>
+    )
+  }
+
+  return { elementKey, instanceKey, wrapper }
+}
+
+describe('useControlInstance', () => {
```
agurtovoy, line 60: Tests for the new hook.

packages/runtime/src/runtimes/react/hooks/use-control-instance.ts
```diff
@@ -0,0 +1,57 @@
...
+  const getChild = (
+    parent: ControlInstance | undefined,
+    childPath: string[],
+  ): ControlInstance | undefined =>
+    childPath.length === 0 || parent == null
+      ? parent
+      : getChild(parent.child(childPath[0]), childPath.slice(1))
+
+  return getChild(parentInstance, childPath) ?? null
+}
```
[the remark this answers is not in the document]
agurtovoy: Uhm, first of all, list prop resolution behavior has nothing to do with the code in question. If a list control doesn't have children when a nested prop is requested, well, then it doesn't have children and we return `null`.

Second of all, the "reasoning" above is circular. In order to call `useControlInstance` with a prop path pointing to a nested list item's prop, we need an instance key. Instance keys are only constructed during control instance creation, which means that by the time `useControlInstance` is called, the instance we're trying to look up already exists.

packages/runtime/src/runtimes/react/hooks/use-control-instance.ts
```diff
@@ -0,0 +1,57 @@
...
+
+import { useControlInstances } from './use-control-instances'
+
+type InstanceConstructor<T extends AnyControlInstance> = abstract new (...args: never[]) => T
+
+export function useControlInstance<T extends AnyControlInstance>(
+  instanceKey: ControlInstanceKey | undefined,
+  instanceClass: InstanceConstructor<T>,
+): T | null {
+  const { elementKey, propPath } = instanceKey ?? {}
```
bc-drew-cutchins, line 12: nit: My personal taste is to return early when handling potentially undefined values, so here (and a couple other places in this file with similar conditions) would have:
```ts
if (instanceKey == null) {
  return null
}
```
to make it immediately obvious that an undefined `instanceKey` will always return a null control instance.

I'm curious if you tend to prioritize a single point of return or find this sort of thing unhelpful?

bc-drew-cutchins (dismissed):

packages/runtime/src/runtimes/react/hooks/use-control-instance.ts
bc-drew-cutchins: Oops nevermind I see the hook called right below it 🤦

agurtovoy: Yep, that's the only reason for not returning right away here, which we do strongly prefer.

agurtovoy: I could have returned after that hook call, that might still be an improvement.

agurtovoy: > I could have returned after that hook call, that might still be an improvement.

Well, not anymore, haha

packages/runtime/src/runtimes/react/hooks/use-control-instance.ts
```diff
@@ -0,0 +1,82 @@
...
+  const subscribe = useCallback(
+    (onStateChange: () => void) => {
+      if (instances == null || propPath == null) return () => {}
+
+      const chain = getInstanceChainByPropPath(instances, propPath)
+      const unsubscribes = chain.filter(isNotNil).map(instance => instance.subscribe(onStateChange))
+      return () => unsubscribes.forEach(unsubscribe => unsubscribe())
+    },
+    [instances, propPath],
+  )
```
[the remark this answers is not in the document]
agurtovoy: Added a test for nested lists that was failing, fixed.

packages/runtime/src/runtimes/react/hooks/__tests__/use-control-instance.test.tsx
```diff
@@ -0,0 +1,268 @@
...
+        sendMessage: () => {},
+      })
+
+      const stylesheet = {
+        child: () => stylesheet,
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      } as any as Stylesheet
+
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      const resolver = {} as any as ResourceResolver
```
agurtovoy, line 181: I'd love to use `makeswift/@controls`' [mocks](https://github.com/makeswift/makeswift/tree/main/packages/controls/src/testing/mocks) here, but we are not exporting them, and properly exporting them under a separate entry point is a task on its own.

packages/runtime/src/runtimes/react/hooks/__tests__/use-control-instance.test.tsx
```diff
@@ -0,0 +1,270 @@
...
+        list,
+        listData,
+        listInstance,
+        resolveValueContext,
+      }
+    }
+
+    test('returns the nested instance once the list resolves its children', async () => {
+      // List item instances are lazily created in a side effect as a part of the list prop resolution;
+      // test that `useControlInstance` correctly picks up these child instances.
```
agurtovoy, line 196: A test specifically for list controls.

arvinpoddar (approved):
Thanks for the additional tests

PR #1386, opened 2026-07-31
feat(controls): ability to subscribe to changes to instance state

agurtovoy:
Enables us to use control instances with `useSyncExternalStore`. Prep for [RSC-enabled `Slot` prop rendering](https://github.com/makeswift/makeswift/pull/1381).

PR #1387, opened 2026-08-03
fix: infinite loop when resubscribing to list instance updates

agurtovoy:
Fix an infinite loop when resubscribing to list instance updates in response to instance update. Following up https://github.com/makeswift/makeswift/pull/1386

### DIFF
--- a/packages/controls/src/controls/list/list-control.ts
+++ b/packages/controls/src/controls/list/list-control.ts
@@ -92,7 +92,11 @@ export class ListControl<
 
     this.itemControlsById = children
     this.itemControlsList = [...children.values()]
-    this.listeners.forEach((listener) => listener())
+
+    // copy the listeners before notifying them to avoid an infinite loop when
+    // user resubscribes to updates in response to receiving one
+    const listeners = [...this.listeners.values()]
+    listeners.forEach((listener) => listener())
   }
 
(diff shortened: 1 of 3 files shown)


packages/controls/src/controls/list/list.test.ts
```diff
@@ -229,6 +229,23 @@ describe('List', () => {
       ])
     })
 
+    test("updating list instance's subscription in response to the instance's state update doesn't cause an infinite loop", async () => {
+      const { listInstance, resolveValueContext } = fixtures()
```
agurtovoy, line 233: Added a test that reproduces the issue.

packages/controls/src/controls/list/list-control.ts
```diff
@@ -92,7 +92,11 @@ export class ListControl<
 
     this.itemControlsById = children
     this.itemControlsList = [...children.values()]
-    this.listeners.forEach((listener) => listener())
+
+    // copy the listeners before notifying them to avoid an infinite loop when
+    // user resubscribes to updates in response to receiving one
+    const listeners = [...this.listeners.values()]
+    listeners.forEach((listener) => listener())
```
agurtovoy, line 99: Duh

bc-drew-cutchins (approved):

arvinpoddar (approved):

PR #1390, opened 2026-08-05
fix: unset body overflow styles to allow scrolling in builder

arvinpoddar:
Addresses a case where setting certain overflow properties on the body results in the body's contents not overflowing to the root element. The runtime then measures the root element height as the same as the window height, resulting in the builder not showing a scroll bar (even if the scroll bar was shown, the root element would also not be scrollable).

### DIFF
--- a/packages/runtime/src/state/middleware/read-write/builder-api/initialize-connection.ts
+++ b/packages/runtime/src/state/middleware/read-write/builder-api/initialize-connection.ts
@@ -100,13 +100,30 @@ function startMeasuringElements(): ThunkAction<() => void, State, unknown, Actio
 
 function lockDocumentScroll(): ThunkAction<() => void, State, unknown, Action> {
   return dispatch => {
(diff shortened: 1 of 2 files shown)


packages/runtime/src/state/middleware/read-write/builder-api/initialize-connection.ts
```diff
@@ -103,10 +103,14 @@ function lockDocumentScroll(): ThunkAction<() => void, State, unknown, Action> {
     const lastDocumentOverflow = window.document.documentElement.style.overflow
     window.document.documentElement.style.overflow = 'hidden'
 
+    const lastBodyOverflow = window.document.body.style.overflow
+    window.document.body.style.overflow = 'initial'
+
```
agurtovoy, line 121: Can we add a comment here explaining how the host/builder scrolling interaction works/why we're overriding these styles? 🙏

agurtovoy (dismissed):
Thanks for getting to the bottom of this 🙏

PR #1391, opened 2026-08-05
feat(controls): Implement POC for options context

bc-drew-cutchins:
## Overview
This PR implements a POC for controls depending on the values of other controls. 

This is implemented using a new `context` parameter passed to `getOptions` that is populated using a provided `requiredOptionsContext` configuration.

See [this cosmos PR](https://github.com/makeswift/cosmos/pull/3077) for the builder implementation enabling this feature.

This PR includes two example use cases of the new context functionality:
### Location Selector
The location selector is a minimal POC of using additional context in getOptions. The city `ComboBox` relies on the selected option in the state `ComboBox`.

https://github.com/user-attachments/assets/ed3f1aea-62eb-4148-9c5f-c89db1e6dd47

### Pokemon Selector
This is a less trivial example based on the @drewpledger's work in [this PR](https://github.com/makeswift/makeswift/pull/1357). This demonstrates dependencies on props nested inside groups and the use of `requiredOptionsContext` on a gallery control.

https://github.com/user-attachments/assets/d676f686-3045-472e-9e79-344217424e92

### Why require a `requiredOptionsContext`?
Without a user specified `requiredOptionsContext`, all options would need to be refreshed whenever any other prop value changes. This would then occur even if users had `getOptions` implementations without any dependency on other prop values. This should also be helpful in providing a small degree of type safety by typing the keys of the context as the keys of the `requiredOptionsContext`. 

The `requiredOptionsContext` could also be used for helpful runtime error reporting to display to users that their queried prop paths are not valid. Otherwise this would only become apparent from errors bubbling up when `getOptions` is run.

### Why not use resolved values?
Currently, the `context` parameter receives the prop data directly from the element tree instead of resolved values. I initially attempted to provide resolved values as context but decided against it due to the following issues:
- Resolving values is a render time feature that is coupled to impure operations such as mounting react nodes and mutating stylesheets. Further refactoring would need to occur to allow for a pure invocation of `resolveValue`.
- Resolved values are not plain data and therefore are not necessarily serializable over a `MessagePort`. This means the builder-side `ControlDefinitions` cannot be used to resolve values before calling `getOptions`. Value resolution would need to happen within the runtime.
- I think it would be reasonable in the future to allow for user provided `resolveValue` callbacks in control configurations. This could be useful for a group with a series of filtering inputs to only resolve as the value of the final selection instead of having to route intermediary values into the react component. Providing resolved values in the options context would then become less useful as it would be impossible for controls in this group to inspect the inputs of other controls in this group.

However, I think requiring the user to directly navigate the structure of the element tree provides an unintuitive api. For example, when querying the value of a prop inside of a `Group`, the user must provide the path `"GROUP_NAME.value.NESTED_PROP_NAME"`, and then when consuming the value of that prop inside their `getOptions` implementation, they must again unwrap the `.value` to access the actual input (as opposed to the `.id` or `.label`).

To improve this, I think it could be useful to add a new `resolveInputValue` method to `ControlDefinition` that would return the unwrapped user input to the control. This would allow for the user to more intuitively navigate the structure of the component props.

### Can the type checking/inference be improved?
Unfortunately the paths provided in `requiredOptionsContext` cannot be checked against the type of their containing component since it would require depending on the type of the value (it's containing component definition) still being checked. Similarly the types provided in the options context cannot be inferred.

The only way to fix this would be requiring the user to predefine the type of the props data and use that to parameterize the props construction and checking of the above values. This could potentially be an option available to users who prefer safety over ergonomics.

Fortunately, this is a builder feature and an error caused in `getOptions` will not cause any issues when rendering the site later, so imperfect type safety is likely more acceptable. 

I do think one area this can be improved is to type the context value passed to `getOptions` as having the keys of `requiredOptionsContext` with `Data` values.


apps/nextjs-app-router/src/components/pokemon-demo/pokemon-demo.makeswift.ts
```diff
@@ -0,0 +1,90 @@
...
+  {
+    type: 'Pokemon Demo',
+    label: 'Custom / Pokémon Demo',
+    props: {
+      className: Style(),
+      pokemon: PokemonSelector(),
+      sprite: unstable_Gallery({
+        label: 'Sprite',
+        requiredOptionsContext: {
+          selectedPokemon: 'pokemon.value.selection',
```
bc-drew-cutchins, line 73: I want this path to be `'pokemon.selection'`

apps/nextjs-app-router/src/components/pokemon-demo/pokemon-demo.makeswift.ts
```diff
@@ -0,0 +1,90 @@
...
+            id: g.toString(),
+            value: g,
+            label: g.toString(),
+          })),
+      }),
+
+      selection: Combobox({
+        label: 'Selection',
+        requiredOptionsContext: {
+          generation: 'pokemon.value.generation',
```
bc-drew-cutchins, line 44: It would be nice to allow for these paths to target their parent (i.e. `'^.value.generation'`) so that custom control compositions like `PokemonSelector` would not need to be aware of their absolute path inside their component's `props`.

apps/nextjs-app-router/src/components/pokemon-demo/pokemon-demo.makeswift.ts
```diff
@@ -0,0 +1,90 @@
...
+    props: {
+      className: Style(),
+      pokemon: PokemonSelector(),
+      sprite: unstable_Gallery({
+        label: 'Sprite',
+        requiredOptionsContext: {
+          selectedPokemon: 'pokemon.value.selection',
+        },
+        getOptions: async (ctx: SpriteOptionsContext) => {
+          const name = ctx.selectedPokemon?.value.name
```
bc-drew-cutchins, line 76: I want this access to simply be:
```ts
const name = ctx.selectedPokemon?.name
```

PR #1395, opened 2026-08-13
WIP: export ServerElementsCache and collectServerElements

fikrikarim:
**Draft — opened primarily to trigger a pkg.pr.new build** (`snapshot-release.yml` runs on `pull_request: [opened, synchronize]`), so `makeswift-stencil-worker` can try this out. Based on `agurtovoy/react-server-components`, not `main`.

## Why

`Slot` is currently the only public way to get `server: true` elements rendered. It is:

```tsx
<RSCSnapshotRenderer snapshot label type={Slot}>
  <ClientSlot label snapshot fallback />
</RSCSnapshotRenderer>
```

`RSCRenderer` (via `RSCSnapshotRenderer`) does the useful part — `collectServerElements` to build an `elementKey -> ReactNode` map, then `ServerElementsCache` to expose it — but the child is hardcoded to `ClientSlot`, and none of those three are exported.

We inject Makeswift regions into an existing page and already have our own region component that renders the tree client-side and takes a `fallback` (the origin's own markup, captured during an HTMLRewriter pass *after* the render). Going through `Slot` means that fallback can't reach the tree, so a region set to show content renders empty.

Since `ElementData` resolves server elements purely by element key out of `ServerElementsCache`, an adapter doesn't need its ancestors to come from flight at all. With these two exports we can build the map ourselves, serialize just the server-element subtrees as our own flight payload, and provide the cache around our existing props-hydrated tree — keeping our fallback and per-region error handling.

## The change

Two re-exports, no behaviour change:

| Export | Subpath | Why there |
| --- | --- | --- |
| `ServerElementsCache` | `unstable-framework-support` | `'use client'` component; sits with `RuntimeProvider` / `Slot` / `RootStyleRegistry`, which adapters already consume |
| `collectServerElements` | `react/server` | Server-only; sits with `ServerElement` / `RSCElementRenderer` / `Slot` |

Verified both land in `dist/esm` and `dist/types`, and that `'use client'` still survives on the `server-elements-cache` leaf module (load-bearing for `@vitejs/plugin-rsc`, which needs a stable client-reference module id).


fikrikarim:
Superseded by #1401, which carries this commit rebased onto the current `agurtovoy/react-server-components` tip plus the `unstable_elementKey` change.

PR #1400, opened 2026-08-14
refactor: client throws on error, retry on rate limit

arvinpoddar:
Update the REST API client to throw when a non-404 error is received. The exception to this is when there's a rate limiting error (429), in which case we retry the request with a backoff policy + jitter applied.

To do this, we switch to using ky for decorating our fetch capabilities.

### DIFF
--- a/packages/runtime/package.json
+++ b/packages/runtime/package.json
@@ -18,7 +18,7 @@
     "unstable-framework-support"
   ],
   "engines": {
-    "node": ">=20.0.0"
+    "node": ">=20.19.0"
   },
   "sideEffects": false,
   "exports": {
@@ -242,6 +242,7 @@
     "is-hotkey": "^0.1.4",
     "js-base64": "^3.7.7",
     "jwt-decode": "^4.0.0",
+    "ky": "^1.14.3",
     "ot-json0": "^1.1.0",
     "parse5": "^7.3.0",
     "path-to-regexp": "^6.2.1",
--- a/packages/runtime/src/api/__tests__/rest-api-client.retries.test.ts
+++ b/packages/runtime/src/api/__tests__/rest-api-client.retries.test.ts
@@ -1,0 +1,141 @@
+import { http, HttpResponse } from 'msw'
+
+import { MakeswiftRestAPIClient, RestApiClientError } from '../rest-api-client'
+
+import { server } from '../../mocks/server'
+import { TestOrigins, TestWorkingSiteVersion, makeSwatch } from '../../testing/fixtures'
+
+const TEST_API_KEY = 'xxx'
+const swatchId = 'mySwatch'
+const resourceUrl = `${TestOrigins.apiOrigin}/v3/swatches/${swatchId}`
+
+function createTestClient() {
+  return new MakeswiftRestAPIClient({
+    fetch: globalThis.fetch,
+    apiKey: TEST_API_KEY,
+    apiOrigin: TestOrigins.apiOrigin,
+  })
+}
+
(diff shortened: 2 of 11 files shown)


arvinpoddar:
cursor review

packages/runtime/src/api/rest-api-client.ts
```diff
@@ -35,14 +61,12 @@ export class MakeswiftRestAPIClient {
 
     if (!response.ok) {
       const failedBody = await failedResponseBody(response)
-      if (response.status !== 404) {
-        console.error(`Failed to get swatch '${swatchId}'`, {
-          response: failedBody,
-          siteVersion,
-        })
-      }
+      if (response.status === 404) return null
```
arvinpoddar, line 68: Here and below, return `null` for 404 responses, otherwise throw a custom error containing the failed response + info on the inputs.

packages/runtime/package.json
```diff
@@ -242,6 +242,7 @@
     "is-hotkey": "^0.1.4",
     "js-base64": "^3.7.7",
     "jwt-decode": "^4.0.0",
+    "ky": "^1.14.3",
```
arvinpoddar, line 245: The latest `ky` requires Node 22+. Sticking to an older version for now to follow our current support for Node 20.19

packages/runtime/src/api/rest-api-client.ts
```diff
@@ -25,7 +32,26 @@ export class MakeswiftRestAPIClient {
...
-    this._fetch = fetch
+    this._fetch = ky.create({
+      fetch,
+      timeout: false,
+      retry: {
+        limit: RetryBackoffConfig.MaxAttempts,
+        statusCodes: [429],
+        backoffLimit: RetryBackoffConfig.MaxDelayMs,
+        jitter: true,
+      },
```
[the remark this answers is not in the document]
arvinpoddar: Not planned, but a desirable behavior IMO.

packages/runtime/src/client/index.ts
```diff
@@ -1040,7 +1036,3 @@ function getPagesQueryParams({
 
   return params
 }
-
-function responseError(response: Response): string {
-  return `${response.status} ${response.statusText}`
-}
```
arvinpoddar, line 1046: Moved to `rest-api-client.ts`

arvinpoddar:
cursor review

packages/runtime/src/api/__tests__/rest-api-client.retries.test.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    // Assert
+    expect(error?.message).toBe("Failed to get swatch 'mySwatch': 429 Too Many Requests")
+    expect(error?.cause).toEqual({ body: 'Rate limited', siteVersion: TestWorkingSiteVersion })
+
+    // The initial request, plus three retries.
+    expect(requestCount()).toBe(4)
+  }, 15_000)
+
+  test('honors the `Retry-After` header', async () => {
```
agurtovoy, line 85: The test name sets the expectation that we verify the full `Retry-After` header's semantic, but we don't really check that we honor the requested delay.

agurtovoy (dismissed):
Nice work, just one minor comment on tests.

packages/runtime/src/api/__tests__/rest-api-client.retries.test.ts
arvinpoddar: Thanks, addressed.

PR #1404, opened 2026-08-14
feat(controls): Add `resolveValueFromData` method to `ControlDefinition`

bc-drew-cutchins:
In order to implement passing control values as context to other control callbacks such as `getOptions`, it must be possible to get the resolved value of controls from within the builder.

However, the existing `resolveValue` function cannot be used for two reasons:
1. It depends on a `ResourceResolver` and `Stylesheet`, using both of these for value resolution and expecting to mutate the stylesheet.
2. `resolveValue` can return values that cannot be serialized, such as React nodes. In this case, it would not be possible for the builder to pass these values to a callback within the runtime.

This PR implements a new method `resolveValueFromData` that returns the control's resolved value only if it can only be derived from the control's `DataType`

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/packages/controls/src/controls/code/code.ts
+++ b/packages/controls/src/controls/code/code.ts
@@ -170,7 +170,9 @@ class Definition<C extends Config> extends ControlDefinition<
     return data
   }
 
-  resolveContextValue(data: DataType<C>): ResolvedValueType<C> | undefined {
+  resolveValueFromData(
+    data: DataType<C> | undefined,
+  ): ResolvedValueType<C> | undefined {
     const value = this.fromData(data) ?? this.config.defaultValue
     return value === undefined ? undefined : ({ value } as ResolvedValueType<C>)
   }
@@ -178,9 +180,7 @@ class Definition<C extends Config> extends ControlDefinition<
   resolveValue(
     data: DataType<C> | undefined,
   ): Resolvable<ResolvedValueType<C> | undefined> {
-    const value = this.fromData(data) ?? this.config.defaultValue
-    const resolved =
-      value === undefined ? undefined : ({ value } as ResolvedValueType<C>)
+    const resolved = this.resolveValueFromData(data)
     return {
       name: Definition.type,
       readStable: () => resolved,
--- a/packages/controls/src/controls/code/code.ts
+++ b/packages/controls/src/controls/code/code.ts
@@ -170,6 +170,11 @@ class Definition<C extends Config> extends ControlDefinition<
     return data
   }
 
+  resolveContextValue(data: DataType<C>): ResolvedValueType<C> | undefined {
+    const value = this.fromData(data) ?? this.config.defaultValue
+    return value === undefined ? undefined : ({ value } as ResolvedValueType<C>)
+  }
+
   resolveValue(
     data: DataType<C> | undefined,
   ): Resolvable<ResolvedValueType<C> | undefined> {
--- a/packages/controls/src/controls/color/color.ts
+++ b/packages/controls/src/controls/color/color.ts
@@ -230,6 +230,10 @@ class Definition<C extends Config> extends ControlDefinition<
     }
   }
 
+  resolveContextValue(): ResolvedValueType<C> | undefined {
+    return undefined
+  }
+
   resolveValue(
     data: DataType<C> | undefined,
     resolver: ResourceResolver,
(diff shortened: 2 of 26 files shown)


packages/controls/src/controls/select/select.ts
```diff
@@ -144,6 +144,12 @@ class Definition<C extends Config> extends ControlDefinition<
     return data
   }
 
+  resolveContextValue(
+    data: DataType<C> | undefined,
+  ): ResolvedValueType<C> | undefined {
+    return this.fromData(data)
+  }
```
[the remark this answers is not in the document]
bc-drew-cutchins: Fixed.

packages/controls/src/controls/list/list.ts
```diff
@@ -176,6 +176,14 @@ class Definition<C extends Config> extends ControlDefinition<
     )
   }
 
+  resolveContextValue(data: DataType<C>): ResolvedValueType<C> | undefined {
+    const itemValues = data?.map(({ value }) =>
+      this.itemDef.resolveContextValue(value),
+    )
+
+    return itemValues
+  }
```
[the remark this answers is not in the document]
bc-drew-cutchins: Fixed.

packages/controls/src/controls/image/image.ts
```diff
@@ -285,6 +285,10 @@ class Definition<C extends Config = DefaultConfig> extends ControlDefinition<
       .otherwise((val) => val)
   }
 
+  resolveContextValue(): ResolvedValueType<C> | undefined {
+    return undefined
+  }
```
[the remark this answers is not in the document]
bc-drew-cutchins: This is preferred to having the value sometimes resolve. Definitions can either always or never resolve context values.

packages/controls/src/controls/color/color.ts
```diff
@@ -230,6 +230,10 @@ class Definition<C extends Config> extends ControlDefinition<
     }
   }
 
+  resolveContextValue(): ResolvedValueType<C> | undefined {
+    return undefined
```
arvinpoddar, line 234: Should we throw if the control isn't supported?

arvinpoddar: Maybe not, that probably complicates things for our composable controls.

arvinpoddar (dismissed):
We can merge as is, has there been any development on whether we want to call this context?

packages/controls/src/controls/color/color.ts
bc-drew-cutchins: Yep exactly, it would require groups to know which nested controls it can safely call resolveContextValue on.

bc-drew-cutchins:
> We can merge as is, has there been any development on whether we want to call this context?

@arvinpoddar I know Aleksey had some opinions on this and I'm happy to do whatever others think is best here.

packages/controls/src/controls/code/code.ts
```diff
@@ -170,6 +170,11 @@ class Definition<C extends Config> extends ControlDefinition<
     return data
   }
 
+  resolveContextValue(data: DataType<C>): ResolvedValueType<C> | undefined {
+    const value = this.fromData(data) ?? this.config.defaultValue
+    return value === undefined ? undefined : ({ value } as ResolvedValueType<C>)
+  }
```
agurtovoy, line 178: I don't love how for most controls the implementation is a copy & paste of `resolveValue`'s `readStable()` (e.g. see lines 181-183 below). 

At the same time, reformulating `readStable` in terms of `resolveContextValue` feels wrong because the name is very specific and tied to the use case for which this method is introduced. What I'm wondering is whether tying the name to the use case is the right call.

What if we focused on our input constraint instead (deriving the resolved value using control data alone, without a resource resolver, stylesheet, or control instance) and called the method `resolveValueFromData`? There might be other use cases for this functionality, and it composes nicely with `readStable`:

```ts
readStable: () => this.resolveValueFromData(data)
```

Thoughts? cc: @arvinpoddar

arvinpoddar: That's much better, good suggestion @agurtovoy

bc-drew-cutchins: Love it, updated.

agurtovoy (dismissed):
Looks great!

agurtovoy:
@bc-drew-cutchins We should update the PR title and description to reflect the new name, though.

agurtovoy:
Also, I think we need a changeset to bump the `controls` package version.

agurtovoy (approved):

PR #1405, opened 2026-08-14
feat(controls, runtime): Implement ContextValue API

bc-drew-cutchins:
Implements the ContextValue API for consuming control values within callbacks inside other controls.

## Example:
```ts
const stateNameContext = ContextValue('stateName').withType<string>()

runtime.registerComponent(
  lazy(() => import('./control-context-demo')),
  {
    type: 'Control Context Demo',
    label: 'Custom / Control Context Demo',
    props: {
      className: Style(),
      stateName: Combobox({
        label: 'State',
        provides: stateNameContext,
        getOptions(query) {
          const lowerCaseQuery = query.toLowerCase()

          return Object.keys(STATE_CITIES)
            .filter((state) => state.toLowerCase().includes(lowerCaseQuery))
            .map((state) => ({
              id: state,
              value: state,
              label: state,
            }))
        },
      }),
      cityName: Combobox({
        label: 'City',
        dependsOn: { selectedState: stateNameContext },
        getOptions(query, context) {
          const lowerCaseQuery = query.toLowerCase()
          const selectedStateName = context.selectedState

          if (!selectedStateName) {
            return []
          }

          const cities: readonly string[] =
            STATE_CITIES[selectedStateName as State]

          return cities
            .filter((city) => city.toLowerCase().includes(lowerCaseQuery))
            .map((city) => ({
              id: `${selectedStateName}-${city}`,
              value: city,
              label: city,
            }))
        },
      }),
    },
  },
)
```

## Type Checking
A component's props are type checked to ensure the following:
- All depended on `ContextValue`'s are provided by a control within the props.
- No controls inside multi-value controls like `List`s provide `ContextValue`s.
- All `ContextValue`s only have a single provider
- A provided `ContextValue`'s type matches the type of the control's resolved value.

### What is the purpose of the string passed to ContextValue?
1. The string is used to infer a generic parameter for PropContext. It's required for typechecking, essentially serving as a type brand for the individual PropContexts. Without it there would not be a way for the type checker to identify the different PropContexts as structurally distinct.
2. It is used in the builder as the unique id used to route the context values between controls.

### Why is a PropContext constructed using `withType<T>` instead of a generic parameter on PropContext?
The use of withType<string>() is to avoid alternatively requiring: PropContext<'stateName', string>('stateName'). Typescript cannot partially infer generic parameters, it has to be all or none. This splits the generic parameterization to two sites so the first can be inferred. Alternatives to this api include:

```ts
const stateNameContext = PropContext('stateName', t<string>())
const stateNameContext = PropContext<string>()('stateName')
const stateNameContext = PropContext<'stateName', string>('stateName')
const stateNameContext = PropContext('stateName').<string>()
```

I think they're all weird for different reasons, and withType was my preference. I'm happy to differ to anyone else's judgement on this one. Without the trailing withType the type checker will complain if we try to use the prop context anywhere.

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


packages/controls/src/controls/context-value/predicates.ts
```diff
@@ -0,0 +1,28 @@
...
+    typeof config === 'object' &&
+    config != undefined &&
+    'provides' in config &&
+    providesSchema.safeParse(config.provides).success
+  )
+}
+
+export function isDependsOnConfig<T>(
+  config: T,
+): config is T & Required<ProvidesConfig<AnyContextValue>> {
```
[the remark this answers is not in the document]
bc-drew-cutchins: Fixed.

packages/controls/src/controls/context-value/index.ts
```diff
@@ -0,0 +1,33 @@
+import type { ContextValue, ContextValueBuilder } from './types'
+
+export * from './predicates'
+export * from './types'
+export * from './schemas'
+export * from './types'
```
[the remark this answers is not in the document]
bc-drew-cutchins: Fixed.

arvinpoddar (dismissed):
Looks good overall. Two things:

1. I'm not qualified to review the Typescript in packages/controls/src/controls/context-value/types.ts, going to have to lean on @agurtovoy for that
2. Before release, we should mark these fields with the `unstable_` prefix on the control definition so it's known that they're not intended for public consumption yet.

packages/controls/src/controls/context-value/index.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export * from './predicates'
+export * from './types'
+export * from './schemas'
+
+/**
+ * Declares a value one control makes available to others in the same component.
+ *
+ * ```ts
+ * const stateName = PropContext('stateName').withType<string>()
```
agurtovoy: `PropContext` -> `ContextValue`

packages/controls/src/controls/context-value/index.ts
```diff
@@ -0,0 +1,32 @@
...
+ */
+export function ContextValue<Id extends string>(
+  id: Id,
+): ContextValueBuilder<Id> {
+  /*
+    The context construction is split across two calls because TypeScript has no partial
+    type argument inference: naming `T` explicitly in a single call would prevent
+    `Id` being inferred. This prevents requiring duplicate declarations of the id.
+  */
+  return { withType: <T>() => ({ id }) as ContextValue<Id, T> }
```
agurtovoy: I still think `ofType` would read more naturally here:

```ts
const stateName = ContextValue('stateName').ofType<string>()
```

packages/controls/src/controls/context-value/index.ts
```diff
@@ -0,0 +1,32 @@
...
+ *   stateName: Combobox({ provides: stateName, getOptions(q) { ... } }),
+ *   cityName: Combobox({
+ *     dependsOn: { selectedState: stateName },
+ *     getOptions(query, context) { ... }, // context: { selectedState: string | undefined }
+ *   }),
+ * }
+ * ```
+ *
+ */
+export function ContextValue<Id extends string>(
```
agurtovoy: To @arvinpoddar's point, since we're declaring this an unstable API, this should probably be `unstable_ContextValue`

packages/controls/src/controls/context-value/index.ts
agurtovoy: Can we reformulate this without a cast, e.g.

```ts
  return { withType: <T>(): ContextValue<Id, T> => ({ id }) }
```

?

packages/controls/src/controls/context-value/types.ts
```diff
@@ -0,0 +1,212 @@
...
+  readonly id: Id
+  readonly __value?: T
+}
+
+export type ContextValueBuilder<Id extends string> = {
+  withType<V>(): ContextValue<Id, V>
+}
+
+// Providing
+export type AnyContextValue = ContextValue<string, any>
```
agurtovoy, line 20: Here and throughout, we should suppress the ESLint warning about the use of `any` (if it's legitimate).

packages/controls/src/controls/context-value/types.ts
```diff
@@ -0,0 +1,212 @@
...
+
+// Dependency
+export type ContextValueDependencies = Record<string, AnyContextValue>
+
+// Runtime
+export type ContextValues<D extends ContextValueDependencies> = {
+  [K in keyof D]: ContextValueOf<D[K]> | undefined
+}
+
+export type ContextIdOf<C> = C extends ContextValue<infer Id, any> ? Id : never
```
agurtovoy: `any` -> `unknown`

packages/controls/src/controls/context-value/types.ts
```diff
@@ -0,0 +1,212 @@
...
+export type ContextValueOf<C> = C extends ContextValue<any, infer V> ? V : never
```
agurtovoy: `any` -> `string`

packages/controls/src/controls/context-value/types.ts
```diff
@@ -0,0 +1,212 @@
...
+  HasConfig<Def> extends false
+    ? never
+    : ConfigType<Def> extends DependsOnConfigUnconstrained<infer D>
+      ? ContextIdOf<NonNullable<D>[keyof NonNullable<D>]>
+      : never
+
+const GROUP_TYPE = 'makeswift::controls::group'
+const SHAPE_TYPE = 'makeswift::controls::shape'
+const LIST_TYPE = 'makeswift::controls::list'
+const STYLE_V2_TYPE = 'makeswift::controls::style-v2'
```
agurtovoy: I think we should invert the dependency here and add some sort of generic marker to composite controls that would allow us to categorize them as "keyed" and "multivalue" containers without enumerating them here or depending on their specific type signatures.

Perhaps something like

```ts
// definition.ts
export abstract class ControlDefinition<
   ...
  ContainerType = never,
> {
  __associatedTypes(_: {
    ...
    ContainerType: ContainerType
  }) {}
...
}

// group.ts
class Definition<C extends Config> extends ControlDefinition<
  typeof Definition.type,
  C,
  DataType<C>,
  ValueType<C>,
  ResolvedValueType<C>,
  InstanceType<C>,
  { kind: 'record'; type: C['props'] }
> {
...
}

// list.ts
class Definition<C extends Config> extends ControlDefinition<
  typeof Definition.type,
  C,
  DataType<C>,
  ValueType<C>,
  ResolvedValueType<C>,
  InstanceType<C>,
  { kind: 'list'; type: C['type']  }
> {
...
}
```

packages/controls/src/controls/context-value/predicates.ts
```diff
@@ -0,0 +1,34 @@
...
+export function isContextValue(val: unknown): val is ContextValue {
+  return contextValueSchema.safeParse(val).success
+}
+
+export function isProvidesConfig<T>(
+  config: T,
+): config is T & Required<ProvidesConfig<AnyContextValue>> {
+  return (
+    typeof config === 'object' &&
+    config != undefined &&
```
agurtovoy: Nit: `config != null` would be more idiomatic for our codebase.

packages/controls/src/controls/context-value/index.ts
bc-drew-cutchins: Fixed.

packages/controls/src/controls/context-value/index.ts
bc-drew-cutchins: Agreed, fixed.

packages/controls/src/controls/context-value/index.ts
bc-drew-cutchins: Updated. 

How do we feel about marking `dependsOn` and `provides` as unstable? I don't love that it would require a new version of the config schemas once these fields were made stable. Is it fair enough to only have the `unstable` label on `ContextValue`?

cc: @arvinpoddar

packages/controls/src/controls/context-value/index.ts
bc-drew-cutchins: Yep, fixed.

packages/controls/src/controls/context-value/types.ts
bc-drew-cutchins: Done.

packages/controls/src/controls/context-value/types.ts
bc-drew-cutchins: Done.

packages/controls/src/controls/context-value/types.ts
bc-drew-cutchins: Done.

packages/controls/src/controls/context-value/predicates.ts
bc-drew-cutchins: Fixed.

packages/controls/src/controls/combobox/combobox.ts
```diff
@@ -58,8 +77,10 @@ class Definition<C extends Config> extends ControlDefinition<
     const config = z.object({
       getOptions: z
         .function()
-        .args(z.string())
+        .args(z.string(), contextValuesSchema)
```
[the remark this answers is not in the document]
bc-drew-cutchins: I think this is desired behavior. This release provides a breaking change where getOptions now must be called with a context parameter by the builder. This call would fail typechecking before the runtime failure.

@agurtovoy Mind double checking my understanding on this one?

packages/controls/src/controls/context-value/index.ts
agurtovoy: > I don't love that it would require a new version of the config schemas once these fields were made stable. 

Good point.

> Is it fair enough to only have the unstable label on ContextValue?

Personally, I think that's fine.

packages/controls/src/controls/gallery/gallery.ts
```diff
@@ -24,24 +34,34 @@ export type GalleryPage<T extends Data = Data> = {
...
-  | Promise<GalleryPage<T>>
+type GetOptionsType<T extends Data, D extends ContextValueDependencies> = (
+  context: ContextValues<D>,
+) => GalleryPage<T> | Promise<GalleryPage<T>>
+
+type Config<
+  T extends Data = Data,
+  D extends ContextValueDependencies = {},
+  P extends AnyContextValue = AnyContextValue,
+> = ProvidesConfig<P> &
```
bc-drew-cutchins, line 43: Currently I only have gallery and combobox able to provide context values to limit implementation scope. However, I wonder if it is better safe than sorry to allow all controls to declare `provides` in case there are any other instances of cross control context when migrating widget templates?

@arvinpoddar

arvinpoddar: Given that this is `unstable`, I'd say keep the surface area small for now. We can add in support later if we need it.

bc-drew-cutchins: Sounds good 👍

packages/controls/src/controls/definition.ts
```diff
@@ -23,13 +23,23 @@ export type Resolvable<T> = ValueSubscription<T> & {
...
+  | Scalar
+
 export abstract class ControlDefinition<
   ControlType extends string = string,
   Config = unknown,
   DataType extends Data = Data,
   ValueType extends Data = Data,
   ResolvedValueType = Data | unknown,
   InstanceType extends AnyControlInstance = AnyControlInstance,
+  ContainerType extends AnyContainerType = AnyContainerType,
```
bc-drew-cutchins, line 42: Because ControlDefinition with default generic parameters must represent the widest type for control definitions, this cannot be a default like `Scalar` or `never`. Instead, definitions can opt in to identifying their container type or correctly reflect that their container is either a keyed, multivalue, or scalar container. 

For now, `Scalar` is not actually used as a parameter, but I think it still makes sense to include it.

packages/controls/src/controls/combobox/combobox.ts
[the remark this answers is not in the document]
agurtovoy: @bc-drew-cutchins Yep, `getOptions` is builder-only API and the context parameter will be provided by the builder. The only tricky compatibility scenario here is a newer builder runtime deserializing control definitions coming from an older host, but that still works out: the builder will call the callback with the additional context arg, which older callbacks will simply ignore.

packages/controls/src/controls/context-value/types.ts
bc-drew-cutchins: Good call, added.

packages/runtime/src/controls/index.ts
```diff
@@ -2,6 +2,7 @@ export * from './control'
 
 export {
   ControlDefinition,
+  ControlDefinitionVisitor,
```
bc-drew-cutchins, line 5: Traversing the definition tree to gather context `dependsOn` and `provides` is easier and more future proof using a `ControlDefinitionVisitor` implementation in the builder. Adding this export for the builder side integration.

packages/controls/src/controls/context-value/schemas.ts
```diff
@@ -0,0 +1,6 @@
+import { z } from 'zod'
+
+export const contextValueSchema = z.object({ id: z.string() })
+export const providesSchema = contextValueSchema
+export const dependsOnSchema = z.record(contextValueSchema)
+export const contextValuesSchema = z.record(z.unknown())
```
agurtovoy: This is minor, but our convention for schemas in this repo is to avoid suffixing each schema var with `Schema` and instead use the `Schema` namespace import (`import * as Schema`) so they can be used as `Schema.contextValue`, `Schema.provides`, etc. -- see, e.g. https://github.com/makeswift/makeswift/blob/main/packages/controls/src/common/types.ts

packages/controls/src/controls/context-value/types.test.types.ts
```diff
@@ -0,0 +1,407 @@
...
+          s: string | undefined
+          z: number | undefined
+        }>()
+        return []
+      },
+    })
+  })
+
+  test('no dependsOn means an empty context', () => {
+    const def = Combobox({
```
agurtovoy, line 104: This and a dozen of other vars below produce the following ESLint warning:

```
'def' is assigned a value but only used as a type. Allowed unused vars must match /^_/u.
```

The rule seems too nitpicky to me, I'm inclined to say that we should disable it altogether (not in this PR). Thoughts? cc: @arvinpoddar

packages/controls/src/controls/context-value/types.ts
```diff
@@ -0,0 +1,191 @@
...
+// Providing
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type AnyContextValue = ContextValue<string, any>
+
+// Dependency
+export type ContextValueDependencies = Record<string, AnyContextValue>
+
+// Runtime
+export type ContextValues<D extends ContextValueDependencies> = {
+  [K in keyof D]: ContextValueOf<D[K]> | undefined
```
agurtovoy, line 27: It's a little unfortunate that all context values can potentially be undefined and thus require checking for `undefined` in dependent controls, even if the source control such as `Select` or `Checkbox` is configured to explicitly disallows undefined values. I think it's okay to merge this as-is for now, but something to consider before we remove the `unstable_` prefix.

agurtovoy (dismissed):
Looking good, just a couple of minor comments.

.changeset/eager-doors-sort.md
```diff
@@ -0,0 +1,6 @@
+---
+'@makeswift/controls': minor
+'@makeswift/runtime': minor
+---
+
+Implement unstable context value API
```
fikrikarim, line 6: Are there any breaking changes? If not, perhaps we can make this a patch bump instead of a minor bump.

The semver semantic is different since we're still on `v0.x.x`

packages/controls/src/controls/context-value/types.test.types.ts
arvinpoddar: Yeah we can definitely drop this rule

packages/controls/src/controls/context-value/types.ts
bc-drew-cutchins: Yep that's fair, I'll think on this more before we implement the version for users.

.changeset/eager-doors-sort.md
bc-drew-cutchins: Yep there are breaking changes for the builder, getOptions for Gallery and Combobox must now get called with a context parameter.

cc: @agurtovoy

fikrikarim: For the builder or for users that are using the package in their codebase? 

I think usually breaking changes are seen from users perspective and not from the builder perspective.

bc-drew-cutchins: Just for the builder use case, though technically the user could be calling their getOptions callbacks arbitrarily, perhaps in unit tests? I'm happy to differ to y'all's judgement here, I had previously discussed with Aleksey and he suggested a minor bump.

packages/controls/src/controls/context-value/schemas.ts
bc-drew-cutchins: Ah, cool, fixed!

.changeset/eager-doors-sort.md
fikrikarim: I'm fine with minor bump since I remember Aleksey also did a minor bump after lot of internal only changes.

But personally I've been following the semver semantic so customers are aware if there's any breaking change that they need to update in their codebase.

packages/controls/src/controls/context-value/schemas.ts
bc-drew-cutchins: Had to import/export this as `ContextValueSchema` to not collide with the already existing `Schema` from `packages/controls/src/common/schema.ts`.

agurtovoy: Did you push your changes? I'm not seeing them.

bc-drew-cutchins: [This commit](https://github.com/makeswift/makeswift/pull/1405/changes/c062d50076d675ae9b17a8017e77e8939e79a9aa)

agurtovoy: Great, thanks! I think it's an improvement.

.changeset/eager-doors-sort.md
agurtovoy: Strictly speaking, by bumping minor version here, we *are* following the semver semantics: we're breaking the public API; it's just it's a public API that we are pretty sure none of our customers are using. That said, I'm okay if we decide to go with patch here and bump the minor after we merge the new CSS runtime. If we don't cut a release in between these two, than what we do here is inconsequential.

bc-drew-cutchins: [This commit](https://github.com/makeswift/makeswift/pull/1405/changes/473a837a1c9c6e576e67c5db08fabb821c29bf76) makes this no longer a breaking change. Now context is only required if the config declares `dependsOn` ... technically I guess they could have arbitrarily included `dependsOn`... but whatever 🤷 ... I'm not a big fan of structural typing 😝.

Updated this to be a patch version.

agurtovoy: Very nice, I think that's a legitimate improvement!

agurtovoy (approved):

PR #1412, opened 2026-08-18
feat(table): update create table records gql mutation for new host api endpoint

rodrigomf24:
## What?

- The built-in `Form` component now creates table records through the host API's `POST /v1/table-records` endpoint instead of the gql `createTableRecord` mutation. The request is proxied through the host's existing `/api/makeswift` route, so the site API key stays server-side.
- Removes the instances of the gql mutation path (`useMakeswiftGraphQLApiClient`, `MakeswiftGraphQLApiClient.createTableRecord`, `CreateTableRecordMutation`) since they were replaced by the new host api endpoint.

https://github.com/user-attachments/assets/d4462cf8-6f14-40c8-a05e-6fe99b961d08

## Why?

This is part of the effort from the Makeswift CDN project, the goal here is to remove GQL from the runtime and have it route requests through the host api.


packages/runtime/src/api/table-record-client.ts
```diff
@@ -0,0 +1,30 @@
+import { type HttpFetch } from './types'
+
+export type TableRecordColumn = { columnId: string; data: unknown }
+
+export class MakeswiftTableRecordClient {
```
rodrigomf24, line 5: note: this was made since the existing implementation of `HostApiResourcesClient` extends `ApiResourcesClient` and this is a typed redux backed cache implementation, which IMO is meant more for reads. 

With this new endpoint that handles writing a table record I consider it would be best to create a separate client, we can perhaps rename it more generic but the goal was to not use a class that is meant for cached back resources.

agurtovoy: Yep, creating a separate client here was the right call.

agurtovoy (dismissed):
Looks great, thank you! 🙏

arvinpoddar (approved):

PR #1413, opened 2026-08-18
fix: prevent host base styles from colliding with rich text

arvinpoddar:
Prevent top-level styles from the host (typically those that target all elements of a particular type) from colliding with text styles set in Makeswift. Due to how react-slate renders editable text + specifity arguments, these top-level styles from the host can often supersede Makeswift styles, preventing the user from editing in the builder.

Original issue repro:

https://github.com/user-attachments/assets/89cc37c6-92d0-457c-b6d2-94a4b1ebdc64

Demo of fix:

https://github.com/user-attachments/assets/e11365d4-1501-403c-b1de-7ada22cebd98

### DIFF
--- a/packages/runtime/src/runtimes/react/controls/use-slate-reset.ts
+++ b/packages/runtime/src/runtimes/react/controls/use-slate-reset.ts
@@ -1,0 +1,43 @@
+import { useStyle } from '../use-style'
+
+/**
+ * `slate-react` renders read-only rich-text within a `<p>` tag containing
+ * multiple spans. Makeswift text-styles are applied to these spans via a class
+ * on the span tags directly:
+ *
+ * <p class="...">
+ *    <span class="mswft-12345">My text</span>
+ * </p>
+ *
+ * HOWEVER: in the builder, `slate-react` renders the same content as editable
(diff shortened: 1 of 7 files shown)


packages/runtime/src/runtimes/react/controls/use-slate-reset.ts
```diff
@@ -0,0 +1,43 @@
...
+ * the Makeswift text styles in terms of specificity. Therefore, setting text
+ * styles won't work in the builder: the top-level host styles will win. On the
+ * live page, the Makeswift styles win since the intermediate DOM elements are
+ * not present.
+ *
+ * These style updates address this by reverting base styles on these spans,
+ * restoring inheritance from Makeswift defined styles, thus matching the live
+ * page where the text is a direct child of the classed leaf span.
+ */
+export function useSlateReset(): string {
```
agurtovoy, line 36: This is subtle. Should we add `isInBuilder: true` test case to `packages/runtime/src/next/components/tests/controls/rich-text-v2-control/client.test.tsx`?

agurtovoy (dismissed):

Codeseph (dismissed):

PR #1415, opened 2026-08-20
fix: prevent host styles from overriding typography styles

arvinpoddar:
Prevent host styles from overriding rich text typography styles by applying the important attribute to all generated typography styles.

Problem:

https://github.com/user-attachments/assets/7f973f12-8283-427d-8d4b-58ff554b9647

Demo of fix:

https://github.com/user-attachments/assets/05e29a3a-f052-4728-91d1-0a1c80aa5976

### DIFF
--- a/packages/runtime/src/runtimes/react/controls/typography.ts
+++ b/packages/runtime/src/runtimes/react/controls/typography.ts
@@ -154,6 +154,14 @@ export default function useEnhancedTypography(
     .filter(isNonNullable)
 }
 
+// Prevent typography styles from being overriden by top level host styles by
+// applying the "!important" modifier.
+function important(styles: CSSObject): CSSObject {
+  return Object.fromEntries(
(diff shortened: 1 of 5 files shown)


packages/runtime/src/runtimes/react/controls/typography.ts
```diff
@@ -174,7 +182,7 @@ function typographyToCssObject(value: EnhancedTypographyValue): CSSObject {
       .join(' ')
   if (value.italic != null) styles.fontStyle = value.italic === true ? 'italic' : 'initial'
 
-  return styles
+  return important(styles)
```
arvinpoddar, line 185: The actual fix: updating all CSS styles to be `!important`

arvinpoddar:
After further discussion, we're probably going to lean towards a different approach that's opt-in, so we avoid breaking existing styles for hosts that may be relying on custom CSS.

PR #1416, opened 2026-08-21
refactor(hono-react)!: `ReactRuntime` no longer registers built-ins

agurtovoy:
`hono-react`'s `ReactRuntime` no longer registers built-in components, allows passing of `apiKey`.

### DIFF
--- a/packages/hono-react/src/runtime.ts
+++ b/packages/hono-react/src/runtime.ts
@@ -1,19 +1,16 @@
-import {
-  type Breakpoints,
-  type StoreKey,
-  ReactRuntime,
-} from '@makeswift/runtime/unstable-framework-support'
+import { ReactRuntimeCore } from '@makeswift/runtime/react/core'
 
-export class HonoReactRuntime extends ReactRuntime {
-  constructor(args: {
-    requestKey?: StoreKey
-    appOrigin?: string
-    apiOrigin?: string
-    breakpoints?: Breakpoints
-  }) {
+type Args = Omit<ConstructorParameters<typeof ReactRuntimeCore>[0], 'fetch'>
+
+/**
+ * Note that this runtime class does not automatically register Makeswift's
+ * built-in components.
+ */
+export class ReactRuntime extends ReactRuntimeCore {
+  constructor(args: Args = {}) {
     super({
       ...args,
-      fetch: (url, init) => fetch(url, init), // TODO: revalidation support
+      fetch: (url, init) => fetch(url, init), // TODO: revalidation support?
     })
   }
 }
--- a/packages/hono-react/src/server/index.ts
+++ b/packages/hono-react/src/server/index.ts
@@ -1,3 +1,3 @@
 export { createApiHandler } from './api-handler'
 export { createPreviewMiddleware } from './preview'
-export { getSiteVersion } from './site-version'
+export { type SiteVersion, getSiteVersion } from './site-version'
--- a/packages/hono-react/src/server/preview.ts
+++ b/packages/hono-react/src/server/preview.ts
@@ -10,8 +10,9 @@ import {
   SearchParams,
   secondsUntilSiteVersionExpiration,
   serializeSiteVersion,
-  ReactRuntime,
 } from '@makeswift/runtime/unstable-framework-support'
+
+import { ReactRuntimeCore } from '@makeswift/runtime/react/core'
 
 import { Makeswift as MakeswiftClient } from '../client'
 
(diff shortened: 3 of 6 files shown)


packages/hono-react/src/server/index.ts
```diff
@@ -1,3 +1,3 @@
 export { createApiHandler } from './api-handler'
 export { createPreviewMiddleware } from './preview'
-export { getSiteVersion } from './site-version'
+export { type SiteVersion, getSiteVersion } from './site-version'
```
agurtovoy, line 3: Export `SiteVersion` type from the server entrypoint.

packages/hono-react/src/runtime.ts
```diff
@@ -1,19 +1,16 @@
...
-    apiOrigin?: string
-    breakpoints?: Breakpoints
-  }) {
+type Args = Omit<ConstructorParameters<typeof BaseReactRuntime>[0], 'fetch'>
+
+/**
+ * Note that this runtime class does not automatically register Makeswift's
+ * built-in components.
+ */
+export class ReactRuntime extends BaseReactRuntime {
```
agurtovoy: Switch to non-registering `ReactRuntime` instances as our default for new packages.

packages/hono-react/src/runtime.ts
```diff
@@ -1,19 +1,16 @@
...
+import { ReactRuntimeCore as BaseReactRuntime } from '@makeswift/runtime/react/core'
 
-export class HonoReactRuntime extends ReactRuntime {
-  constructor(args: {
-    requestKey?: StoreKey
-    appOrigin?: string
-    apiOrigin?: string
-    breakpoints?: Breakpoints
-  }) {
+type Args = Omit<ConstructorParameters<typeof BaseReactRuntime>[0], 'fetch'>
```
agurtovoy: Pass through all of the base class args except for `fetch`.

packages/hono-react/src/runtime.ts
```diff
@@ -1,19 +1,16 @@
...
+/**
+ * Note that this runtime class does not automatically register Makeswift's
+ * built-in components.
+ */
+export class ReactRuntime extends BaseReactRuntime {
+  constructor(args: Args = {}) {
     super({
       ...args,
-      fetch: (url, init) => fetch(url, init), // TODO: revalidation support
+      fetch: (url, init) => fetch(url, init), // TODO: revalidation support?
```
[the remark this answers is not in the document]
agurtovoy: Fixed

packages/hono-react/src/runtime.ts
```diff
@@ -1,19 +1,16 @@
...
+import { ReactRuntimeCore } from '@makeswift/runtime/react/core'
 
-export class HonoReactRuntime extends ReactRuntime {
-  constructor(args: {
-    requestKey?: StoreKey
-    appOrigin?: string
-    apiOrigin?: string
-    breakpoints?: Breakpoints
-  }) {
+type Args = Omit<ConstructorParameters<typeof ReactRuntimeCore>[0], 'fetch'>
```
agurtovoy, line 3: Pass through all of the base class args except for `fetch`.

packages/hono-react/src/runtime.ts
```diff
@@ -1,19 +1,16 @@
...
+
+/**
+ * Note that this runtime class does not automatically register Makeswift's
+ * built-in components.
+ */
+export class ReactRuntime extends ReactRuntimeCore {
```
agurtovoy, line 9: Switch to non-registering `ReactRuntime` instances as our default for new packages.

fikrikarim (approved):

packages/hono-react/src/server/preview.ts
```diff
@@ -40,7 +41,7 @@ export function createPreviewMiddleware<E extends Env>({
   runtime,
 }: {
   apiKey: string
-  runtime: ReactRuntime
```
arvinpoddar, line 43: Is there any reason to not require this to be the `ReactRuntime` class defined in this package? Since this `createPreviewMiddleware` is for `hono`, I could imagine us wanting to be more restrictive on what could be passed here.

arvinpoddar (approved):

fikrikarim:
AFAIK we won't need to update anything on the worker since it's already using ReactRuntimeCore
image

packages/hono-react/src/server/preview.ts
agurtovoy: Great question, I should have commented on this. While the preview middleware is indeed Hono-specific, its implementation only requires the `ReactRuntimeCore` API, and we actually do want to allow passing any `ReactRuntimeCore`-derived runtime here. 

In contrast to say, Next.js, we're not integrating with Hono as an all-encompassing end-to-end framework; allowing any `ReactRuntimeCore`-derived runtime here lets the Hono integration compose with environment-specific packages such as the upcoming `vite-rsc` integration, which brings its own RSC-enabled runtime.

agurtovoy:
@fikrikarim Yep, thank you, I should have mentioned that.

PR #1421, opened 2026-08-21
chore: exports `collectServerElements` and `ServerElementsProvider`

fikrikarim:
I think it's better to let the host use `ServerElementsProvider` directly, instead of shoving it to `ViteRSCFrameworkProvider` like in https://github.com/makeswift/makeswift/pull/1401

### DIFF
--- a/packages/runtime/src/runtimes/react/server/components/renderer.tsx
+++ b/packages/runtime/src/runtimes/react/server/components/renderer.tsx
@@ -3,13 +3,11 @@ import { type PropsWithChildren } from 'react'
 import { type CacheData } from '../../../../api/api-resources-client'
 import { type Document } from '../../../../state/modules/read-only-documents'
 
-import { ClientCSSProvider } from '../css/client-css'
-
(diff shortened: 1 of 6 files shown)


packages/runtime/src/runtimes/react/server/components/renderer.tsx
```diff
@@ -24,11 +22,7 @@ export function RSCRenderer({
 }: PropsWithChildren<{ context: ServerRenderContext; document: Document; cacheData: CacheData }>) {
   const serverElements = collectServerElements(context, document, cacheData)
 
-  return (
-    <ServerElementsCache value={serverElements}>
-      <ClientCSSProvider>{children}</ClientCSSProvider>
-    </ServerElementsCache>
-  )
+  return <ServerElementsProvider elements={serverElements}>{children}</ServerElementsProvider>
```
fikrikarim, line 25: refactored and exported

agurtovoy (approved):

PR #1424, opened 2026-08-22
fix(runtime): do not rely on `requestKey` for server store persistence

agurtovoy:
Move the store into the render context.


packages/runtime/src/runtimes/react/server/render-context.tsx
```diff
@@ -21,6 +21,7 @@ export type ServerRenderContext = {
   client: MakeswiftClient
   siteVersion: SiteVersion | null
   locale: string | undefined
+  store: Store
```
agurtovoy, line 24: Move the store into the render context.

PR #1425, opened 2026-08-22
fix: correct comment typo in rich text control instance

arvinpoddar:
Fixes a typo on a comment within the rich text control instance


arvinpoddar:
So it's on the record, I made this PR (with the changeset) to trigger a release to canary.

PR #1426, opened 2026-08-24
feat(table): replace table query with host api GET endpoint

rodrigomf24:
## What?

- Fetches a table's schema through the Host API's `GET v1/tables/:id` REST endpoint instead of the GraphQL `Table` query.
- Removes the now-unused GraphQL `Table` query path (`MakeswiftGraphQLApiClient.getTable`, the `Table` query document, and its generated types) and the `graphqlApiEndpoint` plumbing into `MakeswiftApiResourcesClient`, which had no other use.

## Why?

Part of the Makeswift CDN effort to remove GraphQL from the runtime and route requests through the host API instead, mirroring the earlier `createTableRecord` migration to `POST /v1/table-records`.

### DIFF
--- a/packages/runtime/src/api/makeswift-api-resources-client.ts
+++ b/packages/runtime/src/api/makeswift-api-resources-client.ts
@@ -16,7 +16,6 @@ import { ApiResourcesClient } from './api-resources-client'
 import { type SiteVersion } from './site-version'
 
 // see the comment in `MakeswiftApiResourcesClient` constructor below
-import { type MakeswiftGraphQLApiClient } from './graphql-api-client'
 import { type MakeswiftRestAPIClient } from './rest-api-client'
 
 /**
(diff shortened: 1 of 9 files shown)


arvinpoddar (dismissed):

packages/runtime/src/api/makeswift-api-resources-client.ts
```diff
@@ -28,28 +27,25 @@ import { type MakeswiftRestAPIClient } from './rest-api-client'
  */
 export class MakeswiftApiResourcesClient extends ApiResourcesClient {
   private restApiClient: Promise<MakeswiftRestAPIClient>
-  private graphQlClient: Promise<MakeswiftGraphQLApiClient>
```
agurtovoy, line 31: Yay!

packages/runtime/src/client/index.ts
```diff
@@ -953,7 +953,19 @@ export class MakeswiftClient extends MakeswiftRestAPIClient {
   }
 
   async getTable(tableId: string): Promise<Table | null> {
-    return this.graphqlClient.getTable(tableId)
+    const response = await this.fetch(`v1/tables/${tableId}`, null)
```
agurtovoy: Since `MakeswiftClient` extends `MakeswiftRestAPIClient`, it inherits its `getTable` implementation, so this method can be simply deleted.

arvinpoddar (approved):

agurtovoy (approved):

PR #1434, opened 2026-08-28
feat(controls, runtime): Improve ergonomics with unconstrained context value declarations

bc-drew-cutchins:
This PR improves ergonomics regarding context value validation of unconstrained control definitions by:
- Opting out unconstrained ContextValues from validation
- Making the default Config `D`(epends on) generic parameter equal to `ContextValueDependencies` instead of `{}`. This is more accurate, since an unconstrained config could depend on any `ContextValueDependencies` and thus it's `getOptions` function should always be called with a context parameter.

This PR will let the following code compile:
```ts
const testContext = unstable_ContextValue('abc').ofType<string>()

// Previously assigning this to a bare `ComboboxDefinition` would be an error, 
// since it's Config would be assumed to contain no dependencies and thus 
// would have different arity for getOptions.
const myCombobox: ComboboxDefinition = Combobox({
  label: 'Selection',
  provides: testContext,
  dependsOn: { value: testContext },
  getOptions: async (query, ctx) => {
    console.log({ ctx })
    return []
  },
})
```

as well as fixing [this issue](https://github.com/bigcommerce/makeswift-stencil-worker/pull/286#discussion_r3872821467).


packages/controls/src/controls/context-value/context-value.test.types.ts
```diff
@@ -344,6 +344,53 @@ describe('duplicate provider detection', () => {
...
+  })
+
+  test('runtime-built prop records validate', () => {
+    // The shape a runtime-generated registry passes to `registerComponent`.
+    const props = {} as Record<
+      string,
+      ReturnType<typeof Combobox> | ReturnType<typeof Checkbox>
+    >
+    register(props)
+  })
```
agurtovoy, line 369: It might be a flawed dichotomy, but only so far we've been classifying tests as "type tests" (`.types` suffix) only if they are using compile-time assertions (i.e. `expectTypeOf`). Under this convention, all other test cases, including the ones utilizing `// @ts-expect-error`, should go into `context-value.test.ts`.

packages/controls/src/controls/context-value/types.test.types.ts
```diff
@@ -344,6 +344,53 @@ describe('duplicate provider detection', () => {
...
+      dependsOn: { selectedState: stateNameContext },
+      getOptions: (_query, context) => [
+        { id: 'a', value: context.selectedState ?? '', label: 'a' },
+      ],
+    })
+    const contextFree: ComboboxDefinition = Combobox({
+      getOptions: (q: string) => [{ id: q, value: q, label: q }],
+    })
+    register({ withContext, contextFree })
+  })
```
agurtovoy: This one should probably go under the `assignability` category in `combobox.test.ts`.

packages/controls/src/controls/context-value/context-value.test.types.ts
agurtovoy: On a related note, `types.test.types.ts` should probably be `context-value.test.types.ts`.

agurtovoy: I obviously missed this during the initial review. You don't have to split this up if you don't think that this distinction is helpful here. I would rename the file as per above, though.

agurtovoy (dismissed):
Looks good, see inline for minor comments on test organization.

packages/controls/src/controls/context-value/types.test.types.ts
bc-drew-cutchins: Moved!

Passing it as a function parameter actually revealed a different typing issue requiring `getOptions` to be declared using method syntax.

packages/controls/src/controls/context-value/context-value.test.types.ts
bc-drew-cutchins: Interesting, I kinda see `register` here as a type of compile-time type assertion... the function doesn't do anything and asserts that the props are validated using its parameter type. I think I would prefer not splitting it up here if that's alright with you.

Renamed the file!

packages/controls/src/controls/gallery/gallery.ts
```diff
@@ -33,18 +33,24 @@ export type GalleryPage<T extends Data = Data> = {
...
   T extends Data = Data,
-  D extends ContextValueDependencies = {},
+  D extends ContextValueDependencies = ContextValueDependencies,
   P extends AnyContextValue = AnyContextValue,
 > = ProvidesConfig<P> &
   DependsOnConfig<D> & {
     label?: string
     description?: string
-    getOptions: GetOptionsType<T, D>
+    // Method syntax to exempt from `strictFunctionTypes` contravariance,
```
bc-drew-cutchins, line 47: I left this (cleaned up) verbose AI comment because I find it unintuitive that changing this from method syntax to property syntax would break type checking.

Happy to delete if others don't find this helpful.

packages/controls/src/controls/combobox/combobox.ts
```diff
@@ -25,18 +25,25 @@ import { ControlDefinitionVisitor } from '../visitor'
 type Option<T extends Data> = { id: string; value: T; label: string }
 type GetOptionsType<T extends Data, D extends ContextValueDependencies> = (
   query: string,
-  ...context: ContextParam<D>
+  context: ContextValues<D>,
```
agurtovoy, line 28: Does this [revert](https://github.com/makeswift/makeswift/pull/1405/changes/473a837a1c9c6e576e67c5db08fabb821c29bf76) us to always requiring `context` regardless of the presence of `dependsOn`?

agurtovoy (approved):

packages/controls/src/controls/combobox/combobox.ts
bc-drew-cutchins: Yes, technically this was already required on unconstrained definitions.

(We followed up offline are aligned on it)

PR #1437, opened 2026-08-31
feat(controls): Add provides to additional control configs

bc-drew-cutchins:
### What
This PR adds support for `provides` declarations to all non-composite control configs whose definitions return a defined value from `resolveValueFromData`.

### Why
This is required to support the various input types referenced by Product Image inputs for MoS.

### Implementation Note
Config types derived from Zod schemas have their provides field typed generically via an intersection with `ProvidesConfig<P>`. This causes toEqualTypeOf to fail when comparing the intersection to the flattened expected type. Flattening the config type fixes this and still tests for assignability.

### DIFF
--- a/.changeset/wide-cougars-press.md
+++ b/.changeset/wide-cougars-press.md
@@ -1,0 +1,5 @@
+---
+'@makeswift/controls': patch
+---
+
+Add provides (context value) field to additional control configs
--- a/.changeset/wide-cougars-press.md
+++ b/.changeset/wide-cougars-press.md
@@ -2,4 +2,4 @@
 '@makeswift/controls': patch
 ---
 
-Add provides (context value) field to additional control configs
+Add `provides` (context value) field to additional control configs
--- a/packages/controls/src/controls/checkbox/checkbox.test.types.ts
+++ b/packages/controls/src/controls/checkbox/checkbox.test.types.ts
@@ -1,7 +1,5 @@
 import { expectTypeOf } from 'expect-type'
 
-import { Flatten } from '../../testing/util-types'
-
 import { ControlDataTypeKey } from '../../common'
 
 import {
@@ -25,7 +23,7 @@ describe('Checkbox Types', () => {
 
     type ConfigType = typeof def.config
 
-    expectTypeOf<Flatten<ConfigType>>().toEqualTypeOf<{
+    expectTypeOf<ConfigType>().toEqualTypeOf<{
       label?: string
       description?: string
       defaultValue: boolean
@@ -46,7 +44,7 @@ describe('Checkbox Types', () => {
     const def = Checkbox()
 
     type ConfigType = typeof def.config
-    expectTypeOf<Flatten<ConfigType>>().toEqualTypeOf<{
+    expectTypeOf<ConfigType>().toEqualTypeOf<{
       label?: string
       description?: string
       defaultValue?: boolean
--- a/packages/controls/src/controls/checkbox/checkbox.test.types.ts
+++ b/packages/controls/src/controls/checkbox/checkbox.test.types.ts
@@ -7,6 +7,7 @@ import {
   type ResolvedValueType,
   type ValueType,
 } from '../associated-types'
+import { unstable_ContextValue } from '../context-value'
 
 import { Checkbox } from './checkbox'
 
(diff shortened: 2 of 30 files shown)


.changeset/wide-cougars-press.md
```diff
@@ -0,0 +1,5 @@
+---
+'@makeswift/controls': patch
+---
+
+Add provides (context value) field to additional control configs
```
agurtovoy: provides -> `provides`

packages/controls/src/testing/util-types.ts
```diff
@@ -0,0 +1 @@
+export type Flatten<T> = { [K in keyof T]: T[K] }
```
agurtovoy: Maybe we can add a comment that shows an example of the type transformation this is performing?

packages/controls/src/controls/checkbox/checkbox.test.types.ts
```diff
@@ -23,10 +25,11 @@ describe('Checkbox Types', () => {
 
     type ConfigType = typeof def.config
 
-    expectTypeOf<ConfigType>().toEqualTypeOf<{
+    expectTypeOf<Flatten<ConfigType>>().toEqualTypeOf<{
```
agurtovoy: > Config types derived from Zod schemas have their provides field typed generically via an intersection with ProvidesConfig<P>. This causes toEqualTypeOf to fail when comparing the intersection to the flattened expected type.

I think we can fix this and avoid flattening (which arguably hides an actual issue), see below.

packages/controls/src/controls/checkbox/checkbox.ts
```diff
@@ -15,7 +20,10 @@ import {
 import { DefaultControlInstance, type ControlInstanceArgs } from '../instance'
 import { ControlDefinitionVisitor } from '../visitor'
 
-type Config = z.infer<typeof Definition.schema.relaxed.config>
+type Config<P extends AnyContextValue = AnyContextValue> = z.infer<
+  typeof Definition.schema.relaxed.config
+> &
+  ProvidesConfig<P>
```
agurtovoy: I'd like us to keep the config type fully derivable from the Zod schema. What if we turned the static `schema` getter into a generic function and did something like this?

```ts
type DefinitionSchema<P extends AnyContextValue = AnyContextValue> = ReturnType<
  typeof Definition.schema<P>
>

type Config<P extends AnyContextValue = AnyContextValue> = z.infer<
  DefinitionSchema<P>['relaxed']['config']
>

...

static schema<P extends AnyContextValue = AnyContextValue>() {
  const provides = ContextValueSchema.provides as SchemaType<P>
  
   ...
  const config = z.object({
    label: z.string().optional(),
    description: z.string().optional(),
    defaultValue: value,
    provides: provides.optional(),
  })

  ...
}
```

agurtovoy: Then `SchemaByDefaultValue` and `NormedConfig` can be defined as

```ts
type SchemaByDefaultValue<
  D extends Config['defaultValue'],
  P extends AnyContextValue = AnyContextValue,
> = undefined extends D
  ? DefinitionSchema<P>['relaxed']
  : DefinitionSchema<P>['strict']

type NormedConfig<
  D extends Config['defaultValue'],
  P extends AnyContextValue,
> = z.infer<SchemaByDefaultValue<D, P>['config']>

```

bc-drew-cutchins: Ah, good call, this is much better! Fixing now.

bc-drew-cutchins: Fixed.

packages/controls/src/controls/checkbox/checkbox.test.types.ts
```diff
@@ -27,6 +27,7 @@ describe('Checkbox Types', () => {
       label?: string
       description?: string
       defaultValue: boolean
+      provides?: undefined
```
agurtovoy, line 31: Here and for all other controls, I'd add a type test (here) and an assignability test (in the sibling test file) for a control definition which provides a context.

packages/controls/src/controls/select/select.test.ts
```diff
@@ -52,6 +54,7 @@ describe('Select', () => {
     // assignTest(Select({ options, defaultValue: undefined }))
     // assignTest(Select({ label: 'Block type', options, defaultValue: undefined }))
     // assignTest(Select({ label: undefined, options, defaultValue: undefined }))
+    // assignTest(Select({ options, provides: unstable_ContextValue('selection').ofType<string>() }))
```
bc-drew-cutchins, line 55: This test is broken because the `SelectDefinition` config generic param is not covariant. (`Config<string>` is not assignable to `NormedConfig<'p' | 'h1', ...>`)

packages/controls/src/controls/checkbox/checkbox.test.types.ts
bc-drew-cutchins: Done!

agurtovoy (approved):

PR #1440, opened 2026-09-02
feat(vite-rsc): `serverOnly` wrapper + Vite plugin

agurtovoy:
Add the `serverOnly` wrapper for server component loaders and a corresponding Vite plugin removing inline server-only loaders from non-RSC module graphs.

Problem statement/demo:

https://github.com/user-attachments/assets/af83cb40-1354-46de-8935-957e5bd19ca7

Best reviewed commit-by-commit.

### DIFF
--- a/apps/vite-rsc-hono-react-cloudflare/src/lib/makeswift/components/rsc-kitchen-sink/register.ts
+++ b/apps/vite-rsc-hono-react-cloudflare/src/lib/makeswift/components/rsc-kitchen-sink/register.ts
@@ -1,5 +1,4 @@
-import { lazy } from 'react'
-import { ReactRuntime } from '@makeswift/vite-rsc'
+import { ReactRuntime, serverOnly } from '@makeswift/vite-rsc'
 
 import {
   Checkbox,
@@ -18,7 +17,7 @@ import {
 
 export const registerRscMarkdownComponent = (runtime: ReactRuntime) =>
   runtime.registerComponent(
-    lazy(() =>
+    serverOnly(() =>
       import('./server').then((mod) => ({ default: mod.RscKitchenSink })),
     ),
     {
--- a/packages/vite-rsc/jest.setup.ts
+++ b/packages/vite-rsc/jest.setup.ts
@@ -1,0 +1,3 @@
+import { name as PACKAGE_NAME, version as PACKAGE_VERSION } from './package.json'
+
+Object.assign(globalThis, { PACKAGE_NAME, PACKAGE_VERSION })
--- a/packages/vite-rsc/package.json
+++ b/packages/vite-rsc/package.json
@@ -23,6 +23,11 @@
       "require": "./dist/cjs/index.js",
       "import": "./dist/esm/index.js",
       "types": "./dist/types/index.d.ts"
+    },
+    "./plugin": {
+      "types": "./dist/types/plugin/index.d.ts",
+      "import": "./dist/esm/plugin/index.js",
+      "require": "./dist/cjs/plugin/index.js"
     }
   },
   "scripts": {
@@ -35,14 +40,16 @@
     "format:check": "prettier --check ."
   },
   "dependencies": {
-    "@makeswift/runtime": "workspace:*"
+    "@makeswift/runtime": "workspace:*",
+    "magic-string": "^0.30.21"
   },
   "peerDependencies": {
     "@types/react": "^19.0.0",
     "@types/react-dom": "^19.0.0",
     "@vitejs/plugin-rsc": "^0.5.34",
     "react": "^19.0.0",
-    "react-dom": "^19.0.0"
+    "react-dom": "^19.0.0",
+    "vite": "^8.0.0"
   },
   "peerDependenciesMeta": {
     "@types/react": {
@@ -58,11 +65,13 @@
     "@types/react": "^19.2.15",
     "@types/react-dom": "^19.2.3",
     "@vitejs/plugin-rsc": "^0.5.34",
+    "acorn": "^8.18.0",
     "concurrently": "^8.2.2",
     "jest": "^29.7.0",
     "react": "19.2.3",
     "react-dom": "19.2.3",
     "tsup": "^8.0.2",
-    "typescript": "^5.9.2"
+    "typescript": "^5.9.2",
+    "vite": "^8.2.2"
   }
 }
--- a/packages/vite-rsc/src/global.d.ts
+++ b/packages/vite-rsc/src/global.d.ts
@@ -1,0 +1,6 @@
+declare global {
+  const PACKAGE_NAME: string
+  const PACKAGE_VERSION: string
+}
+
+export {}
--- a/packages/vite-rsc/src/index.ts
+++ b/packages/vite-rsc/src/index.ts
@@ -1,3 +1,4 @@
 export { type SiteVersion } from '@makeswift/runtime/unstable-framework-support'
 export { ReactRuntime } from './runtime'
 export { ViteRSCFrameworkProvider } from './framework-provider'
+export { serverOnlyNoOp as serverOnly } from './server-only'
--- a/packages/vite-rsc/src/plugin/index.ts
+++ b/packages/vite-rsc/src/plugin/index.ts
@@ -30,7 +30,7 @@ export default function makeswiftViteRSC(): Plugin {
 
         return replaceServerLoaders({
           code,
-          ast: this.parse(code),
+          parse: code => this.parse(code),
           replacementExpr:
             '() => Promise.reject(new Error("Attempt to load server component outside the RSC environment"))',
         })
--- a/packages/vite-rsc/src/plugin/index.ts
+++ b/packages/vite-rsc/src/plugin/index.ts
@@ -1,0 +1,40 @@
+import { type Plugin } from 'vite'
+
+import { MARKER_SYMBOL, replaceServerLoaders } from './transform.js'
+
+export default function makeswiftViteRSC(): Plugin {
+  return {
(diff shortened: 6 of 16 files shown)


packages/vite-rsc/src/plugin/transform.test.ts
```diff
@@ -0,0 +1,76 @@
...
+const REPLACEMENT_EXPR = "() => Promise.reject(new Error('Invalid import'))"
+
+const transform = (code: string) =>
+  replaceServerLoaders({
+    code,
+    ast: parse(code, { ecmaVersion: 'latest', sourceType: 'module' }) as ESTree.Program,
+    replacementExpr: REPLACEMENT_EXPR,
+  })
+
+describe('replaceServerLoaders', () => {
```
agurtovoy, line 15: Tests for the code transform.

packages/vite-rsc/src/plugin/transform.ts
```diff
@@ -0,0 +1,135 @@
...
+  replacementExpr,
+}: {
+  code: string
+  ast: ESTree.Program
+  replacementExpr: string
+}) {
+  const markerSymbols = collectMarkerSymbols(ast)
+  if (markerSymbols.size === 0) return null
+
+  const output = new MagicString(code)
```
agurtovoy, line 33: [`MagicString`](https://www.npmjs.com/package/magic-string) is a funny-named utility for making positional edits to source text and generating a source map for the result.

packages/vite-rsc/src/global.d.ts
```diff
@@ -0,0 +1,6 @@
+declare global {
+  const PACKAGE_NAME: string
+  const PACKAGE_VERSION: string
+}
+
+export {}
```
agurtovoy, line 6: Add type declarations for [this code](https://github.com/makeswift/makeswift/pull/1440/changes/62e421b793e4ad08f8488cc26ec5d0cf52db9a7c#r3919276536) in `tsup.config.ts`.

packages/vite-rsc/tsup.config.ts
```diff
@@ -14,6 +14,7 @@ export default defineConfig(() => {
     sourcemap: true,
     legacyOutput: true,
     define: {
+      PACKAGE_NAME: JSON.stringify(name),
       PACKAGE_VERSION: JSON.stringify(version),
```
agurtovoy, line 18: Expose `PACKAGE_NAME` in addition to `PACKAGE_VERSION`.

packages/vite-rsc/src/index.ts
```diff
@@ -1,3 +1,4 @@
 export { type SiteVersion } from '@makeswift/runtime/unstable-framework-support'
 export { ReactRuntime } from './runtime'
 export { ViteRSCFrameworkProvider } from './framework-provider'
+export { serverOnlyNoOp as serverOnly } from './server-only'
```
agurtovoy, line 4: `serverOnly` never loads anything in non-RSC envs, see below.

packages/vite-rsc/jest.setup.ts
```diff
@@ -0,0 +1,3 @@
+import { name as PACKAGE_NAME, version as PACKAGE_VERSION } from './package.json'
+
+Object.assign(globalThis, { PACKAGE_NAME, PACKAGE_VERSION })
```
agurtovoy, line 3: Provide `PACKAGE_NAME` and  `PACKAGE_VERSION` in tests.

packages/vite-rsc/package.json
```diff
@@ -23,6 +23,11 @@
       "require": "./dist/cjs/index.js",
       "import": "./dist/esm/index.js",
       "types": "./dist/types/index.d.ts"
+    },
+    "./plugin": {
+      "types": "./dist/types/plugin/index.d.ts",
+      "import": "./dist/esm/plugin/index.js",
+      "require": "./dist/cjs/plugin/index.js"
```
agurtovoy, line 30: A separate entry point for the plugin.

packages/vite-rsc/package.json
```diff
@@ -58,11 +65,13 @@
     "@types/react": "^19.2.15",
     "@types/react-dom": "^19.2.3",
     "@vitejs/plugin-rsc": "^0.5.34",
+    "acorn": "^8.18.0",
```
agurtovoy, line 68: Dev dependency, used in tests only.

apps/vite-rsc-hono-react-cloudflare/src/lib/makeswift/components/rsc-kitchen-sink/register.ts
```diff
@@ -18,7 +17,7 @@ import {
 
 export const registerRscMarkdownComponent = (runtime: ReactRuntime) =>
   runtime.registerComponent(
-    lazy(() =>
+    serverOnly(() =>
```
agurtovoy, line 20: We could consider requiring `serverOnly` here when component metadata includes `server: true`, but I didn't go there as it's currently unclear whether we'll be able to provide the same RSC API across all of our supported frameworks.

packages/vite-rsc/src/plugin/index.ts
```diff
@@ -0,0 +1,40 @@
...
+        if (id.includes('/node_modules/') || this.environment.name === 'rsc') {
+          return null
+        }
+
+        return replaceServerLoaders({
+          code,
+          ast: this.parse(code),
+          replacementExpr:
+            '() => Promise.reject(new Error("Attempt to load server component outside the RSC environment"))',
+        })
```
arvinpoddar, line 36: Should we have `replaceServerLoaders` take in the code string and a function that converts the code string to an AST, rather than take in the AST directly? I ask because right now the interface is one of those situations where one input is derived from the other, but you _could_ pass an entirely unrelated AST. 

Obviously not a high priority given how/where this function is being used, this is more of a question for my own learning.

arvinpoddar (dismissed):
Clever solution and nice user-facing APIs, love it

fikrikarim (dismissed):
Nice! I searched for prior art and this feels similar to https://github.com/pcattori/vite-env-only#macros.

I was curious if there's any simpler implementation, and claude suggested one alternative:
```ts
// packages/vite-rsc/src/plugin/index.ts — the whole plugin
const SERVER_ONLY_IMPORT = /(?:^|[;\n])\s*import\s*['"]server-only['"]\s*;?/
const STUB = `throw new Error('A server-only module was imported outside the RSC environment.')`

export default function makeswiftViteRSC({
  serverComponents,
}: {
  serverComponents: Pattern | Pattern[]
}): Plugin {
  return {
    name: '@makeswift/vite-rsc',
    enforce: 'pre',
    transform: {
      filter: { id: serverComponents, code: /['"]server-only['"]/ },
      handler(code) {
        if (this.environment.name === 'rsc') return null
        if (!SERVER_ONLY_IMPORT.test(code)) return null
        return { code: STUB, map: null }
      },
    },
  }
}

// app vite.config.ts
makeswift({ serverComponents: '**/components/**/server.tsx' }),
```

The benefits are simpler plugin implementation and no new public API. The downside is that users need to follow the `'**/components/**/server.tsx'` convention, which might not be intuitive. Just sharing it here for an alternative.

agurtovoy:
> Nice! I searched for prior art and this feels similar to https://github.com/pcattori/vite-env-only#macros.

`vite-env-only` is definitely very similar in spirit, and obviously more comprehensive (at the cost of implementing a full-fledged module rewrite).

> I was curious if there's any simpler implementation [...]
> The benefits are simpler plugin implementation and no new public API. The downside is that users need to follow the '**/components/**/server.tsx' convention, which might not be intuitive. Just sharing it here for an alternative.

@fikrikarim thanks for sharing, it's an interesting take!

Technically, we don't have to impose the `**/components/**/server.tsx` convention, we could simply strip down all modules that import the `server-only` code since, well, they are server-only. I guess the main problem with doing so is that it would change the enforcement semantics: what would previously have been a build-time error could become a potential runtime failure. If the scope of the transformation is all modules, I think it becomes too broad of a decision to make on behalf of our users, which is probably why Claude suggested limiting the transform to `**/components/**/server.tsx`.

Overall, although the simplicity here is appealing, I prefer the explicitness of the approach in this PR. From a DX perspective, the new public API here is IMHO an advantage over relying on the presence of `import 'server-only'` in the component module. Because the "marker" for excluding a component import from non-RSC bundles is tied to the component registration itself, it's highly visible/hard to miss. When someone is copying an example from our docs, it's likely that the `serverOnly` marker will survive their edits; by contrast, the component they are registering is likely something that they already have and they might or might not remember to add `import 'server-only'`. 

Last but not least, we can follow up with compile-time enforcement requiring `serverOnly` when the component metadata includes `server: true`, which would really tie this together.

packages/vite-rsc/src/plugin/index.ts
agurtovoy: Great question! The reason these are passed separately is because `parse` is a method on the plugin itself. As you might have noticed, I had to bring in a [standalone compatible implementation](https://github.com/makeswift/makeswift/pull/1440/changes#r3919296680) of `parse` for the tests. What I could have done here is to pass `parse` as a callback, i.e.:

```ts
replaceServerLoaders({ code, parse: (code) => this.parse(code), replacementExpr: '...' })
```

This would be an improvement; I might actually do it, thanks for bringing this up!

agurtovoy: Done: https://github.com/makeswift/makeswift/pull/1440/changes/9ae8686b5a5adca28b762e9df352f544258004da

arvinpoddar (approved):

PR #1445, opened 2026-09-11
feat(runtime): Integrate with new /content/v1/element-trees endpoint

bc-drew-cutchins:
Integrates with the new /v1/content/element-trees endpoint implemented in [this PR](https://github.com/makeswift/cosmos/pull/3352).

The linked PR is the source of truth for the motivation behind this change.

Note that this PR has been iterated on from an initial /v3/element-trees endpoint, the 3 commits show the design changes.

### DIFF
--- a/.changeset/quiet-regions-return.md
+++ b/.changeset/quiet-regions-return.md
@@ -1,0 +1,5 @@
+---
+'@makeswift/runtime': patch
+---
+
+`getComponentSnapshot` now uses the `v3/element-trees` endpoint, which returns a `200` for regions that have no content yet. This will allow for caching empty regions on Next.js hosts. The experimental `unstable_enforceSuccess` option has been removed, as the new endpoint always returns a success response for well-formed requests.
--- a/packages/runtime/src/client/index.ts
+++ b/packages/runtime/src/client/index.ts
@@ -758,12 +758,10 @@ export class MakeswiftClient extends MakeswiftRestAPIClient {
       siteVersion: siteVersionPromise,
       locale,
       allowLocaleFallback = true,
-      unstable_enforceSuccess = false,
     }: {
       siteVersion: SiteVersion | null | Promise<SiteVersion | null>
       locale?: string
       allowLocaleFallback?: boolean
-      unstable_enforceSuccess?: boolean
     },
   ): Promise<MakeswiftComponentSnapshot> {
     const siteVersion = await siteVersionPromise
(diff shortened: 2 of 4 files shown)


fikrikarim:
@cursor review

.changeset/quiet-regions-return.md
```diff
@@ -0,0 +1,5 @@
+---
+'@makeswift/runtime': patch
+---
+
+`getComponentSnapshot` now uses the `v3/element-trees` endpoint, which returns a `200` for regions that have no content yet. This will allow for caching empty regions on Next.js hosts. The experimental `unstable_enforceSuccess` option has been removed, as the new endpoint always returns a success response for well-formed requests.
```
fikrikarim, line 5: We might want to communicate this with users that we previously asked to implement `unstable_enforceSuccess` in their repo, since it's technically a breaking change on a patch version for them.

fikrikarim (dismissed):
Looks good!

packages/runtime/src/client/index.ts
```diff
@@ -758,89 +758,49 @@ export class MakeswiftClient extends MakeswiftRestAPIClient {
...
-        key,
-        cacheData: CacheData.empty(),
-        meta: {
-          allowLocaleFallback,
-          requestedLocale: locale ?? null,
-        },
-      }
-    }
+    const document =
+      !allowLocaleFallback && documentResponse.locale != locale
```
agurtovoy, line 792: I wonder if we should make `allowLocaleFallback` as deprecated so we can remove it altogether in future versions.

agurtovoy (dismissed):

packages/runtime/src/client/index.ts
agurtovoy: To add more context, it was [initially added in the 0.24.0 release](https://docs.makeswift.com/developer/docs/guides/upgrading/0.24.0#localized-pages-are-online-by-default) as a "backward compatibility" measure to guard against unforeseen consequences of the new fallback behavior. IIRC, the only related use case that has come up since was the desire to have a localized page (e.g. for a region-specific promotion) with no corresponding base page, which can be accomplished by taking the base page offline.

bc-drew-cutchins: That makes sense, marked as deprecated. Thanks for the additional context 👍

agurtovoy (dismissed):

packages/runtime/src/client/index.ts
```diff
@@ -868,6 +828,7 @@ export class MakeswiftClient extends MakeswiftRestAPIClient {
     }: {
       siteVersion: SiteVersion | null | Promise<SiteVersion | null>
       locale?: string
+      /** @deprecated Disallowing locale fallback will be removed in a future release */
       allowLocaleFallback?: boolean
```
agurtovoy, line 831: I was going to say that we should deprecate this on all methods, not just this one, but on second thought, the context I gave [here](https://github.com/makeswift/makeswift/pull/1445#discussion_r4031487061) applies to pages only. I could conceivably see someone wanting to show an empty region instead of a base-locale one on a localized page. Let's revert this for now 🙏

bc-drew-cutchins: Reverted 👍

agurtovoy (approved):

PR #1446, opened 2026-09-16
feat(runtime): Utilize bulk global elements endpoint for introspection

bc-drew-cutchins:
As per [this suggestion in the Runtime Data Fetching Improvements document](https://linear.app/commerce/document/runtime-data-fetching-improvements-55bc483012e4#2-bulk-global-element-retrieval-00b208df), this PR refactors introspection to utilize the new `v1/content/global-elements/bulk` endpoint implemented in [this cosmos PR](https://github.com/makeswift/cosmos/pull/3368).

This is achieved by refactoring the existing depth first traversal of the element trees to defer traversals when a reference to an unfetched global element is found. Once all traversals have either been completed or deferred, the required global elements are fetched in bulk and the deferred traversals are resumed. This continues until all traversals have completed.

### Testing
See the network requests made for the same page load below. Note that the graph lengths are normalized and represent different spans of time.

#### Local Orion - Animal Site
This site has multiple global elements with shallow nesting.
Screenshot 2026-09-18 at 1 17 30 PM

##### Before
Screenshot 2026-09-18 at 11 42 01 AM

##### After
Screenshot 2026-09-18 at 11 39 04 AM

#### Staging Orion - Stacked Rainbow
This site has deeply nested global elements, but they are all mounted at the top level.

Screenshot 2026-09-18 at 1 15 36 PM

##### Before
Screenshot 2026-09-18 at 12 56 15 PM

##### After
Screenshot 2026-09-18 at 12 54 44 PM

#### Staging Orion - Super Rainbow
This site has a deeply nested global element, this is the worst case scenario for bulk fetching and equivalent to fetching individually until the last level is reached containing multiple global elements (each color).

Screenshot 2026-09-18 at 1 14 54 PM

##### Before
Screenshot 2026-09-18 at 1 12 42 PM

##### After
Screenshot 2026-09-18 at 1 11 36 PM


packages/runtime/src/client/index.ts
```diff
@@ -297,161 +297,19 @@ export class MakeswiftClient extends MakeswiftRestAPIClient {
     return { ...unversionedResources, swatches }
   }
 
-  // TODO: Consolidate this method with the introspectMany method once the
```
bc-drew-cutchins, line 300: unstable_getComponentSnapshots is not stable, but introspectMany is a very similar implementation and not using it would requires duplicate implementations of the use of the bulk endpoint... Not sure if we want to hold off on this and would rather duplicate the bulk aggregation logic.

packages/runtime/src/client/index.ts
```diff
@@ -467,125 +325,144 @@ export class MakeswiftClient extends MakeswiftRestAPIClient {
...
-          })
+          }
         }
       }
+
+      // If there are no pendingGlobalElementIds then all elements have been traversed completely
+      if (pendingGlobalElementIds.size === 0) break
+
+      const ids = [...pendingGlobalElementIds]
+      const results = await this.getGlobalElements(ids, siteVersion, { locale })
```
agurtovoy, line 458: If we can the max number of elements in the batch ([which I think we should](https://github.com/makeswift/cosmos/pull/3368/changes#r4031823303)), it'd be better to trigger the fetch as soon as we accumulated that many elements rather than at the end of the iteration.

bc-drew-cutchins: Updated to split into batches. 

You have a better intuition for this data than I do, but I'd be very surprised if fetching as soon as a number of elements were accumulated had a measurable performance difference? I would assume the in memory object traversal would be a drop in the bucket compared to the network request. Sending these requests ahead of time would complicate this code as we would then need to track in flight fetches and add handling for resolving an earlier fetch when blocking on a later one.

If you think this is important though we can build up the abstractions to accomplish this.

agurtovoy: No, you're right, even in the context of hundreds of element trees my concern here might be completely misguided, and I don't understand all the variables well enough to validate/disprove it, so let's go forward with what you have and see how it holds up.

packages/runtime/src/api/rest-api-client.ts
```diff
@@ -139,6 +146,47 @@ export class MakeswiftRestAPIClient {
...
+      const url = new URL(`content/v1/global-elements/bulk`, this.apiOrigin)
+
+      ids.forEach(id => url.searchParams.append('ids', id))
+      if (locale != null) url.searchParams.set('locale', locale)
+
+      const response = await this.fetch(url.pathname + url.search, siteVersion)
+
+      if (!response.ok) {
+        const failedBody = await failedResponseBody(response)
+        // 404 means the requested version has no commit (e.g., site never published)
```
agurtovoy: I don't think the bulk endpoint ever returns 404, does it? Also, this comment (specifically, the "site never published" scenario) makes me question the validity of my comment [here](https://github.com/makeswift/cosmos/pull/3352#discussion_r4039815790).

bc-drew-cutchins: It does not, this was leftover from the individual fetch code. My bad. Removed.

agurtovoy (approved):
Nice work!

packages/runtime/src/client/index.ts
agurtovoy: > // TODO: Consolidate this method with the introspectMany method once the
> // unstable_getComponentSnapshots method is stable and tested in production.

I actually don't understand this comment. AFAIK, `unstable_getComponentSnapshots` is not used in `introspectMany`, is it?

bc-drew-cutchins: Other way around, `introspectMany` is used in `unstable_getComponentSnapshots` (previously only used there). So I believe this means: "Once introspectMany has been proven to work by the use of unstable_getComponentSnapshots, we will trust it's implementation enough to be delegated to by introspect"

agurtovoy: Got it, thanks for clarifying!

fikrikarim (approved):

arvinpoddar (approved):

